Issue #195, opened 2014-10-29
Jenkins support


pjf:
@Sarbian has an [awesome jenkins server](https://ksp.sarbian.com/jenkins/), that builds many important modules, including ModuleManager and MechJeb2.

It has a pretty easy JSON API (just add `/api` to the end of any URL), so it would be possible for us to check for new releases.

@Sarbian, are you happy for us to query your server and index relevant build artefacts into the CKAN? If so, I'd like to do the following:

- [x] Add support for Jenkins to the NetKAN
- [x] Index MechJeb2 via Jenkins query
- [ ] Index ModuleManager via Jenkins query.
- [ ] Check if there are other mods for which build artefacts should be considered "stable" from Jenkins.

Many thanks!

~ Paul

hakan42:
* Work on netkan is done
* MechJeb2 is being indexed via our Jenkins from @sarbian 's Jenkins
* ModuleManager could be indexed next

PR #619, opened 2014-12-17
Add jenkins resource to schema

hakan42:
In preparation for Jenkins inflator support in #195, I want to add a "jenkins" property to the list of possible resources.

Even though this should have no negative impact on existing clients (does ckan.exe validate ckan files before installing?), merging can also wait for 1.4.0


hakan42:
Superceded by https://github.com/KSP-CKAN/CKAN/pull/738

PR #738, opened 2015-01-08
Add ci resource to schema

hakan42:
Supercedes https://github.com/KSP-CKAN/CKAN/pull/619 , as I don't know how to modify PRs from repos that don't exist anymore :blush: 

For use in the Jenkins inflator support in KSP-CKAN/CKAN#195, I want to add a "ci" property to the list of possible resources.

The current netkan.exe reads "x_ci" as a fallback, but having a properly defined property in the schema would make validation possible.

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -96,10 +96,10 @@
             "description" : "Optional list of recommended, but not essential mods",
             "$ref"        : "#/definitions/relationship"
         },
-	"supports" : {
+    "supports" : {
             "description" : "Optional list of supported mods",
             "$ref"        : "#/definitions/relationship"
-	},
+    },
         "conflicts" : {
             "description" : "Optional list of conflicting mods",
             "$ref"        : "#/definitions/relationship"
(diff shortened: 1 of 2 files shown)


AlexanderDzhoganov:
I'm not sure.. why do we need a "ci" resource? Can we do something like "$kref": "#/ckan/jenkins/https://ksp.sarbian.com/jenkins/job/MechJeb2-Dev/"? We already do this for $vref's, I don't see why not encode the path to the CI directly in the kref. Any thoughts?

hakan42:
TODO: Add schema version testing as in https://github.com/KSP-CKAN/NetKAN/commit/8aab50bea29f12613e30fab18d9ff0d0fa189bc3#diff-8f2d300fc53ee5d8176bf1fd4aba1a3b , if and when we revive the perl-based tests

hakan42:
We already use "protocol"-specific kref strings with github and kerbalstuff, the only difference in those two is that the base url is fixed and can be hard-coded in the Inflator class.

If we have a separate resource, this can later on used for other purposes too (like making a nice web interface or whatever thing we cannot think of as yet).

Issue #956, opened 2015-05-25
Formal specification for ckan clients


Ippo343:
Currently, the spec only defines the structure of the metadata.
However, there is no formal specification for how the client must behave with the metadata: as a result, other implementations of the client might end up implementing a different behaviour than the official client.

I propose that we formalize the desired behaviour that the clients should implement: each review of the metadata spec should also come with a specification of the expected behaviour.

pjf:
Oh my, I shuddered at the word "official". Perhaps we can call it extant? :)

> I propose that we formalize the desired behaviour that the clients should implement.

I agree with this. It makes perfect sense to say something like "a client *must not* allow two or more conflicting mods to be installed" when we're talking about conflicts, for example.

This is also nice because it means we can write a test suite that directly exercises the spec.

It's worth noting that a lot of things the clients do will be out of the scope of the spec. Relationship resolution, for example, can be done in *numerous* ways, and it would be detrimental to enforce clients to work one way or the other.

Ippo343:
It's worth noting that a lot of things the clients do will be out of the scope of the spec.

One more reason to formalize what a client must do (and what not) in order to be officially compatible with ckan :)

Personally I would hesitate to say that a client "must not" allow conflicting mods (as this prohibits a --force option if a client wishes to implement that). Later today I will start on a first draft :)

HebaruSan:
Sorry to be a stick in the mud, but what's the value to be provided here in practical terms? It mostly seems to me that this would create distractions rather than advancing the project...

- If such a spec was available, would the CKAN dev team then take on the role of certifying compliant clients?
  - If not, then why should anyone pay attention to it?
  - If we determine that a client isn't compliant, then so what? What consequences would that have other than creating friction with another team, if their app still "worked" in their eyes?
- Would we have to extend the spec (and debate its language) before adding enhancements?
- Does ckan.exe drift in and out of "compliance" as bugs are found and fixed?
- Is anyone actually going to write a fully featured alternative client when CKAN itself is already available? Such ideas come up from time to time (e.g., #2468), but generally go nowhere because the person proposing it just wants to be the leader while other people do all the work (the classic *I'm not a coder but I have this great idea...*).
- Is composing such a spec even possible? We only have one example clause so far (how to handle mod conflicts), and already there is disagreement about whether it's a valid requirement. Aren't we likely to have that exact disagreement about any possible requirement that could be put forward?

politas:
Not likely to happen, even if it was a good idea.

Issue #949, opened 2015-05-23
Saner handling/overwriting of autodetected mods


pjf:
When the CKAN project was first founded, we had the idea of "autodetected" mods that could satisfy dependencies, but not be managed by the client. At the time we had very few mods indexed, and the expectation was that users wouldn't want to uninstall their existing mods just to install them again via the CKAN.

Now things have changed. We have a _lot_ of mods indexed, we have a _lot_ of users, and many folks are exclusively using the CKAN for managing their mods.

We very much need a way to convert "autodetected" mods over to being first class CKAN mods, but this is a relatively hard process. We can't easily detect the version of a mod that the user has installed, and a user may have _additional_ files that the CKAN wouldn't really know about.

For this to work properly, I believe the CKAN would be benefitted by a lot more filesystem awareness, as well as extensions to the spec to make sure we can transition mods correctly. In particular, I believe we need:
- A metadata keyword to indicate files which should be _preserved_ on upgrade.
- A metadata keyword to indicate _additional_ files which should be deleted on upgrade.

These two keywords are useful in their own right, by allowing mods to be upgraded with config intact, or to remove ephemeral files that should disappear during an upgrade. However they _also_ pave the way for sane conversion of pre-installed mods. A migration may consist of:
- Installing the mod as usual, but _overwriting_ files which are in GameData but unowned by CKAN.
- _Except_ where those files are marked as preserved
- In addition, we should delete any files which are specified as needing deletion, provided those files are also unowned.

This means we don't need to know the version of a mod we're coming from, but we _can_ know the version of a mod we're going _to_.

Once we have a sane upgrade process in place, we can work towards _removing_ autodetected mods entirely. Users with existing installs would be required to convert their mods to CKAN control if they were needed as dependencies.

This is all very much fraught with danger; in particular if one mod's metadata suggests it should delete files owned by another mod (which at least we can detect at install time).

---

Edit: Doing this would also make a fine time to examine the relationship between the `Module` and `InstalledModule` classes internally, and if it still makes sense to have them separate.

pjf:
Adding here for the moment, as closely related:

- We should give users a way to scan for untracked files and report them. These can indicate that we've got incomplete metadata, or simply show the presence of manually installed mods.
[...]

PR #961, opened 2015-05-26
Formal client specification.

Ippo343:
This pull requests aims to implement a formal specification for the behaviour of ckan clients.
See also issue #956.

### DIFF
--- a/client-spec/client-specification-v1.md
+++ b/client-spec/client-specification-v1.md
@@ -1,0 +1,322 @@
+# Abstract
+
+This document describes the behaviour that an application must implement
+in order to be a CKAN-compatible mod manager.
+
+Adherence to this document ensures that all mod-management applications
+that consume CKAN metadata behave consistently.
+
+This document refers to version 1 of the CKAN metadata specification.
+
+
+
+# Definitions
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in  [RFC 2119](http://www.faqs.org/rfcs/rfc2119.html).
+
+
+
+# Metadata fields
+
+CKAN metadata consists of a JSON file
+conforming to the official JSON schema.
+
+The following sections define how the information contained in the metadata files
+should be interpreted and handled by ckan clients.
+
+
+## Required Fields
+
+The fields described in this section are required by the ckan metadata specification
+and must be present for all valid metadata files.
+A ckan client may assume that all these fields are present in all valid metadata files.
+Clients may assume that all metadata contained in the official CKAN repositories is valid;
+however, client-side validation against the JSON schema is recommended.
+
+### identifier
+
+This is the unique identifier for a module.
+The same identifier is shared by multiple versions of the same module.
+
+A client must use this identifier to discriminate different modules
+and must not allow the user to install two modules with the same identifier.
+
+### version
+
+A string defining the version of the module.
+The specification allows for any string to be used as a version.
+Since there is no standard regarding versioning schemes, a client should not assume any.
+
+As a significant amount of metadata in the official repository
+is obtained automatically from github, leading 'v' characters are common:
+as such, version comparison according to [PEP 440](https://www.python.org/dev/peps/pep-0440/#version-scheme) is recommended.
+Simple alphabetical ordering will often lead to incorrect comparisons
+and is therefore discouraged.
+
+Clients must not allow the installation of multiple versions
+of the same module at the same time.
+
+### download
+
+The download field contains an url where the module can be downloaded.
+This field can contain any url:
+clients must not assume this field to contain a direct link,
+but they must follow eventual redirects.
+
+The download field will always point to a zip archive:
+therefore clients must be able to process zip archives.
+Clients may assume that the downloaded archive is a single-disk,
+unencrypted archive with no password protection.
+
+### spec-version
+
+This field contains the version of the metadata specification
+that must be used to process the metadata file.
+
+A client must not allow the installation of a file that requires
+a version of the specification that it does not implement completely.
+It is recommended to hide incompatible modules from the list of available modules, if any.
+
+### name
+
+This field contains a user-friendly name that can be used to present the mod to the user.
+This field is not necessarily unique;
+
+clients must not identify a package using its `name`.
+However, it is recommended to use this field to help users
+search for the identifier of a module.
+
+### abstract
+
+This field contains a short description of what a module contains.
+Clients should use the value of this field to provide a description to the user, if any.
+
+### license
+
+This field is required by the metadata specification.
+Clients are not prescribed any behaviour regarding the value of this field.
+
+
+## Optional Fields
+
+The following fields are not required by the specification.
+Clients should not assume they are present in valid metadata;
+however, if they are present, clients must implement
+the behaviours prescribed in the following sections.
+
+### description
+
+The description field contains a longer description of the module
+than the one provided in the `abstract` field.
+Clients may use this field to display a verbose description of the module.
+
+### comment
+
+A `comment` field may appear at any point of the metadata.
+These fields will always contain a string.
+The value of a `comment` should not be displayed to the user,
+but may be employed for any purpose.
+
+### author
+
+Either a string, or a list of strings.
+The content of the `author` field should be shown to the user,
+and it should be used to search for specific modules if present.
+
+### download_size
+
+Size (in bytes) of the file downloaded through the `download` link.
+Showing the total download size to the user before any download
+is recommended, but not required.
+
+### release_status
+
+This field can have the following values:
+- `stable`
+- `testing`
+- `development`
+
+This field may be use to restrict the list of installable packages
+(e.g, to only allow `stable` modules).
+
+If this field is missing, clients must assume its value to be `stable`.
+
+### ksp_version
+
+This field may contain either:
+- `any`
+- a string matching the regular expression:
+		[0-9]+\\.[0-9]+(\\.[0-9]+)?
+
+If the field contains the string `any`,
+clients must allow the installation of the module
+on any version of the game,
+unless the `ksp_version_min` and `ksp_version_max` are also present:
+in this case, the client must respect these constraints.
+
+Otherwise, the client must only allow the installation of the mod
+into a copy of the game with a matching version number.
+
+### ksp_version_min and ksp_version_max
+
+Like the `ksp_version` field, these fields can contain either the string `any`
+or a version number.
+
+If `ksp_version` is `any` and either of these fields is defined,
+the client must enforce these constraints.
+
+If `ksp_version` contains a specific version number,
+these two fields should not be present.
+In this case, the client must disregard their value and only enforce
+the most specific version constraint.
+
+
+## install stanzas
+
+Installation instructions are specified using the `install` field.
+This *optional* field contains an array of install directive
+(called "stanzas" in some of the documentation).
+Each stanza defines where to install a set of files.
+
+### Required: file and install_to
+
+If an `install` stanza is present,
+it must contain at least two fields:
+
+- `file`: path of a file or folder, relative to the top-level of the zip file obtained through the `download` link. The file (or folder) must be extracted to the location specified by the `install_to` field.
+- `install_to`: the location where the file or folder must be extracted. This field can only take one of the following 4 values:
+	. `GameRoot`: the root folder of the KSP installation;
+	. `GameData`: equal to `$(GameRoot)/GameData`;
+    . `Ships`: equal to `$(GameRoot)/Ships`
+    . `Tutorial`: equal to #TODO: where!?
+
+The client must not allow the installation of any file in any directory other than those listed above.
+The file or folder located by the `file` field must be extracted to the destination directory
+preserving the directory tree structure in the zip file.
+Installation of any file outside of the game's directory is strictly forbidden.
+
+### Optional: filters
+
+If a `filter` field is specified,
+it contains one or more files that must be excluded from the installation
+even if they are selected by the corresponding `file` field.
+The `filter` field only affects files in the same install stanza
+(that is, install stanzas are independent from each other).
+
+Analogously, if a `filter_regexp` is specified
+the client must exclude from the installation
+all the files that match one or more of the listed regular expressions.
+
+### File overwriting
+
+File overwriting is not permitted.
+Any attempt to overwrite another file
+must result in an error,
+aborting the installation attempt.
+
+### Rolling back
+
+In case of any error during an installation,
+the client must be able to roll back any change
+restoring the state of the KSP installation
+prior to the installation attempt.
+
+Partial installations are not permitted:
+the whole installation process must be implemented
+as an atomic operation
+(i.e, it either succeeds or fails completely).
+
+
+## Module relationships
+
+Modules can be put in relationship with one another using one or more relationship descriptors.
+
+There are 4 types of relationship fields:
+1. `depends`
+2. `recommends`
+3. `suggests`
+4. `conflicts`
+
+These 4 fields are all optional.
+If present, they contain a list of relationship descriptors with the following fields:
+- `name`: contains the `identifier` of the related module (required);
+- `version`: the specific version of the related module (optional);
+- `min_version` and `max_version`: allows to specify a range of versions (optional).
+
+Version ranges are only valid for the `conflicts` relationship:
+if version ranges are encountered in other relationship descriptors,
+the behaviour is undefined: clients may choose how to resolve the ambiguity.
+
+### depends
+
+Indicates that a module cannot be installed unless all its dependencies are not installed too.
+Clients must not install a module unless all of its dependencies are either installed,
+or are about to be installed alongside the module.
+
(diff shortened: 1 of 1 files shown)


Ippo343:
Current state: I wrote a starting point for v1 specification.
For the moment I am only working on version 1 of the specification: once we will have discussed and accepted version 1 (which is by far the biggest part of the work), I will work on supplementing it with subsequent extensions.

Please see the line comments for the parts where I would need some feedback.

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+
+This document refers to version 1 of the CKAN metadata specification.
+
+
+
+# Definitions
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in  [RFC 2119](http://www.faqs.org/rfcs/rfc2119.html).
+
+
```
Ippo343: TODO: Introducing the definitions for some of the terms using further down.

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+### version
+
+A string defining the version of the module.
+The specification allows for any string to be used as a version.
+Since there is no standard regarding versioning schemes, a client should not assume any.
+
+As a significant amount of metadata in the official repository
+is obtained automatically from github, leading 'v' characters are common:
+as such, version comparison according to [PEP 440](https://www.python.org/dev/peps/pep-0440/#version-scheme) is recommended.
```
Ippo343: PEP 440 is nice... however, it should be noted that not even the official client implements it, so maybe recommending it is a little bold?

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+of the same module at the same time.
+
+### download
+
+The download field contains an url where the module can be downloaded.
+This field can contain any url:
+clients must not assume this field to contain a direct link,
+but they must follow eventual redirects.
+
+The download field will always point to a zip archive:
```
Ippo343: ... will it?

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+
+This field can have the following values:
+- `stable`
+- `testing`
+- `development`
+
+This field may be use to restrict the list of installable packages
+(e.g, to only allow `stable` modules).
+
+If this field is missing, clients must assume its value to be `stable`.
```
Ippo343: Is this correct?

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+This field may contain either:
+- `any`
+- a string matching the regular expression:
+		[0-9]+\\.[0-9]+(\\.[0-9]+)?
+
+If the field contains the string `any`,
+clients must allow the installation of the module
+on any version of the game,
+unless the `ksp_version_min` and `ksp_version_max` are also present:
+in this case, the client must respect these constraints.
```
Ippo343: Is it possible that these fields are found together? Or should they be considered as invalid metadata?

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+Like the `ksp_version` field, these fields can contain either the string `any`
+or a version number.
+
+If `ksp_version` is `any` and either of these fields is defined,
+the client must enforce these constraints.
+
+If `ksp_version` contains a specific version number,
+these two fields should not be present.
+In this case, the client must disregard their value and only enforce
+the most specific version constraint.
```
Ippo343: This looks sane to me, but I'd like to have some feedback regarding this.

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+
+
+## install stanzas
```
Ippo343: What is the proper behaviour if the install stanza is missing?

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+
+If an `install` stanza is present,
+it must contain at least two fields:
+
+- `file`: path of a file or folder, relative to the top-level of the zip file obtained through the `download` link. The file (or folder) must be extracted to the location specified by the `install_to` field.
+- `install_to`: the location where the file or folder must be extracted. This field can only take one of the following 4 values:
+	. `GameRoot`: the root folder of the KSP installation;
+	. `GameData`: equal to `$(GameRoot)/GameData`;
+    . `Ships`: equal to `$(GameRoot)/Ships`
+    . `Tutorial`: equal to #TODO: where!?
```
Ippo343: Where are tutorials installed, exactly? And what should be allowed there?

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+The `filter` field only affects files in the same install stanza
+(that is, install stanzas are independent from each other).
+
+Analogously, if a `filter_regexp` is specified
+the client must exclude from the installation
+all the files that match one or more of the listed regular expressions.
+
+### File overwriting
+
+File overwriting is not permitted.
```
Ippo343: ... ever?

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+Any attempt to overwrite another file
+must result in an error,
+aborting the installation attempt.
+
+### Rolling back
```
Ippo343: This part is a little tricky. Transactions are not a trivial thing and might not be available for every Language.
Should this constraint be forced on *every* client?

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+
+These 4 fields are all optional.
+If present, they contain a list of relationship descriptors with the following fields:
+- `name`: contains the `identifier` of the related module (required);
+- `version`: the specific version of the related module (optional);
+- `min_version` and `max_version`: allows to specify a range of versions (optional).
+
+Version ranges are only valid for the `conflicts` relationship:
+if version ranges are encountered in other relationship descriptors,
+the behaviour is undefined: clients may choose how to resolve the ambiguity.
```
Ippo343: What is the proper behaviour for recommending a range of versions?

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+this mechanism, if implemented, must be explicitly selected by the user
+for each install process.
+
+### provides
+
+Additionally, modules may indicate they provide a name.
+Providing must be used to signal interchangeable modules:
+modules may be in relationships with a provided name.
+
+Version constraints are not valid when referring to a provided name.
```
Ippo343: Is this true?

client-spec/client-specification-v1.md
Dazpoet: "If no install sections are provided, a CKAN client must find the top-most directory in the archive that matches the module identifier, and install that with a target of GameData." fetched from https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#install

client-spec/client-specification-v1.md
Dazpoet: https://github.com/KSP-CKAN/CKAN/issues/949 is probably relevant for for spec_version 1, overwriting is not allowed ever afaik. However if a client would handle updating by overwriting old files somehow... then maybe overwriting is ok? I might have missed it or does this document not handle updating of modules?

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,322 @@
...
+
+If a package conflicts with a provided name,
+clients must consider all the packages that provide that name
+as conflicting and deal with them as described in the `conflicts` section.
```
Dazpoet: I'm pretty sure the CKAN client doesn't do this? If it does why do we add both provides and conflicts to all mods who does this? Sorry for perhaps tangential question.

client-spec/client-specification-v1.md
Dazpoet: No, see e.g. https://github.com/KSP-CKAN/CKAN-meta/blob/master/KAS/KAS-0.4.10.ckan

client-spec/client-specification-v1.md
Dazpoet: Afaik the only 2 examples where both fieldtypes exists are https://github.com/KSP-CKAN/CKAN-meta/blob/master/kribubble/kribubble-1.1.ckan and https://github.com/KSP-CKAN/CKAN-meta/blob/master/krispeeder1/krispeeder1-1.0.ckan. None of them show up in my KSP 0.25 testinstall which makes it appear that `ksp_version_min` trumps `ksp_version`

Looking at the current spec.md and how we gather data from ksp-avc through [`$vref`](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#ckanksp-avc) further enforce that `ksp_version_min` and `ksp_version_max` takes precedence over `ksp_version` in all cases.

client-spec/client-specification-v1.md
Ippo343: Actually it does, if you follow the redirect. KS urls have the same behaviour: the url is that of a webpage, but it redirects you to a zip file.

Dazpoet: Ah I thought you implied that the string in the field would always end in .zip. Thanks for clearing that up :)

client-spec/client-specification-v1.md
Ippo343: Actually, it does: for example FAR [conflicts with AeroDynamicModel](https://github.com/KSP-CKAN/CKAN-meta/blob/master/FerramAerospaceResearch/FerramAerospaceResearch-v0.15.2_Ferri.ckan#L15-L18), which makes it incompatible with any other mod that provides that name (for example, NEAR).

Dazpoet: :+1:

client-spec/client-specification-v1.md
pjf: There seems to be endless opinions on how we *should* handle version comparisons. The metadata spec explicitly mandates that clients *must* use the Debian version comparison rules, and the C# implementation follows these exactly.

I'm hesitant to change how we do version comparisons for two reasons:

1. We index more than 700 mods. Of those, we've only had a fraction where the Debian versioning spec has not worked, and changing the versioning implementation runs a *very real* risk of breaking existing mods.
1. The netkan pre-processor now has additional options for post-processing of versions into standard forms, or bumping the epoch when versions are truly wacky.
1. Changing versioning schemes is *much* harder than everyone seems to think. If we spot a document implementing spec 1.0–1.6, then we should use the old versioning scheme. If we spot a document with the "new" spec, we should use the new versioning scheme. If we're trying to compare *between* routines, we're screwed.

TL;DR: We're not going to change versioning schemes unless someone can provide *very* compelling reasons, a proof that we wont break existing metadata, and a sane migration plan.

TL;DR;TL;DR: We use the Debian spec.

client-spec/client-specification-v1.md
pjf: I wouldn't mandate that it point to a .zip file, I'd really love us to support .tar.gz and .tar.bz2 in the future, as both provide better compression and can probably even be handled with our existing libraries. They're certainly both unencumbered formats.

However clients *should* support zip files, as they're not going to be very functional without that.

client-spec/client-specification-v1.md
pjf: You can have:
-  `ksp_version`, OR
-  `ksp_version_min` and/or `ksp_version_max`

Under the hood, `ksp_version` is effectively a short-cut for setting both `_min` and `_max`. The form `0.23` would set `_min` to `0.23.0` and max to `0.25.MAXINT`, whereas `0.23.5` would have them both be set exactly.

client-spec/client-specification-v1.md
pjf: Not yet. Upgrades currently remove the old mod entirely and install the new one. Overwriting a file owned by another mod is a *hard problem*. Not only we'd have to flip the file ownership in the registry, but we'd also need to figure out what to do if that mod is uninstalled, but the mod it overwrites was not.

I need to see how Debian solves this problem; I have a hunch they require that you can only overwrite another package if you *replace it entirely*, and that this is really special magic for upgrading libc and other critical files which simply can't removed and replaced.

TL;DR: File overwriting is not permitted.

client-spec/client-specification-v1.md
pjf: Clients *must* leave GameData and the registry in a consistent state. How they choose to do that is up to them, but it can include checking for file conflicts before installing anything, or they can take the more lazy approach and use transactions.

So no, you don't have to have transactions, but clients are responsible for making sure things are consistent.

client-spec/client-specification-v1.md
pjf: I'm not sure I understand the question.

client-spec/client-specification-v1.md
pjf: It is! And again, this is straight out of the Debian spec. The idea is that `provides` is almost always for providing functionality¹ where there are one or more choices, but these choices may have entirely different versioning schemes underneath. I can probably dig up the relevant part of the Debian spec, as my mind seems to recall there was another monster that we avoid by doing this, but I can't remember which monster it is.

¹ Although in Debian provides is also used for package replacement, so existing relationships can be preserved when a package is obsoleted in its entirety by another package.

client-spec/client-specification-v1.md
pjf: Aye, Provides + Conflicts is the backbone for "you can have one, but ONLY one, of these options". It's useful for things like aerodynamic models, texture packs, career modes, and all manner of other things.

client-spec/client-specification-v1.md
Ippo343: Ok, that's what we use. What should we prescribe to clients?
1. Clients MUST implement the debian version comparison, or
2. Clients SHOULD implement the debian version comparison

client-spec/client-specification-v1.md
Ippo343: True that; but as of spec 1 (i.e, last november), were other archive formats allowed?
I seem to recall that originally (again, spec 1) only zip files were supported (which posed a problem with Module Manager, that had to be repackaged in a zip).
Again, IIRC, I might not.

client-spec/client-specification-v1.md
Ippo343: Still open after first revision.

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,316 @@
...
+The file or folder located by the `file` field must be extracted to the destination directory
+preserving the directory tree structure in the zip file.
+Installation of any file outside of the game's directory is strictly forbidden.
+
+### Optional: filters
+
+If a `filter` field is specified,
+it contains one or more files that must be excluded from the installation
+even if they are selected by the corresponding `file` field.
+The `filter` field only affects files in the same install stanza
```
Dazpoet: The `filter` field also affects folders in the same install stanza, or is that somehow included in files? Wasn't obvious to me so thought I'd point it out.

pjf:
@Ippo343 : I'm looking to close out old PRs, and I'm not exactly sure what to do with this one. If you're still looking at building a Python core then I'm happy to review and comment, although in terms of ongoing maintenance I feel that integrating this into the existing metadata spec would ensure it continues to be updated going forward.

Alternatively, if you're happy to close this out, then that's one step  closer to my goal of PR-box zero. :)

client-spec/client-specification-v1.md
```diff
@@ -0,0 +1,316 @@
...
+(that is, install stanzas are independent from each other).
+
+Analogously, if a `filter_regexp` is specified
+the client must exclude from the installation
+all the files that match one or more of the listed regular expressions.
+
+### File overwriting
```
plague006: We might want to remove this entry. Issues like https://github.com/KSP-CKAN/CKAN/issues/1311 show that we have thought about (and generally support) a limited method for overwriting so I don't see why we would impose this restriction on others.

Ippo343:
Hey Paul, I've moved on to other things, so (sadly) I don't think I will continue with the python client. This can be closed safely.
________________________________
Da: Paul Fenwick<mailto:notifications@github.com>
Inviato: ‎13/‎12/‎2015 07:18
A: KSP-CKAN/CKAN<mailto:CKAN@noreply.github.com>
Cc: Michele<mailto:m.ippolito@outlook.com>
Oggetto: Re: [CKAN] Formal client specification. (#961)

@Ippo343 : I'm looking to close out old PRs, and I'm not exactly sure what to do with this one. If you're still looking at building a Python core then I'm happy to review and comment, although in terms of ongoing maintenance I feel that integrating this into the existing metadata spec would ensure it continues to be updated going forward.

Alternatively, if you're happy to close this out, then that's one step  closer to my goal of PR-box zero. :)

---

pjf:
> Hey Paul, I've moved on to other things, so (sadly) I don't think I will continue with the python client. 

This is code for "being an awesome guitarist", yes? :)

> This can be closed safely. 

Roger that! One step closer to PR-box zero!

P.S. You're amazing!

Ippo343:
This is code for "being an awesome guitarist", yes? :)

AND skydiver AND cook :P And watch out next year for the paragliding course ;)

    P.S. You're amazing!

Nah, I just have a good role model to follow, that being you :)

PR #986, opened 2015-05-31
Change licence in Module to be a list of licences.

RichardLake:
By the spec a mod can have more than one.
Also make the appropriate changes to the gui and cmdline


pjf:
Testing this now. This is a reminder to myself that we may need to bump the registry version, as older clients may not be able to read the multi-license format.

pjf:
Tests are looking good, and I'm thinking of bumping our client to v1.8. Even though the spec has always specified that multi-license is fine, I suspect any older clients will barf if they try to read a file containing them.

PR #988, opened 2015-05-31
Rel resolver rework

RichardLake:
Moved here from https://github.com/KSP-CKAN/CKAN-core/pull/116

### DIFF
--- a/Core/Relationships/RelationshipResolver.cs
+++ b/Core/Relationships/RelationshipResolver.cs
@@ -208,6 +208,7 @@ namespace CKAN
             foreach (var module in mods)
             {
                 installed_modules.Remove(module);
+                conflicts.RemoveAll(kvp => kvp.Key.Equals(module) || kvp.Value.Equals(module));
             }
         }
 
--- a/Core/Relationships/RelationshipResolver.cs
+++ b/Core/Relationships/RelationshipResolver.cs
@@ -57,7 +57,7 @@ namespace CKAN
     }
 
     // TODO: RR currently conducts a depth-first resolution of requirements. While we do the
-    // right thing in processing all depdenencies first, then recommends, and then suggests,
+    // right thing in processing all dependencies first, then recommends, and then suggests,
     // we could find that a recommendation many layers deep prevents a recommendation in the
     // original mod's recommends list.
     //
@@ -66,6 +66,14 @@ namespace CKAN
 
     // TODO: Add mechanism so that clients can add mods with relationshup other than UserAdded. 
     // Currently only made to support the with_{} options. 
+
+
+    /// <summary>
+    /// A class used to resolve relationships between mods. Primarily used to satisfy missing dependencies and to check for conflicts on proposed installs. 
+    /// </summary>
+    /// <remarks>
+    /// All constructors start with currently installed modules, to remove <see cref="RelationshipResolver.RemoveModsFromInstalledList" />  
+    /// </remarks>
     public class RelationshipResolver
     {
         // A list of all the mods we're going to install.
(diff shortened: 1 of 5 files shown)


Core/Relationships/RelationshipResolver.cs
```diff
@@ -56,63 +57,126 @@ public object Clone()
...
-    // Currently only made to support the with_{} options.
+    // TODO: Add mechanism so that clients can add mods with relationshup other than UserAdded. 
+    // Currently only made to support the with_{} options. 
+
+
+    /// <summary>
+    /// A class used to resolve relationships between mods. Primarily used to satisfy missing dependencies and to check for conflicts on proposed installs. 
+    /// </summary>
+    /// <remarks>
+    /// All constructors start with currently installed modules, to remove <see cref="RelationshipResolver.RemoveModsFromInstalledList" />
```
pjf: You *rock*.

Core/Relationships/RelationshipResolver.cs
```diff
@@ -127,23 +191,24 @@ public class RelationshipResolver
...
                 Resolve(module, options);
             }
 
             var final_modules = new List<Module>(modlist.Values);
-            final_modules.AddRange(registry.InstalledModules.Select(x => x.Module));
+            
 
             if (!options.without_enforce_consistency)
             {
+                final_modules.AddRange(installed_modules);
```
pjf: Uh oh, I'm confused here. We're inside `AddModulesToInstall`, but we're now only adding those modules if we've been asked to enforce consistency.

This seems like the whole method would become a no-op in the case where we're not enforcing consistency, so this is either a bug, or I'm not understanding the code properly.

Core/Relationships/RelationshipResolver.cs
```diff
@@ -396,10 +457,10 @@ public List<CkanModule> ModList()
 
         public bool IsConsistant
         {
-            get { return conflicts.Any(); }
+            get { return !conflicts.Any(); }
```
pjf: Oh my!

I'm also very confused. The only place this seems to be used is in our test suite, which I would expect would have spotted this doing the reverse of what was intended.

Core/Relationships/RelationshipResolver.cs
```diff
@@ -56,63 +57,126 @@ public object Clone()
...
         // A list of all the mods we're going to install.
         private static readonly ILog log = LogManager.GetLogger(typeof (RelationshipResolver));
         private readonly Dictionary<string, CkanModule> modlist = new Dictionary<string, CkanModule>();
-
-        private readonly List<KeyValuePair<Module, Module>> conflicts =
+        private readonly List<CkanModule> user_requested_mods = new List<CkanModule>();
+        private readonly List<KeyValuePair<Module, Module>> conflicts = 
             new List<KeyValuePair<Module, Module>>();
+        private readonly Dictionary<Module, SelectionReason> reasons = 
+            new Dictionary<Module, SelectionReason>(new Module.IdentifierEqualilty());
```
pjf: Sweet! Today I learnt you can pass an equality comparison class to dictionaries. :)

Core/Relationships/RelationshipResolver.cs
RichardLake: The code could be clearer final_modules is only used in the SanityChecker.EnforceConsistency. Actually adding the modules to the installer is done by this point via calls to Resolve and to Add. 

Thinking about it may be clearer to have Resolve() return the modules to add instead of adding them itself.

Core/Relationships/RelationshipResolver.cs
RichardLake: It is used in a single test which this PR enables. This PR is pre repo merge so the GUI wasn't changed to make use of the new helper yet.

Core/Relationships/RelationshipResolver.cs
pjf: Aha! This makes much more sense. :)

PR #992, opened 2015-05-31
Enable checking against a schema.

RichardLake:
From https://github.com/KSP-CKAN/CKAN-core/pull/133


pjf:
@RichardLake : Any chance I can ask you for a rebase? I can do one myself, but our process is cleaner if we keep the same PR number. :)

ayan4m1:
Hey, apologies for resurrecting a zombie here, but I want to move this forward in any way that would be helpful. I can try doing the... extensive rebase here, or if the code has been merged in in another form then we can maybe close this out. Does the schema checking here refer to the CKAN client itself checking the schema of the metadata files against the documented JSON schema?

politas:
Pretty sure this is for the Netkan/Validators to check the .ckan file being created against the schema. I'd have to do some searches for other instances to be sure, but the only other time I can see that being important is the "Install from .ckan" command, which would be the other situation we want to validate a .ckan file. Obviously, the schema included would need to be updated to the latest, and yeah, there's a _lot_ of rebasing required.

Olympic1:
Rebased in #2027

Issue #972, opened 2015-05-28
Users should be able to choose between exporting mods as `recommends` or `depends`


pjf:
Right now all exported mods (eg: `installed-default.ckan`) are in a `depends` relationship. While it would be easy for a user to hand-edit the file, it would be even better if we provided *two* export functions, one as a `depends` relation, and one as a `recommends` relation.

The second is useful for when a user wants to grab the listed mods if they exist, but not have a hard failure if they don't exist. Examples would be having a "favourites" list soon after a KSP version bump. One can just import favourites to automatically install any mods available for the new release.

The "recommends" export should probably have fairly weak version info, either specifiying minimum versions rather than exact versions for the mods, or no version requirements at all (which I think would be best).

I'm not sure of the user-friendly wording for this. "Export favourites list" seems to fit well, especially because you can give your favourites list to someone else and they can choose to not install some mods, whereas they can't choose not to install mods where a hard-depends relationship exists.

(Suggestion from [Tambo_No5](https://www.reddit.com/r/KerbalSpaceProgram/comments/37hj03/ckan_weekly_28_may_2015/crnfd4d))

PR #993, opened 2015-05-31
Serialize current install

RichardLake:
From https://github.com/KSP-CKAN/CKAN-core/pull/139/files

Change depends to recommends.
Allows user customization of the install in the GUI
Do not output provided versions.
They are no real mods.

Closes KSP-CKAN/CKAN#912.
Relevant to KSP-CKAN/CKAN#972.

Currently it does not output a useful .ckan file due to provided modules.


pjf:
Working absolutely as intended! YOU ROCK! :D :D

PR #994, opened 2015-05-31
Make the Installed and Update checkbox columns sortable

RichardLake:
From @slikts

> This is a small usability improvement; sorting the rows helps in a similar way that filtering them does, but may actually be more convenient, since it doesn't hide the other rows.


RichardLake:
@slikts Managed to cause a error, Change the filter to type All and try to sort based on the install column. Looking at datagridview it does have a DataGridView.SortCompare event you can subscribe to in order to provide your own sorting.

PR #995, opened 2015-05-31
Fix KSP icon transparency; add more consistent icons for search and settings

RichardLake:
From @slikts

>Added transparency to the KSP icon and replaced the Search icon with a more consistent looking one; it's from the same icon set as the current Refresh icon.


RichardLake:
@slikts looks good to me. Merging.

PR #996, opened 2015-05-31
Typing nav

RichardLake:
From @slikts at https://github.com/KSP-CKAN/CKAN-GUI/pull/139
> This makes ModList behave more like typical data lists and makes it significantly easier to use because it's possible to go directly to a mod by typing in the first few letters of its name. The delay before it forgets the previous key presses is 1 second. The previous implementation wasn't very useful in cases where a number of mod names share the same first letter.

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -67,6 +67,9 @@ namespace CKAN
 
         private Timer filterTimer;
 
+        private DateTime lastSearchTime;
+        private string lastSearchKey;
+
         private IEnumerable<KeyValuePair<CkanModule, GUIModChangeType>> change_set;
         private Dictionary<Module, string> conflicts;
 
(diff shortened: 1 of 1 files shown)


GUI/Main.cs
```diff
@@ -495,42 +498,82 @@ private void OnFilterUpdateTimer(Object source, EventArgs e)
...
             {
                 var modname = ((GUIMod) row.Tag).ToCkanModule().name;
-                var key = e.KeyChar.ToString();
-                return modname.StartsWith(key, StringComparison.OrdinalIgnoreCase);
+                var row_match = modname.StartsWith(key, StringComparison.OrdinalIgnoreCase);
+                if (row_match && first_match == null) {
+                    // Remember the first match to allow cycling back to it if necessary
+                    first_match = row;
+                }
+                if (row == current_row || (current_match && row.Index < current_row.Index))
```
RichardLake: Why have this in? If I type better I would expect, after bet, to have Better Buoyancy selected and for it to remain on that row instead of cycling through the next couple which also starts with better.

slikts:
I've made more fixes and I'll make a new PR; this one should be discarded

Dazpoet:
closing by request of author, superceded by #1047

Issue #795, opened 2015-03-22
When an upgrade adds new dependencies, include them in the changeset for users to see


MFGookey:
I booted up CKAN for the first time in about a month and ran a refresh and update all install mods through the GUI yesterday.  One of the mods, Orbital Material Science, appeared to have a new package added.  When I attempted to download the changes, the JIT debugger would pop up with a message along the lines of "Cannot update NehemiahScienceCommon because it is not installed".

I was able to fix this by uninstalling then reinstalling Orbital Material Science via CKAN but preferred behavior is probably when the file you want to update doesn't exist, just put the new file into place.

Postremus:
I have the same problem using CKAN 1.6.5.
I' am trying to update Orbital Material Science 0.5.7 to 0.6.1
```
Informationen über das Aufrufen von JIT-Debuggen
anstelle dieses Dialogfelds finden Sie am Ende dieser Meldung.

************** Ausnahmetext **************
CKAN.ModuleNotFoundKraken: Can't upgrade NehemiahScienceCommon, it is not installed
   bei CKAN.ModuleInstaller.Upgrade(IEnumerable`1 modules, NetAsyncDownloader netAsyncDownloader)
   bei CKAN.Main.InstallMods(Object sender, DoWorkEventArgs e)
   bei System.ComponentModel.BackgroundWorker.WorkerThreadStart(Object argument)


************** Geladene Assemblys **************
mscorlib
    Assembly-Version: 4.0.0.0.
    Win32-Version: 4.0.30319.34014 built by: FX45W81RTMGDR.
    CodeBase: file:///C:/Windows/Microsoft.NET/Framework64/v4.0.30319/mscorlib.dll.
----------------------------------------
ckan
    Assembly-Version: 0.0.0.0.
    Win32-Version: 0.0.0.0.
    CodeBase: file:///C:/Users/panze_000/Desktop/ckan.exe.
----------------------------------------
System.Windows.Forms
    Assembly-Version: 4.0.0.0.
    Win32-Version: 4.0.30319.33440 built by: FX45W81RTMREL.
    CodeBase: file:///C:/Windows/Microsoft.Net/assembly/GAC_MSIL/System.Windows.Forms/v4.0_4.0.0.0__b77a5c561934e089/System.Windows.Forms.dll.
----------------------------------------
System.Drawing
    Assembly-Version: 4.0.0.0.
    Win32-Version: 4.0.30319.33440 built by: FX45W81RTMREL.
[...]

PR #1001, opened 2015-05-31
Allow conflicts if mod are to be uninstalled

RichardLake:
From https://github.com/KSP-CKAN/CKAN-GUI/pull/111

Based on https://github.com/KSP-CKAN/CKAN/pull/988

Should allow you to install mods which conflict with mods that will be uninstalled. Such as changing tacls configs.

### DIFF
--- a/Core/KSPManager.cs
+++ b/Core/KSPManager.cs
@@ -21,7 +21,7 @@ namespace CKAN
 
         private readonly SortedList<string, KSP> instances = new SortedList<string, KSP>();
 
-        internal string AutoStartInstance
+        public string AutoStartInstance
         {
             get { return Win32Registry.AutoStartInstance; }
             private set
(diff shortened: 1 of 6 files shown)


dbent:
We should check to see if this fixes #795 and #1124.

dbent:
@RichardLake, thoughts on the best way to apply this to the command line? #1124 still breaks because the `RelationshipResolver` checks for consistency and throws before the mods to remove are even calculated.

RichardLake:
The best way? 
A new constructor in RelationshipResolver which doesn't include the installed mods by default or one which takes a input of ones which will be uninstalled/upgraded. 
I've thrown together another way on https://github.com/RichardLake/CKAN/tree/CmdlineUpgrade_ModsNotInRepo but it is probably easier to make a new constructor.

PR #1002, opened 2015-05-31
Once again allow users to select dependencies when multiple provides exist

### LINKED ISSUE #976: CKAN GUI no longer gives choices of dependencies (TACLS)
Somewhere in the last few releases the GUI stopped giving the user a choice between mods to install.

### Steps to reproduce:

- Start with a fresh v0.90 install
- Download CKAN v1.6.18 or above.
- Tick the TACLS mod in the GUI.

### What should (and used to) happen:

- The GUI should ask if you want stock or reaismoverhaul configs, just like the cmdline does.

### What does happen:

- No changeset appears. The user can manually select a config, but they have to know that's required.

### Relevant tickets (feel free to add to):

- KSP-CKAN/CKAN-core#141

RichardLake:
Fixes #976.

See: https://github.com/KSP-CKAN/CKAN-GUI/pull/137 for some discussion.

### DIFF
--- a/Cmdline/CKAN-cmdline.csproj
+++ b/Cmdline/CKAN-cmdline.csproj
@@ -9,12 +9,15 @@
     <AssemblyName>CmdLine</AssemblyName>
     <StartupObject>CKAN.CmdLine.MainClass</StartupObject>
     <ApplicationIcon>..\GUI\assets\ckan.ico</ApplicationIcon>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkProfile />
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <Optimize>false</Optimize>
     <OutputPath>bin\Debug</OutputPath>
     <WarningLevel>4</WarningLevel>
     <DebugSymbols>true</DebugSymbols>
+    <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="CommandLine">
@@ -63,6 +66,7 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
+    <None Include="app.config" />
     <None Include="packages.config" />
   </ItemGroup>
 </Project>
--- a/Cmdline/app.config
+++ b/Cmdline/app.config
@@ -1,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5"/></startup></configuration>
--- a/Core/Relationships/RelationshipResolver.cs
+++ b/Core/Relationships/RelationshipResolver.cs
@@ -63,8 +63,8 @@ namespace CKAN
     // If we resolved in things breadth-first order, we're less likely to encounter surprises
     // where a nth-deep recommend blocks a top-level recommend.
 
-    // TODO: Add mechanism so that clients can add mods with relationshup other than UserAdded. 
-    // Currently only made to support the with_{} options. 
+    // TODO: Add mechanism so that clients can add mods with relationshup other than UserAdded.
+    // Currently only made to support the with_{} options.
     public class RelationshipResolver
     {
         // A list of all the mods we're going to install.
@@ -111,7 +111,7 @@ namespace CKAN
             {
                 log.DebugFormat("Preparing to resolve relationships for {0} {1}", module.identifier, module.version);
 
-                var module1 = module; //Silence a warning re. closures over foreach var. 
+                var module1 = module; //Silence a warning re. closures over foreach var.
                 foreach (CkanModule listed_mod in modlist.Values.Where(listed_mod => listed_mod.ConflictsWith(module1)))
                 {
                     if (options.procede_with_inconsistencies)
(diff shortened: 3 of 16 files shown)


Cmdline/CKAN-cmdline.csproj
```diff
@@ -9,12 +9,15 @@
     <AssemblyName>CmdLine</AssemblyName>
     <StartupObject>CKAN.CmdLine.MainClass</StartupObject>
     <ApplicationIcon>..\GUI\assets\ckan.ico</ApplicationIcon>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
```
pjf: I'm guessing we didn't have a target framework before? What would it default to without this? (My mind says that we were targeting dotnet40, but I've no idea if that's actually true.)

Cmdline/app.config
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
```
pjf: I'm not familiar with `app.config` files. How do they get used? I'm mostly concerned about having files that our build system may ignore.

Core/Relationships/RelationshipResolver.cs
```diff
@@ -333,6 +334,15 @@ private void Add(CkanModule module, Relationship reason)
             }
         }
 
+        private bool MightBeInstallable(CkanModule module)
```
pjf: - [ ] Some docs on this would be great.

It looks like we recursively chase all dependencies and make sure we can potentially install them all in some iteration. (Although we don't check against conflicts against currently installed mods, since `MightBeInstallable` vs `AbsolutelyInstallable`. :)

pjf:
Under Linux with mono 3.2.8 and 21d2b98, I get a hard lock whenever I select anything that requires a choice to be made. The GUI completely stops responding, the window can't be closed, and I have to kill it from the command-line. :(

pjf:
I can also reproduce the same lock-up under Mono 3.12.1 under Linux.

Cmdline/CKAN-cmdline.csproj
RichardLake: I think the VS default is v4.0 but I'm not 100%

pjf:
Testing 4b5d3e4:

- Selecting Astronomer’s Pack immediately gives me a choice of clouds. Hooray!
- Selecting an option (eg High) and then selecting another without unticking the first *does* change my selection, but causes the continue button to grey out.
- Unselecting an option and then re-selecting it causes the continue button to return to active behaviour.

This isn't a show-stopper (we're way better than we were before), so I'm continuing with review.

pjf:
Same commit, but more serious bug:

- Select Astronomer's Pack
- Select a cloud pack (I picked low)
- Note that we're back on the mod selection list, but there's no Changset tab, nor active "go to changes" button.

Selecting the cloud pack directly in the mod list does the right thing (gives me a change set with dependencies listed).

A more minor behavioural surprise (and also not a merge-blocker) is that if we untick the cloud pack we immediately get asked to choose a cloud pack. That's kind of cool (because it keeps consistency), but users may be surprised by it (eg: either expecting either for the dependant package to be unchosen, or for that to be an option that's presented).

Postremus:
Outerplanetsmod shows a similar behaviour.
Dependencys are not listed, but instead immediatly installed. 
You only see the dependencys in the install screen

Von: "Paul Fenwick" <notifications@github.com>
Gesendet: ‎03.‎06.‎2015 10:05
Betreff: Re: [CKAN] Once again allow users to select dependencies whenmultiple provides exist (#1002)

Same commit, but more serious bug:
Select Astronomer's Pack 
Select a cloud pack (I picked low) 
Note that we're back on the mod selection list, but there's no Changset tab, nor active "go to changes" button. 
Selecting the cloud pack directly in the mod list does the right thing (gives me a change set with dependencies listed).
A more minor behavioural surprise (and also not a merge-blocker) is that if we untick the cloud pack we immediately get asked to choose a cloud pack. That's kind of cool (because it keeps consistency), but users may be surprised by it (eg: either expecting either for the dependant package to be unchosen, or for that to be an option that's presented).

RichardLake:
> Selecting an option (eg High) and then selecting another without unticking the first does change my selection, but causes the continue button to grey out.

Fixed in 2cbcdf1
>. ... (eg: either expecting either for the dependant package to be unchosen, or for that to be an option that's presented).

Long term idea is to continue towards making RelationshipResolver work with multiple calls instead of the current one-shot everything in ctor method. Once this and adding mods to the provider with reasons are in it should be possible to use the reasonFor functionality to extract which mod is the dependant package. At which point it would be easy to change the GUI behaviour. 
> Note that we're back on the mod selection list, but there's no Changset tab, nor active "go to changes" button.

:worksOnMyMachine:. I'll break out the linux VM and check if it is a mono/.Net difference and investigate.

RichardLake:
>Note that we're back on the mod selection list, but there's no Changset tab, nor active "go to changes" button.

Functioning on mono 4.0.1 from a empty install.

RichardLake:
@pjf Can you replicate the missing changeset tab with a empty install?

pjf:
Apologies for my absence over the last two days; life threw me a real curveball.

> @pjf Can you replicate the missing changeset tab with a empty install?

I cannot. And after some tracking, I discovered that I have `PlanetShine-Config-Default` installed, which conflicts causes conflicts. Selecting just the cloud pack doesn't technically depend upon the parent Astronomer’s Pack, so no conflicts.

This is a fault in my testing. Hooray! :D

pjf:
Merging every zig for great justice. :)

PR #1003, opened 2015-05-31
More bad metadata and tests

mgsdk:
We do not have a complete test suite for both the tar.gz and zip formats. Also, I've added some bad KSP versions which failed prior to the merge. Currently they pass without issue, so I'm trying to find the commit which fixed this.

Spiritual successor to https://github.com/KSP-CKAN/CKAN-core/pull/130.


RichardLake:
Could you add the added file to Tests.csproj?

PR #1004, opened 2015-05-31
Main install

RichardLake:
Start to separate out MainInstall so it is not part of Main.

Currently is still tightly coupled as it just passes a Main instance into MainInstall but it is a improvement on what is currently there.

Currently most of the code in the GUI is in the class called Main which, via partial classes, has code spread over

*    Main.cs
*    MainChangset.cs
*    MainModInfo.cs
*    MainRepo.cs
*    MainWait.cs
*    MainInstall.cs
*    Parts of MainModList.cs.

The broad overview of this PR is to extract the code from Main which resides in MainInstall.cs into its own classes, MainInstall and MainInstallGUI. Currently MainInstall contains the code that only interacts with the GUI via the User interface, everything else is in MainInstall.

### DIFF
--- a/GUI/CKAN-GUI.csproj
+++ b/GUI/CKAN-GUI.csproj
@@ -144,9 +144,7 @@
     <Compile Include="MainDialogs.cs">
       <SubType>Form</SubType>
     </Compile>
-    <Compile Include="MainInstall.cs">
-      <SubType>Form</SubType>
-    </Compile>
+    <Compile Include="MainInstall.cs" />
     <Compile Include="MainModList.cs">
       <SubType>Form</SubType>
     </Compile>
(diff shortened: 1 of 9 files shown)


pjf:
I've merged this with current master ( https://github.com/pjf/CKAN/tree/1004_merged aka `v1.6.23-59-g6e663be`) and after installing a mod using the GUI it *doesn't* return me to the mod list, and instead leaves me on the 'Installing mods' screen.

Steps to reproduce for me:

1. Open the GUI
1. Install Better Buoyancy
1. Observe we don't get returned to the modlist on success.

pjf:
Gosh dang it, I really want to merge this (still), especially since we've had additional patches, but I know I'm going to struggle with the merge conflicts due to my unfamiliarity with the code. I'm going to work through the other PRs, and then hopefully grab one of our other fabulous contributors to give me a hand with this. :)

RichardLake:
Way out of date now - I'll recreate later.

PR #1006, opened 2015-05-31
AVC uses -1 as a wildcard since 1.1.5.0.

RichardLake:
Thanks to Tivec for reporting and Blu3wolf for diagnosing the issue.

### DIFF
--- a/Netkan/AVC/JsonAvcToKspVersion.cs
+++ b/Netkan/AVC/JsonAvcToKspVersion.cs
@@ -60,23 +60,22 @@ namespace CKAN.NetKAN
             //AVC uses -1 to indicate a wildcard.
             int integer;
             string version;
-            if (int.TryParse(patch, out integer))
+            if (int.TryParse(major, out integer) && integer == AVC_WILDCARD)
             {
-                version = integer == AVC_WILDCARD
-                    ? string.Join(".", major, minor)
-                    : string.Join(".", major, minor, patch);
+                version = null;
             }
-            else if (int.TryParse(minor, out integer))
+            else if (int.TryParse(minor, out integer) && integer == AVC_WILDCARD)
             {
-                version = integer == AVC_WILDCARD
-                    ? string.Join(".", major)
-                    : string.Join(".", major, minor);
+                version = major;
+            }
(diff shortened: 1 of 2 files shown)


Dazpoet:
Tried this version, got a fragment from @RichardLake on irc, with the only .netkan file I know off that acctually utilizes AVC wildcards, namely [Protractor](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/Protractor.netkan) and it works fine. Resulting file gets:

```
    "ksp_version_min": "1.0.0",
    "ksp_version_max": "1.0"
```
Also tried the fragment on other .netkans just to make sure nothing terrible happened and they included ActiveTextureManagement-x86-Aggressive (uses github asset match), InterstellarFuelSwitch-Core (normal kref to ks), the SCANSat contract pack (which has a vref without wildcards), QuickRevert (kref to ks and vref without wildcards) and it seems I'm unable to break it.

Dazpoet:
What happens if the vref points to a wildcard not in the Patch position? Say we point towards a vref which has
```
"KSP_VERSION_MAX":{
"MAJOR":1,
"MINOR":-1,
"PATCH":2
}
```
Although I agree something with a wildcard in a position not followed by another wildcard makes little sense but... you never know.

I'm unable to test the above case at this time and my code reading is to bad to figure out if this is already covered.

RichardLake:
I've updated the code to support wild-cards in positions other than patch. I choose to treat a wildcard as implying further version information is not relevant. Supporting things such as "-1,-1,3" is silly.

Dazpoet:
I won't argue with that and I also see tests which I've come to learn is a yay \o/.

Since it passes tests I assume all is fine, hard to try in the wild since afaik no mods have wildcards in other places than the patch position yet.

pjf:
User testing form @Dazpoet sounds great, the code looks good to me¹, and the handling of `-1` in different positions seems solid (using the most significant `-1` found, and ignoring everything thereafter). Merging with great thanks.

--

¹ We do have some repeated code, but we've got the *same* amount of repeated code as before the patch, so we're still better than we were before.

PR #1007, opened 2015-05-31
Add in new custom field for kerbalstuff mods.

RichardLake:
"x_screenshot" contains a url to a screenshot used as a icon on kerbalstuff.
A user on IRC requested a field in which mods can have a screenshot. A number of KS mods have such a screenshot accessible via the API so I thought this might encourage use of it.

### DIFF
--- a/Netkan/KS/KSMod.cs
+++ b/Netkan/KS/KSMod.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Text.RegularExpressions;
 using log4net;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 [assembly: InternalsVisibleTo("CKAN.Tests")]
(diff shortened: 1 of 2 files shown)


Postremus:
What if someone wants to add an screenshoot hosted on an external site?

Dazpoet:
I assume it'll just be a matter of hardcoding in a x_screenshot field into the .netkans and/or .ckans just like with any other fields which you want to overwrite with your own rather than the scraped one.

pjf:
> What if someone wants to add an screenshoot hosted on an external site?

@Postremus : Then they do that. We don't actually regulate `x_` fields at all; they're not defined by the spec, we make no promises that they'll work the way anyone expects, and the client ignores them. If you go looking you'll find all sorts of interesting `x_whatever` fields in the metadata for this reason. :)

pjf:
Tested on mods with and without screenshots; merging. :)

PR #1047, opened 2015-05-31
Improve typing nav usability in mod list

slikts:
Mimic the typical behavior of lists in other software; improve usability

 * Start typing the name of the mod in the mod list and it will get selected
 * Pressing the same key will cycle through items whose names start with that key
 * Previous key presses are forgotten in 1 second
 * Toggling the install checkbox by pressing space won't work unless the row is selected
 * Status bar "Not found" message; cleared when a mod is selected
 * Home/End keys go to top/bottom of the list respectively


slikts:
One more thing that would be nice is going to the top/bottom of the mod list with the Home/End keys, but I didn't see an obvious way to implement this; the listener used for searching the list doesn't seem to be fired when Home/End keys are pressed

PR #1049, opened 2015-06-01
Change toLongMin and toLongMax to return new copies.

RichardLake:
Mutating the existing copy will change its hashcode which is not permissible when it is used in a key or a dictionary.


pjf:
This looks fine to me, so merging. Of out interest, have we seen any bugs caused by this?

RichardLake:
@pjf Not as far as I know.

PR #1051, opened 2015-06-01
Add test_data boolean to GetCachedZip

### LINKED ISSUE #1026: The Contents tab freezes the GUI when selecting certain mods
<a href="https://github.com/slikts"></img></a> **Issue by [slikts](https://github.com/slikts)**
_Sunday May 31, 2015 at 05:40 GMT # Sample: Friday Sep 13, 2013 at 22:58 GMT_
_Originally opened as https://github.com/KSP-CKAN/CKAN-GUI/issues/141_

----

The CKAN GUI locks up when selecting certain mods with the Contents tab active; in particular it happens with the Astronomer's Pack mods. The GUI becomes unresponsive for about 5 seconds.

RichardLake:
Currently it would default to true. However the test is expensive - on the order of seconds as compared to retrieving the files names which is on the order of tens of miliseconds.
Given this I've set it to only run the tests before installing.

Closes #1026

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -13,7 +13,7 @@ namespace CKAN
 {
     public delegate void ModuleInstallerReportModInstalled(CkanModule module);
 
-    public struct InstallableFile 
+    public struct InstallableFile
     {
         public ZipEntry source;
         public string destination;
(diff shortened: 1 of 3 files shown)


pjf:
We're still testing the zip file, just not doing the low level test? This seems totally sensible; zip files have their index at the end, so if the file didn't finish downloading we know very quickly because the index is missing.

RichardLake:
I believe that the only thing we are skipping is CRC checks on every file.

pjf:
Absolutely working as intended. Merging with great thanks.

PR #1052, opened 2015-06-01
More robust cURL CA bundle file resolution

dbent:
Currently the cURL bundle file is only used if it exists in the current working directory, which isn't very convenient. This change makes it so that NetKAN will search a list of directories for the file and use the first one found. Specifically it will use in preference:

- Working Directory
- Executable Directory (wherever `netkan.exe` exists)
- `%LOCALAPPDATA%/CKAN`, which on a typical system will correspond to `C:\Users\User\AppData\Local\CKAN`
- `%APPDATA%/CKAN`, which on a typical system will correspond to `C:\Users\User\AppData\Roaming\CKAN`
- `%PROGRAMDATA%/CKAN`, which on a typical system will correspond to `C:\ProgramData\CKAN`

Basically, the more "local" the file is, the higher priority it has. I'm not entirely keen on including the working directory in the search path, but didn't want to change existing behavior.

### DIFF
--- a/Netkan/Web.cs
+++ b/Netkan/Web.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.IO;
+using System.Linq;
+using System.Reflection;
 using System.Text;
 using CurlSharp;
 using log4net;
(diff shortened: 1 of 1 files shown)


pjf:
Haven't had a chance to *test* this, but saying that I definitely believe this to be a good idea. :)

pjf:
This only seems to be doing the checks if we're on Windows, so I imagine clients on non-Windows systems won't find he cURL bundle even if it's in their working directory.

Do the special folders have meaning on non-Windows systems? If they do, it seems pretty reasonable to make the checks non-platform specific.

RichardLake:
From https://ibboard.co.uk/Programming/mono-special-folders.html it would check the following on linux. 
~/.config
~/.local/share
/usr/share

pjf:
Added 2d8ef35 which:

- Runs the cert resolution code on all platforms (not just windows).
- Logs which cert bundle we're using.

It seems to be picking up current working directory and `~/.config/CKAN/` for me. I haven't tested the rest, but this looks good to me, so merging with thanks once travis gives me a green light as well. :)

PR #1054, opened 2015-06-02
Implement programmatic sorting for all columns

slikts:
All columns can be sorted. It also sorts by the mod name column in addition to the primary sort.


GUI/MainModList.cs
```diff
@@ -67,10 +103,7 @@ private void _UpdateModsList(bool repo_updated)
                 }
             }
             mainModList.Modules = new ReadOnlyCollection<GUIMod>(gui_mods.ToList());
-            var rows = mainModList.ConstructModList(mainModList.Modules);
-            ModList.Rows.Clear();
-            ModList.Rows.AddRange(rows.ToArray());
-            ModList.Sort(ModList.Columns[2], ListSortDirection.Ascending);
```
slikts: This part was redundant.

slikts:
A bonus improvement is that changing the filtering settings now persists the selected row, whereas before it would lose the selection and just select the first row.

RichardLake:
@slikts can you either rebase of or merge in master?

PR #1057, opened 2015-06-02
Focuses the mod list when it's updated

### LINKED ISSUE #1056: GUI: Start focus on the big table of mods, not the tab control
As of 3c4f781, launching the CKAN starts with focus on the "Manage mods" tab control itself, however I believe most users will find it useful to start with focus on the big table of mods. This means they can use arrow keys to navigate it, and typing navigation from #1047 to quickly go to mods.

It'd be great if we could start with focus on the mod table itelf. :)

slikts:
Fix #1056


slikts:
Works in mono as well as on Windows

PR #1059, opened 2015-06-02
Temp fix for focus. If it is in Updated Filters it causes issues with…

RichardLake:
… typing.


pjf:
Tested under mono 3.2.8 + Linux. Merging with great thanks!

PR #1077, opened 2015-06-06
Fix edge case in version numbering.

### LINKED ISSUE #1076: The 'compare' command thinks 1.5.01 is both smaller and larger than 1.5.1
Problem initially located in https://github.com/KSP-CKAN/NetKAN/pull/1516

When comparing 2 strings on the form `x.0y` and `x.y` the result is always `-1` when run through the `ckan compare` commmand.

RichardLake:
By the spec 1.01 should equal 1.1 while the code had them non-equal with the order depending on which was called first. Closes #1076.


pjf:
Merging with great thanks!

PR #1079, opened 2015-06-06
Have travis upload "nightly" artefacts.

pjf:
- Fixes a bug where we'd try to upload artefacts multiple times (now
  only the 3.10.0 artefacts are uploaded).
- Uploads `netkan.exe` and `ckan.exe` from any code that's been merged
  into master.
- Kinda lazy because it's using pjf's bucket rather than KSP-CKAN's
  bucket.
- Kinda lazy because it doesn't upload versioned artefacts.
- https://ckan-travis.s3.amazonaws.com/ckan.exe and
  https://ckan-travis.s3.amazonaws.com//netkan.exe will always be the
  latest artefacts from master.
- Example successful run can be found at https://travis-ci.org/KSP-CKAN/CKAN/jobs/65685362

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,36 +23,41 @@ script:
     - nunit-console --exclude=FlakyNetwork build/Tests/bin/Debug/CKAN.Tests.dll
     - prove  # Run all the tests in t/
 
+before_deploy:
+    # S3 doesn't support file for some reason, so we have to put things in
+    # a directory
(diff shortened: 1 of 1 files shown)


pjf:
I'm merging this in my quest to make @Dazpoet happy. Since there are no user-visible changes here we're unlikely to break the internet by doing so. :)

PR #1081, opened 2015-06-07
Target platform v4 (aka .NET 4.5) when building

### LINKED ISSUE #1074: Fatal .Net 4.0 issue on v1.6.21
@pjf 
Due to our use of ILRepack v1.6.21 has a fatal issue on windows machines without .net 4.5 installed. This is due being unable to locate the correct assemblies as they moved between 4.0 and 4.5. The [TypeForwardedTo] system to migrate this doesn't work with our build system. 
* Stop using ILRepack or equivalents. 
* Rewrite away from await/async to the older .net 4 system. 
* Build on a .Net 4.0 windows system
* Have people upgrade to .net 4.5. 

Of these I prefer the last. Adding the "--targetplatform:v4" option to ILRepack generates a exe which will popup a message on start on .net 4.0 which states the need for a upgrade. As it is quite late, and due to lack of testing on mono 3.2.8, I decided against making a new release. A note has been added to the issue and it has been flagged as a pre-release.

pjf:
Closes #1074.

I don't have a good way to test this work as intended, but can confirm it doesn't cause any issues under mono.

@RichardLake, did you say you had a test machine?

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -2186,6 +2186,7 @@ my @cmd = (
     "mono",
     $REPACK,
     "--out:ckan.exe",
+    "--targetplatform=v4",
     "--lib:build/Cmdline/bin/$TARGET",
     "build/Cmdline/bin/$TARGET/CmdLine.exe",
     glob("build/Cmdline/bin/$TARGET/*.dll"),
--- a/bin/build.lean
+++ b/bin/build.lean
@@ -73,6 +73,7 @@ my @cmd = (
     "mono",
     $REPACK,
     "--out:ckan.exe",
+    "--targetplatform=v4",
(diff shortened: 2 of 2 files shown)


RichardLake:
Yes but I will not be home until late Monday.

RichardLake:
Looking at the commit I did test the master at the time with the same
change with build.lean

Postremus:
hey,

tested the build that pjf asked to be tested on twitter.

first tested it on net 4.5, starts normal
net 3.0: asks for net 4.0



net 4.0: starts normal. I have currently no ksp installed on that vm, so no further testing possible from me

Issue #228, opened 2014-11-01
Add `include_only` and `include_only_regexp`


pjf:
From 17405ea:

The CKAN filters (does not install) any file which _matches_ the regexp. The negative look-arounds means we can have a regexp which matches everything _except_ for a given file or path. Yes, that's needlessly complex.

To be honest, in addition to the filter fields, we should probably have their inverse as well, which I propose as `include_only` and `include_only_regexp`. It would be an error to have both filters and include_onlys in the same stanza.

That means we can have `"filter" : "config.cfg"` in the main document, and `"install_only" : "config.cfg"` in the configuration documents, which makes things much easier and more obviously correct.

PR #1089, opened 2015-06-09
find_regexp Directive

dbent:
*Port of KSP-CKAN/CKAN-core#131 from before the great repo merge.*

This extends the `install` stanza to add support for a `find_regexp` directive which is analogous to a `find` directive.

This was inspired by KSP-CKAN/NetKAN#1285 and KSP-CKAN/NetKAN#1288, since the `KWRocketry` distribution contains a top level directory that changes with each version *and* several `KWRocketry` directories throughout the distribution. This makes the behavior using `find` non-obvious and potentially brittle. Hence why I used an exact `file` but which would break with each new release. A `find_regexp` rectifies the situation by making the directory selection as exact as `file` but also as robust to change as `find`.

Take [`KWRocketry-2.7.ckan`](https://github.com/KSP-CKAN/CKAN-meta/blob/2c4f49d965aafe48f1316b3186751dc49d972a99/KWRocketry/KWRocketry-2.7.ckan) and [`KWRocketry-LegacyFairings-2.7.ckan`](https://github.com/KSP-CKAN/CKAN-meta/blob/d66bccb50e133896990983629ff9c3b04edaa846/KWRocketry-LegacyFairings/KWRocketry-LegacyFairings-2.7.ckan) for example. With this change.

```json
{
    "install": [
        {
            "file": "KW Release Package v2.7 (Open this, don't extract it)/GameData/KWRocketry",
            "install_to": "GameData"
        },
        {
            "file": "KW Release Package v2.7 (Open this, don't extract it)/GameData/ModuleAnimateEmissive",
            "install_to": "GameData"
        }
    ]
}
```

becomes:

```json
{
    "install": [
        {
            "find_regexp": "^(.+?/)?GameData/KWRocketry",
            "install_to": "GameData"
        },
        {
            "find_regexp": "^(.+?/)?GameData/ModuleAnimateEmissive",
            "install_to": "GameData"
        }
    ]
}
```

and

```json
{
    "install": [
        {
            "file": "KW Release Package v2.7 (Open this, don't extract it)/Extras/OldFARFairings/GameData/KWRocketry",
            "install_to": "GameData"
        }
    ]
}
```

becomes

```json
{
    "install": [
        {
            "find_regexp": "^(.+?/)?Extras/OldFARFairings/GameData/KWRocketry",
            "install_to": "GameData"
        }
    ]
}
```

Issues linked to from the previous PR:

- KSP-CKAN/CKAN-meta#528
- KSP-CKAN/CKAN#228
- KSP-CKAN/NetKAN#1447


dbent:
@pjf This should be good to go.

pjf:
Looking at this now. The code looks good, but needs some tests, so I'm adding some of those now. :)

pjf:
Tests added. Although I'm a bit sleepy, so they're not the prettiest.

pjf:
All my adjustments are tests and docs, so I'm happy with this. I've bumped the spec to v1.10, so the next CKAN release will be v1.10.0. Thank you very much!

Issue #1095, opened 2015-06-10
Ckan logo is missing in "Select KSP install" dialog


Postremus:
Instead the winforms default is shown.

![](http://puu.sh/ijIRz/26fe393e6c.png)

PR #1098, opened 2015-06-10
Added the ckan icon to select instance and update dialog

Postremus:
see #1095 
Saw that the update dialog also shows the winforms default and changed it to the ckan icon, too.


Postremus:
closed in favour of #1110

PR #1099, opened 2015-06-11
Friendly Mod Export 

### LINKED ISSUE #1094: Exporting all installed mods as a human-readable list from within the gui
Seen this request a few times now so though I'd put it here for tracking.

From what I gather users want a way to export a complete list of mods they can post in e.g. the forums (so not a metapackage).

I normally generate something like this with the command `ckan list --porcelain > list.txt` but that one also has the CKAN specific symbols at the start to show if it's upgradeable, unknown, AD etc so it might not be ideal.

If the reason for wanting these lists are showing what mods are needed for something or for helping in tracking down bugs it would seem to me that the versions of mods might also be pertinent to include. Not sure all users agree though.

dbent:
Fixes #1094

Allow CKAN to export the mod list in multiple formats.

- Plain Text
- Markdown
- BBCode
- CSV
- TSV

This is accomplished in the GUI by expanding the "Save as type" option. In the command line an extra option is added to the `list` command, `--export` that will output the list in the specified format (`text`, `markdown`, `bbcode`, `csv`, `tsv`) to stdout. Users can then redirect to a file as appropriate.

The exact format of the "pretty" formats is just what I came up with, suggestions welcome.

Also, I noticed that the `list` command has a `porcelain` option that's used in the [exact opposite sense that git uses that word](http://stackoverflow.com/questions/6976473/what-does-the-term-porcelain-mean-in-git). We should probably change that?

### DIFF
--- a/Cmdline/Action/List.cs
+++ b/Cmdline/Action/List.cs
@@ -117,7 +117,7 @@ namespace CKAN.CmdLine
             return Exit.OK;
         }
 
-        private ExportFileType? GetExportFileType(string export)
+        private static ExportFileType? GetExportFileType(string export)
         {
             export = export.ToLowerInvariant();
 
--- a/Cmdline/Action/List.cs
+++ b/Cmdline/Action/List.cs
@@ -105,29 +105,8 @@ namespace CKAN.CmdLine
             else
             {
                 var stream = Console.OpenStandardOutput();
-
-                switch (exportFileType.Value)
-                {
-                    case ExportFileType.PlainText:
-                        new PlainTextExporter().Export(registry, stream);
-                        break;
-                    case ExportFileType.Markdown:
-                        new MarkdownExporter().Export(registry, stream);
-                        break;
-                    case ExportFileType.BbCode:
-                        new BbCodeExporter().Export(registry, stream);
-                        break;
-                    case ExportFileType.Csv:
-                        new DelimeterSeperatedValueExporter(DelimeterSeperatedValueExporter.Delimter.Comma)
-                            .Export(registry, stream);
-                        break;
-                    case ExportFileType.Tsv:
-                        new DelimeterSeperatedValueExporter(DelimeterSeperatedValueExporter.Delimter.Tab)
-                            .Export(registry, stream);
-                        break;
-                    default:
-                        throw new ArgumentOutOfRangeException();
-                }
+                new Exporter(exportFileType.Value).Export(registry, stream);
+                stream.Flush();
             }
 
             if (!(options.porcelain) && exportFileType == null)
--- a/Cmdline/Action/List.cs
+++ b/Cmdline/Action/List.cs
@@ -1,7 +1,10 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
+using CKAN.Exporters;
+using CKAN.Types;
 using log4net;
 
 namespace CKAN.CmdLine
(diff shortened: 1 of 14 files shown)


dbent:
This [gist](https://gist.github.com/dbent/9bcbb71cd50ca3b186ec) contains examples of all the formats.

Dazpoet:
I think the examples are awesome and really like them. I think that users might want a plain-text option which doesn't include our identifiers though but only `<mod name> <version>`. However I guess an argument can easily be made that supplying c/tsv makes making one of those on your own an extremely easy task.

Core/Exporters/IExporter.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.IO;
+
+namespace CKAN.Exporters
+{
+    /// <summary>
+    /// Represents an object that can
```
pjf: Are you sure that? ;)

dbent: ... do whatever it puts its mind to because it's beautiful and special.

I'll fix that.

Cmdline/Action/List.cs
```diff
@@ -31,66 +47,117 @@ public int RunCommand(CKAN.KSP ksp, object raw_options)
...
 
-                        if (latest == null)
-                        {
-                            // Not compatible!
-                            bullet = "X";
-                        }
-                        else if (latest.version.IsEqualTo(current_version))
+                    if (current_version is ProvidesVersion)
+                    {
+                        // Skip virtuals for now.
```
pjf: Oh goodness, did we use to display these? Nice catch.

pjf:
I haven't had a chance to test this, but:

- Adding a `CKAN.Exporters` namespace seems solid to me.
- All the code looks good.
- This will make lots of people happy.

In other words, if you're reading this and able to test this is working as intended, I'm totally happy for you to mash the 'merge' button, even if you're not a C# person. :)

pjf:
Also as discussed on IRC, git's use of porcelain is confusing. We're using it as "more machine readable", which is one of the ways in which git uses it. :)

Cmdline/Action/List.cs
dbent: That's just a whitespace change, add `?w=1` to a compare URL to force GitHub to [ignore whitespace in the diff](https://github.com/KSP-CKAN/CKAN/pull/1099/files?w=1).

dbent:
> I'm totally happy for you to mash the 'merge' button, even if you're not a C# person. :)

I'm *totes* a C# person, it's been my professional life for the last six years.

I have one more change I want to make before I mash the merge button: I don't like the duplicate `switch` statement for both the command line and GUI, so I'm going to make a single `IExporter` that takes a `ExportFileType` and performs the `switch` internally. Then both the GUI and command line will just need to know how to determine the `ExportFileType` and instantiate the meta-exporter.

dbent:
A few updates in the latest batch of changes:

  - The `IExporter` interface has been generalized from using a `TextWriter` to using a generic `Stream`. Now the same interface can be used for writing binary formats ([BSON](http://bsonspec.org/) piped over the network anyone?)
  - I implemented a meta `Exporter` class, which consolidates all the switching logic to choose the right `IExporter`. Now the GUI and command line simply instantiate an `Exporter` and call its `Export` method. Yay, for code reuse.
  - I modified the Markdown and BBCode formats to use bold instead of italics on the mod name, I personally think this looks nicer and is more legible.

dbent:
#rebuild

dbent:
I guess the `#rebuild` command doesn't work for Travis? Apparently closing and reopening should work...

dbent:
Interesting, closing and reopening the PR triggered a [new build](https://travis-ci.org/KSP-CKAN/CKAN/builds/66549612), but did not update the status of the PR, which is still linked to the old PR.

techman83:
@dbent you can trigger a new build by adding commits, you just have to wait a little while. You can also go to the Travis page and re-run it as is, there is an action button in the top right of the box.

dbent:
Merging this since the changes since the review are minor.

Issue #1097, opened 2015-06-10
TooManyProvides generated even when an exact mod later in a depends list satsifies it


Postremus:
Hey,

just wanted to get my mods installed on another machine.  
Used the new export feature for this.  
Using on both machines ckan 1.6.22 on windows 8.1, ksp 1.0.2

mod list: http://puu.sh/ijJz3/2d985dcfba.ckan

Exception:
```
Informationen über das Aufrufen von JIT-Debuggen
anstelle dieses Dialogfelds finden Sie am Ende dieser Meldung.

************** Ausnahmetext **************
CKAN.TooManyModsProvideKraken: Too many mods provide AstronomersPack-Clouds:
AstronomersPack-Clouds-High Interstellar.V2
* AstronomersPack-Clouds-Low Interstellar.V2
* AstronomersPack-Clouds-Medium Interstellar.V2
   bei CKAN.RelationshipResolver.ResolveStanza(IEnumerable`1 stanza, Relationship reason, RelationshipResolverOptions options, Boolean soft_resolve)
   bei CKAN.RelationshipResolver.Resolve(CkanModule module, RelationshipResolverOptions options)
   bei CKAN.RelationshipResolver.ResolveStanza(IEnumerable`1 stanza, Relationship reason, RelationshipResolverOptions options, Boolean soft_resolve)
   bei CKAN.RelationshipResolver.Resolve(CkanModule module, RelationshipResolverOptions options)
   bei CKAN.RelationshipResolver..ctor(ICollection`1 modules, RelationshipResolverOptions options, Registry registry, KSPVersion kspversion)
   bei CKAN.Main.InstallList(HashSet`1 toInstall, RelationshipResolverOptions options, NetAsyncDownloader downloader)
   bei CKAN.Main.InstallMods(Object sender, DoWorkEventArgs e)
   bei System.ComponentModel.BackgroundWorker.WorkerThreadStart(Object argument)


************** Geladene Assemblys **************
mscorlib
    Assembly-Version: 4.0.0.0.
    Win32-Version: 4.0.30319.34014 built by: FX45W81RTMGDR.
    CodeBase: file:///C:/Windows/Microsoft.NET/Framework64/v4.0.30319/mscorlib.dll.
----------------------------------------
ckan
    Assembly-Version: 0.0.0.0.
    Win32-Version: 0.0.0.0.
    CodeBase: file:///C:/Users/panze_000/Desktop/ckan.exe.
----------------------------------------
[...]

PR #1108, opened 2015-06-12
Fix TMP error when installing metapackages.

RichardLake:
Current fix is a hack but will fix the common case of  installing from exported modlists.

Helps with #1097 but is not a proper fix. A TMP when the metapackage does not contain a mod which provides the dependency will still cause a exception to be thrown.


Postremus:
Tested it by merging it locally into master, broke cancelling of installation / downloads.

pjf:
@Postremus : You rock. Not merging this for now, since users love to cancel things. :)

Dazpoet:
Antiquated by #1337 so I'm closing this.

Issue #848, opened 2015-05-08
CKAN GUI crashes on cancelled download


obi1kenobi:
Expected result:
- The CKAN GUI should return to the "Manage mods" tab.

Actual result:
- CKAN crashes.

Setup: Win7 x64, .NET 4.0, CKAN v1.6.13

Repro:
1. select any mod (preferably a larger one), click "Go to changes" and start installing it
2. click the "Cancel" button as soon as it appears, and before the mod has finished downloading
3. observe CKAN pop up an error window, freeze and then crash entirely

Exception text:
```
CKAN.CancelledActionKraken: Download cancelled by user
   at CKAN.NetAsyncDownloader.DownloadModules(NetFileCache cache, IEnumerable`1 modules)
   at CKAN.ModuleInstaller.Upgrade(IEnumerable`1 modules, NetAsyncDownloader netAsyncDownloader)
   at CKAN.Main.InstallMods(Object sender, DoWorkEventArgs e)
   at System.ComponentModel.BackgroundWorker.WorkerThreadStart(Object argument)
```

I expect this should be rather simple to fix, as the CKAN GUI already recognizes that the download was interrupted, and simply fails to catch the background worker's exception.

obi1kenobi:
What I expect is a related error happens when attempting to download mods with no Internet connection.

Expected result:
- There is no Internet connection, so the GUI should display an error and then return to the "Manage mods" tab.

Actual result:
- CKAN crashes.

Repro:
1. disable your Internet connection
2. select and attempt to install any mod
3. observe CKAN pop up an error window and then crash.

Exception text:
Uh oh, the following things went wrong when downloading...

```
System.Net.WebException: The remote name could not be resolved: 'github.com'
   at System.Net.HttpWebRequest.EndGetResponse(IAsyncResult asyncResult)
   at System.Net.WebClient.GetWebResponse(WebRequest request, IAsyncResult result)
   at System.Net.WebClient.DownloadBitsResponseCallback(IAsyncResult result)
System.Net.WebException: The remote name could not be resolved: 'github.com'
   at System.Net.HttpWebRequest.EndGetResponse(IAsyncResult asyncResult)
   at System.Net.WebClient.GetWebResponse(WebRequest request, IAsyncResult result)
   at System.Net.WebClient.DownloadBitsResponseCallback(IAsyncResult result)
System.Net.WebException: The remote name could not be resolved: 'kerbalstuff.com'
[...]

PR #1112, opened 2015-06-13
Fix canceling mod downloads. 

### LINKED ISSUE #1023: Canceling a mod install leaves the GUI in an inconsistent state
<a href="https://github.com/savagerose"></img></a> **Issue by [savagerose](https://github.com/savagerose)**
_Monday May 18, 2015 at 23:22 GMT # Sample: Friday Sep 13, 2013 at 22:58 GMT_
_Originally opened as https://github.com/KSP-CKAN/CKAN-GUI/issues/125_

----

1) Select a mod to be installed
2) Click apply on the change tab 
3) Click cancel on the installation tab

The menu bar doesn't reenable as expected.

Also, this might be a separate bug really, but the change list tab doesn't hide even though nothing is check on it, which is what it does when the user clears the list?

It seems like the CancelledActionKraken exception interrupts the PostInstallMods handler. Not sure if that's intended. It seems like that's where it would make sense to do the cleanup when the user cancels, but the exception interrupts it when it tries to get e.Result.

### LINKED ISSUE #1088: This BackgroundWorker is currently busy and cannot run multiple tasks concurrently.
#1023 made me do some testing on my own and I seem to have found a way of reproduction that always leads to the above error, although I will admit it's probably not the most common usecase and is probably resolved by fixing the original issue.

Either way the following have worked for me a few times now to reproduce the error:

1) Have a fresh KSP install (I've tried 0.25, 0.90) and CKAN 1.6.22-0-g007717d, testing done on Win 7
2) Choose to install a big mod (I've used B9, KW or Realism Overhaul) because you need some time after canellation
3) Go to the Changeset tab and go further until the installation tab shows up
4) Click the cancel button
5) Get thrown back to the main tab where Changeset and Installation tabs are now hanging around but cannot be interacted with
6) Choose a mod which has a too_many_provides (RSS or Astronomers was used in my testing)
7) Get the window asking for which TMP you want and pick one
8) You can now add any mods you want to the changeset and when clicking apply you'll get this:

```
System.InvalidOperationException: This BackgroundWorker is currently busy and cannot run multiple tasks concurrently.
   at System.ComponentModel.BackgroundWorker.RunWorkerAsync(Object argument)
   at CKAN.Main.ConfirmChangesButton_Click(Object sender, EventArgs e)
```

### LINKED ISSUE #1102: Refresh dialog shows status of cancelled installation
The refresh dialog is capable of showing the state of the last cancelled installation. Even the mb count of downloads is ticking down.
The problem seems to be in both cases that the installation dialog is not cleared after cancelling.

Steps to reporduce:
- Choice multiple mods. I used around 10.
- "Go to changeset" and then apply them.
- Cancel the installation before the download of any mod begins, and click on refresh to get this:

- Cancel the Installation after the download of any mod begins, and click on refresh to get this:

Tested on windows 8.1, ckan 1.6.22 with ksp 1.0.2

RichardLake:
4ba407e is the actual fix - the cancel was being called on a separate instance of the downloader. 

Closes #1088
Closes #1102 
Closes #1023 
Revisit #848

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -13,7 +13,7 @@ namespace CKAN
 {
     public delegate void ModuleInstallerReportModInstalled(CkanModule module);
 
-    public struct InstallableFile 
+    public struct InstallableFile
     {
         public ZipEntry source;
         public string destination;
@@ -34,7 +34,7 @@ namespace CKAN
         private KSP ksp;
 
         public ModuleInstallerReportModInstalled onReportModInstalled = null;
-          
+
         // Our own cache is that of the KSP instance we're using.
         public NetFileCache Cache
         {
@@ -80,7 +80,7 @@ namespace CKAN
         /// </summary>
         public string Download(Uri url, string filename)
         {
-            User.RaiseProgress(String.Format("Downloading \"{0}\"", url), 0);            
+            User.RaiseProgress(String.Format("Downloading \"{0}\"", url), 0);
             return Download(url, filename, Cache);
         }
 
@@ -144,7 +144,7 @@ namespace CKAN
             return full_path;
         }
 
-        
+
 
         public void InstallList(
             List<string> modules,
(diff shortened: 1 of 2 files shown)


pjf:
Works for me, code looks good, merging with great thanks!

PR #1127, opened 2015-06-19
Catch XML parsing error correctly

gjulianm:
When loading a corrupt configuration file, the exception thrown is not caught as (at least in Windows) what gets thrown is an InvalidOperationException, not a XmlException. I've changed it and also added some additional information so the user can try and repair the file, as I think it can be useful.

Greetings and thanks for your hard work with CKAN!

### DIFF
--- a/GUI/Configuration.cs
+++ b/GUI/Configuration.cs
@@ -73,9 +73,14 @@ namespace CKAN
                 {
                     configuration = (Configuration) serializer.Deserialize(stream);
                 }
-                catch (System.InvalidOperationException)
+                catch (System.InvalidOperationException e)
                 {
-                    string message = string.Format("Error trying to parse \"{0}\". Try to move it out of the folder and restart CKAN.", path);
+                    string additionalErrorData = "";
+
+                    if (e.InnerException != null)
+                        additionalErrorData = ": " + e.InnerException.Message;
+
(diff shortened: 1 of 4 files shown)


pjf:
Hey there @gjulianm! Welcome aboard! :)

I've just tested, and on my system (Linux + mono 3.2.8) it's definitely an `XmlException`, so I don't suppose I can ask you to tweak this PR to catch both?

Many thanks,

~ Paul

gjulianm:
I've added it, that inside "if(e is XXX)" is not pretty but it should do the job. Tell me if it needs more changes.

Greetings

pjf:
Thank you! Works for me under Linux/mono, let's see if I can grab a Windows tester on IRC. :)

Dazpoet:
Makes life easier under Windows, gives a general CKAN.Kraken error with a brief explanation of how I can try to resolve the issue myself.

pjf:
In case I come back here wondering what the `<Service/>` section of the `.csproj` file is doing:

http://stackoverflow.com/questions/18614342/what-is-service-include-in-a-csproj-file-for

pjf:
Merged with many thanks! (And a small tweak to the tests so they remove their tempdirs at the end) You should see this in the 1.8.0 release tonight! :)

gjulianm:
Great, I'm glad it helped :)

PR #1128, opened 2015-06-20
gitignore: Don't track .exe.config files

pjf:
`ckan.exe.config` and related files are auto-generated, so we should
ignore them when it comes to version control.


pjf:
Using my git wizard hat to merge this, as it's non user-visible. :)

PR #1131, opened 2015-06-21
Background worker error

RichardLake:
Should prevent TargetInvocationException from being thrown by the background workers which will aid user support. Also moved the update repo handing out of the DoWork and into the RunWorkerCompleted due to style concerns. This is more a matter of style

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -304,9 +304,10 @@ namespace CKAN
 
             UpdateModsList();
 
+            var old_YesNoDialog = m_User.displayYesNo;
             m_User.displayYesNo = YesNoDialog;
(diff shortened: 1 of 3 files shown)


pjf:
In what cases were a TargetInvocationException being thrown? (Ie: How should I be testing this?)

RichardLake:
Any case with an uncaught exception in the BackgroundWorker.DoWork should trigger it. The ones I knew about have been fixed in #1150 so this PR is a precaution in case the core starts throwing new exceptions.

pjf:
A merge-conflict resolved copy of this is in #1290. Merging that will also close this PR.

pjf:
Closing this as it's a subset of #1290.

PR #1142, opened 2015-06-21
Add command line option to launch debugger

dbent:
Adds support for detecting a `--debugger` command line argument and using `System.Diagnostics.Debugger.Launch()` to launch a debugger. On Windows this opens a dialog prompt to either start or use an existing Visual Studio instance to attach to the process for debugging. Not sure what the exact behavior on Linux is, but from some quick googling it seems it's at least supported to some degree.

This flag is checked as early as possible to ensure maximal utility, it's even done before full command line argument parsing so you could even debug that.

### DIFF
--- a/Cmdline/Main.cs
+++ b/Cmdline/Main.cs
@@ -5,6 +5,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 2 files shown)


pjf:
On Linux it gives me

pjf:
On Linux it gives me a `NotImplementedException`, which is reasonable enough. :)

PR #1150, opened 2015-06-22
Stop existing files from stalling the GUI on install

### LINKED ISSUE #1143: CKAN.FileExistsKraken stalls the GUI when upgrading mods.
Steps to reproduce:

- Install an older version of any mod (for me, RealPlume v8.2.0 on 0.90)
- Create a file that a newer version will try to write to (for me, `RealismOverhaul/SmokeScreen_EffectsCfgs/aies.cfg`)
- Try to upgrade the mod in the GUI (For me, moving to RealPlume v8.4.1)

On the command-line, the exception is shown, but the GUI just stalls forever.

Present since at least v1.6.20 (currently we're on v1.8.1), which is as far back as I can bisect easily.

(Also, I thought we had another bug report of this, but I can't find it.)

RichardLake:
Closes #1143 .

### DIFF
--- a/GUI/MainInstall.cs
+++ b/GUI/MainInstall.cs
@@ -202,13 +202,16 @@ namespace CKAN
                 installCanceled = true;
             };
 
-
+            //Set the result to false/failed in case we return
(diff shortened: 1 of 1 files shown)


pjf:
Code looks good, and this works for me. Merging with great thanks!

PR #1153, opened 2015-06-22
Cleanup

RichardLake:
Removes all build warnings on Travis.

### DIFF
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -27,7 +27,7 @@ namespace CKAN.CmdLine
                 // Oooh! We're installing from a CKAN file.
                 foreach (string ckan_file in options.ckan_files)
                 {
-                    Uri ckan_uri = null;
+                    Uri ckan_uri;
 
                     // Check if the argument if a wellformatted Uri.
                     if (!Uri.IsWellFormedUriString(ckan_file, UriKind.Absolute))
@@ -129,7 +129,7 @@ namespace CKAN.CmdLine
 
                 string message = String.Format("Too many mods provide {0}. Please pick from the following:\n", ex.requested);
 
-                int result = -1;
+                int result;
 
                 try
                 {
--- a/Cmdline/Action/KSP.cs
+++ b/Cmdline/Action/KSP.cs
@@ -16,7 +16,7 @@ namespace CKAN.CmdLine
             Manager = manager;
             User = user;
         }
-        
+
         internal class KSPSubOptions : CommonOptions
         {
             [VerbOption("list", HelpText="List KSP installs")]
@@ -96,7 +96,7 @@ namespace CKAN.CmdLine
                     default:
                         break;
                 }
-            } 
+            }
 
             #endregion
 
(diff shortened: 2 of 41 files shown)


Cmdline/Action/Install.cs
```diff
@@ -129,7 +129,7 @@ public int RunCommand(CKAN.KSP ksp, object raw_options)
 
                 string message = String.Format("Too many mods provide {0}. Please pick from the following:\n", ex.requested);
 
-                int result = -1;
+                int result;
```
pjf: Just to check, wouldn't this change the code below in the case where:

- Our `RaiseSelectionDialog` throws an exception, and
- That exception is not a Kraken?

In those cases we're checking if `result < 0` down at line 145.

Cmdline/Action/KSP.cs
```diff
@@ -245,7 +245,7 @@ private int SetDefaultInstall(DefaultOptions options)
                     keys[0] = Manager.Instances.IndexOfKey(defaultInstance);
                 }
 
-                int result = -1;
+                int result;
```
pjf: Ditto here. :)

GUI/Main.cs
```diff
@@ -979,11 +978,11 @@ private void exportModListToolStripMenuItem_Click(object sender, EventArgs e)
...
         private void selectKSPInstallMenuItem_Click(object sender, EventArgs e)
         {
-            Main.Instance.Manager.ClearAutoStart();
-            var old_instance = Main.Instance.CurrentInstance;
+            Instance.Manager.ClearAutoStart();
+            var old_instance = Instance.CurrentInstance;
             var result = new ChooseKSPInstance().ShowDialog();
-            if (!Equals(old_instance, Main.Instance.CurrentInstance))
-                Main.Instance.CurrentInstanceUpdated();
+            if (result == DialogResult.OK && !Equals(old_instance, Instance.CurrentInstance))
```
pjf: Was there a bug here? We weren't checking to see if the user clicked OK or Cancel?

Netkan/KS/KSAPI.cs
```diff
@@ -11,8 +11,10 @@ namespace CKAN.NetKAN
     /// </summary>
     internal class KSError
     {
+        #pragma warning disable 0649
```
pjf: This disables the 'never assigned to' warning, yes? But I presume we *do* need these attributes; do we know where they *do* get assigned? (If so, we should comment that, otherwise I'll come back and be confused again in the future.)

Cmdline/Action/Install.cs
RichardLake: How could it? Caught exceptions will return without using it. Uncaught will transfer the program flow out of the block and thus send result out of scope.

Netkan/KS/KSAPI.cs
RichardLake: JsonConvert.DeserializeObject<KSError>. I'll add a note

Cmdline/Action/Install.cs
pjf: Oh, of course! I was thinking of another reality where exceptions inside a `try` that weren't explicitly caught were instead suppressed. :)

pjf:
Looks good as of 6688cf0. Hoping to get checks for my line notes (potential bugfix included, and how do KSError attributes get set), but neither of these are blockers. :)

GUI/Main.cs
RichardLake: Not with the current code. On a cancel the instance would not of been changed so the !Equals call would be false.

pjf:
Thank you *so* much! This indeed removes all warnings for me when I type `make`. Merging with great, great thanks!

PR #1157, opened 2015-06-23
Enforce Visual Studio formatting.

pjf:
I suspect during the great merge we ended up with solution files that
work for visual studio, but cause monodevelop to try and do its own
thing.

These get monodevelop to VS formatting for me.


RichardLake:
Tested on VS15 RC

PR #1162, opened 2015-06-23
bin/outdated_metadata.pl: Find and fix vague KSP version strings.

pjf:
We've already run this code and repaired the old data in KSP-CKAN/CKAN-meta#607,
but I figure it'd be good to keep the tool around just in case. This PR adds
the tool to our `bin` directory.


techman83:
Saw it in action + LGTM :-)

PR #1164, opened 2015-06-23
The fancy version name is now shown

Postremus:
In the autoupdate dialog of the gui and cmd,
and also in the settings dialog


pjf:
This is cool! :)

Can I be a stickler and ask for `aka.` to be changed to `aka`? I support the view that acronyms should be fully dotted or not dotted at all. :)

pjf:
Works nicely for me! Merging with great thanks. :)

Issue #1149, opened 2015-06-22
Convert code to use interfaces instead of concete classes. 


RichardLake:
Currently the vast majority of the code does not use interfaces. One effect of this is that tests are mostly integration tests and that unit tests are nearly impossible to write. An example of this is in GUIModTests,  due to being unable to sub in a Registry the tests are forced to create a ksp install on disk.

I can think of a couple of issues related to converting code to interfaces:
*  The file/folder structure - do we keep the actual implementations in the same file or have the interfaces in seperate files.
*  Do we convert existing tests to unit tests or leave them as they are.
*  How fine grained should the interfaces be. For a example the current Registry class could be a single interface or split so that methods which mutate it are separate to methods which just query it.

dbent:
- Very strongly in favor of interfaces in separate files. It's a psychological thing, but keeping the interface in the same file as the main concrete implementation leads to rather leaky abstractions as basically every method just gets added to the interface, without consideration for good design. Keeping the files separate forces you to think "Does this *really* belong on the interface?"
- I'd convert them bit by bit.
- I think just good software design comes in here: if different pieces of code have use for different interfaces then they should be split. *But*, in my experience it's rare to actually need to implement multiple "functional" interfaces (versus say `IEquatable<>`, `IComprable<>`, etc.). In the most extreme case you could literally have a separate interface for each method and property... which is bad.

To make maximal use of this we should also switch to using dependency injection rather than manually constructing classes.  I have a lot of experience with [Autofac](http://autofac.org/), but I'm open to thoughts.

pjf:
* Aye, +1 for separate files.
* Aye, convert them bit by bit. :)
* I have opinions, below. :)

My dynamic language heritage means that interfaces for testing doesn't come naturally (we can do mock objects and duck typing without that), but interfaces to allow different implementations comes *very* naturally.

In the case of the CKAN, one thing I'd love to do is be able move all the KSP-specific bits to use interfaces, and have it so that as much as possible one could compile the CKAN against a *different game* and still have it work. That's absolutely because of my grand scheme to take over the world, but a *lot* of what the CKAN does is pretty game-independent. (`ModuleInstaller` and `KSP` are the big classes that are not.)

[...]

PR #1167, opened 2015-06-23
Start of extracting interfaces.

RichardLake:
See #1149


pjf:
I've been in flux the last 24 hours, but the first 30% of this I've looked through seems excellent. Thank you so much!

Core/Registry/IRegistryQuerier.cs
```diff
@@ -0,0 +1,144 @@
...
+        /// </summary>
+        /// <param name="with_provides">If set to false will not check for provided versions.</param>
+        /// <returns>The version of the mod or null if not found</returns>
+        Version InstalledVersion(string identifier, bool with_provides = true);
+    }
+
+    /// <summary>
+    /// Helpers for <see cref="IRegistryQuerier"/>
+    /// </summary>
+    public static class IRegistryQuerierHelpers
```
pjf: I'm new to the culture, but is `IRegistryQuerierHelpers` the accepted way of doing this? The `I` immediately makes me think this is a virtual (interface) class, except that obviously that's not the case here.

(In Moose we'd compile the virtual class and the helper into a *role*, but obviously we can't do that in C#. :)

pjf:
This is great! Merged with thanks!

Core/Registry/IRegistryQuerier.cs
RichardLake: You are correct on the naming of the helper class - I'm just used to
postfixing helper and forgot about the I prefix. Given that it only
contains extension methods a postfix of extensions may be better then
helper anyway. I'm currently interstate so I can't make a PR untill Sunday.

Issue #374, opened 2014-11-15
NetKAN should accept files without $kref / $vref tokens


pjf:
I thought it did, but recent reports indicate that it doesn't.

It should, however, *warn* the user if it spots a file upon which it performs no expansion.

pjf:
re-opening after repo merge

pjf:
This still exists, with us currently getting:

```
247 [1] FATAL CKAN.NetKAN.Program (null) - Object reference not set to an instance of an object
```

However it strikes me that if we have no `$kref` we can just tell the user:

```
There is no $kref field in {0}, so netkan expansion is not needed or possible.
```

Or something similar. (In short, this is just a better-error-message change.)

plague006:
Having no $kref at all poses no issue, having a broken $kref now gives  a generic JSON deserialization error

techman83:
@plague006 would that be broken JSON? Which should be caught during testing with jsonlint I'd expect. I think this was completed during the refactor #1218 - but I'm not sure. Either way I think this can be closed now.

plague006:
It's not broken JSON as I understand it. It validates with jsonlint without issue. Some excerpts from a test:

````
"$kref": "#/ckan/techman/iscool",
````
gives the output
````
348 [1] FATAL CKAN.NetKAN.Program (null) - JSON deserialization error
````
which is really generic and hard for users to interpret.

**Comment by ayan4m1** [comment]
My first rodeo with the NetKAN codebase. The only place I see that error message in the current codebase is https://github.com/KSP-CKAN/CKAN/blob/master/Core/Types/CkanModule.cs#L297 - having a full stack trace would be handy because that CkanModule ctor is being called somewhat frequently in my limited profiling of the current codebase.

**Comment by HebaruSan** [comment]
Current output with `"$kref": "#/ckan/techman/iscool",`:

```
$ ../CKAN/_build/netkan.exe NetKAN/KSP-AVC.netkan 
343 [1] ERROR CKAN.CkanModule (null) - KSP-AVC missing required field download
354 [1] FATAL CKAN.NetKAN.Program (null) - KSP-AVC missing required field download
```

So it doesn't stop at JSON deserialization anymore, but it waits till later in the processing to complain, when it finally notices it doesn't have any URLs to download. Ideally yes, the error would specify that the `$kref` is invalid.

PR #1169, opened 2015-06-24
Feature: Per  #374 allow files without krefs

RichardLake:
Attempt at #374 . Would like other people to test as I'm not 100% confident on the metadata side of things.


pjf:
Oh dear, the netkan executable has changed so much that we don't even have a `MainClass.cs` any more. I hate not merging code, but I'm closing this as any chance we do make for #374 is likely to be very different in form.

Adding more discussion to #374 on what we should do with a file that has no `$kref`.

Issue #1120, opened 2015-06-16
Character classes in the spec


janbrohl:
There are at least two places in the spec where character classes are mentioned:
Identifiers "may only consist of letters, numbers and dashes."
"The mod_version may contain only alphanumerics and the characters . + (full stop, plus)"
But there are mods in the main repo using "_" in their version-strings (example: FingerboxesCore ksp1.0_release1) or "-"  (example: ActiveTextureManagement-x86-Basic 5-0). 

JSON is unicode so does letters mean everything that is a letter in unicode or just english letters? same for numbers.
Maybe it is just my english but does "dashes" mean "-" or "_", too?
Are "-" and "_" considered alphanumerics? (If not they should explicitly be allowed in the mod_version as otherwise mods in the main repo violate the spec)

pjf:
Thank you *so much* for spotting all these, they're *hugely* appreciated.

I think we need to change our definitions of versions to match what we're
seeing in reality (especially since we're already indexing them!), which
really comes down to "versions are strings". While Debian has the luxury of
enforcing limitations on version strings, I fear we can't usefully do so.

However we *do* have strict requirements on how versions are compared, and
these make the distinction between number sections and non-number sections,
so we absolutely need definitions for those.

For identifiers, I'd encourage use of the 7-bit ascii namespace, again
because is is aimed firmly at machine use. The human visible
title/description *should* be unicode, so "Anpanman" can display as "アンパ
ンマン" if desired.

If we're not mentioning that .ckan files are UTF-8 unicode documents, then
we should mention that as well.

So I think this means:

- [ ] Rewrite versions to match reality
- [ ] Define character classes when discussing version comparison
- [ ] Define allowed identifier characters
- [ ] Explicitly mention we're utf8
- [ ] Make sure our schema agrees with all of the above.

janbrohl:
Simplest way to define allowed characters would be to state the literaly like:
Allowed characters are all in "0123456789" (digits), in "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ" (letters) and in  ".-_+" (no good name).

Wizarth:
Well, strictly speaking, the simplest way is to specify a RegExp (and specify which implementation you're using, ideally one implemented in the .NET/Mono framework).

I would definitely encourage UTF-8 and allowing UTF-8 in descriptive names. Also, I wonder if there are any cases using \" or similar to enable quotes inside quoted JSON entries?

[...]

PR #1170, opened 2015-06-24
More exactly define character classes in spec

janbrohl:
see issue #1120


pjf:
Thank you so much for this, and thank you for your patience with this PR!

PR #1196, opened 2015-06-27
Renamed the Delete Button to Forget

### LINKED ISSUE #1194: "Delete" a KSP install could be interpreted as deleting it on disk
Minor UI issue in this screen...
image

I hesitated before removing an install I had renamed. (Not for long, because 1. it was a backup and 2. CKAN was cautious about deleting stuff elsewhere.) But I think it could confuse other users.

Proposed solution: add a tooltip to the Delete button reassuring the user that their install won't be deleted on disk.

Postremus:
In the "Select KSP Instance" dialog.
Closes #1194


pjf:
Have I mentioned you're awesome recently?

Tested and merging with great thanks. :)

PR #1197, opened 2015-06-27
Implement `x_netkan_override`

### LINKED ISSUE #1160: Allow Netkan per-version overrides
Sometimes we'd like to override what data `netkan.exe` indexes for a single version. For example, if an existing release works with newer KSP releases.

I think a good solution is to update `netkan.exe` so that we can add version-specific directives. I'm going to propose te `x_netkan_override` field, which does an exact version comparison, and if successful uses the override segment to replace existing fields as the final step before writing (ie: after KS, AVC, and other expansions):

```
"x_netkan_override" : {
    "version" : "1.4.2",
    "overide" : {
        "ksp_version_min" : "1.0.0",
        "ksp_version_max" : "1.0.3"
    }
}
```

Then if we see that precise version of a mod being released (which includes reindexing what we already have) then we can change what we spit out. If we see any other version we ignore the override and proceed as normal.

Originally discussed in KSP-CKAN/CKAN-meta#608.

pjf:
This PR implements the `x_netkan_override` syntax described in #1160.

Includes test cases. Tested on live data.

Closes #1160 and brings in a new age of prosperity. :)

### DIFF
--- a/Netkan/CKAN-netkan.csproj
+++ b/Netkan/CKAN-netkan.csproj
@@ -52,6 +52,7 @@
     <Compile Include="AVC\AVC.cs" />
     <Compile Include="AVC\JsonAvcToKspVersion.cs" />
     <Compile Include="Web.cs" />
+    <Compile Include="NetkanOverride.cs" />
   </ItemGroup>
   <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


pjf:
Wishing @Dazpoet to see this, because I'm pretty sure it's the sort of thing Daz would like. :)

pjf:
Also looking hopefully at @dbent for review and merge (if appropriate), as @dbent is the mastermind behind the override syntax. :)

dbent:
:+1: :lollipop:

Issue #839, opened 2015-05-02
Add updates etc to a background thread


KarryonDesigns:
Adding the update feature etc to a background thread would enable to the program window to be drug around the screen while an update is running.  I came across this issue when first loading the program and trying to move it to a secondary monitor and had to wait for the update to finish.  Not a bug more of a feature request to add additional usability

Postremus:
Would like that feature too

----- Ursprüngliche Nachricht -----
Von: "KarryonDesigns" <notifications@github.com>
Gesendet: ‎02.‎05.‎2015 13:10
An: "KSP-CKAN/CKAN" <CKAN@noreply.github.com>
Betreff: [CKAN] Add updates etc to a background thread (#839)

Adding the update feature etc to a background thread would enable to the program window to be drug around the screen while an update is running. I came across this issue when first loading the program and trying to move it to a secondary monitor and had to wait for the update to finish. Not a bug more of a feature request to add additional usability

pjf:
This is a really solid request, adding to our to-do queue.

pjf:
Oh wait, just to check, is this when it client is updating *itself*, or when it's download KSP mods?

Postremus:
The gui is freezed during a refresh, and after starting a update of the client from the settings screen.

ayan4m1:
Need to move those jobs to a BackgroundWorker or if we have access to `async/await` that can also be used to keep the UI responsive during slow operations.

PR #1198, opened 2015-06-27
Client is now movable during the refresh

Postremus:
solves one part of #839

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -366,6 +366,13 @@ namespace CKAN
 
         protected override void OnFormClosing(FormClosingEventArgs e)
         {
+            //Only close the window, when the user has access to the "Exit" of the menu
+            if (!menuStrip1.Enabled)
+            {
+                e.Cancel = true;
+                return;
+            }
+
             m_Configuration.WindowLoc = Location;
 
             // Copy window size to app settings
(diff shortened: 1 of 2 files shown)


pjf:
You are awesome. :)

Tomorrow is sprint day for me, and I'm looking forward to *merging all the things* tomorrow (if I'm not beaten to it first, which is even better!).

GUI/Main.cs
```diff
@@ -366,6 +366,13 @@ protected override void OnLoad(EventArgs e)
 
         protected override void OnFormClosing(FormClosingEventArgs e)
         {
+            //Only close the window, when the user has access to the "Exit" of the menu
+            if (!menuStrip1.Enabled)
```
pjf: Just checking the logic here: if the menu strip *isn't* enabled, then closing the window is treated as a cancel instead?

Postremus: if the menustrip isn't enabled (e.g. the user isn't allowed to click on the menustrip), we disallow closing of the form.

pjf:
Notes for me:

- The client wasn't able to hit 'x' to close the form before this change, so it's *not* a user-visible change that they can now.
- Testing by @plague006 confirms this works as intended under Windows.
- Testing by me confirms this doesn't change behaviour under Linux (which always allows moving windows).

Merging with thanks. :)

PR #1203, opened 2015-06-27
Fixes #1179

### LINKED ISSUE #1179: GUI shouldn't say "About to remove:" when there's nothing to remove.
When performing any operation, the GUI has a message listing what it's "About to remove:", even when there's nothing being removed. While relatively harmless, it'd be nice to skip this spurious message when we're not removing things. :)

martinnj:
Simply check if there is anything in the list of mods to remove before
doing anything else.

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -668,6 +668,12 @@ namespace CKAN
             // Find all the things which need uninstalling.
             IEnumerable<string> goners = registry_manager.registry.FindReverseDependencies(mods);
 
+            // If there us nothing to uninstall, skip out.
+            if (goners.Count() < 1)
+            {
+                return;
+            }
+
             User.RaiseMessage("About to remove:\n");
 
             foreach (string mod in goners)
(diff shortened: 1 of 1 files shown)


martinnj:
Btw, if I missed some steps or procedures you want for PR's, let me know and I'll take another pass at it.

dbent:
Minor change: You can use `!goners.Any()` as the check. The difference being that `Count()` requires the enumeration of the entire enumerable before it can return a value, whereas `Any()` just attempts to enumerate one value before returning `true` or `false`, so linear vs. constant time operation.

I doubt it'd ever make a practical difference in this instance, but a good habit to have.

martinnj:
I've added the change, good tip.

pjf:
You both rock, thank you! :)

The only tweak I'd make in the future is to include a human-friendly description as the PR title, and the issue it relates to in the PR body. This means there's human-friendly merge message for when we're writing release notes, an an easy-to-click on link to the issue when reviewing. (Github isn't great at linking issues in titles)

I've [updated the workflow](https://github.com/KSP-CKAN/CKAN/wiki/Workflow) to mention this now.

Thank you again for the contribution. You're awesome!

martinnj:
I'll keep that in mind for next time. :)

Issue #1191, opened 2015-06-26
CKAN should not remove ship folders if empty on uninstall


pjf:
As reported by @SkyMarhsal in #1189, the CKAN will *remove* subdirectories
(VAB, SPH, etc) if a mod that includes craft files (eg: FAR) is uninstalled
and the directory is then empty.

The CKAN should never be deleting these directories, and realistically we
should probably be *creating* the system Ships subdirectories if they
don't exist.

Hopefully this is a simple patch to `ModuleInstaller`. :)

RichardLake:
Just to be clear is it removing ships/VAB or ships/@thumbs/ and subdirectories? The former should be impossible - the code only calls directory.delete in one place and that has a guard for the directories. However it was never updated with 1.0.0 to include @thumbs and its subdirectories ...

martinnj:
The addition is based on pjf comment "The CKAN should never be deleting these directories, and realistically we should probably be creating the system Ships subdirectories if they don't exist." which I took to mean if the directories are missing for whatever reason, just recreate them, even if they we're manually deleted. If I'm wrong I can close the RP.

RichardLake:
Just spent some time checking - I can get ckan to remove @thumbs/VAB, @thumbs/SPH and @thumbs but not VAB or SPH. If I am correct the "proper" fix is to add the three directories to the guard on line 773 of ModuleInstaller.cs 

Even with a fix it will still be worthwhile recreating them but doing so in uninstall seems a odd location. Having them missing will cause issues when installing modules, or while playing KSP, not while uninstalling.

martinnj:
Right, I'll take another gander at it. Should I let the PR stay open or just close it and create another one once it's fixed?

RichardLake:
As far as I know either way would be fine - it isn't like there is discussion on the PR which would be lost if you made a new one.

Issue #1148, opened 2015-06-22
(De)select all button


RichardLake:
It would be useful to have a button to select/deselect all the mods on the recommend and suggests screens.

Postremus:
such a button is everywhere usefull. for example, when you want to deinstall all installed mods.

RichardLake:
#1211 Adds a delectect all button to the main mod list not the recommends/suggests window.

plague006:
We now have a (de)select all button available for recommends and suggests. We don't have one available in the main mod screen because:

1) wanting to uninstall all mods is very uncommon and we would be better served by a button/function that does explicitly that
2) it's not possible to use it in a "select all" context as you would inevitably create conflicts when trying to install every mod
3) a (de)select all button would almost certainly be mistaken for a "revert all install/uninstall changes" button where a user expects their currently installed mods to remain intact.

With all that established, closing this but certainly open to the idea of a "uninstall all"-type option.

politas:
Select All could be useful if applied to a filtered list based on the "filter by" fields.

plague006:
Can only speak for my own habits, but I usually filter with a single mod in mind. That said I don't represent everyone, so re-opening.

politas:
I can easily see someone deciding they want all of Roverdude's mods, for instance - They generally make a useful combined set.

ayan4m1:
Yes, I would use this with the filter box to select small batches of mods at once using a common search term. I understood that it wasn't trivial to add to this app in particular, but I was missing the ability to find 10 mods and check them all at once.

If you wanted to "uninstall all" then you'd set the filter type to be Installed and then hit the "deselect all shown" button.

From a UI perspective I think changing the "Installed" column to have a checkbox would keep things small. It also means the "de-select all" checkbox can go away assuming the two tabs are using the same grid control.

I think the workaround for the "trying to change the dependency tree state too much at once" issue will be to uncheck each item individually and allow the state to settle without enforcing deps, but that will probably decrease the speed/efficiency of the operation and I haven't looked at the ckan.dll API much yet.

I don't want to bring an opinionated pull request so I am hoping to get a little more feedback from everyone using my half-baked ideas at solving this as a lightning rod.

Issue #977, opened 2015-05-29
Implement `ckan.exe remove --all`


pjf:
This is a common request, and I found myself reaching for it today as well.

Dazpoet:
Currently 'ckan remove --re .` fills that function so a a `remove --all` command could probably just be an alias to trigger it?

PR #1205, opened 2015-06-27
Hopeful contributions

martinnj:
code related to:
 - #1191
 - #977
 - #1148

Note: There is an oddity which occurs when you have a mod installed that have dependencies, and you click the new "deselect all". If you then reselect a mod that have dependencies, it does not install the mods since dependencies are not met.
I've been unable to find where dependencies are resolved, I think that logically it should  be done in `MarkModForInstall`, however I do not see the current resolution logic there.

If you can point me where to implement it I will fix this myself.


pjf:
In the core, relationship resolution happens in the `RelationshipResolver` class, although I'm not personally familiar with how the GUI reaches that point, which is what I think you're after.

I'm working through as many PRs as I can today, so hopefully that includes this one, too (if nobody else gets to it first, which would be awesome).

Also while it's fine to pop all three changes into one PR, we love it even more when they come in separate PRs. That makes reviewing much easier, gives us a nicer changelog (each merge into master usually results in a release notes entry), and means issues with one change doesn't block any others.

Thank you *so much* again for this. And if you haven't found us on #ckan on irc.esper.net, you should. :)

martinnj:
I was on the IRC a couple of hours last night (EU time) was a quiet place at the time.

I'll close this PR, do some voodoo to my fork and create a bunch of new PRs instead.

PR #1206, opened 2015-06-28
netkan.exe: usability: use system-wide CA certificate bundle if custo…

dexen:
…m one was not found

### DIFF
--- a/Netkan/Web.cs
+++ b/Netkan/Web.cs
@@ -28,8 +28,12 @@ namespace CKAN.NetKAN
             easy = new CurlEasy
             {
                 UserAgent = Net.UserAgentString,
-                CaInfo = ResolveCurlCaBundle()
             };
+            string bundle = ResolveCurlCaBundle();
+            if (bundle != null)
+            {
+                easy.CaInfo = bundle;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


dbent:
This is actually a no-op change, the default value of `CurlEasy.CaInfo` is `null`.

dexen:
In CurlEasy's implementation there is difference between passing `null` and not assigning an option at all. Explicitly passing `null` sets the option to zero, and not doing so retains libcurl's default config.

As it stands, netkan.exe explicitly passes null to CurlEasy, which causes libcurl's CA option to be set to `IntPtr.Zero` and thus default CA bundle NOT to be used. With this patch, libcurl's default configuration may be retained.

The problem manifests only when no custom CA bundle is used. For example, when you run *$ make; mono netkan.exe*  in the root directory of the project.

Relevant code fragment from `CurlEasy.cs`:

        private void setStringOption(CurlOption option, ref string field, string value)
        {
            ensureHandle();
            field = value;
            if (string.IsNullOrEmpty(value))
            {
                setLastError(NativeMethods.curl_easy_setopt(_pCurl, option, IntPtr.Zero), option);
             }
             else /* sets the value */

dbent:
Ugh, that's the kind of side effect a property shouldn't have.

PR #1209, opened 2015-06-28
Bad resource links will no longer kill Module construction.

### LINKED ISSUE #1208: Bad links will cause netkan.exe to cry.
{Kerbal Foundries](http://www.kerbalstuff.com/mod/58) has a source code link of `http://included%20in%20the%20download`. This isn't a valid `System.Uri` string, which causes `Module` inflation to fail, which causes `netkan.exe` to bail.

We should log and continue if a resources URL is bad, rather than bailing entirely. I have a fix on the way.

pjf:
Fixes #1208, which is preventing some mods from indexing correctly.

### DIFF
--- a/Core/Converters/JsonIgnoreBadUrlConverter.cs
+++ b/Core/Converters/JsonIgnoreBadUrlConverter.cs
@@ -18,6 +18,9 @@ namespace CKAN
         {
             string value = reader.Value == null ? null : reader.Value.ToString();
 
+            if (value == null)
+                return null;
+
             Uri url;
 
             try
--- a/Core/Converters/JsonIgnoreBadUrlConverter.cs
+++ b/Core/Converters/JsonIgnoreBadUrlConverter.cs
@@ -40,6 +40,15 @@ namespace CKAN
             return false;
         }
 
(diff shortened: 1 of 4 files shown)


pjf:
Found a spammy bug, one moment while I fix. :)

pjf:
And less spammy now. :) Review/merges welcome. :)

Core/Converters/JsonIgnoreBadUrlConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            if (value == null)
+                return null;
+
+            Uri url;
+
+            try
+            {
+                url = new Uri(value);
+                return url;
```
RichardLake: return new Uri(value);
Also why have the declaration outside the try when the catch/following code doesn't use it.

pjf: Because I wrote this when I was very tired. :) Thanks for catching these! :)

pjf:
@RichardLake : Code tidy-ups based upon your feedback is in. :)

PR #1212, opened 2015-06-28
Recreates ship folders if missing.

### LINKED ISSUE #1191: CKAN should not remove ship folders if empty on uninstall

martinnj:
fixes #1191

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -793,6 +793,17 @@ namespace CKAN
                         log.InfoFormat("Not removing directory {0}, it's not empty", directory);
                     }
                 }
+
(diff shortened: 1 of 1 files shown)


martinnj:
Going to make changes as discussed in the [Issue](https://github.com/KSP-CKAN/CKAN/issues/1191).

PR #1204, opened 2015-06-27
Support MetaNetkans

dbent:
This is a rough implementation of #865.

It supports the following:

- Allows you to specify `#/ckan/netkan` `$kref` properties, which are then downloaded and processed automatically.
- The `spec_version` used in the final output `.ckan` is the higher of the two specified in the metanetkan or the target netkan.
- Properties specified in the metanetkan override properties specified in the target netkan.

An example usage:
```json
{
    "spec_version": "v1.6",
    "identifier": "HotSpot",
    "$kref": "#/ckan/netkan/https://raw.githubusercontent.com/KSP-CKAN/NetKAN/master/NetKAN/HotSpot.netkan",
    "license": "restricted"
}
```

Produces the following ckan:
```json
{
    "spec_version": "v1.6",
    "identifier": "HotSpot",
    "depends": [
        {
            "name": "ModuleManager",
            "min_version": "2.6.5"
        }
    ],
    "resources": {
        "homepage": "http://forum.kerbalspaceprogram.com/threads/123967",
        "repository": "https://github.com/Apokee/HotSpot",
        "kerbalstuff": "https://kerbalstuff.com/mod/937/Hot%20Spot"
    },
    "ksp_version": "1.0.4",
    "name": "Hot Spot",
    "license": "restricted",
    "abstract": "Display better thermal data.",
    "author": "dbb",
    "version": "0.4.2",
    "download": "https://kerbalstuff.com/mod/937/Hot%20Spot/download/0.4.2",
    "x_generated_by": "netkan",
    "x_screenshot": "https://kerbalstuff.com/content/dbb_49/Hot_Spot/Hot_Spot-1435028274.4356654.png",
    "download_size": 24906
}
```

Notice that the `spec_version` is `v1.6` and the `license` property is overriden.

PR #1218, opened 2015-06-28
Netkan Refactoring

dbent:
This is a mondo refactoring of NetKAN. Viewing the diff is a pretty useless way to review this, I'd recommend just stepping through the code in sequence. This refactoring does a few different things:

- Refactors the entire "pipeline" into 5 obvious steps:
  1. Read `.netkan`
  2. Validate `.netkan`
  3. Transform `.netkan` into `.ckan`
  4. Validate `.ckan`
  5. Write `.ckan`
- Includes the changes in #1204
- Includes the changes in #1206
- Adds a bunch more tests
- Changed a bunch of stuff to be more idiomatic C#

##### Pipeline
The major refactoring is the introduction of `ITransformer` and `IValidator`. An `ITransformer` has a single method, `Metadata Transform(Metadata)`, that takes a `Metadata` object (a wrapper around a plain `JObject`, that exposes some useful parsed fields) and produces another `Metadata` object. For simplicity's sake `Metadata` objects are immutable. *Every* alternation necessary to turn a `.netkan` into a `.ckan` is implemented as an `ITransformer`. There are 13 "normal" implementations of `ITransformer`, called in this specific sequence:

- `MetaNetkanTransformer` Perform a recursive lookup of another NetKAN file.
- `KerbalstuffTransformer` Looks up data from Kerbal Stuff.
- `GithubTransformer` Looks up data from GitHub.
- `HttpTransformer` Looks up data from an arbitrary HTTP endpoint.
- `JenkinsTransformer` Looks up data from a Jenkins build server.
- `InternalCkanTransformer` Populates data from a CKAN file included in the distribution package.
- `AvcTransformer` Populates version data from an AVC version file included in the distribution package.
- `ForcedVTransformer` Prepends the letter 'v' to the version if necessary.
- `EpochTransformer` Uses a particular version epoch if necessary.
- `VersionedOverrideTransformer` Overrides properties based on the version.
- `DownloadSizeTransformer` Adds the size of the download.
- `GeneratedByTransformer` Adds a property to indicate the program responsible for generating the metadata.
- `StripNetkanMetadataTransformer` Removes metadata useful only to NetKAN.

If a particular `ITransformer` isn't applicable to the current state of the `Metadata`, it simply passes the `Metadata` through the pipeline unchanged.

There is also one "special" `ITransformer`, `NetkanTransformer`, that can transform NetKAN metadata into CKAN metadata, it does this by having knowledge of all the other transformers and calling them in the aforementioned sequence.

NetKAN was effectively already operating as a pipeline but the implementation wasn't explicit. I think this is easier to reason about and change, and probably most importantly: *test*. Every `ITransformer` does *one* thing (for god's sake I have a `GeneratedByTransformer`!) that makes its behavior obvious and testable. This also drastically reduces redundant code; inspecting CKANs in zip files is now done *once*, adding the size of the download is now done *once*, etc.

I've also introduced an `IValidator` interface that has a single method `void Validate(Metadata)` that takes a `Metadata` object and either does nothing if it passes validation or throws an exception if it doesn't. There are two special `IValidator` implementations: `NetkanValidator` which is used to validate the incoming metadata from the `.netkan` file and `CkanValidator` which is used to validate the outgoing metadata before it's written to the `.ckan` file.

##### MetaNetKANs
This folds the changes in from #1204. Basically a `$kref` of `#/ckan/netkan` is used to point to a `.netkan` file hosted somehow which is then consumed as if all its contents were included in the original `.netkan`. The recursion only goes one level, if the target `.netkan` is itself a MetaNetKAN then an exception is thrown.

##### Tests
I've added a bunch of tests for this stuff. Here is the [coverage report](https://dl.dropboxusercontent.com/u/123465/CKAN/NetKANTests-20150630.html) for this branch. I've also added [Moq](https://github.com/Moq/moq4) as a tests-only dependency to mock out interfaced dependencies. I've committed the package to source control like the others, but at some point in the future I'd like to modify the build to remove the committed packages and just use NuGet to restore them on-demand.


dbent:
Just adding a note here that I found redundant implementations of `x_netkan_force_epoch` and/or `x_netkan_epoch` that's only going to exist in `VersionTransformer`.

hakan42:
Do I understand this correctly that NetkanTransformer calls all other known transformer implementations and each of those is responsible to check whether it has to do any work on the input data?

hakan42:
Do we impose any order in which the transformers are being called? If so, I'd love if there would be a comment to that effect in NetkanTransformer

dbent:
> Do I understand this correctly that NetkanTransformer calls all other known transformer implementations and each of those is responsible to check whether it has to do any work on the input data?

Yes.

> Do we impose any order in which the transformers are being called? If so, I'd love if there would be a comment to that effect in NetkanTransformer

The order is given by their order in the list.

dbent:
@pjf This is good to go.

pjf:
I'm reviewing this now. You rock *so hard*. Thank you.

Netkan/Program.cs
```diff
@@ -0,0 +1,175 @@
...
+            if (args.Any(i => i == "--debugger"))
+            {
+                Debugger.Launch();
+            }
+
+            Options = new CmdLineOptions();
+            Parser.Default.ParseArgumentsStrict(args, Options);
+
+            BasicConfigurator.Configure();
+            LogManager.GetRepository().Threshold = Level.Error;
```
pjf: Our default should probably be `Level.Warn` rather than `Level.Error`, but that's the smallest of nits.

Netkan/Services/CachingHttpService.cs
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    internal sealed partial class CachingHttpService
+    {
+        private bool _isDisposed;
+
+        ~CachingHttpService()
+        {
+            Dispose(disposing: false);
```
pjf: I'm not as well versed in C# destructors as I'd like to be. To check my understanding here:

- If  `Dispose` is called directly (eg, by leaving a `using` scope), then we run our curl clean-up, and set `_isDisposed` to true.
- If our `CachingHttpService` is destroyed, we don't touch curl and just set `_isDisposed` to true.
- We can call `EnsureNotDisposed()` to make sure our object hasn't been disposed (but presumably hasn't been destroyed, because if it's destroyed we couldn't have a reference to it).

So mostly I'm checking under what circumstances we'd see the destructor called *without* `Dispose` also being invoked just before it. I feel like I'm missing something with the logic flow here. (I don't know 100% how the .NET garbage collector works, but I presume it won't go destroying something you have an active reference to.)

Netkan/Services/ModuleService.cs
```diff
@@ -0,0 +1,156 @@
...
+                // Okay, there's *really* nothing there.
+                if (files.Count == 0)
+                {
+                    return null;
+                }
+            }
+
+            var remoteIndex = 0;
+
+            if (!string.IsNullOrWhiteSpace(internalFilePath))
```
pjf: Hi, I'm Paul, and I have oodles of jetlag. Am I correct that `internalFilePath` means that we can specify a *particular* `.version` file should be used, which helps us in the case that we find multiple files, or in the situation where we want to be certain the `.version` file is in a well-known location?

Netkan/Sources/Github/GithubApi.cs
```diff
@@ -0,0 +1,76 @@
...
+            Log.Debug("Parsing JSON...");
+            var releases = JArray.Parse(json);
+
+            // Finding the most recent *stable* release means filtering
+            // out on pre-releases.
+
+            foreach (var release in releases)
+            {
+                // First, check for prerelease status...
+                if (reference.UsePrelease == (bool)release["prerelease"])
```
pjf: Oh! This is checking to see if the release status is the same as what we're expecting. In other words:

1. If we're run with `--prerelease`, we'll only ever index a pre-release, even if a more recent stable release exists.
1. If we're run without `--prerelease`, we'll only ever index a stable release.

I guess nobody's complained about that so far? (In a strong multi-repo model I'm using to seeing *all* releases in the "dev" repositories, but I doubt this really has any impact on how our users work.)

Netkan/Services/CachingHttpService.cs
dbent: Pretty much spot on. The destructor will be called if the object is about to be garbage collected without someone previously calling `Dispose()` (usually via a `using` statement). This is pretty much a canonical implementation of the `IDisposable` [pattern](https://msdn.microsoft.com/en-us/library/ms244737.aspx). The missing bit is the clean-up of native resources, since we don't actually have any native resources (The cURL lib presumably does, but we're not responsible for any of that).

Managed resources are only cleaned up when `disposing` is `true`, since if we're being called by the finalizer, you can't assume *any* managed resource still exists.

I am now just seeing that I'm missing the call to `GC.SuppressFinalize(object)`. Not strictly necessary, but a small micro-optimization that prevents the finalizer from bothering to call the destructor if the object was already disposed. I'll add it in a bit.

Netkan/Services/ModuleService.cs
dbent: Yeah, that already existed in previous implementations. I don't think anything actually uses that feature yet...

Netkan/Sources/Github/GithubApi.cs
dbent: Yep, same as the existing behavior.

Although when I first saw `--prerelease` I assumed it meant "include prereleases *in addition to* stable releases" not "include prereleases *instead of* stable releases".

dbent:
I've been meaning to write a small test bench for this that will execute the old and refactored versions of netkan to make sure they generate equivalent outputs, I should get on that...

Netkan/Services/CachingHttpService.cs
pjf: > The destructor will be called if the object is about to be garbage collected without someone previously calling `Dispose()` (usually via a `using` statement)

Time for me to learn about the GC some more. Am I correct the finaliser runs *after* dispose, but as a final pass before the object is completely destroyed?

If that's the case, and we're *not* cleaning up our managed resources, the assumption is that by the time we reach the finaliser then `.Dispose()` will have already been called, yes? In that case I fear I may be misunderstanding the pattern; I would expect us to check to see if our managed resources are still around and dispose of them if so, making sure we don't accidentally leak if our object falls out of scope and wasn't created with `using`.

Netkan/Transformers/OptimusPrimeTransformer.cs
```diff
@@ -0,0 +1,24 @@
...
+        private static readonly ILog Log = LogManager.GetLogger(typeof(OptimusPrimeTransformer));
+
+        public Metadata Transform(Metadata metadata)
+        {
+            var json = metadata.Json();
+
+            JToken optimusPrime;
+            if (json.TryGetValue("x_netkan_optimus_prime", out optimusPrime) && (bool)optimusPrime)
+            {
+                Log.Info("Autobots roll out!");
```
pjf: Oh my goodness,  *I* was going to make this joke somewhere when doing code review! You beat me to it! :)

Netkan/Services/CachingHttpService.cs
dbent: So two scenarios here:

##### Proper Disposal
In this scenario the object is constructed and `Dispose()` is called on it, either manually or automatically by the termination of a `using` block. The order of events is then:

- `Dispose()` is called
- `Dispose(true)` is then called
- We check to see if we were called by the `Dispose()` method by checking the `disposing` flag. This is `true`, so dispose the managed objects we 'own'.
- Regardless of the value of `disposing`, cleanup unmanaged resources. *We don't have any so do nothing*.
- At some indeterminate time in the future, the object will be garbage collected. Just before it's destroyed, the finalizer will check to see if the object has a destructor, it does. It will then check to see if finalization should be suppressed, we called `GC.SuppressFinalize(this)`, so yes. The finalizer destroys the object will no further code execution.

##### Improper Disposal
In this scenario is object is constructed and `Dispose()` is never called on it. At some point in the future all references to it are lost and the object becomes garbage. During garbage collection:

- The finalizer checks if the object has a destructor, it does.
- The finalizer checks if finalization should be suppressed. In this scenario `GC.SuppressFinalize(this)` is never called, so no.
- The finalizer executes the destructor.
- `Dispose(false)` is executed.
- We check to see if we were called by the `Dispose()` method by checking the `disposing` flag. This is `false`, so ignore any managed objects. Why? Because during finalization typical reference rules no longer apply. Any managed objects we hold a reference to could already be finalized and references we hold to them would be `null`, otherwise it would be impossible to garbage collect circular references. We could perform a null check before accessing them, but those objects could have their own references nulled out and they would be in an indeterminate state they did not expect to be in, leading to all sorts of potential exceptions being thrown (and throwing exceptions on the finalizer thread == very bad). Therefore standard procedure is to ignore managed objects and assume their own destructors will handle clean-up.
- Regardless of the value of `disposing`, cleanup unmanaged resources. *We don't have any so do nothing*.

So basically, this scenario is a degenerate case. Nothing happens except `_isDisposed` gets set to `true`. If we happened to have unmanaged resources, we would clean them up, but we don't.

Why bother with a `Dispose()` method at all? Why not just let the destructors handle all clean-up?
- Predictability, you know exactly when the clean-up code will execute, instead of an effectively random time in the future.
- Performance, when everything is disposed properly, destructors should never be executed and the only code running on the finalizer thread is the actual garbage collection.
- Robustness, since `Dispose()` gets called in normal code you can make safe assumptions about the state of the world, versus on the finalizer thread where some objects could still be live, some objects could be destroyed and you can't predict which are which.

pjf: > Because during finalization typical reference rules no longer apply. Any managed objects we hold a reference to could already be finalized and references we hold to them would be null, otherwise it would be impossible to garbage collect circular references.

Hooray! This is where my understanding was flawed. My background is in garbage collectors that hold reference counts to be sacred. That means one *can* leak circular references, but it's also guaranteed an object will never be destroyed until its reference count reaches zero¹. (Indeed, one of the purposes of destructors is to break circular references where weak references are inappropriate.)

So in `Dispose()` for .NET we're guaranteed that our managed resources exist (and are still our problem), but in `~CachingHttpService()` all bets are off regarding other references, so we shouldn't touch them.

Thank you *so much* for your patience and explanations; I've levelled up in .NET garbage collection as a result. :)

---

¹ An interesting aside is that at least in Perl, this also means that garbage collection is 100% deterministic, as it happens the moment an object's reference count equals zero, meaning in almost all circumstances destructors play the same role as `Dispose`.

dbent: :+1:

pjf:
From testing:

```
Sources/Kerbalstuff/KerbalstuffMod.cs(9,20): warning CS0649: Field `CKAN.NetKAN.Sources.Kerbalstuff.KerbalstuffMod.id' is never assigned to, and will always have its default value `0'
Sources/Kerbalstuff/KerbalstuffMod.cs(12,23): warning CS0649: Field `CKAN.NetKAN.Sources.Kerbalstuff.KerbalstuffMod.license' is never assigned to, and will always have its default value `null'
Sources/Kerbalstuff/KerbalstuffMod.cs(13,23): warning CS0649: Field `CKAN.NetKAN.Sources.Kerbalstuff.KerbalstuffMod.name' is never assigned to, and will always have its default value `null'
Sources/Kerbalstuff/KerbalstuffMod.cs(14,23): warning CS0649: Field `CKAN.NetKAN.Sources.Kerbalstuff.KerbalstuffMod.short_description' is never assigned to, and will always have its default value `null'
Sources/Kerbalstuff/KerbalstuffMod.cs(15,23): warning CS0649: Field `CKAN.NetKAN.Sources.Kerbalstuff.KerbalstuffMod.author' is never assigned to, and will always have its default value `null'
Sources/Kerbalstuff/KerbalstuffMod.cs(16,28): warning CS0649: Field `CKAN.NetKAN.Sources.Kerbalstuff.KerbalstuffMod.versions' is never assigned to, and will always have its default value `null'
Sources/Kerbalstuff/KerbalstuffMod.cs(17,20): warning CS0649: Field `CKAN.NetKAN.Sources.Kerbalstuff.KerbalstuffMod.website' is never assigned to, and will always have its default value `null'
Sources/Kerbalstuff/KerbalstuffMod.cs(18,20): warning CS0649: Field `CKAN.NetKAN.Sources.Kerbalstuff.KerbalstuffMod.source_code' is never assigned to, and will always have its default value `null'
Sources/Kerbalstuff/KerbalstuffMod.cs(19,20): warning CS0649: Field `CKAN.NetKAN.Sources.Kerbalstuff.KerbalstuffMod.default_version_id' is never assigned to, and will always have its default value `0'
```

I'm guessing these *do* get assigned to but through the wonders if deserialisation from JSON. Is there a recommended way to suppress these warnings? (Decorate them with JSONy attributes?)

pjf:
Decorating them with `[JsonProperty]` suppresses the warning, and makes it clear they're from deserialisation, so I'm doing that for now. :)

dbent:
Another option would be:

```C#
#pragma warning disable 649
// offending code here
#pragma warning restore 649
```

pjf:
This has passed everything I've thrown at it, and in some cases fixed bugs I didn't even know were in the original. Merging with *great* thanks. This PR made my day.

dbent:
Woo, now I can start using metanetkans for my mods. :smile:

PR #1175, opened 2015-06-25
Fix: Issues with version handling.

RichardLake:
RelRes was not filtering candiade dependancies by the version.
The GUI is passing incorrect versions to the installer.

For a example of the second try to install MechJeb Modules for Far in a clean 1.0.2 install. It will fail. 
Reported by @BasharMilesTeg

Needs testing.

PR #1219, opened 2015-06-29
Fix for uncaught Inconsistent Kraken.

RichardLake:
When ran with procede_with_inconsistencies the exception should never have been thrown. See https://github.com/KSP-CKAN/CKAN/pull/1175#issuecomment-115166104 for details on reproduction steps.
This fix has issues with poor communication to the user regarding why it conflicts but it is better than crashing.


pjf:
Still trying to reproduce the original bug, I suspect all our other patches have been very effective at shielding it.

pjf:
Have had @BasharMilesTeg test and confirm this helps, so I'm merging away. (Apparently BMT can get another exception out of the code, but I don't believe that's a regression at all.)

Issue #844, opened 2015-05-05
Mods with missing/incompatible dependencies should list somewhere, even if they can't be installed


neurovertex:
I've spent some time trying to figure out why some mods weren't showing up, even though they had been updated, and were correctly referenced in CKAN-meta. Turns out when a mod has an incompatible (or uninstallable for any reason) dependency, it won't show in any list, not even 'incompatible'.

Maybe list them in 'incompatible', along with the reason why they are so ? Or add a separate category ?

Postremus:
Any example for an mod where this is the case?

plague006:
This might have been around the time of an update where this effect would have been more readily seen. A way to replicate this now might be: in a custom repo update your ksp rear and a mod that depends on MM to ksp_ver 1.1. That should demonstrate the effect.

ayan4m1:
I realize this issue is the older of the two, but fixing https://github.com/KSP-CKAN/CKAN/issues/844 will likely resolve this enhancement request.

politas:
This looks like a duplicate of #1695 to me. (Or #1695 is a duplicate of this, is probably a more correct way of putting it, but there's more discussion in that, so I'll close this..)

PR #1220, opened 2015-06-29
Show missing dependencies

Ippo343:
It would be very cool to know why a module is not compatible (I'm looking at you, RO).
This PR adds a new section to the output of the show command to list the unsatisfied dependencies for a module.

Currently it's incomplete as the search algorithm only searches for the module name in those available for the current version of KSP (which makes it useless to understand missing dependencies), and it turns out that the version lock is almost hardcoded.

### DIFF
--- a/Cmdline/Action/Show.cs
+++ b/Cmdline/Action/Show.cs
@@ -32,7 +32,7 @@ namespace CKAN.CmdLine
             if (installedModuleToShow != null)
             {
                 // Show the installed module.
-                return ShowMod(installedModuleToShow);
+                return ShowMod(installedModuleToShow, ksp);
             }
 
             // Module was not installed, look for an exact match in the available modules,
@@ -121,7 +121,7 @@ namespace CKAN.CmdLine
         /// </summary>
         /// <returns>Success status.</returns>
         /// <param name="module">The module to show.</param>
-        public int ShowMod(CkanModule module)
+        public int ShowMod(CkanModule module, CKAN.KSP ksp)
         {
             #region Abstract and description
             if (!string.IsNullOrEmpty(module.@abstract))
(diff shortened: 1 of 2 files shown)


pjf:
Still on my quest to close out old PRs, and this one is next on my list. :)

@Ippo343 : I know it wasn't easily possible to find what was missing before, but if you *do* want to try we may have a few options, including:

- Searching for mods by using the YOYO `GameComparator`, although this is by no means reliable, as a mod may refer to other mods that aren't in the user's registry at all. It's also a pretty new feature, and I haven't considered how it might work in this regard.
- Listing mods by their identifier only. This means it doesn't matter if we can't resolve a mod. It's also very straightforward.
- A combination of the above; meaning any display routine would need an enumerable of modules (for which full data can be shown) *and* an enumerable of identifiers (for which resolution failed).

I'm also totally cool closing this PR and opening a new one in the future with a more complete change. :)

P.S. You are awesome!

ayan4m1:
OK, it looks like this is an attempt to fix https://github.com/KSP-CKAN/CKAN/issues/1695 and resolve [this](https://github.com/KSP-CKAN/CKAN/issues/844) request. If so, I can take this ShowMissingDep branch and start working on the bug. Is anyone/everyone OK with closing this PR for now?

politas:
I don't think this has anything to do with #1695, other than accidentally. In fact, it looks like #1695 would make this useless in the GUI, as you'll never see the mods that are missing dependencies.

This looks to me like a CLI change so that when someone does a "list" command, we'll report which dependencies are not currently available for the relevant KSP version.

ayan4m1:
Ah, @politas that sounds much more accurate, I will do a little poking around in the cmdline code to see if there is an obviously place we're limiting a filter to "current KSP version only."

Cmdline/Action/Show.cs
```diff
@@ -120,7 +120,7 @@ public int ShowMod(InstalledModule module)
         /// </summary>
         /// <returns>Success status.</returns>
         /// <param name="module">The module to show.</param>
-        public int ShowMod(Module module)
+        public int ShowMod(Module module, CKAN.KSP ksp)
```
ayan4m1: Why is this method signature changing?

Ippo343:
This is so incredibly old. Farewell, my old PR.

PR #1222, opened 2015-06-30
Remove spurious and harmless warnings.

pjf:
Users often fixate on the warnings produced by our code, but the most
commonly seen warnings are completely harmless. This PR downgrades
two commonly seen warnings (one regarding URL handlers that's seen in
existing releases, and one regarding bad URLs that would be seen in
future releases) to the info level, as the code works perfectly fine
in the conditions which generate these.

No logic changes. Tested by hand.


pjf:
I'm being one of those crabby maintainers who's merging their own PRs because they want to make a release, and because I know changing the logging level of our messages shouldn't introduce any new bugs.

PR #1211, opened 2015-06-28
Adds a deselect all button and a select installed button.

martinnj:
fixes #1148

PR #1223, opened 2015-06-30
Revert "Adds a deselect all button and a select installed button."

pjf:
As mentioned in KSP-CKAN/CKAN#1211, these trigger a bug in mono 3.2.8 that results in any ticky box causing an endless exception spiral. This may or may not be easy to fix (I'm hoping it will be), but since master should always be release-ready, I'm rolling back the changes for now.


pjf:
Can confirm this removes the bug, so I'm merging away.

PR #1226, opened 2015-06-30
The RelationshipResolver can now also resolve with InstalledModules

Postremus:
This is particularly useful for installed .ckan files
Because a InstalledModule can't be a CkanModule, we use Module objects for storage now.
also fixes the issue mentioned in #1223

### DIFF
--- a/Core/Relationships/RelationshipResolver.cs
+++ b/Core/Relationships/RelationshipResolver.cs
@@ -77,8 +77,8 @@ namespace CKAN
     {
         // A list of all the mods we're going to install.
         private static readonly ILog log = LogManager.GetLogger(typeof (RelationshipResolver));
-        private readonly Dictionary<string, CkanModule> modlist = new Dictionary<string, CkanModule>();
-        private readonly List<CkanModule> user_requested_mods = new List<CkanModule>();
+        private readonly Dictionary<string, Module> modlist = new Dictionary<string, Module>();
+        private readonly List<Module> user_requested_mods = new List<Module>();
 
         //TODO As the conflict detection gets more advanced there is a greater need to have messages in here
         // as recreating them from reasons is no longer possible.
@@ -120,20 +120,44 @@ namespace CKAN
         /// <param name="registry"></param>
         /// <param name="kspversion"></param>
         public RelationshipResolver(IEnumerable<string> module_names, RelationshipResolverOptions options, IRegistryQuerier registry,
-            KSPVersion kspversion) :
-                this(module_names.Select(name => CkanModule.FromIDandVersion(registry, name, kspversion)).ToList(),
-                    options,
-                    registry,
-                    kspversion)
+            KSPVersion kspversion) : this(options,registry, kspversion)
         {
-            // Does nothing, just calls the other overloaded constructor
+            List<Module> modules = new List<Module>();
+            foreach (string identifier in module_names)
+            {
+                try
+                {
+                    modules.Add(CkanModule.FromIDandVersion(registry, identifier, kspversion));
+                }
+                catch (ModuleNotFoundKraken kraken)
+                {
+                    // Let's see if the mod was manually installed (via a .ckan, for example)
+                    InstalledModule installed = registry.InstalledModule(identifier);
+                    if (installed != null)
+                    {
+                        modules.Add(installed.Module);
+                    }
+                    else
+                    {
+                        throw kraken;
+                    }
+                }
+            }
+            AddModulesToInstall(modules);
         }
 
         /// <summary>
         /// Creates a new resolver that will find a way to install all the modules specified.
         /// </summary>
         public RelationshipResolver(IEnumerable<CkanModule> modules, RelationshipResolverOptions options, IRegistryQuerier registry,
-            KSPVersion kspversion):this(options,registry,kspversion)
+            KSPVersion kspversion) : this(modules.Select<CkanModule, Module>(m => m), options, registry, kspversion)
+        {
+            // Does nothing, just calls the other overloaded constructor
+        }
+
+        private RelationshipResolver(IEnumerable<Module> modules, RelationshipResolverOptions options, IRegistryQuerier registry,
+            KSPVersion kspversion)
+            : this(options, registry, kspversion)
         {
             AddModulesToInstall(modules);
         }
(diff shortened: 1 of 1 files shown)


Core/Relationships/RelationshipResolver.cs
```diff
@@ -120,20 +120,44 @@ public RelationshipResolver(RelationshipResolverOptions options, IRegistryQuerie
...
+                {
+                    // Let's see if the mod was manually installed (via a .ckan, for example)
+                    InstalledModule installed = registry.InstalledModule(identifier);
+                    if (installed != null)
+                    {
+                        modules.Add(installed.Module);
+                    }
+                    else
+                    {
+                        throw kraken;
```
pjf: Can we get a kraken with a more verbose message by any chance? If this ever ends up in a bug report it would be great to have a message indicating what happened. :)

pjf:
Reviewing this now. Does this mean that previously autodetected mods wouldn't participate in relationship resolution at all, because *that was the whole point of them existing*.

pjf:
@Postremus : Can you give a use case for this, which I presume would be when mods were installed from `.ckan` files but aren't in the upstream repo? In that case I would expect these to still be `CkanModule`s (ie, have full metadata), rather than the base class (which AFAIK is really only for autodetected dlls).

Anything which lets me write a test case that fails without this PR, and passes with this PR would be grand. :)

> Does this mean that previously autodetected mods wouldn't participate in relationship resolution at all, because that was the whole point of them existing.

Answering my own question, they *do* participate. I have a test case I wrote for this, and will merge separately.

Postremus:
@pjf 
```
 Can you give a use case for this, which I presume would be when mods were installed from  .ckan  files but aren't in the upstream repo?
```
yes, that was the main test case.
```
In that case I would expect these to still be  CkanModule s (ie, have full metadata), rather than the base class (which AFAIK is really only for autodetected dlls).
```
no, they are installedmodules. InstalledModules aren't modules iirc, they just have a module property. :(

Core/Relationships/RelationshipResolver.cs
Postremus: done.

pjf:
> no, they are installedmodules. InstalledModules aren't modules iirc, they just have a module property. :(

Gosh dang it, you're right.

Looking at the code, there's almost no difference between the `Module` and `CkanModule` classes any more (`InstalledModule` is a container class), so at some point we really should just merge the two, possibly as part of the work in #949.

I'm reviewing this now. If you're about and have an easy (manual) test case for me, that would be grand. :)

Postremus:
Reproduction Steps:
1) I have following mods installed from the repo. You should install them:
    * Contract Configurator
    * Contract Pack: SCANSat
    * ModularFlightIntegrator
    * Module Manager
    * SCANsat
2) I set my repository to "https://github.com/KSP-CKAN/CKAN-meta/archive/b1a54ad253458a257b79104203df18d300bd12b0.zip". Make sure this is your only repository.
3) Download this .ckan "https://raw.githubusercontent.com/KSP-CKAN/CKAN-meta/master/ContractConfigurator-KerbalAircraftBuilders/ContractConfigurator-KerbalAircraftBuilders-1.1.ckan"
4) This CC Config should not exists with the repo you previously set.
5) install the .ckan
6) Go over to your "Installed" Filter.
7) untick the mods, one mod at the time.
8) enjoy some kraken I can't remember currently

pjf:
@Postremus : You rock! Testing now! :)

pjf:
Hmm, I suspect we're seeing `X` in the mod listing for things that are installed by *unknown* upstream, but that's not related to this (I'm still on master). But so I can note what mods I have for testing:

```
$ ckan.exe list

KSP found at /home/pjf/.steam/steam/SteamApps/common/Kerbal Space Program

KSP Version: 1.0.4

Installed Modules:

* ContractConfigurator 1.6.6
- ContractConfigurator-ContractPack-SCANsat v0.5.2.1
- ContractConfigurator-KerbalAircraftBuilders 1.1
* ModularFlightIntegrator 1.1.1
* ModuleManager 2.6.7
X SCANsat v14.1
X Toolbar 1.7.9

Legend: -: Up to date. X: Incompatible. ^: Upgradable. ?: Unknown 
```

I then update my repo, install the .ckan file, and go unticking the mods, and.... nothing happens. They uninstall fine (without patches applied).

Using v1.14.0-31-geb866a1 .

I'm going to try to follow the instructions again harder, but I can't seem to get the kraken to appear. :/

pjf:
New attempt to reproduce:

1. Download @Postremus' registry file at http://puu.sh/jLgvp/993d5d772a.json
1. Copy file over my existing `registry.json`.
1. Run `ckan.exe list` to see what CKAN thinks I have installed (see below).
1. Launch GUI (`ckan.exe`)
1. Filter by installed.
1. Check I'm sorted by name.
1. Start unticking mods going down the list (Contract Config, Contract Pack: SCANsat, etc). I leave each mod unticked before going to the next.
1. Nothing crashes. Halp!

Screenshot of everything unticked:

uncrashingkan

Result of `ckan.exe list`:

```
$ ckan.exe list

KSP found at /home/pjf/.steam/steam/SteamApps/common/Kerbal Space Program

KSP Version: 1.0.4

Installed Modules:

* ContractConfigurator 1.6.6
? ContractConfigurator-ContractPack-HACKEDINTO v0.5.2.1
- ContractConfigurator-ContractPack-SCANsat v0.5.2.1
- ContractConfigurator-KerbalAircraftBuilders 1.1
* ModularFlightIntegrator 1.1.1
* ModuleManager 2.6.7
X SCANsat v14.1

Legend: -: Up to date. X: Incompatible. ^: Upgradable. ?: Unknown 
```

pjf:
Why we only see this under Windows, a best guess by @pjf:

* `ModList_CellValueChanged` is an `async void` method.
* `async void` methods have trouble throwing exceptions, as they have no `Task` object in which to stash it. Instead it's [raised directly on the SynchronizationContext](https://msdn.microsoft.com/en-us/magazine/jj991977.aspx) that was active.
* Because no exception is available to those without the elder knowledge of Cthulhu, mono probably drops the exception, whereas .NET does not.

pjf:
Also we can't change the return type because it's used by an event handler, that insists that we have `void` methods only.

Postremus:
Closing. This is fixed in the master at least since 1.14.2

PR #1229, opened 2015-07-02
Performance Enhancements

Postremus:
Profiling was started before clicking on Refresh, and stopped after.

Before my changes:
ss 2015-07-10 at 11 42 35

After my changes:
ss 2015-07-10 at 11 38 26

### DIFF
--- a/GUI/MainRepo.cs
+++ b/GUI/MainRepo.cs
@@ -79,8 +79,6 @@ namespace CKAN
 
         private void PostUpdateRepo(object sender, RunWorkerCompletedEventArgs e)
         {
-            SetDescription("Scanning for manually installed mods");
-
             UpdateModsList(repo_updated: true);
 
             HideWaitDialog(true);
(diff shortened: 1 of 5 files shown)


pjf:
puu.sh seems to be having troubles right now, and I'd love to see the performance graphs! :)

Postremus:
Sorry for them not being available. uploaded them to github.

GUI/MainRepo.cs
```diff
@@ -72,7 +72,6 @@ private void UpdateRepo(object sender, DoWorkEventArgs e)
         private void PostUpdateRepo(object sender, RunWorkerCompletedEventArgs e)
         {
             SetDescription("Scanning for manually installed mods");
```
RichardLake: Shouldn't this line also be removed?

Postremus: Why?

Postremus: ah, I see the reason :)

Postremus:
rebased the pr.

PR #1237, opened 2015-07-04
Fix for an invalidcastexception

### LINKED ISSUE #1235: exception when double clicking an autodetected mod
I get an exception when double clicking an autodetected mod. I think it is because you aren't checking if it is indeed a checkbox and not the "AD" text which it is for auto detected mods. This should be at https://github.com/KSP-CKAN/CKAN/blob/master/GUI/Main.cs#L659

'''
System.InvalidCastException: Die angegebene Umwandlung ist ungültig.
   bei CKAN.Main.ModList_CellMouseDoubleClick(Object sender, DataGridViewCellMouseEventArgs e)
   bei System.Windows.Forms.DataGridView.OnMouseDoubleClick(MouseEventArgs e)
   bei System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   bei System.Windows.Forms.Control.WndProc(Message& m)
   bei System.Windows.Forms.DataGridView.WndProc(Message& m)
   bei System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
'''

Postremus:
when double clicking on AD mods
fixes #1235

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -655,7 +655,11 @@ namespace CKAN
                 return;
             }
             DataGridViewRow row = ModList.Rows[e.RowIndex];
-            // Need to change the state here, because the user hasn't clicked on a checkbo
+            if (!(row.Cells[0] is DataGridViewCheckBoxCell))
+            {
+                return;
+            }
+            // Need to change the state here, because the user hasn't clicked on a checkbox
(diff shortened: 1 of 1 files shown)


Postremus:
fixes #1248

plague006:
#1252 fixed as well?

PR #1243, opened 2015-07-05
Allow installation into subdirectories of Ships

### LINKED ISSUE #1134: Spec should say that installing to Ships subfolders is fine, but creating Ships subfolders is not.
Currently we're not allowed to install to subfolders of the Ships folder in accordance with the spec. It would be pretty nice if we could do this since a lot of mods package .craft files directly into their .zips or just smacked into a single folder named CRAFT or Ships or something else we cannot handle.

I'm not sure why this restriction was put in from the start but I think it should be removed now.

dbent:
Fixes #1134, allowing `install_to` directives to specify `Ships/VAB` or `Ships/SPH` to install into those subdirectories. We still don't allow mods to create directories within `Ships` though.

Still needs:
- [x] Tests
- [x] Documentation Update

This would require a bump to the `spec_version` to `v1.12`.

### DIFF
--- a/Core/KSP.cs
+++ b/Core/KSP.cs
@@ -252,6 +252,16 @@ namespace CKAN
             return Path.Combine(GameDir(), "Ships");
         }
 
+        public string ShipsVab()
+        {
+            return Path.Combine(Ships(), "VAB");
+        }
+
+        public string ShipsSph()
+        {
+            return Path.Combine(Ships(), "SPH");
+        }
+
         public string Tutorial()
         {
             return Path.Combine(GameDir(), "saves", "training");
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -426,12 +426,28 @@ namespace CKAN
                 installDir = ksp == null ? null : (KSPPathUtils.NormalizePath(ksp.GameData() + "/" + subDir));
                 makeDirs = true;
             }
+            else if (stanza.install_to.StartsWith("Ships"))
+            {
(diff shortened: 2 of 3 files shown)


dbent:
Tests done. Documentation done in #1244.

This isn't as useful as I thought, because I forgot we can't actually rename files/folders. So you either have to specify all the files explicitly or the package has to have the right Ships/(VAH|SPH) structure. It would be nice to have an `install_as` clause so you could do something like:

```json
{
    "spec_version": "v1.x",
    "identifier": "AwesomeMod",
    "install": [
        {
            "file": "ExampleShips",
            "install_as": "Ships/VAB"
        }
    ]
}
```

Then all the contents of `ExampleShips` would be dumped into `Ships/VAB`.

pjf:
Given how often users manage to delete their SPH/VAB/@thumbs directories, we should re-create those if required. (Not a requirement for this PR, but noting it down because it just came up on IRC...)

pjf:
Are the files in `@thumbs` auto-generated by KSP? Right now I presume we allow installation to it (because it exists, and we'll copy things from `Ships`), so it would make sense to allow it to be specified as an install target as well.

pjf:
Aha, `@thumbs` *is* auto-generated by KSP (thanks @Postremus). This is on my review queue. :)

pjf:
Tested on https://github.com/KSP-CKAN/NetKAN/pull/1837 , and seems to be working fine. Merge in progress. :)

pjf:
Merged with great thanks!

PR #1246, opened 2015-07-06
If KerbalStuff doesn't provide a `homepage`, use the KerbalStuff page instead.

plague006:
A fair number of mods go through the NetKAN addition process without homepages. While it isn't the biggest deal in the world it would be nice if we could reliably provide users with a way to get more information than what's available in the abstract.

What I think the code I wrote does: If we have a homepage provided use that. If we don't and we have a KerbalStuff page, turn that into the homepage. (First time writing anything in C#).

### DIFF
--- a/Core/Exporters/DelimeterSeperatedValueExporter.cs
+++ b/Core/Exporters/DelimeterSeperatedValueExporter.cs
@@ -101,7 +101,7 @@ namespace CKAN.Exporters
             {
                 return QuoteIfNecessary(resources.homepage.ToString());
             }
-            else if (resources != null && resources.homepage = null && resources.kerbalstuff != null)
+            else if (resources != null && resources.homepage == null && resources.kerbalstuff != null)
             {
                 return QuoteIfNecessary(resources.kerbalstuff.ToString());
             }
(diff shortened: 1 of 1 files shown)


Core/Exporters/DelimeterSeperatedValueExporter.cs
```diff
@@ -101,7 +101,11 @@ private string WriteHomepage(ResourcesDescriptor resources)
             {
                 return QuoteIfNecessary(resources.homepage.ToString());
             }
-
+            else if (resources != null && resources.homepage = null && resources.kerbalstuff != null)
```
dbent: Need `==` (equality), `=` is assignment.

plague006:
Added GitHub as well.

dbent:
Ah, actually just read your description. Technically your change only affects the export of the mod list to CSVs or TSVs.

To make the change you want to do, I'd suggest making it a change to the UI. The CLI is fine is this regard I think, as it already shows all the relevant resource information.

For the GUI, the relevant code is [here](https://github.com/KSP-CKAN/CKAN/blob/master/GUI/GUIMod.cs#L70) where it determines what to show for the homepage. Right now it just checks if homepage exists and if it doesn't to use "N/A". You'd want to change that to fallback to the Kerbal Stuff URL, if it exists, the repository URL, if it exists, etc.

plague006:
Closing because this is not what I wanted to change.

Issue #1232, opened 2015-07-03
CKAN shouldn't start two KSP instances on double-click


Icemanau:
I've been using CKAN for a short while and one thing I have noticed is that it is possible to have two instance of KSP start if you double click on the Launch KSP button.

Ideally, CKAN should check for an instance of KSP before starting a new one.
If one is found, it could ask if they want to close it or switch to it.
When one is not found, it just starts it.

Otherwise it's a good program to keep track of KSP mods and add ons.
Regards Brad

plague006:
In CKAN, go into the Settings menu and select KSP command-line. Add `-single-instance` as an argument.

#1255 should address the issue for new installs, and maybe down the road someone will write some code that actually has CKAN itself detect if KSP running (which in itself could be useful to stop people trying to delete mods as the game is using them).

Dazpoet:
Issue has been resolved so I'm closing this.

PR #1255, opened 2015-07-06
Change default config to launch KSP with -single-instance

plague006:
Partially addresses #1232. This won't help existing (or Mac) users, but new users moving forward will be able to click the launch button to their heart's content.

### DIFF
--- a/GUI/Configuration.cs
+++ b/GUI/Configuration.cs
@@ -51,9 +51,9 @@ namespace CKAN
                 var configuration = new Configuration
                 {
                     m_Path = path,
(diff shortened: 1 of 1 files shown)


Postremus:
Does not work with already existing instances.

plague006:
That's correct. I noted that in #1232 but should have mentioned that here. Changing the default config for new instances of CKAN does not affect existing instances of CKAN.

plague006:
To bring another point from #1232: in order to prevent multiple instances someone will need to write some code that actually has CKAN itself detect if KSP running (which in itself could be useful to stop people trying to delete mods as the game is using them).

Postremus:
Tested, works. thanks.

PR #1257, opened 2015-07-07
Added "ckan favourite list" as export possibility

### LINKED ISSUE #972: Users should be able to choose between exporting mods as `recommends` or `depends`

Postremus:
closes #972

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -947,6 +947,7 @@ namespace CKAN
             var exportOptions = new List<ExportOption>
             {
                 new ExportOption(ExportFileType.Ckan, "CKAN metadata (*.ckan)", "ckan"),
+                new ExportOption(ExportFileType.CkanFavourite, "CKAN favourite list (*.ckan)", "ckan"),
                 new ExportOption(ExportFileType.PlainText, "Plain text (*.txt)", "txt"),
                 new ExportOption(ExportFileType.Markdown, "Markdown (*.md)", "md"),
                 new ExportOption(ExportFileType.BbCode, "BBCode (*.txt)", "txt"),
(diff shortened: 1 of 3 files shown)


pjf:
You're awesome. Testing this now. :)

GUI/Main.cs
```diff
@@ -961,15 +962,24 @@ private void exportModListToolStripMenuItem_Click(object sender, EventArgs e)
                 Filter = filter,
                 Title = Resources.ExportInstalledModsDialogTitle
             };
-
+            ExportOption f;
```
pjf: Hmm, this doesn't seem to be used anywhere. Can it be removed?

pjf:
Merged with thanks!

Issue #706, opened 2014-12-28
The CKAN registry should be locked while a client is operating.


pjf:
Since we don't expect the registry to change while a client is open, we should lock it (presuming .NET has the equivalent of the unix flock) to prevent multiple access. I'm guessing this should ideally be done in the `Registry.Manager` class, although it might make sense to do it at the start of the client being open as a whole.

pjf:
Marking this as a bug because it's definitely wrong for two clients to be open at the same time.

PR #1265, opened 2015-07-08
Add support for locking the registry.

### LINKED ISSUE #1259: [GUI] Refresh button can cause issues with the registry if another process installs a mod
It's possible to have CKAN install a module and have it become "orphaned" if an unrelated GUI was launched before the installation and the refresh button is pressed.

2 repro methods:

Have a GUI instance open, install a mod via CLI, refresh GUI. The entries in the registry created by CLI will be lost and the installed mod(s) will be "orphaned".

Have 2 GUI instances open: `A` and `B`. Install a mod in instance `A`, once install is complete refresh instance `B`. The entries in the registry created by `A` will be lost and the installed mod(s) will be "orphaned".

mgsdk:
This change creates a file when the RegistryManager loads a registry. Any other attempts at opening the registry will throw a kraken. Should be implemented in both cmd and gui.

Closes #1259 and #706.

### DIFF
--- a/Core/Registry/RegistryManager.cs
+++ b/Core/Registry/RegistryManager.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Text;
@@ -75,6 +76,11 @@ namespace CKAN
             try
             {
                 lockfile_stream = new FileStream(lockfile_path, FileMode.CreateNew, FileAccess.Write, FileShare.None, 512, FileOptions.DeleteOnClose);
+
+                // Write the current process ID to the file.
+                StreamWriter writer = new StreamWriter(lockfile_stream);
+                writer.Write(Process.GetCurrentProcess().Id);
+                writer.Flush();
             }
             catch (IOException)
             {
--- a/Core/Registry/RegistryManager.cs
+++ b/Core/Registry/RegistryManager.cs
@@ -17,9 +17,8 @@ namespace CKAN
 
         private static readonly ILog log = LogManager.GetLogger(typeof (RegistryManager));
         private readonly string path;
-        private readonly string lock_file_path;
-
-        private bool lock_acquired = false;
+        private readonly string lockfile_path;
+        private FileStream lockfile_stream = null;
 
         private readonly TxFileManager file_transaction = new TxFileManager();
 
@@ -38,14 +37,15 @@ namespace CKAN
             this.ksp = ksp;
 
             this.path = Path.Combine(path, "registry.json");
-            lock_file_path = Path.Combine(path, "registry.json.locked");
-            LoadOrCreate();
+            lockfile_path = Path.Combine(path, "registry.json.locked");
 
             // Create a lock for this registry, so we cannot touch it again.
             if (!GetLock())
             {
-                throw new Kraken("Registry is already in use. If this is not the case, then delete the lock file manually." + lock_file_path);
+                throw new RegistryInUseKraken(lockfile_path);
             }
+
+            LoadOrCreate();
 
             // We don't cause an inconsistency error to stop the registry from being loaded,
             // because then the user can't do anything to correct it. However we're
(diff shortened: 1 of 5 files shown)


Core/Registry/RegistryManager.cs
```diff
@@ -49,6 +60,40 @@ private RegistryManager(string path, KSP ksp)
...
+            ReleaseLock();
+        }
+
+        /// <summary>
+        /// Tries to lock the registry by creating a lock file.
+        /// </summary>
+        /// <returns><c>true</c>, if lock was gotten, <c>false</c> otherwise.</returns>
+        public bool GetLock()
+        {
+            if (File.Exists(lock_file_path))
```
pjf: Oooh! This looks like a race condition to me!

* Process A checks to see if the lockfile exists, and discovers it doesn't.
* Process B creates the lockfile
* Process A calls `File.Create`, but it's already there!  Bam!

Further discussion regarding locking incoming on the main PR thread. :)

pjf:
It looks like we're using lockfiles; while that's not a terrible solution, it means we an leave them lying around by accident (which is a pain to users) and there's a chance they can get deleted even when our registry is in use (which is disastrous for us).

On a Unix system, we'd use the `flock()` system call, which allows us to lock an open file. That guarantees that nothing else can take the lock away from us, and means that when our process exits (or even closes the file) the lock is released.

The equivalent under .NET seems to be the [fileshare permission](https://msdn.microsoft.com/en-us/library/system.io.fileshare\(v=vs.110\).aspx), which allows the `FileShare.None` property be expressed, which prevents any other processes (including threads in the same process) from accessing the file. One possible solution would then be:

- Mark the registry as `:IDisposable` if it's not already.
- Open the file using `FileShare.None`, and for reading *and* writing
- Read the contents, and leave the file open.
- When we need to write to the registry, do so using the same filehandle (so nothing else can steal the lock from us). We may need to do a `truncate()` or equivalent first.
- Close the file as part of `Registry.Dispose()`

This should be race-condition free, and means we don't have any risk of leaving a lockfile around, nor any risk of opening a registry that really truly is locked by another process.

Alternatively, if we* do* stick with lockfiles, is there any chance we can write our process-ID to the file when we open it? That means a human can check for weird edge cases (like potentially hung and invisible) and may even allow future CKAN clients to check to see if a lockfile is stale or not.

mgsdk:
I don't have access to my development setup at the moment, but I'll play
around with fileshare permission when I get back home at the start of next
week. Hopefully Mono has a full implementation of this.

mgsdk:
I see my rebase magic is not up to date. This should hopefully make the case for a race-condition less likely, though acquiring the lock is still not atomic. I tested the `FileShare.None` on Mono, but could easily start two instances, thus I'm still relying on creating and deleting files. Instead of doing the check manually and then creating the file, everything should be done in one go now.

Also added a special check in the GUI to show the error dialog before the main window.

mgsdk:
The process ID is the first of the threads, though I'm not sure this is guaranteed every time the code is executed. On my system I have a total of 9 processes registered to the open CKAN instance.

pjf:
> I see my rebase magic is not up to date. This should hopefully make the case for a race-condition less likely, though acquiring the lock is still not atomic. I tested the FileShare.None on Mono, but could easily start two instances, thus I'm still relying on creating and deleting files. Instead of doing the check manually and then creating the file, everything should be done in one go now.

Hmm, I wonder if this is a limitation of mono or something else. One of the requirements of holding a system lock on the file is that the file must remain open, so if we're doing anything to close the registry file while the process is running, then we'll also forfeit our lock on the file.

pjf:
Oh. Locking the registry file itself is likely to cause issues with the transactional file layer¹, so maintaining a separate lockfile *does* make sense. Continuing to review, as stale locks are one of my pet subjects.

---

¹ I wish all systems used transactional filesystems, instead of no systems at all.

pjf:
Closing this in favour of #1357, which is an extension of this work. (@mgsdk, if you're happy with that work, then do feel free to merge!)

PR #1266, opened 2015-07-08
Fix bad characters in filename in module.

### LINKED ISSUE #1249: `ckan show` doesn't quite show filenames accurately.
`ckan show ModName` displays information about a mod, including what name we'd cache it under. However the filename it displays does not always match what we'd use. For example:

    $ ckan.exe show OuterPlanetsMod | grep Filename

    Filename: 9B1B180E-OuterPlanetsMod-1:1.7.zip

Note the `:` from the epoch. However when we write to disk, it's saved as `9B1B180E-OuterPlanetsMod-1-1.7.zip`, as troublesome characters get re-written to dashes.

Expected behaviour: The `ckan.exe show` filename should match what we actually use on disk.

mgsdk:
@pjf reported a slight issue with show in the cli in #1249. This should fix that by moving the renaming code further up the chain. I cannot test with the case @pjf used, since I can't seem to locate it in my registry.

Closes #1249.

### DIFF
--- a/Tests/Core/Cache.cs
+++ b/Tests/Core/Cache.cs
@@ -56,8 +56,8 @@ namespace Tests.Core
             FileAssert.AreEqual(file, cached_file);
         }
 
-        [Test, TestCase("cheesy.zip","cheesy.zip"), TestCase("Foo-1:2.3","Foo-1-2.3"),
-            TestCase("Foo-1:2:3","Foo-1-2-3"), TestCase("Foo/../etc/passwd","Foo-..-etc-passwd")]
+        [Test, TestCase("cheesy.zip","cheesy.zip"), TestCase("Foo-1-2.3","Foo-1-2.3"),
+            TestCase("Foo-1-2-3","Foo-1-2-3"), TestCase("Foo-..-etc-passwd","Foo-..-etc-passwd")]
         public void NamingHints(string hint, string appendage)
         {
             Uri url = new Uri("http://example.com/");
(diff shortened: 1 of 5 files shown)


mgsdk:
I've butchered one of the testcases now, but not removed it. I don't know if I should remove it, or leave it in, in case we need to re-add it in the future.

Tests/Core/Cache.cs
```diff
@@ -56,8 +56,8 @@ public void StoreRetrieve()
             FileAssert.AreEqual(file, cached_file);
         }
 
-        [Test, TestCase("cheesy.zip","cheesy.zip"), TestCase("Foo-1:2.3","Foo-1-2.3"),
-            TestCase("Foo-1:2:3","Foo-1-2-3"), TestCase("Foo/../etc/passwd","Foo-..-etc-passwd")]
+        [Test, TestCase("cheesy.zip","cheesy.zip"), TestCase("Foo-1-2.3","Foo-1-2.3"),
+            TestCase("Foo-1-2-3","Foo-1-2-3"), TestCase("Foo-..-etc-passwd","Foo-..-etc-passwd")]
```
pjf: Just to confirm, we've changed these tests because the assumption is that we'll always pass hints into `cache.Store` which don't result in invalid filenames?

That seems reasonable, although I might pop an assert into the code just to make it clear this is the case. :)

pjf:
Merged with an `Assert` to document our assumptions regarding what gets passed to the `Store` method. :)

PR #1270, opened 2015-07-09
Small GUI overhaul. Added compatibility column and to right pane, removed homepage column, resized all columns.

plague006:
Most up to date version: http://imgur.com/a/6oJCS#0

This will allow users to see the highest version of KSP that a mod is compatible with. It lists the information as a column in the main mod listing, and as a field in the mod info pane. I removed the homepage column because it wasn't functional, and also an impractical way to view that information. All columns are resized and auto-resize is disabled. I'm running windows and tested at 1080 as well as 720 and it looked the same. I don't have the ability to check how it looks under other OSes.

All the column width changes can be reverted simply with a revert of d630fee if needed.

Test cases specifically for compatibility addition:
ksp_version set long (1.0.4): [X]Science
ksp_version set short (1.0): Action Groups Extended
ksp_version_max (1.0.99): AGroupOnStage
ksp_version_min with no max (0.25): 6S Service Compartment Tubes
no version information: Active Texture Management - X86

Potential issue I wasn't able to test for:
If an installed mod has a new version, will it show the compatibility for the current version or the upgrade?

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -313,7 +313,7 @@ namespace CKAN
             CurrentInstanceUpdated();
 
 
-            ModList.AutoResizeColumns(DataGridViewAutoSizeColumnsMode.AllCells);
+            //ModList.AutoResizeColumns(DataGridViewAutoSizeColumnsMode.AllCells);
             Text = String.Format("CKAN {0} - KSP {1}  --  {2}", Meta.Version(), CurrentInstance.Version(),
                 CurrentInstance.GameDir());
             KSPVersionLabel.Text = String.Format("Kerbal Space Program {0}", CurrentInstance.Version());
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -313,7 +313,6 @@ namespace CKAN
             CurrentInstanceUpdated();
 
 
-            //ModList.AutoResizeColumns(DataGridViewAutoSizeColumnsMode.AllCells);
             Text = String.Format("CKAN {0} - KSP {1}  --  {2}", Meta.Version(), CurrentInstance.Version(),
                 CurrentInstance.GameDir());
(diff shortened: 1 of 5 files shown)


plague006:
If anyone has a better column heading than "Highest compat. KSP version" I'd love to hear it, I'm really not satisfied with what I have.

Murdabenne:
Min KSP Version and Max KSP Version  or perhaps a single column slit into 2 like a spreadsheet, title at top is 
| KSP Versison SUpported |
|       Min        |       max      |
     1.0.2               1.0.4

plague006:
It's only the highest compatible KSP version.

GUI/Main.cs
```diff
@@ -313,7 +313,7 @@ protected override void OnLoad(EventArgs e)
             CurrentInstanceUpdated();
 
 
-            ModList.AutoResizeColumns(DataGridViewAutoSizeColumnsMode.AllCells);
+            //ModList.AutoResizeColumns(DataGridViewAutoSizeColumnsMode.AllCells);
```
Postremus: Please remove the line entirely.

Postremus:
Tested, works. Thanks!

pjf:
Right now this is showing me the maximum KSP version *for the version of the mod the user has installed*. So if the user is on KSP 0.90 using RealismOverhaul, they'll be told the max KSP version is 0.90, even though there's a 1.0.4 release of RO available.

I'm writing a patch show the latest available mod info will be shown (1.0.4 in the example above), and have confirmed with @plague006 this is the intended behaviour. Here's what it currently looks like:

aje

Note that it informs the user that a newer mod version will be required in the `Max KSP` field if required. If the *current* version of a mod is also the latest version, this label is skipped.

plague006:
Love it, definitely what I had intended.

PR #1278, opened 2015-07-12
Stop columns from autoresizing after refresh

plague006:


Postremus:
Tested, works. Thanks!

PR #1290, opened 2015-07-15
BackgroundWorkers merge with adjustments.

pjf:
This is a conflict-resolved version of #1131. I'm not as familiar with the GUI code as I'd like, so I'd appreciate an extra set of eyes. @Postremus, may I call on your expertise?

Closes #1131.

Many thanks!

~ Paul

### DIFF
--- a/GUI/MainRepo.cs
+++ b/GUI/MainRepo.cs
@@ -30,7 +30,6 @@ namespace CKAN
 
             tabController.RenameTab("WaitTabPage", "Updating repositories");
 
-            CurrentInstance.ScanGameData();
 
             try
             {
--- a/GUI/MainRepo.cs
+++ b/GUI/MainRepo.cs
@@ -25,8 +25,8 @@ namespace CKAN
 
         public void UpdateRepo()
         {
-            var old_dialog = m_User.displayYesNo;
-            m_User.displayYesNo = YesNoDialog;
+            var old_dialog = currentUser.displayYesNo;
+            currentUser.displayYesNo = YesNoDialog;
 
             tabController.RenameTab("WaitTabPage", "Updating repositories");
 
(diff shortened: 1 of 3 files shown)


GUI/MainRepo.cs
```diff
@@ -48,25 +54,8 @@ public void UpdateRepo()
 
         private void UpdateRepo(object sender, DoWorkEventArgs e)
         {
-            try
```
pjf: @Postremus reports we don't catch 404s any more, and I'm guessing the removal of this try block may be why.

Postremus: We actually catch them, but we are presenting the user now with the whole WebException, which is a step backward.

pjf:
Old PR request clear time!

- 85b9fab: If we still have a post-repo update, then we want this. I need to check if we still have two; I could have sworn we got rid of one.

- 8aa8abd: We do want this (restoring yes/no handlers)! It's good coding practice.

- b0fa443: I seem to understand what's going on; we're doing more error checks, and we're doing more of them in the `PostUpdateRepo` code. Changes to `MainInstall` seem sensible and straightforward, but I'm concerned the changes in `MainRepo` removes some catch blocks.

I haven't seen @RichardLake in a while (we miss you, Richard!), but I'm going to cherry-pick 8aa8abd immediately. b0fa443 I don't feel comfortable installing as it is without lots of testing, and the line notes from back in July seem to imply we may have some regressions in error handling.

Feedback and thoughts *very* welcome. I know I've been stalling on this ticket for months because I haven't understood when we hit the error it's supposed to fix.

politas:
Speaking of old PR clearing, @HebaruSan, this PR conflicts with some changes you have done. Can you have a look at it? I'm not entirely sure what this change even does.

HebaruSan:
Many parts of this were done in later commits, and the rest of it seems to be adding handling for errors and cancellation for the background jobs that install mods and update the registry. I have no idea how necessary or desirable it would be to merge this, but I've force-pushed what should be a clean rebase of it.

Issue #1238, opened 2015-07-05
One unreachable repository breaks list of multiple repositories


dertroglodyt:
If you have multiple Metadata repositories refreshing breaks if one of them is unreachable.
On start of CKAN everything is fine. After refreshing (and waiting for some minutes) the list of available mods only contains the one that are installed.

Removing the unreachable repository returns CKAN to normal behavior.

Postremus:
Even the cli crashes with:

```
Unbehandelte Ausnahme: System.Net.WebException: Die Verbindung mit dem Remoteser
ver kann nicht hergestellt werden. ---> System.Net.Sockets.SocketException: Ein
Verbindungsversuch ist fehlgeschlagen, da die Gegenstelle nach einer bestimmten
Zeitspanne nicht richtig reagiert hat, oder die hergestellte Verbindung war fehl
erhaft, da der verbundene Host nicht reagiert hat 128.199.40.123:80
   bei System.Net.Sockets.Socket.DoConnect(EndPoint endPointSnapshot, SocketAddr
ess socketAddress)
   bei System.Net.ServicePoint.ConnectSocketInternal(Boolean connectFailure, Soc
ket s4, Socket s6, Socket& socket, IPAddress& address, ConnectSocketState state,
 IAsyncResult asyncResult, Exception& exception)
   --- Ende der internen Ausnahmestapelüberwachung ---
   bei System.Net.WebClient.DownloadFile(Uri address, String fileName)
   bei CKAN.Net.Download(String url, String filename, IUser user) in c:\Users\Ma
rtin\Documents\Visual Studio 2013\Projects\CKAN\Core\Net\Net.cs:Zeile 110.
   bei CKAN.Net.Download(Uri url, String filename, IUser user) in c:\Users\Marti
n\Documents\Visual Studio 2013\Projects\CKAN\Core\Net\Net.cs:Zeile 32.
   bei CKAN.Repo.UpdateRegistry(Uri repo, Registry registry, KSP ksp, IUser user
, Boolean clear) in c:\Users\Martin\Documents\Visual Studio 2013\Projects\CKAN\C
ore\Net\Repo.cs:Zeile 134.
   bei CKAN.Repo.UpdateAllRepositories(RegistryManager registry_manager, KSP ksp
, IUser user) in c:\Users\Martin\Documents\Visual Studio 2013\Projects\CKAN\Core
\Net\Repo.cs:Zeile 89.
[...]

PR #1291, opened 2015-07-15
Catch WebException when downloading metadata

Postremus:
We also continue with downloading metadata from other repositories
fixes part of #1238

### DIFF
--- a/Core/Net/Repo.cs
+++ b/Core/Net/Repo.cs
@@ -131,7 +131,16 @@ namespace CKAN
         {
             log.InfoFormat("Downloading {0}", repo);
 
-            string repo_file = Net.Download(repo);
+            string repo_file = String.Empty;
+            try
+            {
+                repo_file = Net.Download(repo);
+            }
+            catch (System.Net.WebException)
+            {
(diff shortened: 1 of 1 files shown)


pjf:
On Linux/mono 3.2.8 and no internet connectivity hitting "refresh metadata" results in all metadata being lost with this change (no message or anything).

Having said that, we might have the same problem *without* this change, I haven't tested. :)

Postremus:
Yup, same thing happens without this change. Also when one repo failed to update, all would fail before.
This change at least continues to refresh on failure.

Core/Net/Repo.cs
```diff
@@ -131,7 +131,16 @@ internal static void UpdateRegistry(Uri repo, Registry registry, KSP ksp, IUser
         {
             log.InfoFormat("Downloading {0}", repo);
 
-            string repo_file = Net.Download(repo);
+            string repo_file = String.Empty;
```
RichardLake: Why initialise it to a value?

Postremus: You are right, this isn't needed in csharp. That initialization should be removed.

PR #1294, opened 2015-07-16
Consolidate network download code

dbent:
This consolidates much of our network download logic into `Net`, and standardizes behavior such that:

- We always try to use a `WebClient` first
- If that fails, we attempt to use cURL

Also, moves the cURL CA file resolution logic into `Curl` so that everything that uses it benefits from it.

### DIFF
--- a/Core/Net/Curl.cs
+++ b/Core/Net/Curl.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.IO;
+using System.Linq;
+using System.Reflection;
 using CurlSharp;
 using log4net;
 
@@ -11,8 +13,8 @@ namespace CKAN
     /// </summary>
     public static class Curl
     {
-        private static bool init_complete = false;
-        private static readonly ILog log = LogManager.GetLogger(typeof (Curl));
+        private static bool _initComplete;
+        private static readonly ILog Log = LogManager.GetLogger(typeof (Curl));
 
         /// <summary>
         /// Has libcurl do all the work it needs to work correctly.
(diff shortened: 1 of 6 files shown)


pjf:
Either I'm mis-reading the logs, or KerbalStuff is working fine with the default `WebClient` under Linux, which it didn't use to do and was one of the big motivators for having curlsharp support in the first place.

Trying to find more ways to test this. :)

dbent:
I was slightly confused as well. What the code used to do was *always* try with CurlSharp, then catch a `DllNotFoundException` and then re-attempt with `WebClient`. So it is using CurlSharp on Linux, but silently failing on Windows (unless you have cURL installed) and being retried.

dbent:
And to add, my changes standardizes the behavior so that it always attempts with `WebClient` first, then falls back to CurlSharp.

pjf:
Looks like kerbalstuff [supports a number of TLS 1.0](https://www.ssllabs.com/ssltest/analyze.html?d=kerbalstuff.com&s=198.41.204.205&latest) ciphers now, and I'm *sure* their absence was the reason we jumped to libcurl in the first place.

So I'm still testing, trying to find something that forces a failover to libcurl. :)

pjf:
This seems to be working fine with me forcing errors. My only real concern is if mono's `WebClient` class ever results in a timeout due to bugs, which my head says shouldn't happen, but which my gut says has already happened.

However I can't find evidence to support my gut, so I'll be merging this shortly. :)

PR #1295, opened 2015-07-16
GUI: When showing Latest KSP info, do it for the latest mod (for any KSP ver)

pjf:
Updates #1270.

Includes human-readable MaxKSP functionality in `Module.cs`.

I wasn't sure if flipping `UpdateModInfo` to use a `GuiMod` was the
right idea, but it seems to work, so that's awesome!

I would *love* someone to check this over. It's an improved version of #1270, but I'm not well versed in how the GUI works.

### DIFF
--- a/Core/Types/Module.cs
+++ b/Core/Types/Module.cs
@@ -309,6 +309,33 @@ namespace CKAN
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


plague006:
It's an awesome extension and works as expected. :+1:

PR #1299, opened 2015-07-16
More Robust AVC Version Handling

dbent:
This PR makes the handling of version information found in internal `.version` files more robust. In practical experience the internal `.version` files aren't remarkably accurate, so we make the two following changes:

- If the metadata already contains a `version` field, such as from KerbalStuff or a GitHub tag, we **do not override** it with the value found in the `.version` file.
- KSP version information is calculated by "merging" the values that are already in the metadata and what we find in the `.version` file. A common example would be if some mod on KerbalStuff indicates that it's compatible with a version of KSP later than what it specifies in its `.version` file. In that case we calculate a minimum version from the `.version` file and a maximum version from the KerbalStuff metadata.
- Add support for "build" component in AVC version, only added if present.

The `CorrectlyCalculatesKspVersionInfo` test contains a lot of cases. Please add any additional cases you think requires testing.

### DIFF
--- a/Netkan/Transformers/AvcTransformer.cs
+++ b/Netkan/Transformers/AvcTransformer.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 using CKAN.NetKAN.Extensions;
 using CKAN.NetKAN.Model;
 using CKAN.NetKAN.Services;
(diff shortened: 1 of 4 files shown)


pjf:
You are the best.

JonDahm:
Would it be possible to indicate that we'd like for the AVC version to overwrite the GitHub version? I don't think the version tags I've been using in my project on GitHub aren't quite CKAN friendly (ex: "v0.1.1-alpha", rather than "0.1.1"), and it'd be nice if I could correct this via the AVC file I'm creating rather than recreating all the release tags.

plague006:
@JonDahm For what it's worth we see quite a bit of `v` coming and going from one version to the next but thankfully have tools to work around that.

JonDahm:
Just to be clear, you're saying that versions formatted with a v in front of the number work just as well as those without? If so, that's great and I don't have to worry about the version numbers comparing correctly.

plague006:
It's not automatic, we have to add a flag and that flag automatically adds v to every version (so they're uniform). Because it's manual it means someone has to 'catch' that the v is inconsistent. If you have mods that you know use v intermittently just let us know and we can correct it.

JonDahm:
Ah, okay. But so long as I use the v consistently, there shouldn't be a problem, correct?

plague006:
Correct.

dbent:
And if you really want to strip the `v` or other misc parts, #1321 will be an option.

JonDahm:
Got it. Thanks a bunch!

Netkan/Transformers/AvcTransformer.cs
```diff
@@ -52,26 +55,80 @@ public Metadata Transform(Metadata metadata)
...
+                        kspMins.Add(avcKspMin);
+
+                    if (existingKspMax != null)
+                        kspMaxes.Add(existingKspMax);
+
+                    if (avcKspMax != null)
+                        kspMaxes.Add(avcKspMax);
+
+                    var kspMin = kspMins.Any() ? kspMins.Min() : null;
+                    var kspMax = kspMaxes.Any() ? kspMaxes.Max() : null;
```
pjf: This is really clever, thank you! :)

pjf:
@dbent : I'm so sorry this has been hanging so long, I'm still playing catch-up after being overseas, but this looks *fantastic*. Merging with great thanks!

PR #1304, opened 2015-07-16
Sort Properties on Final Netkan Output

dbent:
The order of properties in the final `.ckan` output of Netkan is currently dependent on the order in which code is executed by the various transformers. This PR adds a new `ITransformer`, `PropertySortTransformer` that is added to the very end of the pipeline.

`PropertySortTransformer` has an internal list of known properties and will sort the properties in a well defined order (unknown properties are placed just before the `x_generated_by` property in alphabetical order, with `x_` properties always sorting last).

This means the output of Netkan should remain relatively stable with all sorts of changes.

Sort order was chosen based on what I feel is an intuitive reading order when looking at `.ckan` files, and mirrors the typical order found in handcrafted `.ckan` files.

### DIFF
--- a/Netkan/CKAN-netkan.csproj
+++ b/Netkan/CKAN-netkan.csproj
@@ -78,6 +78,7 @@
     <Compile Include="CmdLineOptions.cs" />
     <Compile Include="Transformers\OptimusPrimeTransformer.cs" />
     <Compile Include="Transformers\VersionEditTransformer.cs" />
+    <Compile Include="Transformers\PropertySortTransformer.cs" />
     <Compile Include="Transformers\VersionedOverrideTransformer.cs" />
     <Compile Include="Transformers\StripNetkanMetadataTransformer.cs" />
     <Compile Include="Transformers\EpochTransformer.cs" />
--- a/Netkan/Transformers/NetkanTransformer.cs
+++ b/Netkan/Transformers/NetkanTransformer.cs
@@ -38,7 +38,8 @@ namespace CKAN.NetKAN.Transformers
                 new DownloadSizeTransformer(http, fileService),
                 new GeneratedByTransformer(),
                 new OptimusPrimeTransformer(),
-                new StripNetkanMetadataTransformer()
+                new StripNetkanMetadataTransformer(),
+                new PropertySortTransformer()
             };
         }
 
--- a/Netkan/Transformers/PropertySortTransformer.cs
+++ b/Netkan/Transformers/PropertySortTransformer.cs
@@ -39,6 +39,17 @@ namespace CKAN.NetKAN.Transformers
             { "x_generated_by", int.MaxValue }
         };
 
+        private static readonly Dictionary<string, int> ResourcePropertySortOrder = new Dictionary<string, int>
+        {
+            { "homepage", 0 },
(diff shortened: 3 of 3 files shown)


plague006:
This time in the right PR: Tried it out with 15ish netkans with no technical errors. But for the sorting.. would it be possible to have the download size at the bottom and download link right above it? I'm really used to that from all the existing .ckans

techman83:
Just be prepared for a netkan commit deluge when the changes go through!

dbent:
> Just be prepared for a netkan commit deluge when the changes go through!

True story, but then after this future changes will be less likely to produce such mass updates, woo.

Dazpoet:
Am I reading this correctly that all comments will now be placed at the top (line 2) in the .ckan files? What about examples like [this](https://github.com/KSP-CKAN/CKAN-meta/blob/6a58dfbfe1f67fba63ebed66d3df85212cecfd56/ModularComputerPackage/ModularComputerPackage-1.0b.ckan) where the comments placement is related to its information?

plague006:
Nested comments remain nested (ex. comments in `resources` or `install` stay inside)

Dazpoet:
Ah makes sense :) I'll admit that my codereading skills are <=0 but this looks good to me and might finally make things formatted in similar ways which would make my ocd very happy

Dazpoet:
Did some testing with some extremely malformed versions of our indexed CommunityResourcePack.netkan file and everything seems to work as it should. Having multiple non-nested versions of the same field in the same file leads to only the one closest to the bottom being listed in the created .ckan. I assumed comments to be the only field where such a thing could possibly be and only found three such cases and one is in a handwritten .ckan which is fixed by https://github.com/KSP-CKAN/CKAN-meta/pull/744

The other two are in [CacEye2.netkan](https://github.com/KSP-CKAN/NetKAN/blob/63c303f23ae6c9e02bf4132d7e351802ef18cfcf/NetKAN/CactEye2.netkan) and [KSIPlacementServices.netkan](https://github.com/KSP-CKAN/NetKAN/blob/63c303f23ae6c9e02bf4132d7e351802ef18cfcf/NetKAN/KSIPlacementServices.netkan) neither of which should present a problem since the comments are only relevant for the .netkan files either way.

I don't speak code so I'll just trust Travis on that one.

To me the above thing with only keeping the bottom most one of non-nested duplicate fields isn't a showstopper so I'd say this is ready for a merge, however I'll let the clever netkan people know before merging and potentially blowing up ckan-meta :smiling_imp:

Dazpoet:
Discussed on irc and the above is not a showstopper so I'm merging this. @techman83 might see a gigantic thing pushed by netkan-bot as an effect if I understand the discussion in this PR correctly.

techman83:
@Dazpoet hahaha, yep! [ALL THE COMMITS!!](https://github.com/KSP-CKAN/CKAN-meta/compare/b88c5d1e9e64a7eeb2e7c79de7e81aff82676d3f...0cf23492cab0fef19e81635af94943b235457df2)

techman83:
And an excerpt from the bot log -> https://gist.github.com/techman83/7cda2f560b4d6cd864b5

pjf:
@techman83 : It looks like the bot is very committed!

PR #1305, opened 2015-07-16
Update KerbalStuff Transformer

dbent:
This makes a few small changes to the `KerbalstuffTransformer`:

- Licenses we receive from KS are trimmed, I've seen a couple of cases of otherwise valid license strings being rejected because they had leading or trailing whitespace.
- The specific cases of `GPLv2` and `GPLv3` are transformed into CKAN compatible license strings, since those are two of the default options for licenses on KS.
- The `x_screenshot` property is moved from being a top-level property to a child of `resources`, which I think makes more sense.


plague006:
Tried it out with 15ish netkans with no technical errors. But for the sorting.. would it be possible to have the download size at the bottom and download link right above it? I'm *really* used to that from all the existing .ckans

dbent:
@plague006 I think that comment was meant for #1304? Done in 23dbb4a.

plague006:
Yes it was. Deleted because it's off-topic.

PR #1308, opened 2015-07-17
Misc Netkan Fixes

### LINKED ISSUE #1307: Strange behaviour from netkan.exe when given a .ckan file to process
So I accidentally had a .ckan run through the most current netkan and netkan didn't reject it. It processed it, downloaded the file to cache* and spat out a perfectly viable .ckan of its own.

I expected it to throw up an error, but under the assumption it's intentional: the cache file it downloaded had an unexpected name `F68501B5-netkan-KerBalloons` (no extension).

.ckan used in test:
```
{
    "spec_version": 1,
    "identifier"  : "KerBalloons",
    "name"        : "KerBalloons",
    "abstract"    : "Adds balloons to KSP.",
    "author"      : "JoePatrick1",
    "version"     : "1.1",
    "ksp_version" : "1.0.2",
    "license"     : "restricted",
    "resources": {
        "homepage": "http://forum.kerbalspaceprogram.com/threads/119297"
    },
    "install": [
        {
            "file"       : "KerBalloons",
            "install_to" : "GameData"
        }
    ],
    "download": "http://addons-origin.cursecdn.com/files/2238/252/KerBalloons%20.zip"
}
```

NetKAN XML data from amazonaws (is it built after every commit to master? I have no idea):
```
<Contents><Key>netkan.exe</Key><LastModified>2015-07-16T21:43:18.000Z</LastModified><ETag>"31983db60b9946fe1a4efa11bc3034ca"</ETag><Size>1734144</Size><StorageClass>STANDARD</StorageClass></Contents>
```

Pinging @dbent since netkan is your baby, and I'm assuming no outstanding PRs cover this.

dbent:
- Makes sure `.zip` is appended to downloaded packages
- Logs a warning if the input file does not end with `.netkan`

Fixes #1307

### DIFF
--- a/Netkan/Program.cs
+++ b/Netkan/Program.cs
@@ -143,6 +143,11 @@ namespace CKAN.NetKAN
 
         private static Metadata ReadNetkan()
         {
+            if (!Options.File.EndsWith(".netkan"))
+            {
+                Log.WarnFormat("Input is not a .netkan file");
+            }
+
             return new Metadata(JObject.Parse(File.ReadAllText(Options.File)));
         }
 
--- a/Netkan/Services/CachingHttpService.cs
+++ b/Netkan/Services/CachingHttpService.cs
@@ -35,7 +35,7 @@ namespace CKAN.NetKAN.Services
             EnsureNotDisposed();
 
             return _cache.GetCachedFilename(url) ?? 
-                _cache.Store(url, CKAN.Net.Download(url), string.Format("netkan-{0}", identifier), move: true);
+                _cache.Store(url, CKAN.Net.Download(url), string.Format("netkan-{0}.zip", identifier), move: true);
         }
 
         public string DownloadText(Uri url)
--- a/Netkan/Services/CachingHttpService.cs
+++ b/Netkan/Services/CachingHttpService.cs
@@ -21,7 +21,7 @@ namespace CKAN.NetKAN.Services
 
             CurlSharp.Curl.GlobalInit(CurlInitFlag.All);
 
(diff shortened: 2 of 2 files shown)


mgsdk:
I do not have the tools to check the code where I am now, I just want to
make sure we do not hardcode the zip extension. We can in theory support
multiple archive formats, so reading the ending from either the downloaded
file type (Magic numbers) or from the URI would be preffered in my mind.

This is also a reminder for myself to check the code when I get back.

dbent:
@mgsdk True story. In my world the cached files would just be saved as the URL hash, no extensions, no other identifiers. I know users get some utility from being able to rapidly identify the cache files, although I'm not sure they get much utility from being able to rapidly *open* the cache files, in which case a totally neutral extension of `.cache` or something could work and be far simpler to implement.

plague006:
I find it very handy to be able to open the cache files as I'm adding metadata. With that said, I don't need it to be .zip. As long as *any* extension is present, I can set a default program to open it.

mgsdk:
After checking the code, I would like if we either:
* Add a TODO mentioning that this has a hardcoded extension.
* Detect the file type using magic numbers and set the ending from this.

Seeing as we do not support anything but zip files now anyway, I would be fine with a TODO. Else, take a look at Core/FileIdentifier.cs for identification of files using magic numbers.

dbent:
@mgsdk d4f7bcb selects file extension by the identified type.

mgsdk:
Looks good to me, and now we are future proof (For whenever we get a mod not stored as a zip) :smile:

PR #1312, opened 2015-07-17
Improve window title when asking yes or no

plague006:
"User input required" looks clunky (not user friendly) on a window. We can stick with "CKAN" and it'll look decent for any application that uses this window.

Image showing example


Postremus:
Tested, works. Thanks!

PR #1320, opened 2015-07-18
Add condition in ForceV for upper case V

plague006:
Paging @dbent because it's your baby =)

This change is to prevent situations like https://github.com/KSP-CKAN/CKAN-meta/commit/f35ff62a2e44795f084da6fb8dea1f1e0af8c178


dbent:
Oy, unfortunately version comparisons are not case insensitive. `v` > `V`. That means if we let something that starts with `V` slip by, it'll sort lower than the previous `v` version, which we'd have to remedy by bumping the epoch, no fun.

I think I may have a nice generalizable solution to this and other wacky version problems. Let me think on it a bit.

plague006:
:+1:

PR #1321, opened 2015-07-18
Version Edit Transformer

dbent:
Spun-off of #1320

This adds a new `ITransformer`, `VersionEditTransformer`, that uses regexes (there's no problem that can't be solved with more regexes) to allow simple editing of version strings.

The full example is:

```json
{
    "spec_version": 1,
    "identifier": "AwesomeMod",
    "x_netkan_version_edit": {
        "find": "^[vV]?(?<version>.+)$",
        "replace": "${version}",
        "strict": true
    }
}
```

`find` is a standard .NET regular expression. `replace` is a .NET regular expression [substitution string](https://msdn.microsoft.com/en-us/library/ewy2t5e0%28v=vs.110%29.aspx).

What the example says is "The version string may start with an uppercase or lowercase `v`, in which case edit the version to be just whatever proceeds the `v`". `strict` toggles whether or not the transformer should throw an exception if `find` fails to match on the version string.

The default value of `replace` is `${version}`. The default value of `strict` is `true`. Therefore the above example can be simplified to:

```json
{
    "spec_version": 1,
    "identifier": "AwesomeMod",
    "x_netkan_version_edit": {
        "find": "^[vV]?(?<version>.+)$"
    }
}
```

If `x_netkan_version_edit` is a string instead of an object, its value is treated as the value of `find`. Therefore the example can be simplified down to:

```json
{
    "spec_version": 1,
    "identifier": "AwesomeMod",
    "x_netkan_version_edit": "^[vV]?(?<version>.+)$"
}
```

Because we're using full regular expressions and substitution strings, we can come up with all sorts of wacky subs:

```json
{
    "spec_version": 1,
    "identifier": "AwesomeMod",
    "version": "1.2.3",
    "x_netkan_version_edit": {
        "find": "^(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>\\d+)$",
        "replace": "${patch}.${minor}.${major}"
    }
}
```

would produce:

```json
{
    "spec_version": 1,
    "identifier": "AwesomeMod",
    "version": "3.2.1"
}
```

My recommendation is that we use this to strip wacky version strings down to their basic value, so we don't have to deal with leading-Vs (or lack thereof), code names in version strings, stability identifiers (alpha, beta), KSP version identifiers, or anything else. Using this would very likely lead to an epoch bump for the mod.


techman83:
That's really cool! [I do like a good regex!](https://xkcd.com/1171/)

Will that cause confusion if we are displaying different version information than what the mod is?

dbent:
> Will that cause confusion if we are displaying different version information than what the mod is?

The idea is to "clean-up" the version so that `KSP-1.0.4-beta-0.7.3__SuperDuperAwesome` just becomes `0.7.3`, not to permute the version into something unrecognizable.

The reversing of the version number I did above, was just an example of the flexibility this offers us, not something I'd really want to see anyone do.

pjf:
I'm locked in a room working on slides so I can give a presentation to a few thousand people on future technologies, so I haven't had a chance to test/review this, but holy smokes this is awesome. :)

(Assuming C# doesn't implement `(?{})` (code executing brackets) in regexps, I can't see any reason why we wouldn't want to do this.) :)

dbent:
> Assuming C# doesn't implement (?{}) (code executing brackets) in regexps

It does not. :smiley:

JonDahm:
Definitely all for this. :+1:

techman83:
:+1:  awesome

Dazpoet:
Passes tests, I see nothing that jumps out as totally crazy in the code, tested with https://github.com/KSP-CKAN/NetKAN/pull/1943 which flew through splendidily, everyone seems to be in agreement that this is amazing == merging.

pjf:
@dbent @Dazpoet : Thank you so much!! I'm having the best day playing catch up and discovering all the best things have been merged! <3

PR #1324, opened 2015-07-19
Fix JObject SafeAdd to make sure underlying value is not null

dbent:
@techman83 Fix for KSP-CKAN/NetKAN-bot#27

Problem was that converting a `null` to a `JToken` still produces a `JToken` object, but one that represents a `null` value. This makes sure that if the `JToken` is not `null`, that its underlying object is also not `null`.

Adds a regression test that fails prior to the change and passes after the change.


techman83:
@dbent a fix and a test. Awesome, love your work. LGTM

PR #1325, opened 2015-07-19
wrap the remove, upgrade and installation in a transaction

### LINKED ISSUE #454: GUI: Hitting cancel doesn't cancel an install
Steps to reproduce:

1. Select something to install with lots of dependencies (eg: B9) and which you already have cached (not sure if essential, but I have everything cached).
1. Accept the changes.
1. Hit 'cancel' while it's installing.

Note that the status bar at the bottom will continue to show installation progress.

This is almost certainly my fault, as I've removed the ability to signal to ModuleInstaller that we should cancel a process.

Luckily, we *may* have a good work-around. Because `InstallList` is transactionally aware, one can create a transaction and call `InstallList`. When it returns if the user has cancelled the transaction then we simply do a rollback, and if not we perform a commit.

I have no idea what happens if we exit the transaction *while* `InstallList` is running. However if we can terminate the thread running it and then abort the transaction then we might be fine. (The registry will restore its old state on transaction abort, and any file changes will be cleared up.)

Postremus:
This allows the user to cancel them,
and also reverts the changes made to the filesystem and registry
fixes #454

### DIFF
--- a/GUI/MainInstall.cs
+++ b/GUI/MainInstall.cs
@@ -203,42 +203,53 @@ namespace CKAN
                 installCanceled = true;
             };
 
-            //Set the result to false/failed in case we return
-            e.Result = new KeyValuePair<bool, ModChanges>(false, opts.Key);
-            SetDescription("Uninstalling selected mods");
-            if (!WasSuccessful(() => installer.UninstallList(toUninstall)))
-                return;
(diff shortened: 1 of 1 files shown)


pjf:
Thank you!!

PR #1328, opened 2015-07-20
Update spec.md to clarify the identifier's case sensitivity

### LINKED ISSUE #1119: Is "identifer" case-sensitive?
In the main ckan-repository there are mods providing identifiers which are the same as the identifier of existing mods if case is ignored.
In my opinion it would be strange if "MyMod" is distinct from "mymod" and both coexist in the index. So I would suggest to define identifiers as case-insensitive.

Please define in the Spec if identifiers are case-sensitive.

plague006:
I'm relatively content with how I have it phrased, but I'd be happy to have someone express it better.

Closes #1119

### DIFF
--- a/Spec.md
+++ b/Spec.md
@@ -136,9 +136,15 @@ A short, one line description of the mod and what it does.
 ##### identifer
 
 This is the gloablly unique identifier for the mod, and is how the mod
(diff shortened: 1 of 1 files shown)


techman83:
Clear enough, LGTM.

PR #1329, opened 2015-07-21
Fix NullReferenceException when pressing keys on an empty modlist

Postremus:
Reproduction steps:
1) Search for a mod that does not exists. Make sure your modlist is empty now
2) Focus the modlist
3) press some key.
4) CKAN now should throw a NRE at you

With this fix, CKAN does not throw a NRE at you anymore. :)


pjf:
Reviewed and tested. Merging with great thanks!

PR #1344, opened 2015-07-31
Download default metadata in .tar.gz rather than .zip format

pjf:
At some point, we gained the ability to read `.tar.gz` files, but the client continued to use `.zip` to download metadata. This change flips us to use tarballs, which are *an order of magnitude smaller* than the .zip files. This can make a big difference, especially on slow connections.

This change will affect both new users, and existing users if and only if they're using the default repository settings.

### DIFF
--- a/Core/Registry/Registry.cs
+++ b/Core/Registry/Registry.cs
@@ -168,10 +168,10 @@ namespace CKAN
             // custom-added by our user.
 
             Repository default_repo;
-            var OldDefaultRepo = new Uri("https://github.com/KSP-CKAN/CKAN-meta/archive/master.zip");
-            if (repositories.TryGetValue(Repository.default_ckan_repo_name, out default_repo) && default_repo.uri == OldDefaultRepo)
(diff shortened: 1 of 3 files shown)


Core/Registry/Registry.cs
```diff
@@ -163,6 +163,18 @@ private void DeSerialisationFixes(StreamingContext context)
                 }
             }
 
+            // If we spot a default repo with the old .zip URL, flip it to the new .tar.gz URL
+            // Any other repo we leave *as-is*, even if it's the github meta-repo, as it's been
+            // custom-added by our user.
+
+            Repository default_repo;
+            var OldDefaultRepo = new Uri("https://github.com/KSP-CKAN/CKAN-meta/archive/master.zip");
```
dbent: Local variables are typically always `camelCase`, so `defaultRepo` and `oldDefaultRepo`.

pjf:
@dbent : Thank you so much! Adjusted code to contain more camels. :)

PR #1352, opened 2015-08-07
Green checkmark and clarify the functions of some buttons.

### LINKED ISSUE #1349: "Mark All Updated" is really bad wording.
I only recently discovered that this button is actually "Mark All TO BE UPDATED". I assumed it was to forcefully ignore the files and mark them as being already updated outside CKAN (so I didn't dare touch it!); only when I hit it accidentally did I notice it just checked all the updatable mods - a great feature.

I suggest removing this button and just adding a checkbox in the "Update" menu title which has the same effect, and is familiar to people from other programs.

At the very least, just change it to "Update All" (the user will immediately discover this just checks the mods and they still need to "Go to Changes" (which really should be "Apply Changes" as that is the normal wording of such an operation).

plague006:
Along the lines of https://github.com/KSP-CKAN/CKAN/issues/1349, this updates the text of some buttons, and makes the check mark green. In a *very* scientific study I have determined that check marks should be green. (Citation: [very scientific/official source](https://www.google.com/search?tbm=isch&q=check+mark))

Jenkins gives this PR a green check mark.. think about it.

Closes #1349


Postremus:
Tested, works. Thank you.

PR #1355, opened 2015-08-09
RelationshipResolver: Additional tests against autodetected mods

pjf:
Originally written as part of #1226, this passed even without those
changes. Just additional tests to make sure that autodetected mods can
satisify dependencies, which I didn't want to go to waste.

This commit only consists of test and whitespace changes.


techman83:
LGTM.

PR #1359, opened 2015-08-09
The AutoUpdater now shows that something is being done

### LINKED ISSUE #1347: Auto-update Install action should show progress of download
When I get an auto-update notice on starting CKAN, I click "Install" and... nothing. No indication that it's downloading or that there may be a problem; the interface basically locks up. Has it crashed? is it working away in the background? I just have no clue, until eventually it finishes the download (when it's working). Please add a download progress dialog for the auto-update, or some indication that the mouse click has been accepted at least. In general, when a user clicks on an action, they should see some indication that the application has registered the click and is doing something.

Postremus:
Instead of just doing nothing to indicate the Progress.

Made the autoupdater show the refresh tab on update 
also made the settingsdialog show the release notes before update
Closes #1347

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -414,6 +414,7 @@ namespace CKAN
 
                     ResetProgress();
                     ShowWaitDialog(false);
+                    SetEnabledState(false);
 
                     ClearLog();
                     m_TabController.RenameTab("WaitTabPage", "Updating CKAN");
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -272,21 +272,7 @@ namespace CKAN
                 try
                 {
                     log.Info("Making autoupdate call");
-                    AutoUpdate.Instance.FetchLatestReleaseInfo();
-                    var latest_version = AutoUpdate.Instance.LatestVersion;
-                    var current_version = new Version(Meta.Version());
-
-                    if (AutoUpdate.Instance.IsFetched() && latest_version.IsGreaterThan(current_version))
-                    {
-                        log.Debug("Found higher ckan version");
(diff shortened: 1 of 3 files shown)


Postremus:
fixes #839

politas:
Not seeing any sign that it fixes #1347 yet. Which release should this have gone into?

politas:
The release notes did appear, though, and that is good.

pjf:
Sorry, I've been *terrible* at merging this, because of all the things I'm worried about screwing up, the auto-updater is the one that I'm afraid of breaking the most, so there has always been 'lower hanging' fruit in our PR queue. 

@politas : If you've tested a build with this PR applied then let me know on this ticket. :)

politas:
Ok, so it's in for the next release, now? Which means it can be tested at the time of the release after that.

politas:
I'd love to test this out, but I don't know where to start. It would need at least two builds in order to be able to find a 'newer' release to install, right?

Postremus:
@politas here is a build with that feature enabled:
https://dl.dropboxusercontent.com/u/46071056/CKAN/Debug%20ckan%20%231359.zip

That build will check the releases on my repository for updates, and therefore should find https://github.com/Postremus/CKAN/releases/tag/v1.50.0

politas:
Cool, I'll test it tonight!

politas:
Beautiful. Selecting install made the update window vanish, and brought up the "installing" page with the moving bar. Then the new version launched, taking me back to the KSP Install select dialog.

Thanks, @Postremus! That is a completely successful test.

Postremus:
closing this, since #1692 was merged first.

PR #1362, opened 2015-08-09
Moved duplicate installation code into seperate methods

Postremus:


plague006:
Played around with this for a while and in 200+ mod installs ran into no issues (that aren't in the current master).

pjf:
Thank you so much for this! Merging with great thanks!

PR #1369, opened 2015-08-11
Fixed marking of mods as new if they are uninstalled

### LINKED ISSUE #1368: CKAN marks all uninstalled mods as "New in repository" on auto refresh on startup
This is new behaviour in 1.12.0.

Whenever I refresh my repositories to check for updates, if an update is found CKAN will then mark all uninstalled mods as being new in the repo.

Here is an image of the bug in action.
ckanbug

If there is some extended debug info that would be useful, please specify what you need.  I can't seem to find a log file anywhere for CKAN.

Postremus:
on auto refresh at startup
fixes #1368

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -312,12 +312,12 @@ namespace CKAN
 
             ApplyToolButton.Enabled = false;
 
+            CurrentInstanceUpdated();
+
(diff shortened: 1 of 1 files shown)


pjf:
Tested and reviewed! Merging with thanks!

PR #1384, opened 2015-08-15
Remember the last used filter

### LINKED ISSUE #1377: Make the mod filter the gui is starting with configurable
From Tekener, on the ckan [Thread](http://forum.kerbalspaceprogram.com/threads/100067-The-Comprehensive-Kerbal-Archive-Network-%28CKAN%29-Package-Manager-v1-12-0-30-Jul-2015?p=2131583&viewfull=1#post2131583)
"Is there a way to set the default filter on startup or maybe add a setting for this?
 CKAN is always starting with "Compatible" while I prefer to have "Installed" in order to see which of my used mods need an update."

And a part of another [post](http://forum.kerbalspaceprogram.com/threads/100067-The-Comprehensive-Kerbal-Archive-Network-%28CKAN%29-Package-Manager-v1-12-0-30-Jul-2015?p=2132922&viewfull=1#post2132922) from him for clarification:
"[....]. If some of them are upgradeable I can see it easily as they all fit on one one page. That why I suggested to have the user decide which filter should be the default one after starting."

Postremus:
closes #1377

### DIFF
--- a/GUI/Configuration.cs
+++ b/GUI/Configuration.cs
@@ -17,6 +17,8 @@ namespace CKAN
         public bool RefreshOnStartup = true; // Defaults to true, so everyone is forced to refresh on first start
         public bool RefreshOnStartupNoNag = false;
 
+        public int ActiveFilter = 0;
+
         // Sort by the mod name (index = 2) column by default
         public int SortByColumnIndex = 2;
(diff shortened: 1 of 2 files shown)


pjf:
Tested and works nicely! Merging with thanks!

PR #1392, opened 2015-08-18
We can't update ad mods

### LINKED ISSUE #1387: "Mark all updated" button is active even if there aren't any to update
It used to be that when I did an update and there weren't any mods which needed updating, that this button would be disabled.

Now it is always enabled, even if there is nothing to update

Postremus:
fixes #1387

### DIFF
--- a/Core/Registry/IRegistryQuerier.cs
+++ b/Core/Registry/IRegistryQuerier.cs
@@ -125,6 +125,7 @@ namespace CKAN
 
         /// <summary>
         /// Is the mod installed and does it have a newer version compatible with version
+        /// We can't update AD mods
         /// </summary>
         public static bool HasUpdate(this IRegistryQuerier querier, string identifier, KSPVersion version)
         {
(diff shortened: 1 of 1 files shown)


plague006:
Tested:
1 & 2 AD mod in fresh install.
1 & 2 AD mods in normal install.
1 & 2 AD mods in install with updates available.
Each tested yielded the expected results (Button unavailable, unavailable, and available respectively). In the cases of 2 AD mods I had 1 with a version number and 1 without.

While I happily did the testing and was happy with it, I don't really understand the code so I'd rather someone else give it a once-over and make sure it makes sense.

pjf:
linq

pjf:
Reviewed code, which is adding `!new List<string>(querier.InstalledDlls).Contains(identifier)` to the list of tests (ie: we *don't* have a DLL/AD mod with that name). @plague006 has done testing, so I'm merging this with thanks. :)

PR #1393, opened 2015-08-18
removed the priorities from the repository list in the settings

Postremus:
The list is already sorted, so you won't need to know them


pjf:
Tested, and I can't find any bugs that don't already exist in the code without this patch, so merging with thanks!

Issue #1313, opened 2015-07-17
Searching mods by their abbreviation / alternate names


mongoose11235813:
Is it possible to make a search capable of finding mods by their abbreviations? E.g. Kerbal Attachment System - KAS,  Kerbal Inventory System - KIS and so on.
Some people don't even know/remember whole names of mods because everyone on forums uses abbreviations instead.

plague006:
A full implementation would require a new metadata field (probably inside `resources`) to list alternate names. A much more 'cheaty' (but incomplete) implementation would be adding the alternate names in parentheses to the existing `name` field.

Dazpoet:
One might also consider making the `"provides"` field searchable seeing as some mods acctually have that one. When I first started ckanning I remember thinking it weird that I couldn't search for FAR or NEAR when clearly their metadata include those as alternate names.

HebaruSan:
GUI already does this by pre-calculating the abbreviations:

https://github.com/KSP-CKAN/CKAN/blob/8746077501dcf1edb365230af270c451a0f645e3/GUI/GUIMod.cs#L242-L243

https://github.com/KSP-CKAN/CKAN/blob/8746077501dcf1edb365230af270c451a0f645e3/GUI/MainModList.cs#L745-L750

Closing as fixed by #1430.

PR #1394, opened 2015-08-20
Search mods by their abbreviation

Postremus:
this implementation just splits the name of the mod after every space,
and uses the first letter of every word.
Kerbal Inventory System will be for example KIS
related to #1313

### DIFF
--- a/GUI/MainModList.cs
+++ b/GUI/MainModList.cs
@@ -469,7 +469,10 @@ namespace CKAN
 
         private bool IsNameInNameFilter(GUIMod mod)
         {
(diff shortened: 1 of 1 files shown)


Dazpoet:
Got an artifact from @pjf and tested this hoping it'd choke on things like `(RW` but I failed to break it so I'll merge this based on the fact that I think the code is easy enough for me to understand and review :+1:

Postremus:
Thank you for The fast merge. :)

Von: "Willhelm Rendahl" <notifications@github.com>
Gesendet: ‎20.‎08.‎2015 10:16
Betreff: Re: [CKAN] Search mods by their abbreviation (#1394)

Merged #1394.

PR #1396, opened 2015-08-21
Show latest version of incompatible mod

Postremus:
in the Incompatible filter


plague006:
Tested across multiple installs and it does what's expected: gives a latest version even when mods are incompatible.

Postremus:
Thanks for The quick merge. :)


Gesendet: ‎21.‎08.‎2015 12:35
Betreff: Re: [CKAN] Show latest version of incompatible mod (#1396)

Merged #1396.

Issue #1136, opened 2015-06-21
Add size column


linuxgurugamer:
I have a rather large KSP installation.  It would be nice to be able to see the list of mods sorted by size.  Preferably the size of the unzipped file, but even the size of the zipped file would be an indication of how big it is.

Dazpoet:
fixed in #1399 all hail @Postremus !

PR #1399, opened 2015-08-22
Added a size column

Postremus:
this column shows the download_size in KB
if no download_size is given, it shows 0 KB
implements #1136 

This is how the gui Looks now:
ss 2015-08-22 at 08 02 28


Dazpoet:
When I see a KB column I don't think I'd be sure what it entails. Many people have requested a, near impossible, memory footprint column and this might be misinterperated as such a thing? Is it possible to squeeze in something like "Download (kB)" rather than just KB in the name of the column or does that make it to clunky?

Postremus:
ss 2015-08-22 at 08 46 06

pjf:
I love this so much!!

PR #1401, opened 2015-08-22
Show name and version of the mod we are (un)installing

Postremus:
previously we only showed the identifier,
which can be quit different from the actual mod name


pjf:
You rock! Merged with thanks! :)

Issue #1129, opened 2015-06-20
Refreshing should not wipe list of selected mods


joflashstudios:
The refresh button is quite close to the filter by mod name box. If someone is working on a large changeset and accidentally clicks the refresh button when trying to search, it wipes out their changeset. This can be quite frustrating on large installs, when the user has to go back and re-select 100+ mods again.

And no, of course this didn't happen to me... I'm not that clumsy... :)

pjf:
This is an absolutely solid feature request. Putting it on our list of things to do. Thank you very much!

HebaruSan:
I believe this was done in #2552 (or earlier).

PR #1402, opened 2015-08-22
Refreshing does not wipe the list of selected mods anymore

Postremus:
This changes only the behaviour on a refresh
in every other case, we still do a full refresh of the modlist
for example, the changeset page' clear button needs this
for unticking checkboxes (sigh)
implements #1129


pjf:
Merged with thanks! I've also applied b489c81ecc6fe834b9544552d7838b8ea5049935 which is a documentation-only change for some of the code (if my understanding of how things work is incorrect, then please change it!).

PR #1403, opened 2015-08-22
ckan --verbose" now starts the GUI with verbose debugging

### LINKED ISSUE #437: "ckan --verbose" should start the GUI with verbose debugging
`ckan` without arguments starts the GUI, one would expect that `ckan --verbose` will start the GUI with verbose logging, instead we get the default help message. I expect this is also true for `--debug` and other such arguments.

Postremus:
We just inject gui into the command :)
fixes #437


pjf:
Fabulous! Tested and merging now.

PR #1405, opened 2015-08-22
Don't remove empty ship folders on uninstall 

Postremus:
also check for severall other directories
which should not be removed

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -797,7 +797,9 @@ namespace CKAN
                         if (directory == ksp.Tutorial() || directory == ksp.ShipsVab()
                             || directory == ksp.ShipsSph() || directory == ksp.Ships()
                             || directory == ksp.Scenarios() || directory == ksp.GameData()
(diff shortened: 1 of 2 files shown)


Postremus:
fixes #1191

Core/ModuleInstaller.cs
```diff
@@ -793,8 +793,12 @@ private void Uninstall(string modName)
                     if (!Directory.EnumerateFileSystemEntries(directory).Any())
                     {
                         // Skip Ships/VAB ans Ships/SPH
-                        if (directory == KSPPathUtils.ToAbsolute("VAB", ksp.Ships())
-                            || directory == KSPPathUtils.ToAbsolute("SPH", ksp.Ships()))
+                        string vab = ksp.ShipsVab();
+                        string sph = ksp.ShipsSph();
```
pjf: These two strings never get used!

pjf:
This is excellent. Any chance we can get `@thumbs` protected as well?

Postremus:
@pjf done

PR #1408, opened 2015-08-23
Add a quotes/fortune file to the repo.

pjf:
@Dazpoet suggested it would be fun to keep a list of memorable CKAN quotes. This does exactly that.

If you're on a unix flavoured system, you can `make fortune` and then `fortune ./quotes.txt` to get a random quote.


techman83:
This makes me exceptionally happy :D

PR #1410, opened 2015-08-23
Cache the result from the Autoupdate calls

Postremus:
and also do less of them
the AutoUpdate class is now a singleton


pjf:
Delighted with this, and have submitted #1453 for counter-review (it's all your changes, but with extra tests and bugfix of a very old bug on top).

Issue #1268, opened 2015-07-09
Changeset page: Repurpose the  Description Column 


Murdabenne:
The changeset "Description" column is currently unused.

Propose the following

1) on adds, the mods the the mod-to-be-added depends on are listed (forward dependencies).  This accomplishes the task of informing the end use just which mod is causing the install of the additional mods listed.  Also, changeset line for added dependencies would list the reverse dependency as the reason why it was installed ex: "[modname] added because [mod you are adding] depends on it" (CKAN #589).  Per CKAN #978, also the things that were installed by user request  (not via dependency) get tagged as being independently installed, and flagged against optional removal.

2) on removes, the mods that depend on the mod-to-be-removed are listed (reverse dependencies).  This accomplishes the task of informing the user a list of the mods that this will cause to be removed with it.  Also automatically removed mods listed would have in their description "[modname being autoremoved] removed because it depends on [modname you are removing]". Also, per CKAN #978, the user should be offered an option to uninstall any mods that were installed a a dependency that have no other mods depending on them - but not ones that are flagged as being deliberately installed by the user (those would have to be deliberately removed by the usual way of the user unchecking the box on the main screen).

In the case of something with >5 mods depending on it (modulemanager, firespitter, FAR, etc), the "these mods depend on this" would be too much data to put on a single line of text, so instead the text would say "removing this mod will cause XX other mods to be removed". 

Relationships: 

A depends on B.  This means A has a dependency (needs) on B, and B has a reverse dependency (is needed by) on A.

Examples to clarify (All these are on the changeset screen)
Mods A, B C, and X are as follows:
A depends on B and C
B is independent
C is independent
X depends on B

The user installs X, B is noted as dependency and installed along with it as well as being on the notification line 

The user then installs mod A, which causes mod C to be installed as a dependency, and mod B is updated to have mod A as a reverse dependency. 

Now the user removes mod A - which causes Mod C to be offered for uninstall since the only mod it has as a reverse dependency is Mod A.  B is not offered because removing the reverse for B to A still leaves reverse dependency of B to X

Next the user removes mod X, which causes Mod B to be offered for uninstall since the only remaining reverse dependency is mod X.

reset, have both X and A installed, meaning A B C X are all there.

First, remove X - nothing is offered since A and C are uninvolved, and B still has a reverse dependency B to A after eliminating B to X.

next remove A - both B and C are offered for removal since neither have any reverse dependencies other than the B to A and C to A that are being cut. 

[...]

PR #1412, opened 2015-08-24
Show why a mod was picked for a change

Postremus:
The column "Reason for Action" shows why an action was picked for a mod change.
The columns are sorted by action in the order Remove, Update, Install.
Install changes are also sorted based on dependencies.
This pr implements a part of part 1) of #1268. 

How it looks currently:

ss 2015-08-27 at 06 44 59

### DIFF
--- a/Core/Relationships/RelationshipResolver.cs
+++ b/Core/Relationships/RelationshipResolver.cs
@@ -525,18 +525,23 @@ namespace CKAN
         /// <returns></returns>
         public string ReasonStringFor(Module mod)
         {
+            var reason = ReasonFor(mod);
+            var is_root_type = reason.GetType() == typeof (SelectionReason.UserRequested)
+                || reason.GetType() == typeof(SelectionReason.Installed);
+            return is_root_type
+                ? reason.Reason
+                : reason.Reason + ReasonStringFor(reason.Parent);
+        }
+
+        public SelectionReason ReasonFor(Module mod)
+        {
             if (mod == null) throw new ArgumentNullException();
             if (!ModList().Contains(mod))
             {
                 throw new ArgumentException("Mod " + mod.identifier + " is not in the list");
             }
 
-            var reason = reasons[mod];
-            var is_root_type = reason.GetType() == typeof (SelectionReason.UserRequested)
-                || reason.GetType() == typeof(SelectionReason.Installed);
-            return is_root_type
-                ? reason.Reason
-                : reason.Reason + ReasonStringFor(reason.Parent);
+            return reasons[mod];
         }
     }
 
@@ -544,7 +549,7 @@ namespace CKAN
     /// Used to keep track of the relationships between modules in the resolver.
     /// Intended to be used for displaying messages to the user.
     /// </summary>
-    internal abstract class SelectionReason
+    public abstract class SelectionReason
     {
         //Currently assumed to exist for any relationship other than useradded or installed
         public virtual CkanModule Parent { get; protected set; }
--- a/Core/Relationships/RelationshipResolver.cs
+++ b/Core/Relationships/RelationshipResolver.cs
@@ -615,7 +615,7 @@ namespace CKAN
 
             public override string Reason
             {
-                get { return "  To satisfy dependancy from " + Parent.identifier + ".\n"; }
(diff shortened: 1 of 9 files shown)


Core/Relationships/RelationshipResolver.cs
```diff
@@ -610,7 +615,7 @@ public Depends(CkanModule module)
 
             public override string Reason
             {
-                get { return "  To satisfy dependancy from " + Parent.identifier + ".\n"; }
+                get { return "  To satisfy dependency from " + Parent.name + ".\n"; }
```
pjf: Thank you! :)

GUI/MainChangeset.cs
```diff
@@ -22,22 +22,59 @@ public void UpdateChangesDialog(List<KeyValuePair<GUIMod, GUIModChangeType>> cha
...
+                    description.Text = String.Format("Update to version {0}", change.Mod.LatestVersion);
+                }
 
                 item.SubItems.Add(sub_change_type);
+                item.SubItems.Add(description);
                 ChangesListView.Items.Add(item);
             }
         }
 
+        private void CreateInstallList(IEnumerable<ModChange> changes, ModChange parent=null)
```
pjf: Any chance I can ask for method documentation on `CreateInstallList`? It's not immediately clear to me what's going on without reading the code. (And even then I find myself looking for docs to confirm my understanding.)

GUI/MainChangeset.cs
```diff
@@ -22,22 +22,59 @@ public void UpdateChangesDialog(List<KeyValuePair<GUIMod, GUIModChangeType>> cha
                 return;
             }
 
+            m_Changeset = new List<ModChange>();
+            m_Changeset.AddRange(changeset.Where(change => change.ChangeType == GUIModChangeType.Remove));
+            m_Changeset.AddRange(changeset.Where(change => change.ChangeType == GUIModChangeType.Update));
+
+            IEnumerable<ModChange> left = changeset.Where(change => change.ChangeType == GUIModChangeType.Install);
```
pjf: `left` being everything which is left after we've excluded removes and updates? I feel that would be more clear if our linq-code was everything in `changeset` excluding everything in `m_Changeset`, which means `left` is forward-consistent even if we have new changetypes in the future (eg: a "reconfigure" changetype).

Otherwise I'd be more inclined to call this `install_list` or similar. :)

GUI/ModChange.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CKAN
+{
+    public class ModChange
```
pjf: This is fabulous, thank you.

pjf:
This looks awesome. I've added some stylistic requests as line-notes (which will also help me understand the code a little better). Doing human testing now. :)

GUI/MainChangeset.cs
Postremus: pjf wrote at 8:33 german time:
```
left being everything which is left after we've excluded removes and updates? I feel that would be more clear if our linq-code was everything in changeset excluding everything in m_Changeset, which means left is forward-consistent even if we have new changetypes in the future (eg: a "reconfigure" changetype).

Otherwise I'd be more inclined to call this install_list or similar. :)
```
(Repost of his comment, since i was to dumb and pressed delete instead of cancel..)

plague006:
Had a look at the latest image and poked at the build as well. Everything looks good so far for installs and upgrades. Currently for removal we don't tell users when mods are removed due to broken dependencies. To use an example in the image if a user only selected to remove Contract Configurator it would be nice if we told them why the Contract Packs are being uninstalled.

Postremus:
Will ne done in one of The next pr's. The changeset creation needs to be redone completly first.


Gesendet: ‎27.‎08.‎2015 07:21
Betreff: Re: [CKAN] Show why a mod was picked for a change (#1412)

Had a look at the latest image and poked at the build as well. Everything looks good so far for installs and upgrades. Currently for removal we don't tell users when mods are removed due to broken dependencies. To use an example in the image if a user only selected to remove Contract Configurator it would be nice if we told them why the Contract Packs are being uninstalled.

PR #1415, opened 2015-08-24
NetKAN: Add support for `--version` diagnostic switch.

pjf:
This shows the version number we're built from and exits.


techman83:
LGTM!

PR #1417, opened 2015-08-24
Set (de-)select mods checkbox to a more sensible value when processing recommends

### LINKED ISSUE #1416: In the recommended mods part of installation the (De)-select mods checkbox should default to checked
It's awkward to have users click the button twice. I used it for the first time today and didn't realize I had to double click until the mod was installing.

Dazpoet:
Hoping this is the right way, got some well needed guidance from @plague006 to find the right place to put it :)

closes #1416

### DIFF
--- a/GUI/MainInstall.cs
+++ b/GUI/MainInstall.cs
@@ -443,6 +443,7 @@ namespace CKAN
                     "The following modules have been suggested by one or more of the chosen modules:";
                 RecommendedModsListView.Columns[1].Text = "Suggested by:";
                 RecommendedModsToggleCheckbox.Text = "(De-)select all suggested mods.";
+                RecommendedModsToggleCheckbox.Checked=false;
                 m_TabController.RenameTab("ChooseRecommendedModsTabPage", "Choose suggested mods");
             }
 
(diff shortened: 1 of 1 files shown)


plague006:
Basically a case of Image of two blind men trying to write code

Postremus:
Xd


Gesendet: ‎24.‎08.‎2015 15:08
Betreff: Re: [CKAN] Update MainInstall.cs to fix #1416 (#1417)

Basically a case of

GUI/MainInstall.cs
```diff
@@ -434,6 +434,7 @@ private void UpdateRecommendedDialog(Dictionary<string, List<string>> mods, bool
                     "The following modules have been recommended by one or more of the chosen modules:";
                 RecommendedModsListView.Columns[1].Text = "Recommended by:";
                 RecommendedModsToggleCheckbox.Text = "(De-)select all recommended mods.";
+                RecommendedModsToggleCheckbox.Checked=true;
```
Postremus: Problem in this line: The recommended and suggested page are the same. They are just renamed.
Therefore the Checkbox is also on in the suggested page, although it should be off.

Dazpoet: Shot in the dark here, can I fix that by setting a similar but false line after the else switch below?

Postremus:
Tested, works. Merging with my thanks!

PR #1426, opened 2015-08-28
Added a cached filter

### LINKED ISSUE #1423: Filter by Cached Mods
Having the possibility to set the filter to currently cached mods would make re-installing mods (after a crash-at-launch) scenario easier. This could lead to a possible feature to remove specific mods in the cache (opposed to clearing them all in the settings) via the same tick and apply format currently set up.

Postremus:
this filter only shows already cached mods
Closes #1423

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -24,7 +24,8 @@ namespace CKAN
         NewInRepository = 3,
         NotInstalled = 4,
         Incompatible = 5,
-        All = 6
+        All = 6,
+        Cached = 7
     }
 
     public enum GUIModChangeType
@@ -741,6 +742,11 @@ namespace CKAN
             Filter(GUIModFilter.InstalledUpdateAvailable);
         }
 
+        private void cachedToolStripMenuItem_Click(object sender, EventArgs e)
+        {
+            Filter(GUIModFilter.Cached);
+        }
+
         private void FilterNewButton_Click(object sender, EventArgs e)
         {
             Filter(GUIModFilter.NewInRepository);
@@ -773,6 +779,8 @@ namespace CKAN
                 FilterToolButton.Text = "Filter (Installed)";
             else if (filter == GUIModFilter.InstalledUpdateAvailable)
                 FilterToolButton.Text = "Filter (Upgradeable)";
+            else if (filter == GUIModFilter.Cached)
+                FilterToolButton.Text = "Filter (Cached)";
             else if (filter == GUIModFilter.NewInRepository)
                 FilterToolButton.Text = "Filter (New)";
             else if (filter == GUIModFilter.NotInstalled)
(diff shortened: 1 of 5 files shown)


GUI/Main.cs
```diff
@@ -21,10 +21,11 @@ public enum GUIModFilter
         Compatible = 0,
         Installed = 1,
         InstalledUpdateAvailable = 2,
-        NewInRepository = 3,
-        NotInstalled = 4,
-        Incompatible = 5,
-        All = 6
+        Cached = 3,
```
pjf: I know that we save our filter in as `ActiveFilter` in `GUIConfig.xml`, which meas that re-ordering these means that users may mysteriously have their default filter change if they're coming from an older version.

Is there any chance we can add `Cached` to the end of the `enum` rather than in the middle?

Postremus: I actually forgot we save The current filter (although i wrote The code for it). Will change it.

Von: "Paul Fenwick" <notifications@github.com>
Gesendet: ‎01.‎09.‎2015 09:32
Betreff: Re: [CKAN] Added a cached filter (#1426)

In GUI/Main.cs:
> @@ -21,10 +21,11 @@ public enum GUIModFilter
>          Compatible = 0,
>          Installed = 1,
>          InstalledUpdateAvailable = 2,
> -        NewInRepository = 3,
> -        NotInstalled = 4,
> -        Incompatible = 5,
> -        All = 6
> +        Cached = 3,
I know that we save our filter in as ActiveFilter in GUIConfig.xml, which meas that re-ordering these means that users may mysteriously have their default filter change if they're coming from an older version.
Is there any chance we can add Cached to the end of the enum rather than in the middle?

Postremus:
@pjf done

pjf:
This works great, but I've hit a performance hiccup. I have about 200 mods in my cache, and so it takes a very noticeable amount of time for the client to start as we check each mod to see if it's present.

The delay appears to be in how many times we're hitting the directory structure. When checking the cache, we read the whole directory into memory, and then walk through it seeing if we can spot the file we want. With about 900 mods in the metadata index, and about 200 mods in my cache, that's in the order of O(900×200) = O(180,000) operations.

There's a lesser cost in checking the zip doesn't have any obvious malformations, too.

I've managed to eliminate both of these in a branch I'll be uploading soon. The main gotcha I can see is that we only set `.IsCached` once, and I don't know when the `GUIMod` objects get discarded and rebuilt. (So we may not pick up on someone hitting the "download" button that appears in the contents view, for example.)

In any case, I have code coming soon that may bring joy to all these things. :)

Postremus:
I am not able to reproduce the Performance issues you mention. How many files do you have in your Cache Directory?

Postremus:
@pjf i guees some os related Caching is going on here. starting ckan the first time after a pc reboot is slow with this pr. every next start is fast.

pjf:
@Postremus : I have 774 files in my cache directory (Linux/btrfs). The first run is definitely the slowest, but subsequent runs still have a noticeable slow-down compared to before the filter-by-cache. (Although #1454 fixes this, if you fancy reviewing it!)

pjf:
Merged with a tweak that we use `IsMaybeCachedZip` rather than `IsCachedZip`, which is faster (but doesn't check the zip file is absolutely 100% valid). Thank you!! :)

PR #1427, opened 2015-08-29
Don't show suggested/recommended for mods that can't be installed

Postremus:
We don't show an empty suggested and recommended list if we can't install any of the mods.
Also don't show the recommended / suggested page if the user already cancelled the Installation.

### DIFF
--- a/GUI/MainInstall.cs
+++ b/GUI/MainInstall.cs
@@ -185,11 +185,13 @@ namespace CKAN
                 selectable.Remove(item);
             }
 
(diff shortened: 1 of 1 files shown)


politas:
I'm trying to merge this, and it seems to lose the buildID checking (and possibly other things. Can you re-base this? (I tried rebasing it myself and got some build errors)

Postremus:
done

politas:
Merged with thanks!

PR #1431, opened 2015-08-30
reordering of repos isn't weird anymore

### LINKED ISSUE #1409: GUI: Reordering repos is weird
We seem to order our repos by giving them a number (I don't know why) and having the GUI increase/decrease this number when we hit the up/down buttons in the GUI settings page. However rather than making the number one bigger or smaller than the largest or smallest number, we seem to just increment/decrement it, meaning we can get number conflicts and repos seem to jump around in ordering.

Steps to reproduce: Add at least three repos. Try moving the top one down, or the bottom one up a few times. Observe that it will sometimes "jump" an extra space.

What we should be doing: Not have them jump around. I'm not sure why the numbers are there at all, since lists are ordered in nature anyway.

Discovered in testing of #1393.

Postremus:
fixes #1409

### DIFF
--- a/Core/Types/Repository.cs
+++ b/Core/Types/Repository.cs
@@ -24,6 +24,13 @@ namespace CKAN
             this.uri = new Uri(uri);
         }
 
+        public Repository(string name, string uri, int priority)
+        {
+            this.name = name;
+            this.uri = new Uri(uri);
+            this.priority = priority;
+        }
+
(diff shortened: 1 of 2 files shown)


pjf:
You are awesome.
This works great with me testing as a human, I understand the code, and you are wonderful.
Merging with great thanks!

PR #1432, opened 2015-08-30
The user is allowed to close ckan while applying a changeset

Postremus:
the transaction will make sure nothing is changed if the user clicks on
either cancel or on the X to close the window
this restores the functionality of 1.10.0

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -372,8 +372,11 @@ namespace CKAN
 
         protected override void OnFormClosing(FormClosingEventArgs e)
         {
-            //Only close the window, when the user has access to the "Exit" of the menu
-            if (!menuStrip1.Enabled)
+            // Only close the window, when the user has access to the "Exit" of the menu
+            // and if he is allowed to cancel the current action
+            // This will for example close the windows during an installation
+            // while it will stop the closing during a refresh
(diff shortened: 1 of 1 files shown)


pjf:
I can't test this under Linux, as Linux lets me close anything I feel like. :)

But I built [this executable](https://dl.dropboxusercontent.com/u/9702672/KSP/ckan-v1.14.0.24-g5bc00ad.exe) if I can recruit any Windows users (@Dazpoet?) to try closing the CKAN client and hitting the 'x' button to to close the window (we should cancel the install and return everything to as it was before).

plague006:
Closing during installation of multiple mods will not revert all completed installs, but rather you'll end up with a split: some mods installed and others not. Worse still the mods that have been installed aren't registered in CKAN's registry. http://puu.sh/k37fk/35fefa61b5.png

pjf:
@plague006 : Yikes! That's not what we want. I wonder if 'x' kills our process without the ability for our transaction system to run a rollback¹? Just to confirm, those directories weren't empty; they were actually partially/fully installed mods?

---

¹ While actual OS guaranteed transaction filesystems exist, nobody seems to have them, so our transaction layer still depends upon user-space software to perform operations.

plague006:
They were not empty, though if I recall correctly I didn't check each mod's integrity (one of them might have been partially installed).

Issue #1433, opened 2015-08-30
ckan:// link handling


ddevault:
A while ago I set down a list of concerns for integrating CKAN and Kerbal Stuff more closely together. This would come in the form of a `ckan://` link being added to mod pages for easier installation. I had two concerns:
- I couldn't get `ckan://` to work at all on Linux - the protocol handler installation needs to be consistent and work on all platforms
- Clicking a `ckan://` link will open CKAN but not with the mod in question selected for install. I would like to see it open a window with the mod details and a button to install it, nice and simple.

Postremus:
Any idea how the "window with the mod details and a button to install it" would look?

SirCmpwn:
![asdf](https://cloud.githubusercontent.com/assets/1310872/9567783/1dcbcd22-4f05-11e5-891e-64cbb02e04c2.png)

Mod title on top, brief details along the side, and two buttons - install and cancel.

politas:
Spacedock integration is taking a different path, so I'm closing this.

politas:
@Postremus, if there's something left from this issue that you are working on, please re-open.

Postremus:
The specific installation screen @SirCmpwn is still left todo.

HebaruSan:
Duplicate of #723.

PR #1434, opened 2015-08-30
Fixed protocol handler registration on linux (at least ubuntu)

Postremus:
Tested on Ubuntu 14.04
-If we have no existing mimeapps.list, create it
-start ckan with mono
-make the .desktop file executable
-correctly resolve the home Directory
-focus the mod we got in the modlist
related to #1433

### DIFF
--- a/Core/Net/AutoUpdate.cs
+++ b/Core/Net/AutoUpdate.cs
@@ -137,19 +137,7 @@ namespace CKAN
 
             // run updater
 
-            // mark as executable if on Linux or Mac
-            if (Platform.IsUnix || Platform.IsMac)
-            {
-                // TODO: It would be really lovely (and safer!) to use the native system
-                // call here: http://docs.go-mono.com/index.aspx?link=M:Mono.Unix.Native.Syscall.chmod
-
-                string command = string.Format("+x \"{0}\"", updaterFilename);
-
-                ProcessStartInfo permsinfo = new ProcessStartInfo("chmod", command);
-                permsinfo.UseShellExecute = false;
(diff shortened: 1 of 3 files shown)


SirCmpwn:
Does not work for me here on Arch Linux. I don't get a desktop file and it's not added to mimeapps.list. How do I enable logging?

Postremus:
logging can only be activated with the Cmdline.exe, which will then start the gui:
Cmdline.exe --verbose
or
Cmdline.exe --debug

SirCmpwn:
3926 [1] INFO CKAN.URLHandlers (null) - Trying to register URL handler
    3927 [1] INFO CKAN.URLHandlers (null) - Skipping URL handler: Parsing Error: Could not parse file ~/.local/share/applications/mimeapps.list

My mimeapps.list: https://sr.ht/Xz8G.txt

SirCmpwn:
:+1: works for me, I just made a mistake when testing it.

pjf:
Just checking if f81069e24cbf5cd5a8c5c22d065e5533ae551df1 is intentionally or accidentally on this branch. :)

Postremus:
intentionally, because it also replaces some part of the url handler code.

pjf:
@Postremus : Oh! That makes sense to me now. :) In the process of testing. :)

pjf:
This works for me as well! Although if I've already got the CKAN open, then clicking on a `ckan://` link means I then have *two* (as opposed to the original being signalled somehow to change its focus).

Still, that's way better than what we had before, so I'm giving the code another glance over and then merging. Thank you very much! :)

PR #1436, opened 2015-08-31
Fix a castexception when installing metapackage

Postremus:
The install logic still  used a KeyValuePair<GuiMod, GuiModChangeTyp>
while everything else uses ModChange object's now


pjf:
You rock! :)

PR #1438, opened 2015-09-02
Select full row

Postremus:
this marks the whole row blue
instead of only the cell you clicked
This change is done for the suggested, changeset and tmp list


pjf:
Woah, I had no idea we didn't do this! Merging with thanks!

PR #1442, opened 2015-09-03
Merged CkanModule and Module to a single class

### LINKED ISSUE #1440: Merge CkanModule and Module to a single class
Despite the fact that I was the one who wrote the original `Module` and `CkanModule` classes, I've long been puzzled as to why. After some searching I finally unearthed 9b52fa82454cad484860cfb7627d06833c6377ee, which reveals the reason they're different classes:

> We now differentiate between full modules (`CkanModules` from CKAN files) and merely bundled mods (`Modules`, which do not).

However bundled modules were *removed* from the CKAN spec well before the 1.0 release, meaning that `Module` and `CkanModule` are separate classes *for no reason at all*.

Since these cause a huge amount of bother and confusion, we should merge them into a single class. I suggest we merge them into `CkanModule`, since that's the most functional class.

This will also fix a bug I recently noticed where we'd lose install stanzas in the registry as part of serialising `InstalledModules`. While harmless to our users, it means that we lose the ability to tell exactly what metadata was used for a module's install (which is one of the reasons for stashing it).

Thank goodness for version control letting me find where the classes first split. :)

Postremus:
the spec_field needed some extra Attention. it was not required for module's, but for ckanmodule's..
closes #1440

### DIFF
--- a/Cmdline/Action/Show.cs
+++ b/Cmdline/Action/Show.cs
@@ -120,7 +120,7 @@ namespace CKAN.CmdLine
         /// </summary>
         /// <returns>Success status.</returns>
         /// <param name="module">The module to show.</param>
-        public int ShowMod(Module module)
+        public int ShowMod(CkanModule module)
         {
             #region Abstract and description
             if (!string.IsNullOrEmpty(module.@abstract))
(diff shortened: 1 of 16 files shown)


pjf:
zOMG I LOVE YOU SO MUCH!!!

pjf:
Merged so hard! Thank you! :)

PR #1443, opened 2015-09-04
Handle double click on the contentTree

Postremus:
.net opens the windows explorer
mono uses xdg-open to open the directory


pjf:
This is awesome, works wonderfully on my system, handles all the edge cases I could think of (eg: cached mods), and is a *really neat* addition. Merging with thanks!

PR #1444, opened 2015-09-05
Improve Steam install detection

LarsOL:
If the steam game isn't found in the default spot. Parse the steam config file to get the path.

### DIFF
--- a/Core/KSPPathUtils.cs
+++ b/Core/KSPPathUtils.cs
@@ -84,19 +84,12 @@ namespace CKAN
         }
 
         /// <summary>
-        /// Finds the KSP path under Steam. Returns null if the folder cannot be located.
+        /// Finds the KSP path under a Steam Libary. Returns null if the folder cannot be located.
         /// </summary>
+        /// <param name="steam_path">Steam Libary Path</param>
         /// <returns>The KSP path.</returns>
-        public static string KSPSteamPath()
+        public static string KSPDirectory(string steam_path)
         {
-            // Attempt to get the Steam path.
-            string steam_path = SteamPath();
-
-            if (steam_path == null)
-            {
-                return null;
-            }
-
             // There are several possibilities for the path under Linux.
             // Try with the uppercase version.
             string ksp_path = Path.Combine(steam_path, "SteamApps", "common", "Kerbal Space Program");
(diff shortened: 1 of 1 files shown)


pjf:
Oh cool! I always knew that in theory we could go parsing Steam's config files, but I didn't realise it could be this straightforward. :)

LarsOL:
I am a bit curious (don't take this as criticism i am a new software engineer and just don't understand) but why do pull requests take so long to merge?

You seem to have a Continuous Integration system setup, and from my understanding merging often means dev's branches are less likely to conflict with each other, reducing the admin work.

If it's to do with stability, shouldn't there be a "Dev" branch that is continuous integration, with the "Dev" branch being merged to a release branch when it is deemed stable enough?

Thanks,
Lars

pjf:
> I am a bit curious (don't take this as criticism i am a new software engineer and just don't understand) but why do pull requests take so long to merge?

This is an *excellent* question. And the answer is nothing to do with stability, and all to do with human psychology! While everyone who has a collaboration bit has the ability to merge PRs, I think there's been a bit of a hesitancy to do so. I'm sure there are many reasons for this: merging a PR can make you feel responsible if you've missed any bugs in it, you may find reviewing code less fun than writing code, and the good old fashioned [bystander effect](https://en.wikipedia.org/wiki/Bystander_effect) where it's easy to believe that someone else will do it.

In our case, I suspect it's *mostly* bystander effect. I have [scheduled sprints](https://patreon.com/pjf0) where I work on the CKAN, so it's easy to feel that for any given PR, I'll review and merge it if nobody else does. And that belief is indeed correct, except that we've built up a backlog of PRs as a result, especially in the last couple of months where I've been on the conference circuit and pouring hours into [The Kerbal Book](http://www.amazon.com/Kerbal-Space-Program-Jon-Manning/dp/1491913053?tag=pjf-20) so we can hit our Christmas deadline.

I'd actually love for us to move towards something similar to what [Rust](https://github.com/rust-lang) do, where a *bot* assigns PRs to humans as soon as they come in.

pjf:
Moving discussion to #1481, which is @LarsOL's excellent work with the results of code review on top (ie, a guard to make sure we don't choke on a bad config file). :)

PR #1445, opened 2015-09-05
show progess of cache download

Postremus:
when user pressed the "Download" button
We use the netasyncdownloader for this


pjf:
This is awesome, and *way* better than what we had before!

PR #1447, opened 2015-09-06
New dependencies of upgraded mods are scheduled for install by the GUI

### LINKED ISSUE #795: When an upgrade adds new dependencies, include them in the changeset for users to see

Postremus:
for a upgrade of a mod to the changeset
fixes #795

### DIFF
--- a/GUI/MainModList.cs
+++ b/GUI/MainModList.cs
@@ -315,6 +315,7 @@ namespace CKAN
                 {
                     case GUIModChangeType.None:
                         break;
+                    case GUIModChangeType.Update:
                     case GUIModChangeType.Install:
                         //TODO: Fix
                         //This will give us a mod with a wrong version!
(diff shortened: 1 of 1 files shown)


pjf:
I don't have a handy mod with new dependencies to test on, but the code looks solid, and it lets me upgrade fine with mods that don't have new dependencies, so I'm merging this with great joy.

PR #1448, opened 2015-09-07
Allow installation to the @thumbs folders

### LINKED ISSUE #1404: Create thumbs directory in Ships
I just wanted to install FAR in a minimal dummy install. I had no Ships Folder, and therefore no ```@thumbs``` subdirectory.
Ckan didn't want to automaticly create this directory, and the Installation failed.

Postremus:
fixes #1404

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -442,6 +442,15 @@ namespace CKAN
                     case "Ships/SPH":
                         installDir = ksp == null ? null : ksp.ShipsSph();
                         break;
+                    case "Ships/@thumbs":
+                        installDir = ksp == null ? null : ksp.ShipsThumbs();
+                        break;
+                    case "Ships/@thumbs/VAB":
+                        installDir = ksp == null ? null : ksp.ShipsThumbsVAB();
+                        break;
+                    case "Ships/@thumbs/SPH":
+                        installDir = ksp == null ? null : ksp.ShipsThumbsSPH();
+                        break;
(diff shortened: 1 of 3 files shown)


pjf:
@Postremus : Will this actually fix #1404? My understanding is that FAR would have installed fine if the `@thumbs` directory either existed or CKAN would create it; however this looks like we're allowing `@thumbs` subdirs as targets (with `makeDirs` still set to `false`), which feels like a fairly rare case.

Wouldn't we instead want something that checks all the directories we expect are in place, and makes them if not? (Either as part of the install, or as a separate step.)

Postremus:
@pjf yes, it fixes 1404. yes, a blacklist like System would be much better, especially since we already allow Installation to every directory in the ksp folder.

pjf:
I can't get this to fix #1404, or at least not as I understand it...

Steps to reproduce:

1. Start with a fresh (or fresh enough) KSP install.
2. Delete the `Ships/@thumbs/VAB` folder.
3. Attempt to install FAR:

> `Could not find a part of the path "/home/pjf/.steam/steam/SteamApps/common/Kerbal Space Program/Ships/@thumbs/VAB/FAR Shuttle 'Montauk'.png".`

Re-creating the `@thumbs/VAB` folder has it work fine.

Using 28b2635.

pjf:
While I can't get this to fix #1404 (but now fixed in #1525), this does allow us more flexibility in installing files. Merging with thanks!

PR #1452, opened 2015-09-10
Add logging to file

LarsOL:
Add log4net as a extension in visual studio project file.
Add logging to file "log.txt"


LarsOL:
Used this tutorial: https://stackoverflow.com/questions/9805281/writing-logs-to-file
We may need to tweak the config file, currently at debug lvl it fills up the logs very quickly.
maybe 5MB max file size and 3 log backups?

LarsOL:
Oh something is broken, ill have a look

pjf:
Ahoy, and welcome! Logging is something I adore, so I'm delighted to see this PR! :)

>  We may need to tweak the config file, currently at debug lvl it fills up the logs very quickly.

At debug level we're supposed to produce lots of output, and is most useful when we turn on debug for individual classes (or usually in my case, for individual tests in our test suite).

Currently by default we show the user everything with `warn` and above, although you can start CKAN with the `--verbose` or `--debug` switches to lower that to `info` and `debug` respectively.

I definitely like the ability for us to be able to log to files, although I wouldn't have this start anywhere more detailed than `info`, which will hopefully capture important operations without an immensity of spam.

(I haven't yet looked at the commit itself; I may have more feedback once I've done so.)

Many thanks again!

~ Paul

LarsOL:
Currently it logs to whatever switch is enabled. All i have really done is enable the config file and pointed it to a default one i found in the above link.

Once it is set up properly you should be able to configure everything to do with log4net in that file. From my brief overview the log4net config format is extremely flexible, so it should be a matter of editing that file only for logging changes and not having to edit the code much anymore.

AutoUpdate/CKAN-autoupdate.csproj
```diff
@@ -32,6 +32,10 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="log4net, Version=1.2.13.0, Culture=neutral, PublicKeyToken=669e0ddf0bb1aa2a, processorArchitecture=MSIL">
```
LarsOL: This seems to be the error in mono, It can parse this line (repeated for each project file). But this was generated by Visual Studio so i am hesitant to edit it back without knowing what all of it does? 

Anyone with more expertise than me tell me if its fine to change back to:
`<Reference Include="log4net">`

Olympic1:
Try adding this line in the .csproj files between `<Reference>` and `<HintPath>` and see if that works `<SpecificVersion>False</SpecificVersion>`

ayan4m1:
Lars, I have a rebased branch that contains file logging using a log4net config and some other upstream changes - if you'd like to rebase this and continue working on it, feel free, but if you don't have the time/wherewithal with it or we get the other PR merged in I will close this out to keep things tidy.

PR #1453, opened 2015-09-10
AutoUpdate makes less calls, has better tests.

pjf:
This is an extension of #1410, and includes all of its commits. In addition, it:

- Has more documentation
- Has extra tests
- Fixes a bug where we could crash on malformed release notes
- Changes release notes handling so that we use everything after the first line of dashes.

I'm very happy with @Postremus' work, so I'm happy for this to be merged if anyone (including Postremus) is happy with [my extensions](https://github.com/KSP-CKAN/CKAN/compare/ab836f3...eefbdc8).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ All notable changes to this project will be documented in this file.
 
 ### Internal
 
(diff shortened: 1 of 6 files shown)


Postremus:

PR #1454, opened 2015-09-10
NetFileCache now caches the cache directory so you can cache while you cache.

pjf:
`[YoDawg.jpg]`

In the past, `NetFileCache` would read the entire cache directory each time we checked to see if we had a file. That was okay, because we didn't check very often. However #1426 adds a filter on cached status, and with 900+ potential mods it can be very slow indeed.

This PR implements a number of changes to improve our cache handling, in particular:

- We now cache the list of files we read from the disk, so if we're performing lots of look-ups we are much faster.
- We clear that cache if we ever knowingly change the directory, such as by adding or removing a file.
- We use the `FileSystemWatcher` class to *also* clear the cache, so if anything else modifies the directory we should know about it.

Gotchas:

- `FileSystemWatcher` isn't necessarily reliable under mono. It's supposed to recruit the kernel for help, on my system it doesn't. That's okay, because users can probably deal with "you need to restart the CKAN if you manually mess with the cache under Linux/OSX". (With `MONO_MANAGED_WATCHER=1 mono ckan.exe` it *does* work for me, as this causes mono to emulate the filewatcher with polling.)
- We have to ask for some special permissions in the code, and I haven't checked to see if they cause the user to get an annoying pop-up under Windows. **Please test this** dear Windows users.
- I may have this subscribed to too many events. I'm pretty sure that `NotifyFilters.LastAccess` isn't going to be needed in the end.

Merging this makes #1426 run *oodles* faster, and even moreso if we change it to use `IsMaybecachedZip` (ff24fcefeb0e10ecbe932653c94b18b61bb60761).

Attn @Postremus, as without this change #1426 can be very slow on start-up.

### DIFF
--- a/Core/KSP.cs
+++ b/Core/KSP.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -15,7 +16,7 @@ namespace CKAN
     /// <summary>
     ///     Everything for dealing with KSP itself.
     /// </summary>
-    public class KSP
+    public class KSP : IDisposable
     {
         public IUser User { get; set; }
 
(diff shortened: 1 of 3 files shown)


Core/KSP.cs
```diff
@@ -335,6 +349,9 @@ public KSPVersion Version()
         /// </summary>
         public void CleanCache()
         {
+            // TODO: We really should be asking our Cache object to do the
+            // cleaning, rather than doing it ourselves.
+            
             log.Debug("Cleaning cahce directory");
```
Postremus: Little typo.

Postremus:
Tested, works. thank you!

PR #1460, opened 2015-09-16
Change the displayed text of WaitTabPage

plague006:
For consistency, "Installing mods" doesn't really make sense since we list updates and removals in this tab as well.

Picture of a hippo giving a child a piggyback ride


Postremus:

PR #1464, opened 2015-09-19
NullReferenceException opening .ckan file.

zman97211:
When opening a .ckan file previously exported, a NullReferenceException can happen. This fixes it.


Postremus:
The value of GuiMod.isCached is used for calculation the "Cached" view, and showing if the mod is cached in the metadata panel.
Could you please fix the nre, without Setting it always to false?

zman97211:
What does it mean for an exported list of other mods to be cached? Doesn't that depend on whether each of the other mods is cached? To be honest, I use CKAN and wanted to load my list of mods and just did a quick fix. I don't have time to investigate a "real" fix, unfortunately. Also, I don't know what you mean by "the nre".

PR #1473, opened 2015-09-23
*chucklesnort* adding for posterity

techman83:
@pjf 
chuckle-snort

### DIFF
--- a/quotes.txt
+++ b/quotes.txt
@@ -7,7 +7,7 @@ deleting our entire CKAN-meta repository]
 This is less than ideal.
 
     -- dbent, #979
-
+%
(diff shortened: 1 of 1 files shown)


Postremus:
Windows, you know, the safe will crash later this day.

pjf:
@techman83: Woah! That's right, we have a quotes file! I'm pretty sure we don't need review to add to it, but since I'm here you'll need a `%` on a line by itself for `fortune` and friends to process it. :)

@Postremus : Or you'll go to lock your sandwich in a safe, but your anti-virus software will have locked it in its own safe first. :)

techman83:
@pjf haha, I'm glad it was reviewed :)

Issue #1109, opened 2015-06-12
OSX can't locate installations


beldougie:
I have all of my KSP installs (and all of my Steam games to be fair) on an external drive that is accessed via `/Volumes/{drivename}/STEAM_LIBRARY/steamapps/common/`. I can run CKAN using mono, no problem, but when I click the 'add new' button, the path tree is wrong. It seems to think it's on Windows or something, but it starts at Desktop and then won't let me traverse higher in the tree (my desktop is some 4 or 5 nodes below the root of the drive.

This makes it unusable because I can't access the installations I have. Would love to try this out at some point though if you can address this. Cheers,

beldougie:
However, I have just installed with homebrew and with the command line I'm able to access the correct installations ok - it's just running from `mono ckan.exe` seems to nerf the path detection.

Hope that helps

craigmaslowski:
I'm having a similar problem. My Steam games are installed on a second internal hard drive. I cannot see that volume regardless of whether I run `mono ckan.exe` or install the homebrew version and run it. I had to use Terminal and the command line options to get my install recognized.

`ckan ksp add Main /volumes/data/SteamLibrary/SteamApps/common/Kerbal\ Space\ Program`

pjf:
Adding @McJones to this issue, who's also reported this.

`KSPPathUtils.cs` contains the Steam detection code. I'd love a patch that *does* find KSP under OSX. :)

pjf:
Closing this with the application of #1480; our next release (today) should show some relief. :)

PR #1480, opened 2015-09-25
Fixed OS X cmdline not finding a valid KSP install

McJones:
I believe this fixes issue #1109 or at least it did on my install of OS X...

Looking at the code I changed, I have a sneaking suspicion this issue may have been more than OS X and impacted every OS, but I have no means of testing that theory currently.

I am not a C# programmer so my bad if I have made some sort of C# faux-pas.


Postremus:

pjf:
Oh goodness, this bug would have been around for a very long time! Thank you so much for catching it!

PR #1481, opened 2015-09-27
Improved steam detection with guards

pjf:
This is #1444, but I've added some guards so that if our config file is corrupted we won't throw an exception from running off the end of an array.

I don't have a machine handy with a non-standard install of Steam, but the code looks right to me, so if you're reading this and anyone has reported it's working as intended, then please merge. :)


pjf:
I've [built a ckan.exe](https://dl.dropboxusercontent.com/u/9702672/KSP/ckan-v1.14.2-19-g0dd553b.exe) from this branch to make it easier to test. :)

LarsOL:
Can confirm that it works on secondary library location.

pjf:
Works for @LarsOL , and looks good to me, so I'm doing the merge thing!

pjf:
Thank you @LarsOL ! My apologies that took such a long time to get through. You rock!

PR #1496, opened 2015-10-16
Sort Conflicts Relationship

dbent:
Somehow missed adding an explicit sort for `conflicts` relationships.


pjf:
This looks fabulously safe! Merging with thanks!

Issue #1173, opened 2015-06-24
How do we sensibly handle "forced" installs?


Murdabenne:
There is no straightforward way to do this.  Currently you have to specify the version and do this from the command line.   When you have 35+ apps that have yet to have their metadata updated, this is pretty tedious business, having to first go into the metadata to find what exact version info you need, then to individually install each of these.  It pretty much defeats the purpose of having this tool.

The GUI app already has the latest version and the KSP version available to it, so it can handle this if given an override. Here is how I propose to make it work:
1. Go to Options, have a checkbox or button there that says "Allow Forced Install". Click it. 
   1a. A popup appears explaining the dangers, etc. And a warning that this is a temporary setting, and will not persist between CKAN runs.  OK/Cancel are the only options 
2. Got to filters, select incompatible.
   2a.The list changes to show the same thing it does now, except they have checkboxes. This is the ONLY filter that will allow the option to make checkboxes, and conversely the only time you will ever see checkboxes on the incompatibles screen.
3. Proceed as usual for installs, overriding the version checks, simply grabbing the latest version available.
4. Forced Installs show in the list as "F" (like "AD") and cannot be manipulated by CKAN other than being uninstalled.

Making the setting it non-persistent prevents someone from accidentally screwing up by leaving it on the incompatibles screen and mistakenly thinking they were on the normal screen.  it is a safety measure.

This will help users when we get a bugfix (x.x.BUGFIX) KSP release and authors that have a tonne of apps arent able or are unwilling to update the KSP version, or (the most likely case) they simply haven't gotten around to it but the app works and the users want it.

This issue dovetails somewhat with the version checking issue - that might provide a way for stricter version checking to prevent outdated installs, while this would provide users the ability to override that.

Murdabenne:
Later uses for this could be expanded to perhaps allow an app to be "pinned" to a specific version (and prevent unwanted updates from showing up - like AD, this would show as P).

Or possible using the force version install methods from the command line, wrapping it for GUI use - which I do realize opens a whole can of worms for expanding the database to track older versions of the metadata, if we want to give the user a list of previous versions from which to choose.

sarbian:
I would just add an "ignore minor" option so a mod marked for 1.0.3 would install on 1.0.4.
Minor updates that break mods are rare and an option would allow CKAN to still be useful when we have a KSP minor patch slave.

pjf:
[...]

PR #1499, opened 2015-10-20
Add fuzzy versioning to the CKAN.

pjf:
The buzz is that 1.0.5 will be out soon-ish, and will be mostly compatible
with 1.0.4. However the last time we had a highly compatible KSP release,
[it wasn't fun for CKAN users](https://www.reddit.com/r/KerbalSpaceProgram/comments/3beh00/why_are_there_so_few_ckan_mods_for_104/). This PR hopes to fix that.

These changes all relate to #1173, and include:

**IGameComparator**

Rather than having only a single way to test if a mod is compatible, we now have a whole family that conform to the `IGameComparator` interface. The classic mode the CKAN used is the 'Strict' version, but I've now adjusted the default to be the 'Generally Recognised As Safe' (GRAS) version, which will allow installing mods from 1.0.4 on 1.0.5 unless they're flagged otherwise.

Also included for completeness is 'You're On Your Own' (YoYo) comparisons, which let you do anything.

At the moment there's no way for the user to switch comparators, but basics of doing that are here.

**Autofac**

Autofac is an inversion of control / dependency injection framework. It means that when we create instances Autofac can supply them with dependencies. We only use this for `IGameComparator` at the moment, but it has potential to make our tests better, and for us to set a variety of options at start-up in a more sane fashion.

Right now we'reusing the [Service locator (anti-)pattern](https://en.wikipedia.org/wiki/Service_locator_pattern), which isn't my favourite, but it means we didn't need to re-achitecture everything.

**Schema updates**

We introduce a new keyword `ksp_version_strict` into the spec, which defaults to false. If set to true, it says that says that comparisons must always be strictly performed. The idea is we can flip this on for any 1.0.4 mods we know don't work in 1.0.5.

**Feedback?**

This isn't a complete implementation, and I'd love feedback. Right now the user can't switch comparison modes, and there's a risk that we may try to choose an *older* mod as being more compatible than a more recent mod. However I think that's a separate change that's independent to this work, so I hope to throw in a separate PR for it.


pjf:
I've also made sure I'm pulling from the KSP-CKAN repo, so feel free to add extra commits as needed with revisions if needed. :)

@dbent : If you have a chance, can you look over this for sanity? I know you're the Wizard of Autofac, so I'd love to know if I'm off in a completely wrong direction.

pjf:
**Potential gotcha**

At the moment, we default strict mode to false, meaning all 1.0.4 mods will be 1.0.5 installable unless marked otherwise. However imagine the following:

* FictionalKerbalHeatMod has four releases for KSP 1.0.4, but is *not* 1.0.5 compatible.
* We mark it as `ksp_version_strict` in the netkan file, but only the *latest* release gets that flag.
* When 1.0.5 comes out, the CKAN tries to install an *old* version of FictionalKerbalHeatMod.
* Derp.

**Unsatisfying solution**

When we need to enforce strict mode, we simply tag all documents that target KSP 1.0.4 that don't work with 1.0.5. It requires human effort, and we'd have to bump to `spec_version` at the same time, but it's not a disaster. I can do this on the command-line with perl; so no matter what, we have a solution, even if it's not a satisfying one.

**Satisfying solution**

In an ideal world, as `AvailableModule.Latest()` looks back through our list of versions to install, it should halt if a version is rejected because our KSP version is "too new". This means we should never pick an older version because it's *more* forward-compatible than a newer version. (I think @Dazpoet or @plague006 made this point once, and I'm really starting to agree with it.)

Unfortunately, I've spent a good hour trying to find a decent way to say "hey, stop looking" in `AvailableModule.Latest()`, but don't have one yet. Admittedly I am on a two day sprint, and after 20-something hours of coding I think whatever neurotransmitters I use for logic are severely depleted.

**Other solutions**

I'm sure they're out there. Suggestions welcome.

**TL;DR:** We may have a little extra metadata finangling for mods we want to place into strict mode, but *way less* than the amount of effort needed when 1.0.5 comes out and is "mostly compatible" with 1.0.4.

dbent:
Cool, I'll probably look at this tomorrow (~12 hours from now).

Core/ServiceLocator.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                if (_container == null)
+                {
+                    Init();
+                }
+
+                return _container;
+            }
+
+            set
```
dbent: Is there any need to make this `set`-able?

Core/ServiceLocator.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                _container = value;
+            }
+        }
+
+        public static void Init()
```
dbent: This can be `private` and probably should be.

Core/ServiceLocator.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace CKAN
+{
+    /// <summary>
+    /// This class exists as a really obvious place for our service locator (ie: Autofac container)
+    /// to live.
+    /// </summary>
+    public static class ServiceLocator
+    {
+        private static IContainer _container;
+        public static IContainer container
```
dbent: Public properties should be `PascalCase`, so `Container`.

Core/ServiceLocator.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            get
```
dbent: It's probably worth noting in a comment that this isn't thread-safe, not that it needs to be, but if CKAN ever goes super-parallel it's something to watch out for.

Core/Types/CkanModule.cs
```diff
@@ -194,6 +201,9 @@ public class CkanModule : IEquatable<CkanModule>
         [JsonProperty("install")]
         public ModuleInstallDescriptor[] install;
 
+        // Used to see if we're compatible with a given game/KSP version or not.
+        private IGameComparator _comparator;
```
dbent: If `_comparator` is only ever set in a constructor (which I think it is) it can also be made `readonly`.

Core/Types/CkanModule.cs
```diff
@@ -381,65 +470,17 @@ public static void ToFile(CkanModule module, string filename)
...
-            {
-                throw new UnsupportedKraken(
-                    String.Format(
-                        "{0} requires CKAN {1}, we can't read it.",
-                        newModule,
-                        newModule.spec_version
-                    )
-                );
-            }
+            comparator = ServiceLocator.container.Resolve<IGameComparator>();
```
dbent: Any reason not to join the variable declaration and assignment?

Core/Types/GameComparatorGRAS.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+
+namespace CKAN
+{
+    /// <summary>
+    /// Test to see if the user's game is "Generally Recognised As Safe" (GRAS) with a given mod,
+    /// with extra understanding of which KSP versions are "safe" (ie: 1.0.5 mostly works with 1.0.4 mods).
+    /// If the mod has `ksp_version_strict` set then this is identical to strict checking.
+    /// </summary>
+    public class GameComparatorGRAS : IGameComparator
```
dbent: So [acronyms longer than two characters are `PascalCased`](http://stackoverflow.com/questions/2109171), so `GameComparatorGras`. (This also means the top level namespace really ought to be `Ckan`...) Also, descriptors and adjective modifiers typically occur before the noun in a class name, following regular english conventions. So `GrasGameComparator`. Unfortunately this means similar classes are not sorted next to each other in an alphabetical list, but a common namespace could do instead.

Core/Types/GameComparatorGRAS.cs
```diff
@@ -0,0 +1,37 @@
...
+            // If we're in strict mode, and it's not strictly compatible, then it's
+            // not compatible.
+            if (module.ksp_version_strict)
+                return false;
+
+            // Otherwise, check if it's "generally recognise as safe".
+
+            // If we're running KSP 1.0.5, then allow the mod to run if we would have
+            // considered it compatible under 1.0.4
+            if (gameVersion.Equals("1.0.5"))
```
dbent: So hardcoding this is fine for now... but what I'd like to see in the long run is a probably something like a `config.json` in the top level of CKAN-meta that would contain dynamic configuration for clients that don't apply to any particular mod. So it could contain something like:

```json
{
  "ksp_version_compatability": {
    "1.0.4": [ "1.0.5" ]  
  }
}
```

Which would read as "Any mod that says its compatible with `1.0.4` is also compatible with `1.0.5` by the default (GRAS) scheme.

By making it part of the metadata we can update it through PRs and make sure older clients get the necessary information without relying upon people upgrading the client.

Core/Types/GameComparatorStrict.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+
+namespace CKAN
+{
+    /// <summary>
+    /// Test to see if a module is compatible with the user's installed game,
+    /// using strict tests.
+    /// </summary>
+    public class GameComparatorStrict : IGameComparator
```
dbent: `StrictGameComparator`

Core/Types/GameComparatorYOYO.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+
+namespace CKAN
+{
+    /// <summary>
+    /// You're On Your Own (YOYO) game compatibility comparison.
+    /// This claims everything is compatible with everything.
+    /// </summary>
+    public class GameComparatorYOYO : IGameComparator
```
dbent: `YoyoGameComparator`

dbent:
:+1: with comments.

Core/ServiceLocator.cs
pjf: Yes, for testing! And for user preferences!

Core/Types/CkanModule.cs
pjf: No reason. I had them separated for debugging, and this is an artefact of that.

Core/Types/GameComparatorGRAS.cs
pjf: I'd love this as well, along with the ability for the user to override it if they really want to.

pjf:
@dbent : You do the best code reviews! Shall add some commits that address your comments! Thank you so much!

Dazpoet:
These things sound pretty freaking amazing :+1:

Also anything making CKAN stop from using older versions due to being more forward compatible would be pretty amazing but last I checked (which admittedly was quite some time ago) @plague006 was on the case and hunting down any mod daring to be non-conforming to his rigorus demands of uniformity :)

Olympic1:
POKE

plague006:
Some mods that will need the strict version flag (are 1.0.5 incompatible): `ContractConfigurator`, `ChuteSafetyIndicator`, `ContractConfigurator-AnomalySurveyor`, `Trajectories`, `WaypointManager`, `ModuleAnimateEmissive`, `SCANsat`, `DMagicOrbitalScience`, `RemoteTech`, `AGExt`, `RasterPropMonitor`, `IntakeBuildAid`, `TweakScale`, `Kopernicus`.

The preceding were all .netkans, the following are .ckans only:
`TweakableEverything-EngineStagingToggle`, `TweakableEverything-FairingStagingToggle`,`ScienceAlert`, `MechJeb2`.

nobodyhasthis:
There is no point in going for another pull request just yet to do the same as #1513. However can we look at this for the opposite reason given in the request?

Can we be given an option to force application of ksp_version_strict on an single mod where the assumed fuzzy version metadata turns out to be potentially wrong. Almost all mods are 1.0.5 compatible and are now appearing as such. However there are some on the allowed list that should not be there ...

[1.0] CrossFeedEnabler v3.3 May 11 (approved for 1.0.5. Is it not in stock now? - Asking becuase it done by the man that can best answer the question)
[1.0.x - OBSOLETE] ClampsBeGone - Stop clamps following your vessels into flight! [Aug 17] (yes it is obsolete in 1.0.5 so can't be 1.0.5 approved)

Y3mo:
How about doing nothing of the sort and just enabling users to install all mods instead of only the ones in the "compatible" list?
Then the responsibility for installing not specifically marked compatible versions is where it actually belongs, with the users installing them.
If users want a "controlled environment", they can just use the default "compatible" list.

lamont-granquist:
How about you just let users force install a mod and ignore the kerbal version on that install, and do no depsolving of deps (either fail due to deps not being there and tell users which deps need to be force installed first, or just install for the user and warn that deps haven't been installed).

This PR seems to be getting lost trying to solve a very hard problem that has a ton of edge conditions and needs a lot of clever magic to try to make it work.  You can simplify the problem a *lot* by just rejecting the depsolving problem--which is complicated a lot by old packages that necessarily lacking metadata that couldn't have been added to them because packages don't know when they're written what their "real" maximum kernel version will be.

But really all I want right now is a button that lets me install the v1.0.5-compatible Adjustable Landing Gear mod.  If you force me to do my own depsolving, that is fine.  Just blow up if there's a dep that isn't satisfied and make me manually go force install every dep and do the validation myself that the whole tree is really v1.0.5 compatible.

And I offer this advice as a software developer who has fought with both the rubygems depsolver and the depsolver for chef cookbooks for the past 4 years.  Depsolving is a really hard problem, and you're asking your depsolver to try to mitigate the problem of past metadata not being able to predict the future.

nobodyhasthis:
@lamont-granquist. I am neither a contributor or a programmer. So please take the following with a pinch of salt. What I do is help with the financial funding a little bit and have watched the project though it's up and downs. That is all. So please forgive me if I give a silly answer here. 

Your are asking for forced installs. After a somewhat detailed discussion #1173. We ended up here. Up to now compatability checking has be included with the mod spec_version. With the mod being thrown out if it did not strictly match. That was ok when every update of KSP broke things. Now with stable builds the opposite is true. 

Mods can now be assumed to be compatible. With the new default setting letting you instal mods as KSP updates. The 'Generally Recognised As Safe' (GRAS) option. Right now installing the v1.0.5-compatible Adjustable Landing Gear mod should be as simple as ticking it. The default settings list it as compatible along with the dependency in the Animation Modules. Max KSP version in the current metadata might be 1.0.4 but the default filter option is now GRAS. At least that how I see it CKAN 1.14.3

However mod owners know best and can stop this default. CKAN serves two groups and the first is obvious: mod users. The second is indirect but no less important: the mod authors. If they say a new version of KSP always breaks things so it goes. This only affects a few mods which can use the 'Strict' option of version control on request. The user community can also help here by giving feedback on what has become incompatible. See #2475 as an example.

Of course the end user can still throw caution to the wind and force installs. The completeness of adding in the 'You're On Your Own' (YoYo) option. Which what I think your asking for. Can be done but there no reason to throw out the first two solutions.

Logically there is several outcomes depending on how the filtering is chosen by the end user or mod author. This is an attempt to come up with a solution that tries to please everyone. However difficult that may be. It is worth trying.

plague006:
Just throwing it out there that the 1.0.4 -> 1.0.5 window has basically closed. Anything compatible with 1.0.5 (should) be marked as such by now. But this can still be useful for the 1.0.X -> 1.1 transition. (or 1.1.0 -> 1.1.X).

nobodyhasthis:
A quick question how much does it matter when we get odd version numbers? 

Having problems getting it to work in a 1.0.4 build. I am looking at some "exotic" version numbering system by CKAN standards with EVE. What I am seeing in the CKAN file search is.

Running KSP 1.0.4 

EVE version number is 7-4
With low resolution config version number also 7-4 installed it works.
High resolution file is listed as EVE-1.05-1 but it does not work. ( i think the bot has found the wrong config)

When running with KSP 1.0.5 

EVE version is now 1:EVE-1.05-4 and there is only one config file also marked 1:EVE-1.05-4

I am think the character string version numbers are messing up the CKAN bot. A possible candidate for very fuzzy version control?

plague006:
@nobodyhasthis 

> EVE is odd in KSP 1.0.4.

This PR isn't about mod version numbers but rather about ksp_version compatibility fuzziness. With that said the EVE situation you're describing arose when a new version of EVE was released for 1.0.5 but things got "messy". This should be fixed with https://github.com/KSP-CKAN/CKAN-meta/pull/859.

pjf:
Oh my, catching up on *my own* pull requests here.

@plague006:

> Just throwing it out there that the 1.0.4 -> 1.0.5 window has basically closed. Anything compatible with 1.0.5 (should) be marked as such by now. But this can still be useful for the 1.0.X -> 1.1 transition. (or 1.1.0 -> 1.1.X).

I wasn't too stressed about this missing 1.0.5, because it turned out that 1.0.4 and 1.0.5 had a lot of incompatible changes between them, and before the release I had a polite request from those close to Squad that 1.0.5 be considered a significant new release.

Having said that, I still think these features are definitely something we want in the client. It would have been great for 1.0.3 -> 1.0.4, and my software sense tells me there's a high likelihood of it being useful with 1.1.0 -> 1.1.1.

@nobodyhasthis : Thank you for your awesome explanations! You rock!

@lamont-granquist : As it happens, this patch *also* provides the foundation to turn off version checks entirely, if that's what you want. Just engage the `YoyoGameComparator` and you're golden. In the GUI that will probably be a "restart in the DANGER ZONE" style option; on the command-line we'll probably have a switch `--versioning=yoyo` or similar.

In any case I'm going to finish my changes to @dbent's review, since I believe this is something we want in the core even if we haven't got a front-end for it.

P.S. You all rock!

Core/ServiceLocator.cs
pjf: Fixing now!

Core/ServiceLocator.cs
pjf: Done!

Core/ServiceLocator.cs
pjf: Done!

Core/Types/CkanModule.cs
pjf: Oh! I now finally understand readonly! Thank you!

Core/Types/GameComparatorGRAS.cs
pjf: Roger this! Fixing now. :)

pjf:
I've made by updates based upon @dbent's changes. I've made 1.0.4 GRAS with 1.0.3 mods, so users won't see anything yet, but this now gives us a framework to let users run the client with different comparison routines, along with a way to handle hotfixes nicely in the future. :)

About to merge as all my changes except 757d9cb as either renames or documentation only changes. (If anyone wants to verify 757d9cb looks sensible, that would be grand!)

PR #1503, opened 2015-10-27
Bring the schema up to date

plague006:
#1243 gave us the ability to install into SPH and VAB, this will reflect that in the schema (I think..)


techman83:
Enum would have worked too. I hadn't tested it that way when I suggested the regex to @plague006

techman83:
Travis liked it, tested it on previously failing file and various others for completeness. LGTM, thanks @plague006 !

Issue #1241, opened 2015-07-05
Allow find directives to pick files


dbent:
Came up in KSP-CKAN/NetKAN#1821 and KSP-CKAN/NetKAN#1822. Any particular reason it only matches on directories?

pjf:
"Historical reasons".

The `find` directive was essentially exposing the function that did the "default" install when no install stanza was specified. It always searched for a directory name, and hence the `find` directive did the same.

dbent:
Cool, should be fairly straight forward to allow files as well.

plague006:
As #1512 is merged I think this can safely be closed.

PR #1512, opened 2015-11-10
Better Jenkins Support

dbent:
@pjf @techman83 @BasharMilesTeg @Olympic1 @Dazpoet @plague006 @sarbian

So netkan has had support for Jenkins CI servers for a long while, but the support always required specifying so much custom data that it was pointless to be used with netkan as each `.ckan` generated required manual editing.

What I've done:

- I've modified the `jenkins` `$kref` to take a URL to a Jenkins job, not some random name, this is a backwards incompatible change.
- Using the URL the `JenkinsTransformer` will now look for the most recent build for that job, and it will then find the appropriate artifact to use for that specific build (*not* the lastestBuild URL).
- I've introduced an `x_netkan_jenkins` object which allows the following properties to be specified:
  - `build` (string, defaults: `stable`) specifies what kind of build to use. The options are `any`, `completed`, `failed`, `stable`, `successful`, `unstable`, `unsuccessful`. A lot of these don't make any sense to use in practice (`unsuccessful`??), but are there just for completeness.
  - `asset_match` (regex, defaults: `\.zip$`) specifies a regex for matching what artifact to use. By default it will use any single artifact that ends with `.zip`. This is basically identical to what the `GithubTransformer` does.
  - `use_filename_version` (bool, defaults: `false`) specifies if the `JenkinsTransformer` should use the filename as the `version` property of the `.ckan`. Combined with `x_netkan_version_edit` this allows use to extract the version from the filename itself. Otherwise the expectation is that the archive will have an AVC `.version` file which will be used to generate the `version` by the `AvcTransformer`.
- I've modified the `find` and `find_regexp` install directives to also match on files as well as directories (#1241). **This likely requires bumping the spec version.**

All this combined allows the following `.netkan`:

```json
{
    "spec_version": "v1.10",
    "identifier": "ModuleManager",
    "$kref": "#/ckan/jenkins/https://ksp.sarbian.com/jenkins/job/ModuleManager/",
    "x_netkan_jenkins": {
        "use_filename_version": true
    },
    "x_netkan_version_edit": "^ModuleManager-(?<version>\\d+\\.\\d+\\.\\d+)\\.zip$",
    "name": "ModuleManager",
    "abstract": "Modify KSP configs without conflict",
    "author": [ "ialdabaoth", "Sarbian" ],
    "license": "CC-BY-SA",
    "install": [
        {
            "find_regexp": "ModuleManager.*\\.dll$",
            "find_matches_files": true,
            "install_to": "GameData"
        }
    ]
}
```

To produce the following `.ckan`

```json
{
    "spec_version": "v1.10",
    "identifier": "ModuleManager",
    "name": "ModuleManager",
    "abstract": "Modify KSP configs without conflict",
    "author": [
        "ialdabaoth",
        "Sarbian"
    ],
    "license": "CC-BY-SA",
    "resources": {
        "ci": "https://ksp.sarbian.com/jenkins/job/ModuleManager/"
    },
    "version": "2.6.13",
    "install": [
        {
            "find_regexp": "ModuleManager.*\\.dll$",
            "find_matches_files": true,
            "install_to": "GameData"
        }
    ],
    "download": "https://ksp.sarbian.com/jenkins/job/ModuleManager/101/artifact/ModuleManager-2.6.13.zip",
    "download_size": 28055,
    "x_generated_by": "netkan"
}
```


dbent:
Added support for automatically populating the `ci` resource.

Olympic1:
LGTM, this will also be good for MechJeb

hakan42:
Excellent, thank you very much for taking the JenkinsPopulator where I always wanted to if I had had the time :smile:

plague006:
Being able to `find` files will be great!

Dazpoet:
This looks like the best thing since whatever the last thing I said was the best thing that ever happened to netkan was!

dbent:
Updated to add an explicit `find_matches_files` property to `install` stanza to preserve backwards compatibility. @pjf

pjf:
Merging now with tweaks to the schema describing the new `find_matches_files` property. Thank you!!

(And thank you for your patience!)

Olympic1:
@dbent, this method only works for builds that only have 1 `.zip` file (ex. [ModuleManager](https://ksp.sarbian.com/jenkins/job/ModuleManager/lastStableBuild/)), but this doesn't work for builds that have multiple `.zip` files (ex. [GCMonitor](https://ksp.sarbian.com/jenkins/job/GCMonitor/lastStableBuild/)) because it contains too many artifacts.

Is there a way to get around that problem or does @Sarbian need to include only one `.zip` file from now on?

dbent:
@Olympic1 well there's the `asset_match` property, but that wouldn't really help in the case of GCMonitor where you'd need to determine what the latest version is to select correctly. We could certainly add additional logic to do that but I think it makes more sense for a build to only contain the assets built specifically for that particular build like with ModuleManager.

@sarbian is there a particular reason GCMonitor is setup that way and would it be too onerous to change?

sarbian:
Nope, I just need to check the box to clean the workspace before each build and rebuild. I just did it on all the Jenkins project. 
Tell me if something else is needed

dbent:
Great, thanks!

PR #1525, opened 2015-11-25
Rebuild Ships Directory on Load

### LINKED ISSUE #1404: Create thumbs directory in Ships

Wetmelon:
Small addition that allows CKAN to rebuild any directory structure
within the current KSP game directory on load, if the directory doesn't exist.

Currently only supports Ships/VAB, Ships/SPH, Ships/@thumbs/VAB, and
Ships/@thumbs/SPH, but can be easily extended to other paths.

Fixes #1404

### DIFF
--- a/Core/KSP.cs
+++ b/Core/KSP.cs
@@ -236,6 +236,20 @@ namespace CKAN
 
             throw new NotKSPDirKraken(directory, "Could not find KSP version in readme.txt");
         }
+        
+        /// <summary>
+        /// Rebuilds the "Ships" directory inside the current KSP instance
+        /// </summary>
+        public void RebuildKSPSubDir()
+        {
+            string[] FoldersToCheck = { "Ships/VAB", "Ships/SPH", "Ships/@thumbs/VAB", "Ships/@thumbs/SPH" };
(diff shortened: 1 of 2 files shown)


pjf:
This makes me really happy, because it's so easy for these to get deleted by accident!

pjf:
I'm going to tweak this to also rebuild the top level `Ships` directory if it's not there, but otherwise merging. This works on my system and the code is fabulously straightforward. Thank you!

Wetmelon:
Technically it will already create the entire chain of folders by default, but explicitly defining them can't hurt.

pjf:
@Wetmelon : Oh really? Let me test that, merging without me adding changes makes me super happy!

pjf:
Tested! Merging without changes.

pjf:
Merging when I'm back from lunch because I'm on a strict hackathon schedule. :)

pjf:
Merged with many thanks!!

PR #1528, opened 2015-12-03
Fix typo bool->boolean

NathanKell:


Olympic1:
Thanks @ThomasKerman for the fix
Thanks @NathanKell for making a PR

PR #1538, opened 2015-12-13
Restore old displayYesNo instead of setting to null

pjf:
This is a cherry-pick of 8aa8abd from #1290. Instead of assuming the `YesNoDialog` default to null, it instead caches what it previously was and restores it. At the moment this is effectively a no-op, but it may save us hassles in the future.

Code courtesy of @RichardLake, with cherry-picking and integration by me. :)

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -308,9 +308,10 @@ namespace CKAN
             m_CacheWorker.RunWorkerCompleted += PostModCaching;
             m_CacheWorker.DoWork += CacheMod;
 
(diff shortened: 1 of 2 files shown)


Postremus:
tested, works.
merging with my thanks!

pjf:
@Postremus : Thank you! :)

politas:
So this will maintain selections after a failed install attempt? Cool.

pjf:
@politas : Alas no! It just means if we change how we do some dialogs in the future we won't uncover a bug. :)

Issue #1327, opened 2015-07-20
`Max KSP Version` isn't using the most recent mod version


plague006:
Easily seen with `RealPlume` which requires a forced epoch. Its most recent version is `1-v0.4.1` but we catch `v10.0.0` for the sake of `Max KSP Version`

Postremus:
It uses correctly 1-v0.4.1 of RealPlume for me.

plague006:
Ack you're right the bug doesn't exist as I described it. I was forcing an epoch to the mod and got ahead of myself. `1-v0.4.1` didn't exist at the time I reported the bug, `v0.4.1` was the max version at that time, and the GUI displayed `v10.0.0`

I created https://github.com/KSP-CKAN/CKAN-meta/tree/kspmaxverbug which reverts the creation of the forced epoch `1-v0.4.1`. If you replace the default repo with that you'll see the bug as it was when I reported it.

http://puu.sh/j7kYP/ca53b12c21.png

mgsdk:
Another case is InfernalRobotics. Latest version states 1.0.4 compability in the metadata, but the shown result is from an earlier release stating 1.0.2.

Postremus:
NFS has for ksp 1.0.4 the version 0.5.3. One release however has the version 0.5.21. The "max compatible" column uses registry.latest() to get the newest release for any ksp version. This returns 0.5.21.

pjf:
I'm not sure this is a bug in the GUI per se, rather an artefact stemming from the fact that mod version numbers don't always follow our version comparison rules, with 0.5.21 being a perfect example.

We should be able to happily epoch-bump these mods to solve the issues. If we hide the epoch in the display (which I think would provide a nicer user experience) then the users won't even notice. :)

politas:
Epoch-bumping is not the best solution. We should populate the "Max KSP" column data from the same version we get the "Latest" version for the mod from. Grabbing the list of .ckan files on Github in alphabetical order is absurd, and we'll be constantly bumping epoch numbers every time a mod author adds a layer to their version numbering. Epoch bumping is the wrong way to fix this. The GUI should handle it better.

politas:
Ok, I've gone through all the instances I could find of mods listing as compatible with 1.0.5 but showing Max_KSP as a lower version and done something to all of them. Not sure if all the workarounds will be successful, but I think most of them will at least.

politas:
[...]

PR #1554, opened 2015-12-28
Fixing version sorting of _ vs .

distantcam:
The NavUtilities versions were not being sorted properly. Currently the sort is

- 0.5_RC_3
- 0.5.1_RC_1_repacked
- 0.4.3

It should be

- 0.5.1_RC_1_repacked
- 0.5_RC_3
- 0.4.3

This change modifies the sort so that '.' takes precedence over all other non-digit characters. That way version parts are considered higher than other random characters tacked onto the end of the version.

All the other version tests pass.

### DIFF
--- a/Core/Types/Version.cs
+++ b/Core/Types/Version.cs
@@ -216,13 +216,20 @@ namespace CKAN {
 
             // Then compare the two strings, and return our comparison state.
             // Override sorting of '.' to higher than other characters.
-            if (!str1.StartsWith(".", StringComparison.Ordinal) && str2.StartsWith(".", StringComparison.Ordinal))
+            if (str1.Length > 0 && str2.Length > 0)
             {
-                comp.compare_to = -1;
-            }
-            else if (str1.StartsWith(".", StringComparison.Ordinal) && !str2.StartsWith(".", StringComparison.Ordinal))
-            {
-                comp.compare_to = 1;
+                if (str1[0] != '.' && str2[0] == '.')
+                {
+                    comp.compare_to = -1;
+                }
+                else if (str1[0] == '.' && str2[0] != '.')
+                {
+                    comp.compare_to = 1;
+                }
+                else
+                {
+                    comp.compare_to = String.CompareOrdinal(str1, str2);
+                }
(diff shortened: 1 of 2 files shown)


plague006:
Tested this some, and it does correctly bring forward the correct version for NavUtils. It also brings forth Deep Space Utilities properly which closes https://github.com/KSP-CKAN/CKAN-meta/pull/919. This also addresses many cases of https://github.com/KSP-CKAN/CKAN/issues/1327.

@politas as you've been poking around with sorting for a little while, would you mind trying the build out to see if it generates and/or fixes any outstanding issues? ([.zip with debug](http://www.mediafire.com/download/7p1w6pp1w48r807/b423152ce7747c7ace8927ebfaab24d9ffb25fb5.zip) for convenience, you can just extract the folder and run ckan.exe)

distantcam:
I found another example that could possibly break the versioning. I've updated this PR to solve that scenario too.

`ModularFlightIntegrator` had versions:
- 1.1.2.0
- 1.1.1
- 1.1.0.0
- 1.0.repackaged0
- 1.0

I added a test for the uneven version lengths (1.1.0.0 vs 1.1.1) and fixed the case where a patch number could cause the sort to fail (for example, if there had been a 1.0.1 version in the list above.

politas:
Ok, I'm trying to use that version, but I don't seem to be getting any updates to metadata when I refresh using it. NavUtilities  0.5.1_RC_1_repacked does show up as the latest version, but its saying its Max KSP is v1.0.4, and since the CKAN file for that version definitely says 1.0.5 now, I can't trust that I'm properly testing anything. Is there a way to clear the local CKAN metadata store while not losing the installed mods data?

distantcam:
@politas Could you please try refreshing again? The NavUtilities metadata is up-to-date for me when I refresh.

politas:
Ok. Looks fine. The few remaining cases where it's picking up obsoleted versions are not ones that a _/. sorting change will fix.

Olympic1:
I can also confirm that this works as intended, it shows the correct version.

PR #1577, opened 2016-02-10
Added include_only and include_only_regexp

### LINKED ISSUE #228: Add `include_only` and `include_only_regexp`

Zane6888:
This PR adds two fields which practically work as negative filter/filter_regexp. 
Also Includes a test and changes to the CKAN.schema to make sure there are only filter or include_only fields but never both. 

I hope I didn't miss anything, don't really know the code-base yet.

Closes #228

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -192,7 +192,31 @@
                     "filter_regexp" : {
                         "description" : "List of regexps that should filter files from this install",
                         "$ref" : "#/definitions/one_or_more_strings"
+                    },
+                    "include_only" : {
+                        "description" : "List of files and directories that should not be excluded from the install",
+                        "$ref" : "#/definitions/one_or_more_strings"
+                    },
+                    "include_only_regexp" : {
+                        "description" : "List of regexps that should include files in this install",
+                        "$ref" : "#/definitions/one_or_more_strings"
                     }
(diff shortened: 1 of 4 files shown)


plague006:
I'm confused on the use-case for this since `find` and `find_regexp` only match to a single file or folder.

Dazpoet:
This requires a rebase now after #1578 was merged it seems.

ayan4m1:
LGTM, @politas can you give some insight on what direction the project should be taking w/r/t this schema change?

politas:
I think it's a great idea. It would make some mods much easier to configure installs for. To implement, we'd need to make the next release 1.23.0, and have this feature require spec_version v1.23

PR #1578, opened 2016-02-10
Changed schema to enforce structure of install direcitves

Zane6888:
Found a JSON Schema friendly way of specifying that excactly one of 'file', 'find' and 'find_regexp' must be present.

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -195,7 +195,17 @@
                     }
                 },
                 "required" : [ "install_to" ],
-                "comment" : "One must have either file or find, is there a JSON Schema friendly way of specifying that?"
+                "oneOf": [
+                    {
+                        "required": [ "file" ]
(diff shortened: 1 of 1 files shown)


pjf:
Thank you so much, @Zane6888 ! It looks like we're behind on PRs; so thanks for your patience with the delay. I'm testing this now. :)

Dazpoet:
Looking over this I feel safe merging it, however I don't know if this will undo the changes made since the creation of this PR in #1597 or if it will just work by some github magic?

politas:
Wouldn't it say there was a merge conflict in that case?

Dazpoet:
I shall apply the magic of TIAS with the knowledge that it can easily be reverted in the case of breakage.

Dazpoet:
It worked :D

PR #1580, opened 2016-02-13
[NetKAN] Tolerate Malformed Author URLs from KS/SD

dbent:
From KSP-CKAN/NetKAN#2980. This PR changes the datatype of KerbalStuff's `website` and `source_code` properties into `String` from `Uri`, since we have no guarantee that they're actually well-formed `Uri` strings. Now when "normalizing" the URIs (renamed the method from `Escape` to `Normalize` since we were doing a lot more in it than merely escaping characters) we do a final check to make sure the string is a well formed URI. If it is, we proceed as normal, if it isn't we spit out a warning and return `null`, preventing it from being added to the metadata (the `SafeAdd` extension method will ignore `null` inputs).

@politas @plague006 @pjf

### DIFF
--- a/Netkan/Sources/Kerbalstuff/KerbalstuffMod.cs
+++ b/Netkan/Sources/Kerbalstuff/KerbalstuffMod.cs
@@ -12,8 +12,8 @@ namespace CKAN.NetKAN.Sources.Kerbalstuff
         [JsonProperty] public string short_description;
         [JsonProperty] public string author;
         [JsonProperty] public KSVersion[] versions;
-        [JsonProperty] public Uri website;
-        [JsonProperty] public Uri source_code;
+        [JsonProperty] public string website;
+        [JsonProperty] public string source_code;
         [JsonProperty] public int default_version_id;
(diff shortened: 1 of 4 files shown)


plague006:
Trying to run the debug output leads to a problem with windows not having curl, and I don't have access to a linux machine to properly build a netkan.exe. If anyone would link one, I'd happily start using/testing it.

Olympic1:
Tried out the debug build from @plague006 with KSP-CKAN/NetKAN#2980 and `netkan.exe` succesfully created a `.ckan` file without problems. Looking at the debug screen it checks the link and removes the broken link.

Will test further for a couple of days with existing and new netkans.

plague006:
This PR seems moot with recent events, but I have to imagine the changes can, will, and should be applied to spacedock.

dbent:
Yes, I'll have to update it for Spacedock.

dbent:
Updated for SpaceDock

Olympic1:
Weird, it only added the Kerbalstuff ones, merging manually then.

PR #1581, opened 2016-02-13
Update Spec with NetKAN information

dbent:
The spec is woefully out of date with regards to all the nifty features we've added for NetKAN. This PR documents all of the ones I can think of.

### DIFF
--- a/Spec.md
+++ b/Spec.md
@@ -600,14 +600,12 @@ If any options are not present their default values are used.
 An example `.netkan` excerpt:
 ```json
 {
-    // ...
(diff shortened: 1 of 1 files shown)


politas:
The information on all the extra Netkan fields should be added to wiki/Adding-a-mod-to-the-CKAN, too.

Dazpoet:
Yes, yes for the love of all that is holy YES! Trying to return to checking metadata requests this summer was a chore because all these newfangled things just existed without explanation. Just having read this PR just increased my understanding by ~200%

PR #1584, opened 2016-02-14
Fix #1548 - general case of switching back to tab automatically

### LINKED ISSUE #1548: Scrolling the mod list with the mouse wheel doesn't work initially
When I start CKAN and move the mouse cursor over the mod list, then turn the mouse wheel, the modlist doesn't scroll. Not even when I first click on it. However, when I click outside the CKAN window, then click back on it, the mouse wheel scrolling starts working as expected.

I'm using CKAN v1.16.0-0-g7206ce5. I'm on Ubuntu 15.10.

ChucklesTheBeard:


Olympic1:
@ChucklesTheBeard this works in the beginning when starting `CKAN.exe` up. But when you change to another tab and back to the modlist and hover over it, it doesn't scroll untill you click in the modlist. The same happens if you click the `Refrech` button.

ChucklesTheBeard:
Yep, I noted the first thing in...

...

Ok I know I wrote it down somewhere. Ah well. 

I can't reproduce it not scrolling after hitting the refresh button, but I think I have a better fix anyway. BRB

ChucklesTheBeard:
Try that, @Olympic1

ChucklesTheBeard:
I cannot reproduce the original issue in either 68e97cd or 7ffe21e under Windows 10 Pro x64, building CKAN-GUI for .NET 4.5 via Visual Studio 2015 Community Edition.

@Olympic1 what is your platform?

Olympic1:
Windows 8.1 x64 - Visual Studio 2015 Enterprise build CKAN-GUI with .NET 4.5

The only problem left is that it doesn't scroll after clicking the `Refresh` button untill I change a tab or click in the modlist.

Olympic1:
This works on Windows 7+ and Linux (Tested by me and @ChucklesTheBeard). Merging with my thanks

PR #1589, opened 2016-02-16
Use Cake for build and overhaul/cleanup build

dbent:
@pjf [Cake](http://cakebuild.net/) (_C# Make_) is a build system that uses C#  as its scripting language. This PR allows CKAN and NetKAN to be built using Cake.

On any system to build CKAN:
```
>./build ckan
```

On any system to build NetKAN:
```
>./build netkan
```

On any system to build both:
```
>./build
```

## New
- `build.cake` this is the actual Cake build script. I wrote it to do pretty much what `bin/build` was doing. It performs the following operations:
  - Import ILRepack from NuGet
  - Read some arguments from the command line:
    - `--target`: What task to execute, if not specified defaults to `Build`.
    - `--configuration`: The build configuration to use, if not specified defaults to `Debug`.
    - `--solution`: The solution file to build, if not specified defaults to `CKAN.sln`
    - Define four tasks and their dependencies:
      - `BuildDotNet`, this generates `Meta.cs` and runs `DotNotBuild` which executes `msbuild` on Windows platforms and `xbuild` on Unix-like platforms.
      - `BuildCkan`, this builds `ckan.exe`
      - `BuildNetkan`, this builds `netkan.exe`
      - `Build`, a wrapper tasks that does nothing itself but depends on `BuildCkan` and `BuildNetkan`.
    - Run the specified target using `RunTarget(string)`
    - Define a couple of helper methods, `IsStable()` and `GetGitVersion()`
- `build.ps1` and `build` are shims to run Cake. `build.ps1` is written in PowerShell for Windows systems and `build` is written in POSIX-compliant shell script (not bash) for Unix-like systems. Both scripts are basically identical and perform the same general steps:
  - Get a copy of NuGet if it's not already available locally
  - Install Cake from NuGet (the version of Cake to use is stored in `packages.config`, a standard NuGet package configuration and is pasred by the shims)
  - Build the arguments that should go to Cake, since Cake does not yet support getting positional arguments and it's nice to just be able to do `> ./build <target>` parse the arguments to make that possible.
- Travis now uses Cake to build.
- Add environment variables for build configuration. Only use the `Debug` configuration for now until I can properly configure the existing projects.

## Differences
- I'm using ILRepack 2.0.12 (the latest) because the version 1.25.x that was being used had a bug (gluck/il-repack#58) that was causing an issue on Windows.
- Instead of copying the projects over I'm leaving them in place for simplicity. Because of this I don't want to edit `Meta.cs` in place, so instead I have a `Meta.cs.in` template file which produces a `Meta.cs` on build. `Meta.cs` is then removed from source control and added to `.gitignore`.
- This builds puts some things in a `.build` top-level directory. The eventual goal is to get *all* build artifacts into this directory so that performing a clean build is a simple `rm -r .build` and only one entry needs to be in our `.gitignore`
- Get rid of all extraneous build scripts I could find (`Makefile`, `build.sh`, etc.)
- Remove all NuGet packages from source control. All packages are now stored in the (unversioned) `.build/lib/nuget` directory and all project references have been fixed to refer to that directory. A `nuget.config` file has also been added which will install packages in that directory by default. Yay for no more duplicate NuGet packages! (`packages/` has also been removed from `.gitignore`).
- Create `lib/` for CurlSharp since it doesn't seem to have an official NuGet release. I want to eventually get away from CurlSharp anyway and consolidate all network code (working on that in another branch...)
- Remove all project specific solution files, there is now only one top level solution file (`CKAN.sln`).
- All build output is now directed to `.build/out` and the ILRepack'd executables are now directed to `.build/repack`. The Travis CI configuration has been updated to reflect these changes.
- Building in both `Debug` and `Release` configurations is now properly supported. `Debug` is default, to build in `Release` use `./build --configuration=Release`. Travis CI builds in both but only deploys in `Release`.
- Move all duplicate assets to top level `assets` directory
- Add CKAN icon to NetKAN executable

## Motivation
- Right now building CKAN and NetKAN on Windows is a bit of a pain, Visual Studio can obviously build it but doesn't do the ILRepack steps to produce a consolidated executable. Setting up a POSIX-like environment to get the standard build script to run is also non-trivial. This should make it trivial to do so by executing `./build` in PowerShell.
- CKAN contributors will certainly have to know C#, but may not know Perl or Python to help with the build system. This makes our build system use the same language as our codebase so anyone contributing to the codebase can also improve the build.
- There's a lot of cruft and rough edges in the way the build is currently done, this is a first step in consolidating _everything_ into One Build To Rule Them All™. Once this in place I can work on getting packages and tools out of source control, create a semantic changelog, and reorganize the projects and their output for clarity.


Dazpoet:
I can only talk about the reasoning in this PR since code isn't really my forté and henceforth I cannot discuss the potential coding issues that might/might not arise down the line.

The reasoning does seem on point to me. Lowering the barrier to entry for contributing to *all parts of CKAN* should be something of a prime directive for us and would work in the best interest of our users since it means more contributors and thereby more fresh ideas and added features.

As a Windows user I've also long envied how easy it seems to build the expected .exe files in a POSIX environment and would love the same ease in my own preferred OS.

mathuin:
If the goal is to easily build CKAN on Windows, I have a PR #1652 which uses Docker to build and run CKAN.  Docker has made great strides in usability on Windows and Mac -- https://docs.docker.com/engine/installation/windows/ shows how it's done.  The Docker Compose file that I use to actually run CKAN would need tweaking for Windows but all you need to do is use the location of your target KSP installation in place of the Linux default which is in that file!

politas:
Ok, so this renames thhe existing "build.sh" used by Gnu make to "build", and adds the build.ps1 script used to kick  off cake on Windows, but for starting cake on Llinux/MacOS, there should be a build.sh (or some renamed version).

Ruedii:
Microsoft now ships a version of GNU Make for Visual Studio as part of an optional package.

We could utilize that to unify the build scripts, and use a tiny shim so that Microsoft's standard make tools hand off the job to GNU Make.  

We might want to (mostly just for fun) see if Mono or Visual Studio .Net do a better job with the build process for official binary.   However, we should probably chose whichever one builds with fewer bugs.

dbent:
> Ok, so this renames the existing "build.sh" used by Gnu make to "build", and adds the build.ps1 script used to kick off cake on Windows, but for starting cake on Llinux/MacOS, there should be a build.sh (or some renamed version).

`build` is that script. It does exactly the same thing as `build.ps1`.

I've updated the PR with some changes including removing every extraneous build script I could find including makefiles and shell scripts and cutting over Travis to use cake for the build.

An "unfortunate" consequence of this is that the most recent versions of Cake require at least a .NET Framework 4.5 compatible environment to work which means Mono>=4.0. I think this is fair and reasonable; Mono 4.0 is two years old and anyone developing for .NET at this point should really be on it. Also, this is the way the entire .NET ecosystem is heading, with the new [.NET Standard](https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/), .NET Framework *4.6.1* is going to become the basis on which every other runtime is based (.NET Framework, .NET Core, Mono, etc.), and in the long term CKAN should really be able to be run on .NET Core.

Note this is **only** a *build* requirement. AFAIK we still target .NET Framework 3.5 for the assemblies so CKAN should continue maybe sorta working on ancient versions of Mono.

politas:
So, in either environment, the command to build is 

    ./build

Is that right? I'll try it out. Sorry to leave this hanging so long; I have been a bit intimidated by build scripting, it's all been a black box to me.

dbent:
> So, in either environment, the command to build is
>    
>    
>    ./build
>    

Yes.

dbent:
Some additional changes:

- Remove all NuGet packages from source control. All packages are now stored in the (unversioned) `.build/lib/nuget` directory and all project references have been fixed to refer to that directory. A `nuget.config` file has also been added which will install packages in that directory by default. Yay for no more duplicate NuGet packages! (`packages/` has also been removed from `.gitignore`).
- Create `lib/` for CurlSharp since it doesn't seem to have an official NuGet release. I want to eventually get away from CurlSharp anyway and consolidate all network code (working on that in another branch...)
- Remove all project specific solution files, there is now only one top level solution file (`CKAN.sln`).
- Update the ILRepack version to the most recent.
- Change the default target name from `Build` to `Default`
- Added alias tasks `Ckan` and `Netkan` so you can just type `./build ckan` or `./build netkan` to build one or the other. `./build` still builds both.

dbent:
Some additional changes:

- All build output has been moved to `.build/out` and Travis CI configuration has been updated accordingly.
- The ILRepack'd executables have been moved to `.build/repack` and Travis CI configuration has been updated accordingly.
- Building in both `Debug` and `Release` configurations is not properly supported. Travis builds in both but only deploys in `Release`. `Debug` is default, to build in `Release` use: `./build --configuration=Release`.
- Updated `.gitignore` to remove entries that are no longer necessary.

politas:
Trying under Windows Powershell:

```
PS C:\Users\polit\Source\Repos\CKAN> .\build.ps1
Program 'NuGet.exe' failed to run: The specified executable is not a valid application for this OS platform.At line:1
char:1
+ C:\Users\polit\Source\Repos\CKAN/.build/tools/nuget/2.8.6/NuGet.exe i ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~.
At line:1 char:1
+ C:\Users\polit\Source\Repos\CKAN/.build/tools/nuget/2.8.6/NuGet.exe i ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ResourceUnavailable: (:) [], ApplicationFailedException
    + FullyQualifiedErrorId : NativeCommandFailed

C:\Users\polit\Source\Repos\CKAN/.build/lib/nuget/Cake.0.17.0/Cake.exe : The term
'C:\Users\polit\Source\Repos\CKAN/.build/lib/nuget/Cake.0.17.0/Cake.exe' is not recognized as the name of a cmdlet,
function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the
path is correct and try again.
At line:1 char:1
+ C:\Users\polit\Source\Repos\CKAN/.build/lib/nuget/Cake.0.17.0/Cake.ex ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\Users\polit\...0.17.0/Cake.exe:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

mathuin:
The Dockerfile currently builds CKAN using these two lines:

```
RUN nuget restore -NonInteractive
RUN xbuild /property:Configuration=Release /property:OutDir=/build/
```

@dbent could you please include a patch to the Dockerfile to replace these lines with whatever is required to build in the brave new Cake world?

dbent:
> Trying under Windows Powershell:

What does...
```
> Get-Item .\.build\tools\nuget\2.8.6\NuGet.exe
```
...output? It should have a length of 1686528.

Or if you're on a recent PS version...
```
> Get-FileHash .\.build\tools\nuget\2.8.6\NuGet.exe
```
...should spit out a SHA256 of A6BE87794A542E3F518B718437443C6C29D379D12CC94BEA7DE787BF05BABF83.

I suspect your download of NuGet borked. `rm -r .build` and try again.

dbent:
> @dbent could you please include a patch to the Dockerfile to replace these lines with whatever is required to build in the brave new Cake world?

I don't use Docker, so it'd just be a blind change with no testing.

The command is simply `./build` in the source root. The ILRepack'd executable would be in:
`.build/repack/<configuration>/ckan.exe` where `<configuration>` is `Debug` by default or `Release` if `./build --configuration=Release` is used.

dbent:
Small additional change:

- Move all duplicate assets to top level `assets` directory
- Change all project references to use this directory
- Add CKAN icon to the NetKAN executable

dbent:
Things I want to finish before taking away the *WIP* tag:

- [ ] Use built-in Cake functions to access git rather than starting to process manually
- [ ] Get rid of `Meta.cs` and replace it with dynamically generated (and standard) `AssemblyInfo.cs` file(s).
- [ ] Add a `Test` target to `build.cake` and make Travis use that instead of calling NUnit directly
- [ ] Add a smoke test for the built executables to make sure they function

Longer term things I want to do but *not* in this PR:

- Update all referenced libraries to latest versions
- Move all projects to underneath a top-level `src` directory
- Make  project paths, names, assembly names, and root namespaces consistent

mathuin:
@dbent I changed the Dockerfile to use ./build and it started well.  Cake was downloaded and installed, and so ere some other packages.  It failed with the `BuildDotNet` task:

```
BuildDotNet
An error occurred when executing task 'BuildDotNet'.
Error: ApplicationName='git', CommandLine='describe --tags --long', CurrentDirectory='/source', Native error= Cannot find the specified file
ERROR: Service 'ckan' failed to build: The command '/bin/sh -c ./build' returned a non-zero code: 1

```

politas:
>I suspect your download of NuGet borked. rm -r .build and try again.

Yep, that was it. Still getting used to the VS way of handling git, too.

dbent:
Some major changes:

Version information is now read from `CHANGELOG.md`. It looks for the first H2 line (one starting with `##` and uses that as the current version number). Therefore instead of using `Unreleased` you should use whatever the next anticipated version number is followed by a prerelease marker, e.g. `v1.22.2-dev`, or `v1.22.2-alpha` or whatever. When getting ready to do a release switch to `v1.22.2`, do release stuff, and then create a new entry for the next anticipated version `v1.22.3-dev` (or whatever). If major changes go in and you anticipate the next version will be `v1.24` then replace the current unreleased version number with `v1.24.0-dev`.

This is the single-source-of-truth for version information and is what is used to generate `GlobalAssemblyVersionInfo.cs` which populates all the .NET metadata with version information. Currently, AssemblyVersion is the *Major.Minor*, AssemblyFileVersion is *Major.Minor.Patch*, and AssemblyInformationalVerrsion is *Major.Minor.Patch[-prerelease][+commit_hash]*. The only constraint on the version string is that it has to be a [semver 2.0.0](http://semver.org/) string. Also the old Meta.cs is dead dead dead.

Also, even/odd minor versions have no significance for indicating "stability" as SemVer includes provisions for such information with the prerelease tag.

> How do you run the tests?

Tests can now be executed with `./build test`, which executes the default build and then runs the NUnit tests. Some of the tests have never been particularly robust when executing in my environment... they run on Travis though. A brief look at the cache tests reveals that there seems to be an assumption that the tests are run sequentially which is bad. Maybe I'll fix those up. Also Travis has been cutover to just execute `./build test --configuration=$BUILD_COFIGURATION --exclude=FlakyNetwork`.

politas:
Previously, the release process was taking the version number from the GitHub Release tag; you've replaced that code? This would involve updating https://github.com/KSP-CKAN/CKAN/wiki/Releasing-a-new-CKAN-client

I tried `./build test` and it failed. From what I saw of the Cake demos, they seem to suggest that you can make a "release" build dependent on a successful "test" run? That sounds like a good idea.

dbent:
@politas

> Previously, the release process was taking the version number from the GitHub Release tag; you've replaced that code?

Yes.

> I tried `./build test` and it failed.

Not surprising, like I said quite a number of tests aren't particularly robust.

> From what I saw of the Cake demos, they seem to suggest that you can make a "release" build dependent on a successful "test" run? That sounds like a good idea.

Travis builds the cross product of mono versions (latest, 4.2.4, 4.0.5) and build configurations (Debug, Release) at all times. It will only *deploy* to GitHub when the follow conditions are met:

- The build was successful. This includes all tests passing.
- The commit has been tagged.
- The build configuration is `Release`
- The mono version is the most recent (currently hardcoded at 4.6.2), but I'd either like to make that more dynamic or fix the specific versions of Mono we test so we always know exactly which versions are being used.

All this is in the `deploy` section of `.travis.yml`. I'll write a comprehensive guide for building when I'm done with this PR.

@mathuin 

> I changed the Dockerfile to use ./build and it started well. Cake was downloaded and installed, and so ere some other packages. It failed with the BuildDotNet task:

Looks like it can't find git in the PATH. Git was a build time dependency to get the version information that was baked into the assemblies. I've changed it so now it's not *strictly* necessary, but git is still used to add a short commit hash to the version information, which is useful. Try again, it shouldn't error out if git isn't present but it's still good to have.

dbent:
I've added some basic smoke tests for the CKAN and NetKAN executables that just run the executables with the version command `ckan version` and `netkan --version` and ensures that they execute successfully and that the outputted version corresponds to the expected value.

I've also switched from using mono `latest` to an explicit `4.6.2` so the build should be entirely predictable. So when new versions of Mono are released we can add it to the list and update the version checks in the deploy section of `.travis.yml`.

I consider this complete.

politas:
@techman83, have you been keeping up with this? I'd like to get your input before merging, but this is all looking pretty good to me.

techman83:
@politas tangently. I'll take a quick look see this arvo, but in general it seems quite sane.

dbent:
Added some [documentation](https://github.com/KSP-CKAN/CKAN/blob/feature/cake/doc/building.md) for the build. Got pretty tired at the end there, not entirely sure it makes sense, but eh.

dbent:
I should probably note somewhere that building in Visual Studio/MonoDevelop won't work until at least one `./build` is done from the command line (or at the very least `./build generate-globalassemblyversioninfo` if you want to be super pedantic).

Makefile
```diff
@@ -1,29 +0,0 @@
...
-
-test: build
-	nunit-console --exclude=FlakyNetwork build/Tests/bin/Debug/CKAN.Tests.dll
-
-# This likely only works under Unix, but then again Unix folks are the only
-# ones likely to have `fortune` installed. This target is for amusement purposes
-# only.
-fortune: quotes.txt
-	strfile quotes.txt
-
```
techman83: Awww, we might need to re-implement our fortune :-)

doc/building.md
```diff
@@ -0,0 +1,222 @@
...
+- The built executables (`ckan.exe` and `netkan.exe`) are uploaded to Amazon S3 if the following conditions are met:
+    - All previous steps were successful
+    - Its on the `master` branch
+    - The build configuraton is `Release`
+    - The mono version used is the latest (currently `4.6.2`)
+- A GitHub release is generated if the following conditions are met:
+    - All previous steps were successful
+    - The commit is tagged
+    - The build configuration is `Release`
+    - The mono version used is the latest (currently `4.6.2`)
```
techman83: <3 Documentation

politas:
Okay, this is very strange. running ./build on Linux:

```
$ rm -r .build
$ ./build
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1647k  100 1647k    0     0   168k      0  0:00:09  0:00:09 --:--:--  254k
Installing 'Cake 0.17.0'.
Successfully installed 'Cake 0.17.0'.
Error: Could not locate nuget.exe.
$ ls -l ./.build/tools/NuGet/2.8.6/
total 1648
-rw-rw-r-- 1 politas politas 1686528 Jan 14 22:57 NuGet.exe
```
I cannot figure out why it throws that error.

dbent:
Try again, the build was accidentally dependent on NuGet being in PATH.

politas:
The output seems excessive to me; the signal is buried in the noise.
http://pastebin.com/ceiJJiyd

And it doesn't seem to have created new .exe files.

```
politas@copper:/.../CKAN$ ls .build/out/CmdLine/Debug/bin/
Autofac.dll               CKAN-GUI.exe        CmdLine.exe.mdb              INIFileParser.dll
ChinhDo.Transactions.dll  CKAN-GUI.exe.mdb    CommandLine.dll              log4net.dll
CKAN.dll                  CmdLine.exe         CurlSharp.dll                log4net.xml
CKAN.dll.mdb              CmdLine.exe.config  ICSharpCode.SharpZipLib.dll  Newtonsoft.Json.dll
politas@copper:/.../CKAN$ ls .build/out/
AutoUpdater/ CKAN/        CKAN-GUI/    CKAN.Tests/  CmdLine/     NetKAN/      
politas@copper:/.../CKAN$ ls .build/out/CKAN/Debug/bin/
Autofac.dll               CKAN.dll      CommandLine.dll  ICSharpCode.SharpZipLib.dll  Newtonsoft.Json.dll
ChinhDo.Transactions.dll  CKAN.dll.mdb  CurlSharp.dll    log4net.dll
politas@copper:/.../CKAN$ ls .build/out/NetKAN/Debug/bin/
Autofac.dll               CKAN.dll.mdb        CurlSharp.dll                NetKAN.exe         Newtonsoft.Json.dll
ChinhDo.Transactions.dll  CommandLine.dll     ICSharpCode.SharpZipLib.dll  NetKAN.exe.config
CKAN.dll                  curl-ca-bundle.crt  log4net.dll                  NetKAN.exe.mdb
politas@copper:/.../CKAN$ ls -l
total 8404
...
-rwxrwxr-x  1 politas politas 6364160 Jan 13 05:47 ckan.exe
-rw-r--r--  1 politas politas     169 Jan 13 05:47 ckan.exe.config
...
drwxr-xr-x 12 politas politas    4096 Jan 15 02:54 Netkan
-rwxrwxr-x  1 politas politas 1649152 Jan 13 05:47 netkan.exe
...
-rw-r--r--  1 politas politas  172919 Jan 13 05:47 TestResult.xml
drwxr-xr-x  9 politas politas    4096 Jan 15 02:54 Tests
```

politas:
Ah, in the .build/repack/Debug directory - That's a bit hard to find.

Ok, now I'm getting my head around the philosophy, can I suggest that the top level directory be called just "build" rather than ".build"? Making it a hidden directory is confusing for Linux users

dbent:
> The output seems excessive to me; the signal is buried in the noise.

Eh, I'd rather have useful information at a glance. Plus if the build errors it'll pretty much stop right at the error. All green at the end means its good. This is par for course in the .NET world, although I know the Unix philosophy is "no output means success".

> Ah, in the .build/repack/Debug directory - That's a bit hard to find.

I'd be open to a final build step that copies them directly below `.build`. I certainly don't think they belong in the top-level though like the build was previously doing.

> can I suggest that the top level directory be called just "build" rather than ".build"?

That'd conflict with the `build` script. How about `_build`?

Dazpoet:
If it's just for POSIX, can't you just name the output Build since capital letters are differentiated? Or will that screw over us poor Win users?

politas:
_build sounds fine. Build would conflict for Windows.

politas:
Weird error on Windows:

```
Test-UnitTests+Only
An error occurred when executing task 'Test-UnitTests+Only'.
Error: NUnit: Unrecognised error (exit code -2146232576).
```

And then I got a dialog asking whether I wanted to install .Net 4.5, which is odd for a Windows 10 installation with Visual Studio running.

politas:
We're definitely moving forward from  `Error: The target 'test' was not found.` which was what I was getting when trying to build test on Windows before!

politas:
With the .Net 4.5 installation, it's running the tests.

15 failed tests on Windows, 0 failed tests on Linux. Looks like it's largely down to  \ versus / in paths. That can be fixed later.

mathuin:
I am almost done with a patch for the Dockerfile.  @dbent would you prefer I give you the new lines here or would you rather a PR against your fork?

mathuin:
Spoke too soon. :-(

```
Unhandled Exception:
System.IO.FileNotFoundException: Could not load file or assembly 'CKAN, Version=1.22.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies.
File name: 'CKAN, Version=1.22.0.0, Culture=neutral, PublicKeyToken=null'
[ERROR] FATAL UNHANDLED EXCEPTION: System.IO.FileNotFoundException: Could not load file or assembly 'CKAN, Version=1.22.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies.
File name: 'CKAN, Version=1.22.0.0, Culture=neutral, PublicKeyToken=null'
```

Any idea what's wrong?

Makefile
politas: Hmm, is it worth importing a library? - https://github.com/juusimaa/FortuneSharp

politas:
>`System.IO.FileNotFoundException: Could not load file or assembly 'CKAN, Version=1.22.0.0, `
>Any idea what's wrong?

Why is it trying to build 1.22.0? This branch is at 1.22.1 (or 1.22.2 Pre)

dbent:
> 15 failed tests on Windows, 0 failed tests on Linux. Looks like it's largely down to \ versus / in paths. That can be fixed later.

Both forward and back slashes are valid path separators on Windows, anything with a `/` should work fine on Windows.

> I am almost done with a patch for the Dockerfile. @dbent would you prefer I give you the new lines here or would you rather a PR against your fork?

PR.

> Any idea what's wrong?

No idea without knowing the exact operation(s) being performed. Is this trying to execute `.build/repack/ckan.exe`?

> Why is it trying to build 1.22.0? This branch is at 1.22.1 (or 1.22.2 Pre)

That number is the [`AssemblyVersion`](https://msdn.microsoft.com/en-us/library/system.reflection.assemblyversionattribute%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396) which is used by the runtimes to determine compatibility for dependency resolution. If we follow semver versioning rules (which we really should) then patch releases (ones that only change the 3rd version number) should not change the public API of an assembly. Therefore all 1.22.x releases should be programmatically compatible and interchangeable. To get this behavior the `AssemblyVersion` only stores the major (1st) and minor (2nd) version numbers. The [`AssemblyFileVersion`](https://msdn.microsoft.com/en-us/library/system.reflection.assemblyfileversionattribute(v=vs.110).aspx) stores the major, minor, and patch (3rd) version numbers. The [`AssemblyInformationVersion`](https://msdn.microsoft.com/en-us/library/system.reflection.assemblyinformationalversionattribute(v=vs.110).aspx) which is just a freeform string stores the "full" version: major, minor, patch, prerelease, and build metadata (short commit hash).

Or:

- `AssemblyVersion`: 1.2.0.0
- `AssemblyFileVersion`: 1.2.3.0
- `AssemblyInformationalVersionAttribute`: 1.2.3.0-dev+abcdef

The [`Version`](https://msdn.microsoft.com/en-us/library/system.version.aspx?f=255&MSPPError=-2147217396) class stores a 4th number, build, which is not used in semver so is always zero.

dbent:
New changes:

- I've changed the build directory from `.build` to `_build`
- The cake script will now spit out a random quote from `quotes.txt` after running.

dbent:
Pushed another change which will copy `ckan.exe` and `netkan.exe` to directly below `_build` after repack.

This is **only** for user convenience. Automated scripts should still get the files from `_build/repack/$CONFIGURATION`.

mathuin:
Figured out that error.  The Docker container relies on `CmdLine.exe`, which does not appear to be built anymore.  Has it been replaced with `ckan.exe`?  A cursory test makes me think yes, but I would like to know for sure before I send you a pull request.

politas:
ckan.exe has always been the actual final executable after the ilrepack.

mathuin:
Ah, TIL.  I found the other one and it always worked for me! :-)

Created that pull request #1988 -- let me know if there's anything else I can do.

dbent:
I've made some minor changes that make some problematic test fixtures more robust. As far as I can tell they should not be 100% reliable through ReSharper test runner and the CLI test runner. This means you should be able to `./build test` on Windows to your heart's content (keeping in mind that unit testing is about 50% of the build time).

> Created that pull request #1988 -- let me know if there's anything else I can do.

Cool.

doc/building.md
```diff
@@ -0,0 +1,221 @@
...
+    - The `--exclude=FlakyNetwork` argument is used to not execute tests that rely upon a network connection and could
+      thus behave nondeterministically (frustrating behavior for a CI system).
+- The built executables (`ckan.exe` and `netkan.exe`) are uploaded to Amazon S3 if the following conditions are met:
+    - All previous steps were successful
+    - Its on the `master` branch
+    - The build configuraton is `Release`
+    - The mono version used is the latest (currently `4.6.2`)
+- A GitHub release is generated if the following conditions are met:
+    - All previous steps were successful
+    - The commit is tagged
```
politas: Can you expand on this? Is this a tag we put on the GitHub PR when merging, a tag on the git commit, or a tag in CHANGES.md? I can't find anything that controls this in travis.yml (though I admit to not fully understanding the syntax)

It might be good to be able to create prereleases on GitHub as well as full releases. (Though we'd need to add an option in the client Auto-update code to opt-in to pre-releases) - I'm cool with adding this functionality later.

politas:
#1989 seems relevant to this, too. I'm not sure what we'd need to do to include this for Mac users.

doc/building.md
dbent: > Is this a tag we put on the GitHub PR when merging, a tag on the git commit

These are the same thing. The actual mechanics of a release haven't changed. The only thing different is that the version number is now taken from CHANGES.md rather than the tag itself. The tag can literally be anything so long as it exists. But for consistency it ought to match the version in CHANGES.md.

dbent:
> #1989 seems relevant to this, too. I'm not sure what we'd need to do to include this for Mac users.

No idea, last time I used a Mac in earnest was before the turn of the millennium.

doc/building.md
dbent: > It might be good to be able to create prereleases on GitHub as well as full releases.

Just tag a prerelease.

politas: The [current method](https://github.com/KSP-CKAN/CKAN/wiki/Releasing-a-new-CKAN-client) is to go to the releases page and create a new release there. So we still do that, but now the version number reported by the client will come from CHANGELOG.md. Ok, I get it. At some point I must grokify how that actually happens.

politas:
@dbent, are you going to merge in #1988? I think I'm cool to merge the whole PR once that's in. I'm looking forward to rebasing my #1971 so I can debug the failing test.

dbent:
> @dbent, are you going to merge in #1988?

Done.

mathuin:
Docker works as expected.  LGTM!

politas:
Ooof! That was more of a headache than I expected. Must remember to merge upstream changes first!

PR #1592, opened 2016-02-17
Added support for spacedock to netkan

Zane6888:
This is basically just the old Kerbalstuff code refactored to spacedock. 

I'm not sure if there is something else to fix to allow CKAN to install mods from spacedock like it used to  do for kerbalstuff, but this PR covers the netkan.exe.

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -152,6 +152,11 @@
                     "type" : "string",
                     "format" : "uri"
                 },
+                "spacedock" : {
+                    "description" : "Project on Spacedock",
+                    "type" : "string",
+                    "format" : "uri"
+                },
                 "manual" : {
                     "description" : "Mod's manual",
                     "type" : "string",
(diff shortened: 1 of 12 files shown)


dbent:
Probably want to also copy `Tests.NetKAN.Transformers.KerbalstuffTransformerTests` and `Tests.NetKAN.Sources.Kerbalstuff.KerbalstuffApiTests`.

Olympic1:
#1593 will add these changes + some extras

pjf:
@Zane6888 : Thank you *so much* for this. I can't tell you how delighted we are to have two independent PRs come in to add support for Spacedock.

While I've merged the changes in #1593, as they come with updates to the spec and other supporting files, I absolutely want to encourage you to continue to submit contributions, and thank you very much for your efforts in this one.

Keep being amazing,

~ Paul

Issue #1595, opened 2016-02-18
Kerbal Stuff


Ricochetv1:
Ummm.... you been to Kerbalstuff lately? It's gone. He left a torrent link behind for all the mods on there, but it needs 62GB... We should probably find a way someone else can host those files and redirect ckan to that store for now...

plague006:
Being addressed in https://github.com/KSP-CKAN/CKAN/issues/1586.

Shuudoushi:
Do people even look in IRC before posting this stuff?

Ricochetv1:
People still use IRC?

politas:
Yeah, expecting people to check IRC (when they quite possibly have never heard of it) is a bit unreasonable. Checking existing issues, on the other hand, is a perfectly reasonable expectation.

PR #1596, opened 2016-02-18
Create an issue template

plague006:
If someone wants to change something feel free, I just saw that [GitHub added this](http://venturebeat.com/2016/02/17/github-launches-templates-for-new-issues-and-pull-requests/) and after closing https://github.com/KSP-CKAN/CKAN/issues/1595 this felt cathartic.


Dazpoet:
Sounds amazing to me :)

Will people just get this at the end of their issues thoguh so we will get spammar with this template forever after a merge?

plague006:
I haven't had occasion to test it, but from the article it looks like every issue opened contains the text in the field. If we repeatedly see people not removing the text we can always just delete the template.

Dazpoet:
Sounds good to me

plague006:
With it merged, it is indeed real text.. shame that it's not HTML5 placeholder text.

Dazpoet:
Better than nothing imo :+1:

Issue #1152, opened 2015-06-22
Match Spec with Wiki


janbrohl:
On https://github.com/KSP-CKAN/CKAN/wiki/Sharing-a-modlist-%28metapackages%29 metapackages are mentioned as added in v1.6 while there is nothing about that in the Spec which is inconsistent if metapackages should be valid ckan-files.

Is it right that one of "download" and "kind" is required? Why not just make "download" optional?

PR #1597, opened 2016-02-18
Add `kind` to the spec

plague006:
Because it's a.. kind thing to include everyone. (I'm so sorry, I'm really tired)

Closes https://github.com/KSP-CKAN/CKAN/issues/1152

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -212,6 +212,18 @@
         "license",
         "version"
     ],
+    "oneOf": [
+        {
+            "required": [ "download" ]
+        },
+        {
+            "properties": {
+                "kind": {
+                    "enum": [ "package", "metapackage" ]
+                }
+            }
+        }
+    ],
     "definitions" : {
         "identifier" : {
             "description" : "Unique identifier for this mod.",
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -209,7 +209,6 @@
         "name",
         "abstract",
         "identifier",
-        "download",
         "license",
         "version"
     ],
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -219,7 +219,7 @@
         {
             "properties": {
                 "kind": {
-                    "enum": [ "package", "metapackage" ]
+                    "enum": [ "metapackage" ]
                 }
             }
         }
--- a/Spec.md
+++ b/Spec.md
@@ -152,7 +152,7 @@ with any version and the `.dll` suffix removed.
 ##### download
 
 A fully formed URL, indicating where a machine may download the
-described version of the mod.
+described version of the mod. Note: This field is not required if the `kind` is `metapackage`.
 
 ##### license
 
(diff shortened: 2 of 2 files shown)


plague006:
A quick and easy line to drop in the changelog for whoever merges this:
````
- [Spec] Updated Spec with the `kind` field which was introduced in v1.6. (#1597 by: plague006; reviewed: )
````

**Comment by Zane6888** [comment]
You might want to update the CKAN.schema as well, any file not containing a download field will currently fail to validate against the schema.

**Comment by plague006** [comment]
Good idea. I don't suppose there's a way to specify `download or kind`?

**Comment by Zane6888** [comment]
There is a way using `"oneOf"`, you can even go one step further and specify `download or kind: "metapackage"` by adding the following code to the schama: 

` "oneOf":[{required:["download"]},{properties: {kind: {enum: ["metapackage"]}}}]`

**Comment by Olympic1** [comment]
@Zane6888 I edited your post to make the code more readable.

Also doesn't an `enum` method need at least 2 variables?

**Comment by Zane6888** [comment]
The spec states that the array "MUST have at least one element".

[http://json-schema.org/latest/json-schema-validation.html#anchor76](url)

**Comment by Dazpoet** [comment]
After some dwelling about in documentation this makes sense to me and seeing as this has been around since, not kidding, [Laika](https://github.com/KSP-CKAN/CKAN/releases/tag/v1.6.0) I'm sure @Ippo343 just forgot about adding it back then so I'm merging this.

**Comment by techman83** [comment]
The schema change seems to be breaking valid metadata.

Validating ../NetKAN/KerbodynePlus-1.051.ckan.. Failed! 
```
{
    "spec_version": "v1.4",
    "identifier": "KerbodynePlus",
    "name": "Kerbodyne Plus",
    "abstract": "An end-game parts pack.",
    "author": "BonusEventus",
    "license": "CC-BY-NC-SA-3.0",
    "resources": {
        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/130809-105-kerbodyne-plus",
        "spacedock": "https://spacedock.info/mod/300/Kerbodyne%20Plus",
        "repository": "https://github.com/bonus-eventus/kerbodynePlus",
        "x_screenshot": "https://spacedock.info/content/BonusEventus_872/Kerbodyne_Plus/Kerbodyne_Plus-1460182481.8785317.jpg"
    },
    "version": "1.051",
    "ksp_version": "1.1.0",
    "conflicts": [
        {
            "name": "KerbodynePlus01Alpha"
        }
    ],
    "install": [
        {
            "file": "kerbodynePlus/GameData/kerbodynePlus",
            "install_to": "GameData",
            "filter": ".DS_Store"
        },
        {
            "file": "kerbodynePlus/README.md",
            "install_to": "GameData/kerbodynePlus"
        }
    ],
    "download": "https://spacedock.info/mod/300/Kerbodyne%20Plus/download/1.051",
    "download_size": 34089620,
    "x_generated_by": "netkan"
}
```

Netkan currently isn't indexing anything, I'll revert this for now. http://status.ksp-ckan.org/

**Comment by Zane6888** [comment]
I found what kept Netkan from indexing most files. `"oneOf"` works like xor and `"properties"` only defines fields but doesn't state anything about there existence, therefore a .ckan containing a download field would match against both schema and the `"oneOf"` would fail to match as a whole.


Explicitly requiring `"kind"` and `"not"` requiring `"download"`should fix the problem.
```
"oneOf": [
    {
        "required": [ "download" ]
    },
    {
        "properties": {
            "kind": {
                "enum": [ "metapackage" ]
            }
        },
        "not": {
            "required": [ "download" ]
        },
        "required": [ "kind" ]
    }
]
```

Issue #1591, opened 2016-02-16
Suggestion: Torrent based cache sharing plugin


raziel420:
Using torrents has shown time and time to reduce load on centralized servers (such as kerbal stuff, sorry for the hug of death).  Many people are against using torrents mainly due to piracy accusations associated with torrents.

I think torrents would be a great way to distribute load and reduce bandwidth usage on servers, the hash function of torrents are typically good enough to maintain versioning.  Using a centralized server for listing the torrent files (or more commonly used magnet links these days) can keep files accurate.  Using caches of people who already have it downloaded of course will mean less data the central servers give out.

Due to concerns of people over torrent usage, rather then have it even in the default client I advise against, but ckan does support plugins, so a plugin designed to utilize torrents available would be a good alternate resource for those willing to use them.  I would even suggest that if the plugin is intalled it uses torrents as a priority over the central repository.

ChucklesTheBeard:
I have started in on something like this, feature branch is here:
https://github.com/ChucklesTheBeard/CKAN/tree/feature/torrent

From what I can tell, it's not possible to do this entirely in a plugin, because it requires extra metadata, and the classes it would need to rewrite aren't overloadable.

In order to download a torrent, you need, at minimum, a hash of the .torrent file and the size of the target (this is how `magnet:` links work, with xt=urn:btih:<hash>&xl=<bytes>). That means either adding an optional hash field to the .ckan spec, or keeping a separate repository of hashes matched to .ckan files.

There are no up-to-date / actively developed implementations of the bittorrent protocol in C#, but there are an abundance of bittorrent clients on every platform. Why write a bunch of complicated netcode into CKAN, when others have already done it better? I'm planning on generating a magnet link, then calling `Process.Start("magnet link here")` and watch the (user-configured) directory for the finished downloads. Http sources from the download field can be added to magnet links as a backup when there are no seeders.

Ghostt1:
[...]

Issue #783, opened 2015-03-05
Torrent downloads


Ippo343:
I was wondering how hard would it be to add torrent support within ckan.
The idea is to ease the bandwidth and infrastructure costs from community hosting like KerbalStuff.
It also provides an easy and painless solution to setting up multiple mirrors.
The download field could be extended to accept a magnet link.

I wonder how stupid this idea really is.

erendrake:
I dont think its stupid. it would be kinda cool to give people the ability to quickly host a mirror

Ippo343:
Update: it appears that Mono has an X11/MIT torrent libray we could use.
I will give it a more in-depth look as soon as I can :)

TeddyDD:
That would be really awesome!

pjf:
The idea is sane. However #935 is achievable at the moment, especially since we have infrastructure funding, so I'd suggest we head towards that first.

I'm working on getting  our tickets under control, so I'm closing this out with the expectation that #935 would solve most or all of our bandwidth and throttling issues. However please do feel free to re-open if you feel we should be keeping this around for the future. :)

PR #1604, opened 2016-02-21
WIP: Feature/torrent

ChucklesTheBeard:
Needs review + multi platform testing.

Ideally, it also needs unit testing, but I'm not sure how to implement that for torrents... maybe some checks against a known-good .torrent file, I guess...?

---

This PR is an implementation of a torrent-based download feature that has been requested at least twice in issues (#783, #1591), and multiple times in IRC. With this feature, mods can be distributed in the absence of any file hosts, as long as any one person will seed them.

---
### When NetKAN generates the `.ckan` for a mod,

it can optionally (`-T...` / `--generate-torrents=...`) generate everything needed to download the file over the BitTorrent protocol - including the download field as an http seed, if available.
1. A hex string to be used in a magnet link, stored in the `.ckan` file as "btih" (BitTorrent Information Hash).
2. A .torrent file to a configurable directory, to allow for immediate seeding (this isn't strictly necessary for the official NetKAN instance, but it allows users to set up automated seedboxes if they want).
### When a user goes to install a mod,

if the feature is enabled, a magnet link will be generated (and opened) for any mods with a `btih` field, and the rest will be downloaded the same way they always have. Torrent clients generally let you specify where to put completed files, which works out swimmingly - users have the option of configuring where CKAN looks for completed downloads in the GUI, on the settings page. Once the download completes, it's added to the cache and everything proceeds as usual.
### Opt-In

Everything about this is opt-in - mods without licenses which allow redistribution are ignored. If a user doesn't want to share bandwidth, they can configure their external torrent client however they'd like. It doesn't even run unless a user sets the "download complete" directory.
### FAQ:

Q: How do we prevent users from getting malware? / Is this secure?

A: The `btih` string is a SHA-1 cryptographic hash of the .torrent file, which can be reproducibly generated from the .netkan (well, as long as the download url stays static). If you can break SHA-1, you have better things to do with your time than upload malware. It's significantly more secure (as in, "the resulting file is the same as what I asked for") than downloading over HTTP, which does no verification whatsoever.

Q: Don't you need a tracker or something?

A: Trackers speed up the process somewhat, but any client with a [DHT feature](https://en.wikipedia.org/wiki/Mainline_DHT) doesn't need them. Technical spec for the DHT used in the BitTorrent protocl is [here](http://www.bittorrent.org/beps/bep_0005.html), for the curious.

Q: Will this be fast?

A: Depends on how many people use it. More peers = more available upload bandwidth. The .torrent files have a "web-seed" with the original download link, so as a fallback, if the client can't find any peers, the torrent client can still get the file from the original host.

Q: Who will host the original .torrent files?

A: Anyone with access to the netkan metadata and netkan.exe can reproduce the .torrent, which is ideal for seedboxes, because generating the .torrent does require downloading the original file anyway. Generally, clients which have already downloaded the file (and leaves the client open) also supply a copy of the .torrent over the DHT.

I think I can supply .torrent files without seeding actual content, but I'm not sure.

---
#### Todo list
- [x] https -> http for web seed compatibility
- [ ] Discuss: should netkan / ckan spec versions be bumped?
- [ ] Write unit tests
- [ ] add monotorrent.dll to netkan build scripts
  - Assistance required - ilrepack apparently hates monotorrent.dll...
    - apparently monotorrent tries to look at assembly information, which isn't available?
- [x] ckan.exe crash on updating new ksp install - https://gist.github.com/ChucklesTheBeard/8fcdff49fcd96e848ba9


ChucklesTheBeard:
rebased onto master because this PR contained some commits which were addressed in #1602.

Dazpoet:
I am amazed someone finally wrote a PR for this :+1: 

I think this could be one of the ways we could alleviate the effekts of a potential host shutdown in the future.

techman83:
@ChucklesTheBeard the internet archive generates Torrent links for all files uploaded to it. The added benefit is that we would then be only re-distributing mods that allow it.

https://archive.org/details/SpaceshuttleSRB-0.5

miwilc:
@ChucklesTheBeard can't you use SHA-256 for torrent checking?

PR #1608, opened 2016-02-23
Add Curse as a $kref source

airminer:
I implemented a basic Curse scraper using the http://widget.mcf.li/ Json API.
The syntax for the new $kref is:
"$kref" : "#/ckan/curse/**curseID**"
where **curseID** is the number that can be seen in the curse.com urls of the mods.  

For example: the curse.com url of MechJeb is:
[http://www.curse.com/ksp-mods/kerbal/**220221**-mechjeb](http://www.curse.com/ksp-mods/kerbal/220221-mechjeb)
thus its $kref is:
"$kref" : "#/ckan/curse/**220221**"
and its API-link is
[https://widget.mcf.li/project/**220221**.json](https://widget.mcf.li/project/220221.json)

The API I'm using doesn't expose all information present on the web interface (eg. Only one compatible version is reported, no source reported, etc.), and this plus information could be relatively easily extracted using the (Ms-PL licensed) [Html Agility Pack](https://www.nuget.org/packages/HtmlAgilityPack), however this plus information is generally non-essential, and may not be worth the additional effort.

Some quite important information however is not reported on either interface, namely the name and version of the mods. The name is not reported in the sense that many authors change it to indicate the most recent version.

I think I managed to relatively reliably isolate the name (from the title) and version (from the names of the zip files), however my approach in both cases is rather primitive.

To support the license choices available on Curse, I also added the "AFL-3.0", "AGPL-3.0", "APSL-2.0", "MPL-1.0", "Ms-PL" and "Ms-RL" OSI approved open-source licenses to the accepted licenses list.

This PR probably needs more testing and review, as I wasn't able to test it much aside from some sample .netkan files on windows in visual studio.

Any and all feedback is welcome.


Dazpoet:
I can't comment on the code but to err on the side of safety shouldn't custom license default to restricted so we don't end up with a less restricted string than the intention is by the author?

Olympic1:
Everything looks good but `netkan.exe` gives an error while executing [CurseTransformer.cs - Line 38](https://github.com/KSP-CKAN/CKAN/pull/1608/files#diff-49337ec655542f797e82ec67856458f6R38)

```
2519 [1] INFO CKAN.NetKAN.Transformers.CurseTransformer (null) - Executing Curse transformation with #/curse/220221
log4net:WARN Exception while rendering format [Found Curse Mod: {0} {1}]
System.FormatException: Index (zero based) must be greater than or equal to zero and less than the list of arguments.
   at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.Format(IFormatProvider provider, String format, Object[] args)
   at log4net.Util.SystemStringFormat.StringFormat(IFormatProvider provider, String format, Object[] args)

11867 [1] INFO CKAN.NetKAN.Transformers.CurseTransformer (null) - <log4net.Error>Exception during StringFormat: Index (zero based) must be greater than or equal to zero and less than the list of arguments. <format>Found Curse Mod: {0} {1}</format><args>{MechJeb}</args></log4net.Error>
```

Other than that `netkan.exe` succesfully transforms the `.netkan` file into a `.ckan` file.

**This `.netkan` file**
```
{
    "spec_version": "v1.4",
    "identifier": "MechJeb2",
    "$kref": "#/ckan/curse/220221",
    "abstract": "Curse doesn't have an abstract or mod description.",
    "install": [
        {
            "find": "MechJeb2",
            "install_to": "GameData",
        }
    ]
}
```
**Gives this `.ckan` file**
```
{
    "spec_version": "v1.4",
    "identifier": "MechJeb2",
    "name": "MechJeb",
    "abstract": "Curse doesn't have an abstract or mod description.",
    "author": "r4m0n",
    "license": "LGPL-3.0",
    "resources": {
        "curse": "http://www.curseforge.com/projects/220221/",
        "x_screenshot": "https://media-curse.cursecdn.com/attachments/110/952/18b6dda728d709420f4b8959464e32ba.png"
    },
    "version": "2.5.5.0",
    "ksp_version": "1.0.5",
    "install": [
        {
            "find": "MechJeb2",
            "install_to": "GameData"
        }
    ],
    "download": "http://addons-origin.cursecdn.com/files/2269/153/MechJeb2-2.5.5.0.zip",
    "download_size": 3011096,
    "x_generated_by": "netkan"
}
```

airminer:
@Olympic1 Thanks for the pull request. Unfortunately, I couldn't reproduce the error you encountered. I assume your pull request fixed it?

dbent:
Good job, a few things:

1. Add a redirect limit to `ResolveRedirect`, we can potentially enter an infinite loop there. Worse yet one that could be triggered by a 3rd party. Really, we shouldn't permit more than 3 redirects before erroring out.
2. As far as users and mod authors are concerned widget.mcf.li is an untrusted 3rd party, if we're going to use it for data we have a duty to at least add some modicum of verification to the most sensitive part: the download. My recommendation is to add a whitelist of acceptable hostnames for the download URL to have and to throw a big fat exception if it isn't on that list. Use [`Uri.Host`](https://msdn.microsoft.com/en-us/library/system.uri.host%28v=vs.110%29.aspx) don't bother trying to parse it yourself.
3. Don't bother trying to parse the version or the name intelligently. Modders use *wayyyy* too many different schemes to make it particularly reliable. Use the strategy we used for Jenkins in #1512, add an option to dump the `name` verbatim as the value of `version` and then rely on `x_netkan_version_edit` to strip it. Otherwise we expect and prefer AVC `.version` files.

DuoDex:
@dbent That widget is maintained by a Curse employee (though in an unofficial capacity). If it starts giving sketchy links, there is probably something going on elsewhere - it's just as reliable as KS was. Though I agree with the idea of a domain whitelist.

dbent:
@DuoDex The difference with KS is that at least KS was in the "trust chain", users trust the modder, the modder trusts KS. widget.mcf.li (even if maintained by someone at Curse) is unknown to both modders and users. If KS started throwing back sketchy things CKAN's involvement would be a non-issue. If widget.mcf.li starts throwing back sketchy things user and modders would blame us (*and rightly so*) for trusting some third party they weren't involved with at all.

DuoDex:
@dbent Put it in an obvious place that Curse indexing is powered by that widget, and link to its source.  There is absolutely no reason to suggest that KS/Spacedock is any better than this widget. They're both open-source, community maintained initiatives.

dbent:
@DuoDex The only thing I'm suggesting is that neither the end user nor the modder are associated with widget.mcf.li in any way, that's not disputable. Therefore the responsibility is on *us* to make sure the data we get from it is accurate. If KS/SpaceDock throws back garbage we're just using the source the modder chose. If GitHub throws back garbage we're just using the source the modder chose. If the modder's own Jenkins server throws back garbage we're just using the source the modder chose. If widget.mcf.li throws back garbage we're involving a third party neither the user nor modder know anything about.

We already have a credibility problem with a large contingent of modders, adding in another potential source of errors that modders themselves are wholly ignorant of won't help the situation. A disclosure won't help because the people that would care (mod authors wary of CKAN's auto-indexing) won't see it.

DuoDex:
> A disclosure won't help because the people that would care (mod authors wary of CKAN's auto-indexing) won't see it.

If someone isn't reading your disclosures, you have no business indexing their mods.

airminer:
@DuoDex @dbent Guys, this is a moot point, because I can generate the download links from just the modID and fileID. In fact, download will fail if the widget supplies a fileID that is not associated with the project in question, so even if it goes rouge, it will at most break the download link, but *never* download anything other than the mod author uploaded. Will submit code in a bit.

As for redirecting, Curse seems to be quite fond of redirecting you all over the place. With the link we're currently using, it's around 6 redirects to get to the project page, but I can get it down to around 3 by generating the base link myself. However, I would still put the redirect cap a bit higher, just in case Curse decides to add even more redirections.

airminer:
Now, as for versioning, I will add an option to force an `x_netkan_version_edit` friendly version output, however since my default parsing algorithm seems to work for most mods on Curse, I wouldn't remove it. It's a lot nicer to have a default behaviour that works in most cases, than to have to replicate this work in every single .netkan file.

airminer:
Also, speaking of `x_netkan_version_edit`, someone should *really* write up all those netkan and ckan tags that are not in the ckan spec.

dbent:
They are in the [Spec](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md).

airminer:
@dbent Huh, sorry my bad. I have been blind.

pjf:
This is totally awesome. Plus I have contacts in Curse who *want* us to index mods from their site, so the timing in perfect. I'll both be reviewing this code and pinging in my contacts from Curse today, because I don't want it to be a surprise when we use this to index *all the things*. :)

pjf:
OMG, we detect curse-styled licenses and convert them to KSP-CKAN spec-friendly ones. This keeps getting better! Thank you!!

Olympic1:
@airminer If you could add some tests, this can then be merged.

airminer:
@dbent Ok, I added three options for controlling the version string as in [](https://github.com/KSP-CKAN/CKAN/pull/1512):

```
"x_netkan_curse": {
    "use_download_name_version": true    //This sets the version string to the name of the download
    "use_filename_version": true                 //This sets the version string to the filename of the download
    "use_curse_id_version": true                 // This sets the version string to "0curse" + the curse id of the file
}
```

These options conflict with each other, so I throw an exception of more than one is used.

The default option does the parsing as before, since it usually arrives at the correct version.

doktorjet:
So... any progress here?

DuoDex:
Let's get this finished!

techman83:
@airminer @dbent @pjf 

Have had a recent review of this, the version extraction from files could do with a couple of tests, but aside from that seems to do what it says on the tin. The code is building download urls and only using the widget for finding the most recent version, if it goes rogue it'll generate invalid links limiting exposure to the users.

It'll need a rebase as DownloadSizeTransformer has been replaced by DownloadAttributeTransformer.

```
{
    "spec_version": "v1.4",
    "identifier": "MechJeb2",
    "name": "MechJeb",
    "abstract": "Curse doesn't have an abstract or mod description.",
    "author": [
        "r4m0n",
        "sarbian",
        "Anatid"
    ],
    "license": "LGPL-3.0",
    "resources": {
        "curse": "http://kerbal.curseforge.com/projects/220221",
        "x_screenshot": "https://media-curse.cursecdn.com/attachments/110/952/18b6dda728d709420f4b8959464e32ba.png"
    },
    "version": "2.5.7.0",
    "ksp_version": "1.1-prerelease",
    "install": [
        {
            "find": "MechJeb2",
            "install_to": "GameData"
        }
    ],
    "download": "http://addons-origin.cursecdn.com/files/2294/994/MechJeb2-2.5.7.0.zip",
    "download_size": 3324519,
    "download_hash": {
        "sha1": "80A944CC31C3E1480663D0276F805E59CC47807F",
        "sha256": "EE9C3449F716FFBD13392A476DBA00D8EFCABFF5B66823D5F6D7280A802B483F"
    },
    "x_generated_by": "netkan"
}
```

Netkan/Transformers/CurseTransformer.cs
```diff
@@ -0,0 +1,263 @@
...
+{
+    /// <summary>
+    /// An <see cref="ITransformer"/> that looks up data from Curse.
+    /// </summary>
+    internal sealed class CurseTransformer : ITransformer
+    {
+        private static readonly ILog Log = LogManager.GetLogger(typeof(CurseTransformer));
+
+        private readonly ICurseApi _api;
+
```
techman83: Will need a name for `x_netkan_override`

```
public string Name { get { return "curse"; } }
```

Netkan/Sources/Curse/CurseFile.cs
```diff
@@ -0,0 +1,111 @@
...
+using Newtonsoft.Json;
+using System;
+using System.Text.RegularExpressions;
+
+namespace CKAN.NetKAN.Sources.Curse
+{
+    public class CurseFile
+    {
+        private static readonly ILog log = LogManager.GetLogger(typeof (CurseFile));
+
```
techman83: This logging declaration isn't utilised, resulting in a warning.

Netkan/Transformers/CurseTransformer.cs
airminer: Done.

politas:
This has been merged, but it seems the usual matching up by GitHub didn't work. Thanks, @airminer!

airminer:
@politas I guess it didn't match up because you rebased it instead of merged?

Anyway, thanks for merging it though I guess there is still some documentation to do. (eg. x_netkan_curse)

politas:
I think it was because I made a local copy of the branch, rebased that, and then merged. I'm still learning all the intricacies of git!

PR #1609, opened 2016-02-24
Make "CKAN favourites list" the default export option (no hard version requirements)

plague006:
Most users don't care about specific versions and aren't getting what they might expect when using the default option. So let's put the more advanced option below while also trying to clarify the difference.


Dazpoet:
This sounds very resonable to me :+1:

pjf:
I agree, this should be the default. Reviewing now.

pjf:
Thank you!!

PR #1613, opened 2016-02-25
Extract the GitHub repository name

Olympic1:
Extract the repository name and transform it to a usable `name` for a `.ckan` file.

GitHub supports `_`, `-` and `.`. When a symbol is invalid it changes to a `-`. If the repo name contains any of those symbols it will replace them with a "space", if it doesn't contain any of those symbols it will search for capital letters and split them using a "space".

**For example this `.netkan`:**
```
{
    "spec_version" : "v1.4",
    "identifier"   : "TalisarParts",
    "$kref"        : "#/ckan/github/Olympic1/TalisarParts",
    "license"      : "CC-BY-SA-3.0",
    "ksp_version"  : "1.0.5",
    "install": [
        {
            "find"       : "TalisarParts",
            "install_to" : "GameData"
        }
    ]
}
```
**Generates this `.ckan`:**
```
{
    "spec_version": "v1.4",
    "identifier": "TalisarParts",
    "name": "Talisar Parts",
    "abstract": "Update of Talisar's Spherical and Toroidal Tanks pack.",
    "author": "Olympic1",
    "license": "CC-BY-SA-3.0",
    "resources": {
        "repository": "https://github.com/Olympic1/TalisarParts"
    },
    "version": "v0.4.2",
    "ksp_version": "1.0.5",
    "install": [
        {
            "find": "TalisarParts",
            "install_to": "GameData"
        }
    ],
    "download": "https://github.com/Olympic1/TalisarParts/releases/download/v0.4.2/CargoTransportationSolutions_v0.4.2.zip",
    "download_size": 1764922,
    "download_hash": {
        "sha1": "30641E5F4A21DE84D68D581379065C9BDDF6D6BC",
        "sha256": "E6DE08AA3F61F2B36FC82E0649130D1BB05933A9EC5E6BC7357B1AB303D49E97"
    },
    "x_generated_by": "netkan"
}
```


pjf:
So to summarise I understand this correctly:

- We can autogenerate human-readable names form github based upon the repo name.
- We do this by:
    - Replacing underscores with spaces, if the repo has underscores OR
    - Replacing dashes with spaces, if the repo has spaces OR
    - Replacing dots with spaces, if the repo has dots OR
    - Splitting on camelCase/PascalCase capitalisation, otherwise.

This seems like a pretty good time-saver, and since we're using `SafeAdd` we're not going to overwrite something that already exists. In the worst case we end up with a silly looking human-readable name and we just add one manually.

Would it be too tricky to add some tests to this code? Because this is the sort of straightforward transforms which are beautifully testable. :) (I'm happy to merge without tests, but ecstatic to merge with them!) 

Many, many thanks for this!

politas:
Merged with thanks and apologies for taking so long.

PR #1593, opened 2016-02-17
Support SpaceDock

Olympic1:
Added `$kref` support for SpaceDock mods. Tested with multiple mods on SpaceDock (new and existing).

`netkan.exe` succesfully transformed `.netkan` files into `.ckan` files. Also compared the existing KS mods with their equivalent on SD.

The new `$kref` also shows up in the GUI.

~~The only thing I'm not sure about is [VersionedOverrideTransformer.cs - L25](https://github.com/KSP-CKAN/CKAN/pull/1593/files#diff-9c0e20a6467ef46703f2a452a0afac1cR25)~~

Closes #1592

PR #1615, opened 2016-02-25
Remove KerbalStuff

Olympic1:
Remove everything KerbalStuff related out of CKAN except for the testing data.

Part of #1593


politas:
Could you do a similar job for Spaceport? Ref #1621

Olympic1:
We don't have anything spaceport related

politas:
I'm pretty sure the Transformer is hanging around. Try that example file in #1621. Netkan goes off searching for a mod on Spaceport, then gives a weird error.

Olympic1:
No, `JSON deserialization error` means there is an error in the JSON file (in this case a `.netkan`).

politas:
Ah, ok. It's just that we have no specific tests for valid kref identifiers, then.

pjf:
This looks good to me. The only thing I can think of that might cause hiccups is if a .netkan file has an explicit `kerbalstuff` URL in it, in case case I'm not sure how it will sort, but I'm not too stressed abut that.

Testing now and then merging. thank you!

PR #1627, opened 2016-03-13
Change Mod Info window label for repository

politas:
We have sources other than github, and the field definition is not GitHub specific


pjf:
This looks awesome! :) Testing now.

pjf:
And it works! Merged with great thanks! :)

PR #1648, opened 2016-04-04
Catch ValueErrors as in KSP-CKAN/CKAN-meta#1035

techman83:
A small modification to 'ckan-validate.py' which is used in our testing. Gets rid of the big Traceback, we can probably do more with it.

### DIFF
--- a/bin/ckan-validate.py
+++ b/bin/ckan-validate.py
@@ -45,6 +45,11 @@ def main():
                 print e
                 error = 1
                 continue
+            except ValueError as e:
+                print 'Failed! This error will be cryptic, but often a JSON or property error'
+                print e
+                error = 1
+                continue
             print 'Success!'
 
     sys.exit(error)


Dazpoet:
After a bit of googling I think I've gotten myself a good enough understanding of this to merge it in good conciense, also it passes travis.

techman83:
It's a pretty safe one, really only breaks Jenkins + Indexing and easy to revert in that case.

Dazpoet:
After looking a few more times, because who doens't love learning? I wonder if ValueError should've also been added to https://github.com/KSP-CKAN/CKAN/blob/master/bin/ckan-validate.py#L6 or have I, once again, totally misunderstood things?

techman83:
@Dazpoet hmm, Python is something I only dabble in. To my eyes you import from to access the method/function directly. 

For example you can do this
```
import json
from jsonschema import validate

validate(json.load(ckan_file), schema)
```
versus
```
import json
import jsonschema

jsonschema.validate(json.load(ckan_file), schema)
```

I haven't fully grok'd pythons error handling, but I don't think those even need to be exported. It certainly catches ```ValueError``` without doing as such.

Dazpoet:
Thanks for expanding on the issue, I'm doing my best to learn :)

PR #1649, opened 2016-04-04
Compare operation should check for negative/positive not -1/+1

dbent:
Otherwise a comparison of `1_0` to `1.0` will fail.


Dazpoet:
- [x] Passes travis
- [x] Looks simple enough so that Daz can understand it
- [x] Adds proper documentation of what it does in the code and PR
- [x] Merged

Issue #1660, opened 2016-04-13
TCA upgradability not correctly detected when switching from 1.0.5 to 1.1-prerelease


mathuin:
**CKAN Version:** 
development (beta) -- HEAD of master

**Operating System:** 
Ubuntu 14.04.4 LTS with CKAN running within Docker as per my dockerize-2 branch

**The issue you are experiencing:** 
TCA has a version for 1.1-prerelease on CKAN but CKAN can't tell that it's an upgrade for me.

**How to recreate this issue:**

1. Install KSP 1.0.5.

2. Install CKAN pointing at that installation of KSP.
  - Run it any way you like, Docker not required.

3. Install the TCA mod with CKAN.
  - Note that the mod should be version v3.0.1.

4. Tell Steam you want to switch to 1.1-prerelease.

5. Use CKAN to list mods.
  - Note that TCA has a bullet of '*' which is not part of the legend.

**CKAN error codes (if applicable):** 
None displayed.

When I ran `ckan list --debug`, I got the following output:

```
2066 [1] DEBUG CKAN.Registry (null) - Finding latest available for ThrottleControlledAvionics
2066 [1] DEBUG CKAN.AvailableModule (null) - Our dictionary has 13 keys
2066 [1] DEBUG CKAN.AvailableModule (null) - Choosing between 13 available versions
2066 [1] DEBUG CKAN.CkanModule (null) - Testing if ThrottleControlledAvionics v3.0.1 is compatible with KSP 1.1.0
2066 [1] DEBUG CKAN.CkanModule (null) - Testing if ThrottleControlledAvionics 3.0.1-KSP-1.1-BETA is compatible with KSP 1.1.0
2066 [1] INFO CKAN.CmdLine.List (null) - Latest ThrottleControlledAvionics is ThrottleControlledAvionics 3.0.1-KSP-1.1-BETA
* ThrottleControlledAvionics v3.0.1
```

That bullet is not listed in the legend.

From `CmdLine/Action/List.cs`, lines 52-103:

```
                var installed = new SortedDictionary<string, Version>(registry.Installed());

                foreach (KeyValuePair<string, Version> mod in installed)
                {
                    Version current_version = mod.Value;

                    string bullet = "*";

                    if (current_version is ProvidesVersion)
                    {
                        // Skip virtuals for now.
                        continue;
                    }
                    else if (current_version is DllVersion)
                    {
                        // Autodetected dll
                        bullet = "-";
                    }
                    else
                    {
[...]

PR #1664, opened 2016-04-15
Add Asterisk to legend in CLI list command

politas:
An * for status shows up whenever a mod falls through all the status tests without matching anything, thus implying an inability for CKAN to intelligently handle the mod. Since "Unknown" has been used to refer to mods that have been installed through CKAN but are now unable to be found in the registry, I have selected "Broken" as a term to describe the status of these mods.

### DIFF
--- a/Cmdline/Action/List.cs
+++ b/Cmdline/Action/List.cs
@@ -111,7 +111,8 @@ namespace CKAN.CmdLine
 
             if (!(options.porcelain) && exportFileType == null)
             {
-                user.RaiseMessage("\nLegend: -: Up to date. X: Incompatible. ^: Upgradable. ?: Unknown ");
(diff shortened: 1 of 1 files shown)


politas:
Ref #1660, for which this is only a partial fix.

politas:
Deleted the wrong branch

politas:
#rebuild

politas:
So it's failing on Mono 3.2.8? I didn't think we supported that version of Mono.

plague006:
All you did was modify a string and add a comment, the failure was unrelated to your changes.

politas:
I figured as much.

pjf:
This looks very sensible to me. Merging now. :)

As an aside, CKAN shows me '*' on the cmdline for some mods which can be upgraded via the GUI.

PR #1662, opened 2016-04-14
Metapackage schema

techman83:
Tested @Zane6888's correction in #1597 and it works for ```"download": "url"``` and ```"kind": "metapackage"```.

PR #1665, opened 2016-04-15
Update quotes.txt

Dazpoet:
@plague006 comment in #1662 made me remember the good ol BaconLabs.netkan incidient so I've added it to the quotes aswell as the plague006s hilarious, and accurate, comment about my ability to break things.

We totally need another guy doing more screw-ups around here T_T


plague006:
Metadata now has a dedicated team instead of just you watching over it all. It was a different time back then, lol.

PR #1684, opened 2016-04-26
Omnipresent VersionedOverrideTransformers

### LINKED ISSUE #1674: NetKAN doesn't override install field if $vref defined
**CKAN Version:** CKAN 1.16.1-0-g2e91715 (beta), NetKAN 1.0.5950.1397
**Operating System:** Windows 10 64-bit

I am updating my mod's NetKAN config to prepare for a breaking change in the format of the install zip. The new `install` directive is being placed in the main NetKAN file; the old one is in `x_netkan_override` with version "< [next version]". When I run NetKAN on the file, however, I get the following error:
> 578 [1] FATAL CKAN.NetKAN.Program (null) - No files found in [new directory] to install!

Temporarily replacing the new install directive with the old one confirms that other fields, like `depends`, are correctly overridden in the output .ckan file.

In trying to create a minimum working example, I discovered that this issue only shows up if `$vref` is defined. If I remove it, NetKAN runs without error, and the resulting .ckan file has `install` correctly overridden for the old file format.

Download [demo.netkan.txt](https://github.com/KSP-CKAN/CKAN/files/232344/demo.netkan.txt) and run it using NetKAN. NetKAN will refuse to create a file for v1.2.0 unless the `$vref` line is removed.

**CKAN error codes (if applicable):** 578 [1] FATAL CKAN.NetKAN.Program (null)

dbent:
@pjf @Starstrider42 @Dazpoet

Prompted by #1674. The issue was that the `x_netkan_override` was overriding the `install` stanza. However, the `AvcTransformer` uses the `install` stanza to determine which files in an archive will be installed so that it can pick the correct `.version` file to use. The problem was that the `AvcTransformer` executes before the `VersionedOverrideTransformer` so the `AvcTransformer` didn't think *any* files would be installed because the appropriate override hadn't occurred yet.

What this PR does is basically allow you specify *when* an override should execute, through the use of `before` and `after` properties in the override.

To accomplish this there are three main changes:

- Each `ITransformer` must have a name which identifies it uniquely.
- `VersionedOverrideTransformer` takes a list of names of `ITransformer` implementations that it executes directly before and executes directly after.
- The top-level `NetkanTransformer` has been modified to inject an instance of `VersionedOverrideTransformer` before and after *every other `ITransformer`*.

So now the transformation pipeline looks like:

```
VersionedOverrideTransformer: before: first
FirstTransformer
VersionedOverrideTransformer: before: second, after: first
SecondTransformer
VersionedOverrideTransformer: before third, after: second
ThirdTransformer
VersionedOverrideTransformer: after: third
```

There are also two "psuedo-transformers", `$none` and `$all` (should be self-explanatory) so the pipeline really looks like:

```
VersionedOverrideTransformer: before: first, $all, after: $none
FirstTransformer
VersionedOverrideTransformer: before: second, after: first
SecondTransformer
VersionedOverrideTransformer: before third, after: second
ThirdTransformer
VersionedOverrideTransformer: after: third, $all, before: $none
```

How does this fix #1674?

Well given this `.netkan` (@Starstrider42 changed the archive to workaround the issue in NetKAN#3691 so to make it reproducable I modified from the one given in #1674 by inverting the install stanzas and override condition).

```json
{
    "spec_version": 1,
    "identifier": "CustomAsteroids-Pops-Stock-Inner",
    "$kref": "#/ckan/spacedock/210",
    "$vref" : "#/ckan/ksp-avc",
    "install": [
        {
            "file": "GameData/CustomAsteroids",
            "install_to": "GameData",
            "filter_regexp": "(?<!Basic Asteroids\\.cfg)$"
        }
    ],
    "x_netkan_override" : [
        {
            "version": ">=v1.3.0",
            "override" : {
                "install": [
                    {
                        "file": "Standard Setup",
                        "install_to": "GameData",
                        "filter_regexp": "(?<!Basic Asteroids\\.cfg)$"
                    }
                ]
            }
        }
    ],
}
```

NetKAN compalains:
```
1076 [1] FATAL CKAN.NetKAN.Program (null) - No files found in GameData/CustomAsteroids to install!
```

However using this `.netkan` (notice the addition of `"before": "avc"`)

```json
{
    "spec_version": 1,
    "identifier": "CustomAsteroids-Pops-Stock-Inner",
    "$kref": "#/ckan/spacedock/210",
    "$vref" : "#/ckan/ksp-avc",
    "install": [
        {
            "file": "GameData/CustomAsteroids",
            "install_to": "GameData",
            "filter_regexp": "(?<!Basic Asteroids\\.cfg)$"
        }
    ],
    "x_netkan_override" : [
        {
            "version": ">=v1.3.0",
            "before": "avc",
            "override" : {
                "install": [
                    {
                        "file": "Standard Setup",
                        "install_to": "GameData",
                        "filter_regexp": "(?<!Basic Asteroids\\.cfg)$"
                    }
                ]
            }
        }
    ],
}
```

NetKAN successfully produces the following `.ckan`:

```json
{
    "spec_version": 1,
    "identifier": "CustomAsteroids-Pops-Stock-Inner",
    "name": "Custom Asteroids",
    "abstract": "Lets users control where asteroids appear",
    "author": "Starstrider42",
    "license": "MIT",
    "resources": {
        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/72785-/",
        "spacedock": "https://spacedock.info/mod/210/Custom%20Asteroids",
        "repository": "https://github.com/Starstrider42/Custom-Asteroids",
        "x_screenshot": "https://spacedock.info/content/Starstrider42_1134/Custom_Asteroids/Custom_Asteroids-1461528691.6712377.png"
    },
    "version": "v1.3.0",
    "ksp_version": "1.1.0",
    "install": [
        {
            "file": "Standard Setup",
            "install_to": "GameData",
            "filter_regexp": "(?<!Basic Asteroids\\.cfg)$"
        }
    ],
    "download": "https://spacedock.info/mod/210/Custom%20Asteroids/download/v1.3.0",
    "download_size": 77254,
    "x_generated_by": "netkan"
}
```

To remain backwards compatible with existing overrides, there is a hardcoded `VersionedOverrideTransformer` that occurs `before` and `after` `null`.

The following is the list of all possible values of `before` and `after`. It exposes some of the internal architecture of NetKAN, but I think this is the best possible compromise to solve this particular issue.

```
$none
$all
avc
download_size
epoch
forced_v
generated_by
github
http
internal_ckan
jenkins
metanetkan
optimus_prime
property_sort
spacedock
strip_netkan_metadata
version_edit
versioned_override
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
 - [spec] ckan.schema now enforces structure of install directives (#1578 by: Zane6888; reviewed: pjf, Daz)
 - [NetKAN] Catch ValueErrors rather than printing the trace (#1648 by: techman83; reviewed: Daz )
 - [NetKAN] Catch ksp_version from SpaceDocks newly implemented game_version (#1655 by: dbent; reviewed: -)
+- [NetKAN] Allow specifying when an override is executed (#1684 by: dbent; fixes: #1674)
 
 ## v1.16.1
 
--- a/Netkan/Transformers/AvcTransformer.cs
+++ b/Netkan/Transformers/AvcTransformer.cs
@@ -19,6 +19,8 @@ namespace CKAN.NetKAN.Transformers
         private readonly IHttpService _http;
         private readonly IModuleService _moduleService;
 
+        public string Name { get { return "avc"; } }
+
         public AvcTransformer(IHttpService http, IModuleService moduleService)
         {
             _http = http;
--- a/Netkan/Transformers/DownloadSizeTransformer.cs
+++ b/Netkan/Transformers/DownloadSizeTransformer.cs
@@ -15,6 +15,8 @@ namespace CKAN.NetKAN.Transformers
         private readonly IHttpService _http;
         private readonly IFileService _fileService;
 
+        public string Name { get { return "download_size"; } }
+
         public DownloadSizeTransformer(IHttpService http, IFileService fileService)
         {
             _http = http;
--- a/Netkan/Transformers/EpochTransformer.cs
+++ b/Netkan/Transformers/EpochTransformer.cs
@@ -12,6 +12,8 @@ namespace CKAN.NetKAN.Transformers
     {
         private static readonly ILog Log = LogManager.GetLogger(typeof(EpochTransformer));
 
+        public string Name { get { return "epoch"; } }
+
         public Metadata Transform(Metadata metadata)
         {
             Log.Debug("Fixing version strings (if required)...");
--- a/Netkan/Transformers/ForcedVTransformer.cs
+++ b/Netkan/Transformers/ForcedVTransformer.cs
@@ -12,6 +12,8 @@ namespace CKAN.NetKAN.Transformers
     {
         private static readonly ILog Log = LogManager.GetLogger(typeof(ForcedVTransformer));
 
+        public string Name { get { return "forced_v"; } }
+
         public Metadata Transform(Metadata metadata)
         {
             var json = metadata.Json();
--- a/Netkan/Transformers/GeneratedByTransformer.cs
+++ b/Netkan/Transformers/GeneratedByTransformer.cs
@@ -12,6 +12,8 @@ namespace CKAN.NetKAN.Transformers
     {
         private static readonly ILog Log = LogManager.GetLogger(typeof(GeneratedByTransformer));
 
+        public string Name { get { return "generated_by"; } }
+
         public Metadata Transform(Metadata metadata)
         {
             var json = metadata.Json();
--- a/Netkan/Transformers/GithubTransformer.cs
+++ b/Netkan/Transformers/GithubTransformer.cs
@@ -17,6 +17,8 @@ namespace CKAN.NetKAN.Transformers
         private readonly IGithubApi _api;
         private readonly bool _matchPreleases;
 
+        public string Name { get { return "github"; } }
+
         public GithubTransformer(IGithubApi api, bool matchPreleases)
         {
             if (api == null)
--- a/Netkan/Transformers/HttpTransformer.cs
+++ b/Netkan/Transformers/HttpTransformer.cs
@@ -11,6 +11,8 @@ namespace CKAN.NetKAN.Transformers
     {
         private static readonly ILog Log = LogManager.GetLogger(typeof(HttpTransformer));
 
+        public string Name { get { return "http"; } }
+
         public Metadata Transform(Metadata metadata)
         {
             if (metadata.Kref != null && metadata.Kref.Source == "http")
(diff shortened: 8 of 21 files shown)


Starstrider42:
From your opening explanation I take it you guys need to support multiple `.version` files in the same archive... :astonished:

I'm afraid the whole *Transformers/pipeline business is a bit over my head... so in the example you gave, the >=v1.3.0 override is applied before KSP-AVC is invoked. Is that chosen based on the version number from SpaceDock? Should I (as a user, not a CKAN insider) think of `"before": "avc"` as "pretend, for this override only, that there was no KSP-AVC `$vref`"?

If you're concerned about exposing implementation details, perhaps you can trim the list of allowed values a bit. I certainly can't guess what all of them mean off the top of my head... (but I'll bite: is there actually an `optimus_prime` in NetKAN)?

Dazpoet:
Yes, there is an easter egg which triggers an Optimus Prime quote ;)

dbent:
> From your opening explanation I take it you guys need to support multiple .version files in the same archive... :astonished:

It's necessary because a lot of mods distribute their dependencies in their own archive (bleh), meaning we need some mechanism to decide which of multiple possible `.version` files actually correspond to the mod we're installing. The default mechanism is to only look at files that will actually be *installed*, hence why the value of the `install` stanzas are important to the `AvcTransformer`. (It supports selecting a `.version` file via a regex expression if necessary, but that's rarely used).

> I'm afraid the whole *Transformers/pipeline business is a bit over my head... so in the example you gave, the >=v1.3.0 override is applied before KSP-AVC is invoked. Is that chosen based on the version number from SpaceDock? Should I (as a user, not a CKAN insider) think of "before": "avc" as "pretend, for this override only, that there was no KSP-AVC $vref"?

NetKAN works by applying a series of transformations on its input metadata. This series of transformations is the "pipeline". The metadata at the start of the pipeline should be a valid `.netkan` file and the metadata at the end of the pipeline should be a valid `.ckan` file.

One of these transformations is applied by the `AvcTransformer` which will read a `.version` file and set the `version` property (if not already set) and update the `ksp_version` properties. At the point in the pipeline when the `AvcTransformer` is invoked the only version information that exists is what we retrieved from SpaceDock (by way of the `SpacedockTransformer`).

So what `before` and `after` do is tell the `VersionedOverrideTransformer` (which processes the override stanzas) *when* to apply the override. In this case it tells it to apply the override *before* the `AvcTransformer`, so that when the `AvcTransformer` is invokved, the `install` stanzas have been overridden and it knows the correct set of files to look for the `.version` file in.

> If you're concerned about exposing implementation details, perhaps you can trim the list of allowed values a bit. I certainly can't guess what all of them mean off the top of my head...

I'm not too concerned, like I said in the documentation update in this PR, the *vast vast* majority of overrides should never have to use a `before` or `after` property, just for special cases like this.

>  (but I'll bite: is there actually an optimus_prime in NetKAN)?

Isn't there a little bit of [Optimus Prime](https://www.youtube.com/watch?v=2ZqEor-ey0M) everywhere?

techman83:
I've had a good read over this and it looks really good. I've also tested across all our NetKANs and tested with the example NetKAN. The indexer will pickup the new version on the next run.

Autobots roll out!

Issue #1689, opened 2016-04-30
Force Installation of incompatible mods


fat-lobyte:
Hi!

While the compatibility check of the mods highest KSP version vs the KSP version is generally a good idea, I believe there should be a way to force the installation of mods, on responsibility of the user.

This has been demonstrated very well with the 1.1.1 patch to KSP, where suddenly most mods were uninstallable, even if they are perfectly compatible in reality.

I suggest either a check box in the settings: "Disable KSP version check" or a right-click menu option for each mod "Force installation". Maybe a warning message or so should pop up, but I do think such a thing is necessary.

plague006:
Closing this due to / in favour of the referred PR/issue/commit.

https://github.com/KSP-CKAN/CKAN/issues/1173

PR #1690, opened 2016-04-30
GrasComparator: regard 1.1.1 and 1.1.2 as safe

fat-lobyte:
KSP 1.1.1 and 1.1.2 are Hotfixes to 1.1.0, so we allow installation of
mods for this version.

This alleviates issue #1689 and ties into PR #1499 .

### DIFF
--- a/Core/Types/GameComparator/GrasGameComparator.cs
+++ b/Core/Types/GameComparator/GrasGameComparator.cs
@@ -11,6 +11,8 @@ namespace CKAN
     {
         static readonly StrictGameComparator strict = new StrictGameComparator();
         static readonly KSPVersion v103 = new KSPVersion("1.0.3");
+        static readonly KSPVersion v110 = new KSPVersion("1.1.0");
+        static readonly KSPVersion v111 = new KSPVersion("1.1.1");
 
         public bool Compatible(KSPVersion gameVersion, CkanModule module)
         {
@@ -29,6 +31,10 @@ namespace CKAN
             // considered it compatible under 1.0.3 (as 1.0.4 was "just a hotfix").
             if (gameVersion.Equals("1.0.4"))
                 return strict.Compatible(v103, module);
+				
+			// 1.1.1 and 1.1.2 are hotfixes to 1.1.0
+            if (gameVersion.Equals("1.1.2") || gameVersion.Equals("1.1.1"))
+                return strict.Compatible(v110, module) || strict.Compatible(v111, module);
 
             return false;
         }


dewiniaid:
Considering even hotfixes can break mods (Example: 1.1.2 breaks the current ModuleManager, and thus almost everything else), this seems like a bad idea?

I think it'd be better for mod authors to make that call -- either by claiming compatibility with 1.1.0 - 1.1.99, or not.

fat-lobyte:
@dewiniaid 
Suit yourself, I'm just a user who tries to get his install to work, and wanted to share a quick-fix.

I understand your viewpoint as a developer: better be sure than to break something. It makes sense.

Now consider the users standpoint:
"I just updated KSP, now I can't install mods? What the heck? It's just a hotfix, it's probably fine. Now I have to go and download the mod manually! What's the point of CKAN anyway?"

Now consider the mod developers standpoint:
"Wow, so many users complaining about my mod, just because of CKAN? Why do I always have mess with that metadata for no apparent reason? My mod works just fine!"

That last point kind of violates the philosophy to not put additional burden on the modders. There are already plenty of modders who really don't care about CKAN and get annoyed by it, because it always causes them support issues.

I'm not even saying you should merge this. It's just a suggestion to solve a problem.

dewiniaid:
A mod developer who doesn't want to continually update their metadata can just say ksp_version_max 1.1.99 and not need to update until 1.2, regardless of how many patches come along.

A mod developer who wants more control over things can set that to the current version and update as they see fit.

A better option here, IMHO, would be to allow version checks to be overridden (with appropriate warnings), but still hide incompatible mods by default.

jakkarth:
Now consider the users standpoint:
"I just updated KSP, all of my mods are listed as compatible, but nothing is working! Wait, ModuleManager actually *isn't* compatible, but CKAN installed it anyway? What's the point of CKAN anyway?"

Now consider the mod developers standpoint:
"Wow, so many users complaining about my mod, just because of CKAN? I specified which versions my mod was compatible with, why is it ignoring all that work I put into writing correct metadata?"

As a regular user just like you, I absolutely believe the existing CKAN behavior is what I'd prefer.

fat-lobyte:
@dewiniaid 
> A better option here, IMHO, would be to allow version checks to be overridden (with appropriate warnings), but still hide incompatible mods by default.

I agree, but apperently doing that is so complicated, that it hasn't been implemented in almost a year: #1173 

@jakkarth 

> "Wow, so many users complaining about my mod, just because of CKAN? I specified which versions my mod was compatible with, why is it ignoring all that work I put into writing correct metadata?"

I believe that the mod can set the version comparison policy to "strict" which prevents that kind of stuff.

> As a regular user just like you, I absolutely believe the existing CKAN behavior is what I'd prefer.

I disagree, which is fine, because it only shows that it shold be a toggle-able option.

Either way, is there any indication that the case between 1.0.3 -> 1.0.4 is any different than 1.1.0->1.1.1->1.1.2? Because to me, it seems like it's exactly the same thing and in that case, such behaviour was fine.

plague006:
> Either way, is there any indication that the case between 1.0.3 -> 1.0.4 is any different than 1.1.0->1.1.1->1.1.2?

Not a modder myself but from RealChute's stupid_chris

````
[09:48] (stupid_chris|code): turns out 1.1.2 fucked up PopupDialog.SpawnPopupDialog
[09:48] (stupid_chris|code): turns out we all have to recompile
````

And from ModuleManager's Sarbian

````
[10:24] (Sarbian): MM is out and a required update
[10:27] (@plague006): Sarbian: is the new version of MM compatible with 1.1.0 and 1.1.1 or just 1.1.2 and up?
[10:27] (Sarbian): just 1.1.2
[10:28] (Sarbian): and up
````

Given their statements and the fact that the patch is only a few hours old (meaning very few modders have had a chance to look at the changes) I feel comfortably saying 1.1.1 -> 1.1.2 is different than 1.0.3 -> 1.0.4.

nobodyhasthis:
@fat-lobyte is correct. If we agree that a minor change has taken place. We currently can't make a decision on this as it too early to call.. The options are still

**If it is minor change?**
We have been around the block a few times on this already. After a long discussion elsewhere the best policy we can come up with is to allow GRAS by default. With an open mind toward individual cases where "strict" can be applied on request. So PR is good.
( #1173 ). 

**However this is starting to look major change?**
Historically speaking minor version number changes are not consistent. Some are indeed trivial and some actually turn out to major upsets for mod authors. Breaking MM is pretty much catastrophic and would indicate everyone is now playing catch up. So the PR is bad. If we assume that compatibility has been compromised across the board. Minor version changes should be easier to dealt with but based on passed experience we have to judge them carefully.   

I think we should wait this one out before considering a merge. To give mod authors time to take a a better look.

politas:
If we're going to have this usable, it needs to be in metadata, not baked into the client. Releasing a new client version isn't a simple task, and if we're going to mess around with GRAS KSP versions, we need to be able to change that status quickly, both to account for new hotfix KSP releases, and to revert the GRAS status if it becomes clear that too much stuff is getting broken.

politas:
My understanding is that 1.1.1 -> 1.1.2 was not a safe move for mods, and the issue now is all the 1.1.2 mods that aren't updated for 1.1.3, anyway. If someone would like to implement a metadata approach for GRAS, I'd certainly support that. This change seems unnecessary and possibly problematic.

Issue #1658, opened 2016-04-11
Crash when reinstalling a mod after cleaning cache


trakos:
**CKAN Version:** 
1.16.1

**Operating System:** 
Windows 7 x64 Professional SP1

**The issue you are experiencing:** 
CKAN crashes when reinstalling the mod after the cache has been cleaned.

**How to recreate this issue:** 
1. Click "Download" on Contents tab of a mod (or just install and remove a mod).
2. Go to Settings -> CKAN settings and click "Clear cache"
3. Check the same mod and try to install it

**CKAN error codes (if applicable):** 
```
System.IO.FileNotFoundException: Could not find file 'C:\Program Files (x86)\GalaxyClient\Games\Kerbal Space Program\CKAN\downloads\2BBA8A7C-xScience-4.14.zip'.
File name: 'C:\Program Files (x86)\GalaxyClient\Games\Kerbal Space Program\CKAN\downloads\2BBA8A7C-xScience-4.14.zip'
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
   at ICSharpCode.SharpZipLib.Zip.ZipFile..ctor(String name)
   at CKAN.NetFileCache.GetCachedZip(Uri url, Boolean test_data)
   at CKAN.ModuleInstaller.GetModuleContentsList(CkanModule module)
   at CKAN.Main._UpdateModContentsTree(Boolean force)
   at CKAN.Util.Invoke[T](T obj, Action action)
   at CKAN.Main.UpdateModContentsTree(CkanModule module, Boolean force)
   at CKAN.Main.ModList_SelectedIndexChanged(Object sender, EventArgs e)
   at System.Windows.Forms.DataGridView.OnSelectionChanged(EventArgs e)
   at System.Windows.Forms.DataGridView.FlushSelectionChanged()
   at System.Windows.Forms.DataGridView.ClearSelection(Int32 columnIndexException, Int32 rowIndexException, Boolean selectExceptionElement)
[...]

PR #1694, opened 2016-05-01
Fixed IOException "Directory is not empty" errors in Test

RichardLake:
Had issues on my system (win10 VS15 dotCover) with some of the tests due to incorrect disposals. This caused the program to call Directory.Delete(dir, true); with outstanding handles to directories within the dir that was to be deleted. This resulted in an IOException with "Directory is not empty" as the message.

### DIFF
--- a/Tests/Core/Cache.cs
+++ b/Tests/Core/Cache.cs
@@ -23,6 +23,7 @@ namespace Tests.Core
         [TearDown]
         public void RemoveCache()
         {
+            cache.Dispose();
             Directory.Delete(cache_dir, true);
         }
 
--- a/Tests/Core/KSP.cs
+++ b/Tests/Core/KSP.cs
@@ -24,6 +24,7 @@ namespace Tests.Core
         [TearDown]
         public void TearDown()
         {
+            ksp.Dispose();
             Directory.Delete(ksp_dir, true);
         }
 
--- a/Tests/Core/KSPManager.cs
+++ b/Tests/Core/KSPManager.cs
@@ -25,6 +25,7 @@ namespace Tests.Core
         [TearDown]
         public void TearDown()
         {
+            tidy.DisposeOfAllManagedKSPs(manager);
             tidy.Dispose();
         }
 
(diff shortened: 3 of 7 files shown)


RichardLake:
Found a case I didn't fix.

trakos:
I have same problems (win7 VS15 with ReSharper). With your fix and some tweaks I can get all tests to pass but one: Tests.Core.ModuleInstaller.CanUninstallMod.

The problems are caused by FileSystemWatcher, if you comment it everything works. For some reason FileSystemWatcher is known to lock parent directory - but I'm not sure why it's not disposing fully in this test - probably something with deleting mod blocking it.

I wonder if it's related with FileSystemWatcher not reacting to deleting files ( https://github.com/KSP-CKAN/CKAN/issues/1658 ).

RichardLake:
Could you rerun the tests on your machine? That is one of the tests I missed and it wasn't being triggered on my system making the solution hard to test.

trakos:
I had one (new?) error:

````
Core (218 tests) [0:08.133] Failed: 1 test failed
  KSPManagerTests (15 tests) [0:00.522] Failed: One or more child tests had errors: 1 test failed
   GetPreferredInstance_WithEmptyAutoStartAndMultipleInstances_ReturnsNull [0:00.022] Failed:

TearDown : System.IO.IOException : The directory is not empty.

--TearDown
   at System.IO.Directory.DeleteHelper(String fullPath, String userPath, Boolean recursive, Boolean throwOnTopLevelDirectoryNotFound)
   at System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
   at Tests.Data.DisposableKSP.Dispose() in C:\home\coding\c#\ckan\Tests\Data\DisposableKSP.cs:line 45
   at Tests.Core.KSPManagerTests.TearDown() in C:\home\coding\c#\ckan\Tests\Core\KSPManager.cs:line 29
````

politas:
I'm not seeing any errors like that when I `make test`, so my ability to do anything regarding this is limited.

ayan4m1:
Assigning myself so that I will remember to test this out in VS/Win10 and in Linux.

ayan4m1:
Fixed in #1880.

PR #1696, opened 2016-05-02
HttpTransformer Resolves Redirects

dbent:
@pjf @techman83 @plague006 @Ezriilc

This PR modifies the `HttpTransformer` to resolve redirects and output the resolved URL as the 
`download` property of the `.ckan` file, instead of value specified in the `$kref`.

[Graphotron](https://github.com/KSP-CKAN/NetKAN/blob/33eafb20f717b278e5376847e988dea1d93bfed6/NetKAN/Graphotron.netkan) and [HyperEdit](https://github.com/KSP-CKAN/NetKAN/blob/30241029a0fa7862875b97b37249452ec504222d/NetKAN/HyperEdit.netkan) are the only two `.netkan` files using the `HttpTransformer` at the moment.

Since they use a fixed URL this makes caching problematic. @Ezriilc can make it so that instead of returning the download directly in the response it can produce an HTTP redirect to a version specific URL. This should prevent caching issues (or at least reduce caching issues to the kind we already have to deal with), keep us from having to intervene with updates to Graphotron and HyperEdit, and make sure clients get a version specific URL for backwards compatibility with old versions of KSP.

This example `.netkan` points to some test redirects I setup on my server.

`https://www.bent.io/tmp/redirect1` redirects to `https://www.bent.io/tmp/redirect2` which redirects to `https://www.bent.io/tmp/redirect3` which redirects to `https://www.bent.io/tmp/PlaneMode-1.3.1.zip`. (The code has a hardcoded limit of up to 5 requests before it gives up).

```json
{
    "spec_version": 1,
    "identifier": "PlaneMode",
    "$kref": "#/ckan/http/https://www.bent.io/tmp/redirect1",
    "version": "1.3.1",
    "license": "MIT"
}
```

Produce the following `.ckan`:
```json
{
    "spec_version": 1,
    "identifier": "PlaneMode",
    "license": "MIT",
    "version": "1.3.1",
    "download": "https://www.bent.io/tmp/PlaneMode-1.3.1.zip",
    "download_size": 17530,
    "x_generated_by": "netkan"
}
```

Notice that the `download` property is the final destination of the redirects starting at `https://www.bent.io/tmp/redirect1`.

### DIFF
--- a/Core/Net/Net.cs
+++ b/Core/Net/Net.cs
@@ -151,12 +151,76 @@ namespace CKAN
             }
         }
 
+        public static Uri ResolveRedirect(Uri uri)
+        {
+            const int maximumRequest = 5;
+
+            var currentUri = uri;
+            for (var i = 0; i < maximumRequest; i++)
+            {
+                var client = new RedirectWebClient();
+
+                // The empty using is so that we dispose of the stream and don't block
+                // We don't ACTUALLY attempt to download the file, but it appears that if the client sees a
+                // Content Length in the response it thinks there will be a response body and blocks.
+                using (client.OpenRead(currentUri)) { }
+
+                var location = client.ResponseHeaders["Location"];
+
+                if (location == null)
+                {
+                    return currentUri;
+                }
+                else
+                {
+                    Log.DebugFormat("{0} redirected to {1}", currentUri, location);
+
+                    if (Uri.IsWellFormedUriString(location, UriKind.Absolute))
+                    {
+                        currentUri = new Uri(location);
+                    }
+                    else if (Uri.IsWellFormedUriString(location, UriKind.Relative))
+                    {
+                        currentUri = new Uri(currentUri, location);
+                        Log.DebugFormat("Relative URL {0} is absolute URL {1}", location, currentUri);
+                    }
+                    else
+                    {
+                        throw new Kraken("Invalid URL in Location header: " + location);
+                    }
+                }
+            }
+
+            return null;
+        }
+
         private static WebClient MakeDefaultHttpClient()
         {
             var client = new WebClient();
             client.Headers.Add("User-Agent", UserAgentString);
 
             return client;
+        }
+
+        // HACK: The ancient WebClient doesn't support setting the request type to HEAD and WebRequest doesn't support
+        // setting the User-Agent header.
+        // Maybe one day we'll be able to use HttpClient (https://msdn.microsoft.com/en-us/library/system.net.http.httpclient%28v=vs.118%29.aspx)
+        private sealed class RedirectWebClient : WebClient
+        {
+            public RedirectWebClient()
+            {
+                Headers.Add("User-Agent", UserAgentString);
+            }
+
+            protected override WebRequest GetWebRequest(Uri address)
+            {
+                var webRequest = (HttpWebRequest)base.GetWebRequest(address);
+                webRequest.AllowAutoRedirect = false;
+
+                webRequest.Method = "HEAD";
+
+                return webRequest;
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


techman83:
I get an error with your test ckan, but Graphotron + HyperEdit inflate correctly.

```
203 [1] DEBUG CKAN.NetKAN.Transformers.HttpTransformer (null) - Input metadata:
{
  "spec_version": 1,
  "identifier": "PlaneMode",
  "$kref": "#/ckan/http/https://www.bent.io/tmp/redirect1",
  "version": "1.3.1",
  "license": "MIT"
}
1298 [1] FATAL CKAN.NetKAN.Program (null) - Error: SendFailure (Error writing headers)
1299 [1] FATAL CKAN.NetKAN.Program (null) -   at System.Net.WebClient.OpenRead (System.Uri address) <0x40fcb8b0 + 0x003a3> in <filename unknown>:0 
  at (wrapper remoting-invoke-with-check) System.Net.WebClient:OpenRead (System.Uri)
  at CKAN.Net.ResolveRedirect (System.Uri uri) <0x40fc9b20 + 0x00063> in <filename unknown>:0 
  at CKAN.NetKAN.Transformers.HttpTransformer.Transform (CKAN.NetKAN.Model.Metadata metadata) <0x40fc2a40 + 0x00153> in <filename unknown>:0 
  at CKAN.NetKAN.Transformers.NetkanTransformer.<Transform>m__0 (CKAN.NetKAN.Model.Metadata transformedMetadata, ITransformer transformer) <0x40fc1790 + 0x00023> in <filename unknown>:0 
  at System.Linq.Enumerable.Aggregate[TSource,TAccumulate] (IEnumerable`1 source, System.Linq.TAccumulate seed, System.Func`3 func) <0x40fc1540 + 0x0009b> in <filename unknown>:0 
  at CKAN.NetKAN.Transformers.NetkanTransformer.Transform (CKAN.NetKAN.Model.Metadata metadata) <0x40fc1480 + 0x00093> in <filename unknown>:0 
  at CKAN.NetKAN.Program.Main (System.String[] args) <0x40f67750 + 0x00383> in <filename unknown>:0 
```

Am a bit sleepy, so nothing obvious is jumping out at me just yet.

dbent:
@techman83 

What about?
```json
{
    "spec_version": 1,
    "identifier": "PlaneMode",
    "$kref": "#/ckan/http/http://www.zengei.com/tmp/redirect1",
    "version": "1.3.1",
    "license": "MIT"
}
```

Mono's current TLS implementation doesn't support the elliptic curve cipher suites my server uses.

techman83:
Aha, I had an inkling that would be the case. Is libcurl usable here? It's no biggie for now, the 2 netkans that depend on this transformer work correctly.

```
{
    "spec_version": 1,
    "identifier": "PlaneMode",
    "license": "MIT",
    "version": "1.3.1",
    "download": "http://www.zengei.com/tmp/PlaneMode-1.3.1.zip",
    "download_size": 17530,
    "x_generated_by": "netkan"
}
```
LGTM, merging with thanks!

dbent:
Yeah, I'd rather not deal with using libcurl. I'm betting on Mono's [new TLS implementation](https://github.com/mono/mono-tls) becoming the default before we have need of resolving redirects through TLS that Mono's current implementation doesn't support.

techman83:
@dbent Fair point. Cross the bridge if we need to. As an aside, KerbalTek doesn't look to be performing redirects though.

dbent:
Not yet, that's something @Ezriilc will implement.

techman83:
Ahh, awesome.

Ezriilc:
I just realized that I never commented here to say that I've completed the redirect for our mods (HyperEdit & Graphotron).  So, yea.  I did that and it's working as expected.

PR #1700, opened 2016-05-02
Avoid unexpected null reference exception

keyspace:
While traversing possible nested exceptions (added because of #182) it is possible to reach a "leaf" where there are no more `InnerException`s. If at this point the exception hasn't been `handle`d, then `log.ErrorFormat` will try to access `exception.Message`. However, due to the `while` loop just above it, `exception` is `null`.


keyspace:
Please see the referencing issue for more details.

dbent:
I think I ran into this issue at some point and forgot about it. Thanks for fixing.

Issue #1682, opened 2016-04-26
Upload FOSS mods to the Internet Archive, allow clients to fallback 


techman83:
# Proposal

Similar to #935, we can upload FOSS mods to the Internet Archive. There is example code over at KSP-CKAN/NetKAN-bot for uploading to the IA, the sticking point is the resulting uploaded file.

An [Item](https://archive.org/about/faqs.php#1084) per Mod Version made sense to me, but open to ideas. As the metadata per version could change.

The big rub is the filename. We have options, that have consequences.
1. Use similar method to cache, Generate hash from URL and derive the filename
   - :+1: We already do it in a similar manner, simple to implement
   - :+1: Easy
   - :+1: Requires no altering of the metadata
   - :-1: Client needs to derive the url
   -  :-1: Download URL changes, mod gets re-mirrored
   - :-1: Major mod hosting site goes offline, _a lot_ of mods get re-mirrored.
   - :-1: If a file changes it won't get re-uploaded
2. Generate a hash from the identifier/version, derive the filename
   - :+1: Easy
   - :+1: Requires no altering of the metadata
   - :+1: Download url changes, file is not re-uploaded.
   - :-1: Client needs to derive the url
   - :-1: If a file changes it won't get re-uploaded
3. Generate a hash on the downloaded file and add it to the metadata
   - :+1: We have a hash of the file
   - :+1: Client doesn't have to derive the filename
   - :-1: Requires altering the metadata
   - :-1: Requires spec change
   - :+1: :-1: If a file changes we could re-upload it and add it to the metadata. This may not always be desirable.

I came up with 2 whilst I was writing this.. It's the option I'm currently leaning towards. Pinging @dbent @pjf being more across internals of the Client code base and @Dazpoet @plague006 @politas being across the metadata.
# Implementation

Option 3 was decided on.

**Phase 1**
- [x] Get NetKAN to produce the field
- [x] Start uploading things. 95% of the code is already done, just add a webhook shim and watch CKAN-meta for changes.
- [x] Scan for any hooks we missed daily with git diff $(git rev-list -n1 --before="yesterday" master) --name-only KSP-CKAN/NetKAN-bot#49

**Phase 2**
- [x] Implement whatever we decide with CKAN-meta/NetKAN amalgamation - #3890
- [ ] Rearchitect the indexer (I have some solid ideas).
- [ ] Crawling head/api checks all day
- [ ] Full inflation daily on NetKAN_repo/NetKAN/
[...]

Issue #62, opened 2014-10-13
Support sha1/md5 fingerprints of downloaded files


pjf:
Right now we just blindly trust any file we download is what the user's after. It would be great if we supported a fingerprint (sha1/md5) field, allowing the ckan client to test that what it downloaded matches what we expect. This would catch incomplete downloads, incorrect versions, and upstream attackers using KSP mods to deliver code to a target machine.

These fields should be optional.

dbent:
The more I think about this the more I realize having a checksum is fine so long as it's fairly strong (SHA256 at least). In this case users would have to trust the CKAN team as an intermediary and the mod author as the originator... but they already do, implicitly, if they're using CKAN. It would at least protect against a 3rd party compromising the upstream host and sending malicious content down.

janbrohl:
Not sure if this belongs here but at least when generating hashes for #/ckan/http/:url you can only make sure netkan and you download the same file - not that it is the one created by the author

A solution woud be to add the possibility to store an author's public key (not even a full certificate) in the netkan files and verify the downloaded zip with it.

PR #1703, opened 2016-05-03
[PROPOSAL] Add download hash to Metadata

techman83:
**NOTE** Merging this will update all the current metadata on the next NetKAN-bot run. This is to aid in progressing #1682

@dbent - This is pretty much a duplicate of the DownloadSize Transformer. It generates a SHA1 of the downloaded file.

Tested on the redirect netkan
```
{
    "spec_version": 1,
    "identifier": "PlaneMode",
    "$kref": "#/ckan/http/http://www.zengei.com/tmp/redirect1",
    "version": "1.3.1",
    "license": "MIT"
}
```
Produces
```
{
    "spec_version": 1,
    "identifier": "PlaneMode",
    "license": "MIT",
    "version": "1.3.1",
    "download": "http://www.zengei.com/tmp/PlaneMode-1.3.1.zip",
    "download_size": 17530,
    "download_hash": "1F4B3F21A77D4A302E3417A7C7A24A0B63740FC5",
    "x_generated_by": "netkan"
}
```
Which matches what I can produce with shasum
```
leon@beast /tmp $ shasum C942169E-netkan-PlaneMode.zip | awk {'print $1}' |tr '[:lower:]' '[:upper:]' 
1F4B3F21A77D4A302E3417A7C7A24A0B63740FC5
```

I'm not hugely tied to the field name or the method of generation if there are better suggestions.


dbent:
A few things:

- You can stick the new hash method in `IFileService` the intent of that is to have all file related operations in an single interface that can be easily mocked for testing purposes.
- As per #62 we may use the hash for security purposes and SHA1 is on its way out for those purposes. I'd recommend the 256-bit variant of SHA2.
- I'd add the *type* of hash to the metadata in some capacity so we can change it later on, something like:

```json
  {
     "download_hash": {
       "type": "sha256",
       "digest": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
     }
  }
   ```

Other than that looks good.

**Comment by dbent** [comment]
Also, I was indifferent towards the casing of the digest but [RFC 4648](https://tools.ietf.org/html/rfc4648#section-8) specifies A-F for its alphabet so upper case is the way to go!

**Comment by dbent** [comment]
And, I just realized you need the sha1 for the Internet Archive, so I'd recommend:
```json
"download_hash": [
  {
    "type": "sha1",
    "digest": "1F4B3F21A77D4A302E3417A7C7A24A0B63740FC5"
  },
  {
    "type": "sha256",
    "digest": "E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855"
  }
]
```

techman83:
Having an array of hashes is a little more future proof as well. We can get the client to prefer stronger and fallback if in the future SHA256 is broken.
```
{
  "spec_version": 1,
  "identifier": "PlaneMode",
  "license": "MIT",
  "version": "1.3.1",
  "download": "http://www.zengei.com/tmp/PlaneMode-1.3.1.zip",
  "download_size": 17530,
  "download_hash": [
    {
      "type": "sha1",
      "digest": "1F4B3F21A77D4A302E3417A7C7A24A0B63740FC5"
    },
    {
      "type": "sha256",
      "digest": "7AB22F6E427B8FC96CF4635AF76515DE7EBBC8CAEDD2627FFFD7229A02EAA62D"
    }
  ],
  "x_generated_by": "netkan"
}
```

dbent:
LGTM, just need to update `Spec.md` for the new structure.

dbent:
@techman83 Sorry, I just realized, it would be easier to work with and more compact if we did:

```json
"download_hash": {
  "sha1": "1F4B3F21A77D4A302E3417A7C7A24A0B63740FC5",
  "sha256": "E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855"
}
```

Then if we want the `sha256` hash we can just index off the `download_hash` object rather than iterate through the array looking for the appropriate `type`. This also makes it impossible to have duplicate hashes of the same type.

techman83:
Yep, that's easier for me too. A little less awkward to build the object as well!

```
{
  "spec_version": 1,
  "identifier": "PlaneMode",
  "license": "MIT",
  "version": "1.3.1",
  "download": "http://www.zengei.com/tmp/PlaneMode-1.3.1.zip",
  "download_size": 17530,
  "download_hash": {
    "sha1": "1F4B3F21A77D4A302E3417A7C7A24A0B63740FC5",
    "sha256": "7AB22F6E427B8FC96CF4635AF76515DE7EBBC8CAEDD2627FFFD7229A02EAA62D"
  },
  "x_generated_by": "netkan"
}
```

techman83:
Closing this as the new changes go beyond just a 'download_hash'.

dbent:
Closing this as I assume @techman83 just forgot to actually do the deed. :wink:

techman83:
:dancers:

Issue #1628, opened 2016-03-13
ckan.exe picks Steam directory over current directory


pjf:
Reported in [this forum post](http://forum.kerbalspaceprogram.com/index.php?/topic/90246-the-comprehensive-kerbal-archive-network-ckan-package-manager-v1160-13-dec-2015/&do=findComment&comment=2454837) , using CKAN v1.16.0.

I haven't verified this myself, but it seems plausible. We _should_ be checking same directory as `ckan.exe` first, then everything else from there.

HebaruSan:
I think this was a false alarm. The code is pretty clear that `PortableDir` (ckan.exe's folder) is checked first, and `FindAndRegisterDefaultInstance` (Steam) is last:

https://github.com/KSP-CKAN/CKAN/blob/e825bdef9d52a4117e37acf6b6b74da216ee4567/Core/KSPManager.cs#L70-L105

A few ways the problem report could have been mistaken:

- Installing ckan.exe into GameData instead of game root
- Running a different copy of ckan.exe via a shortcut
- Symbolic links?

No details have been added in 2 years, so we're unlikely to see them now; closing.

Issue #723, opened 2015-01-02
URL handlers for ckan://


AlexanderDzhoganov:
https://github.com/KSP-CKAN/CKAN-GUI/pull/19 adds handlers for ckan://<identifier> for Windows and Linux. Not sure if this will work on OSX, maybe we need a separate config for it. The handler itself will run `ckan.exe gui %u` where ckan.exe is the location of the most recently used executable and %u is the identifier part.

AlexanderDzhoganov:
Using [this release](https://github.com/KSP-CKAN/CKAN-nightly/releases/tag/jenkins-CKAN-158) you'll be able to click [here](ckan://FerramAerospaceResearch) and have CKAN open.

pjf:
Marking this as closed, because we both have ckan:// handlers, and because they either work or fail in a non-user-provoking way. :)

Dazpoet:
From what I gather these where added so that KerbalStuff (or any other modwebsite for that part, KerbalTopia and KerbalX come to mind) would be able to have a button which led directly to CKAN client starting and marking the mod for installation. Do these handlers acctually work on all OS now or is it an unfinished feature? The question about these are raised from time to time on irc by @SirCmpwn

SirCmpwn:
Yes, they weren't usable last time anyone bothered to answer my questions about them.

Dazpoet:
Reopening so we can evalute the scope of this before opening a new issue with more relevant information. I'm not sure who'd know more save for @AlexanderDzhoganov though.

I assume the initial scope was to add the ability to click something like "Install mod through CKAN" on KerbalStuff which might help KS with https://github.com/KerbalStuff/KerbalStuff/issues/5 and give us a more obvious connection to where users would go looking for usage stats such as which mods are the most popular right now.

pjf:
I've not seen them work on my (Linux) machine, the ini parser library is unable to parse my `mimeapps.list` file. On other platforms I can't comment as to whether they work or not.

Halbann:
A short while later...

I'm interested in implementing this for a dependency helper within Lazy Painter. Some design questions:

- Is this actually something anyone still wants besides myself?
- Would simply `ckan://identifier/install` be an appropriate design for this? It's tempting to find a way to give more control (in a way that's backwards compatible with `ckan://identifier`), but maybe that's not desired.
[...]

PR #1704, opened 2016-05-03
Quote path to executable in URL handler .desktop file

reavertm:
Quote path to executable in URL handler .desktop file, likely fixes issue #1628.


reavertm:
Hello? Without proper path quoting, installing .desktop file is kinda pointless :)

politas:
This actually fails for me because the working directory is not a KSP installation (my CKAN is not stored in any of my KSP installs). Can you remove the `gui` directive? It launches fine without it. That would need some testing on Windows, I guess.

politas:
This may fix #723, actually. Reavertm, can you check that the mimeapps.lst entry is being correctly handled, too?

pjf:
Oh my! We absolutely need path quoting if the user's path may have spaces or other characters in it.

On my system things launch fine with the "gui" directive (on Linux), and it was there before this change, so this PR is making things objectively better than they were before, so I'm merging.

@reavertm : Thank you so much with your patience in us taking a little while to add this, and thank you so much for your contribution! this will be going out in today's release. :)

reavertm:
Glad I could help!

PR #1722, opened 2016-05-09
Add ksp_version_strict to property sort order

dbent:


plague006:
Counting from 16 to 27? I can review this!

PR #1723, opened 2016-05-09
GithubTransformer Enhancements

dbent:

This PR allows the GitHub transformer to write `abstract` based upon the repository description and `resources.homepage` based upon the repository website.

### DIFF
--- a/Netkan/Sources/Github/IGithubApi.cs
+++ b/Netkan/Sources/Github/IGithubApi.cs
@@ -2,6 +2,7 @@ namespace CKAN.NetKAN.Sources.Github
 {
     internal interface IGithubApi
     {
+        GithubRepo GetRepo(GithubRef reference);
         GithubRelease GetLatestRelease(GithubRef reference);
     }
 }
--- a/Netkan/CKAN-netkan.csproj
+++ b/Netkan/CKAN-netkan.csproj
@@ -52,6 +52,7 @@
     <Compile Include="Sources\Github\GithubApi.cs" />
     <Compile Include="Sources\Github\GithubRef.cs" />
     <Compile Include="Sources\Github\GithubRelease.cs" />
+    <Compile Include="Sources\Github\GithubRepo.cs" />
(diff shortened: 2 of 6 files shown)


politas:
I can understand most of what's happening here, but there are calls to a "GetRepo" that I can't find. Is it an external function?

Netkan/Sources/Github/IGithubApi.cs
```diff
@@ -2,6 +2,7 @@ namespace CKAN.NetKAN.Sources.Github
 {
     internal interface IGithubApi
     {
+        GithubRepo GetRepo(GithubRef reference);
         GithubRelease GetLatestRelease(GithubRef reference);
```
techman83: @politas GetRepo is made available here. IGithubApi gets passed in during the object construction of GitHubTransformer.

techman83:
LGTM. Tested some NetKANs, Merging with Thanks!

politas:
Would it be possible to scan the text associated with a release to look for a "KSP <version>" string and use that for `"ksp_version"`? Should probably limit the searching to the current patch (1.1.2), Release level (1.1), and say the last three patches (1.1.1, 1.1.0, 1.0.5)

dbent:
@politas I'm generally opposed to heuristics like that because of the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment); modders would likely be *very* surprised that we're extracting meaningful information from the title of their releases, and annoyed if metadata breaks because of that (*"Wait, you were relying on me adding the KSP version to the title of each of my releases for everything to work? Wat?"*).

I'd rather gentle push modders to add `.version` files or host their own metadata.

politas:
You don't think that some of them are thinking "but the release notes say it is for version y, why haven't you fixed it to work on that version," now?

techman83:
Whilst I use regex like a hammer at times, I'm with @dbent on this one. I think we could raise issues based on something like this, but I'd be pretty hesitant to change metadata based on it. That's something the indexer could be doing.

PR #1726, opened 2016-05-10
GitHubTransformer shouldn't add empty data

dbent:

GitHub API returns an [empty string for the homepage in some cases](https://github.com/KSP-CKAN/CKAN-meta/commit/9805b7fe4edea89d0c7352d230d66c34d8546956), guard against that and adding empty descriptions as the abstract.


politas:
LGTM. Once in a while, you guys make some code changes I can understand!

PR #1728, opened 2016-05-11
install as

dbent:

This PR adds supports for a feature similar to the [previously discussed](https://github.com/KSP-CKAN/CKAN/pull/1243#issuecomment-118680302) `install_as` allowing directories or files to be renamed.

Unlike the previously discussed implementation, there is no separate `install_as` which replaces `install_to`. Instead there is an optional extra property `as` which is the name to give to the matching directory or file when installed. This was done to minimize the number of changes required in the code base, especially in an area as sensitive as the install code.

For example, given the following ZIP structure:
- Foo/Bar/File1.dll
- Foo/Bar/File2.dll
- Foo/Bar/Baz/File3.dll

This install structure:
```json
{
  "file": "Foo/Bar",
  "install_to": "GameData/Foo",
  "as": "Bar-Moved"
}
```

Would produce the following installed files relative to the KSP root:
- GameData/Foo/Bar-Moved/File1.dll
- GameData/Foo/Bar-Moved/File2.dll
-  GameData/Foo/Bar-Moved/Baz/File3.dll

Or this install structure:
```json
{
  "file": "Foo/Bar/File1.dll",
  "install_to": "GameData/Foo2/Bar2",
  "as": "FileOne.dll"
}
```
Would produce the following installed files relative to the KSP root:
- GameData/Foo2/Bar2/FileOne.dll

For a more practical example, take this `.ckan` file:
```json
{
    "spec_version": "v1.18",
    "identifier": "Kerbalism-Profile-TAC",
    "name": "Kerbalism - TAC Profile",
    "abstract": "TAC profile for Kerbalism",
    "author": "ShotgunNinja",
    "license": "public-domain",
    "resources": {
        "homepage": "https://github.com/ShotgunNinja/Kerbalism/wiki",
        "spacedock": "https://spacedock.info/mod/533/Kerbalism",
        "repository": "https://github.com/ShotgunNinja/Kerbalism",
        "x_screenshot": "https://spacedock.info/content/ShotgunNinja_2282/Kerbalism/Kerbalism-1460950521.2737155.png"
    },
    "version": "0.9.9.7",
    "ksp_version": "1.1.2",
    "provides": [
        "Kerbalism-Profile"
    ],
    "depends": [
        {
            "name": "Kerbalism"
        },
        {
            "name": "CommunityResourcePack"
        }
    ],
    "conflicts": [
        {
            "name": "Kerbalism-Profile"
        }
    ],
    "install": [
        {
            "file": "GameData/Kerbalism/Profiles/TAC.cfg-disabled",
            "install_to": "GameData/Kerbalism/Profiles",
            "as": "TAC.cfg"
        }
    ],
    "download": "https://spacedock.info/mod/533/Kerbalism/download/0.9.9.7",
    "download_size": 5020915,
    "download_hash": {
        "sha1": "778492A3846D0DC9E87084BD187475D2B3BA3FE3",
        "sha256": "1FBA3441A5A52CBBE7BE09F4C3A407D8CFFF2C87AB8659A77398AAB16B7E0246"
    },
    "x_generated_by": "netkan"
}
```

Which allows use to rename the **TAC.cfg-disabled** file to **TAC.cfg**, enabling it when the `Kerablism-Profile-TAC` package is installed.

**Because this affects installation behavior, using this requires a spec version bump and new release.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- [Core] `install` stanzas can have an `as` property allowing directories and files to be renamed/moved on installation. (#1728 by: dbent)
 - [GUI] Added "filter by description" search box. (#1632 by: politas; reviewed: pjf)
 - [CLI] `compare` command now checks positive and negative rather than -1/+1 (#1649 by: dbent; reviewed: Daz)
 - [GUI] In windows launch KSP_x64.exe by default rather than KSP.exe. (#1711 by plague006; reviewed: )
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -203,6 +203,11 @@
                         "description" : "Where file should be installed to",
                         "$ref"        : "#/definitions/install_to"
                     },
+                    "as": {
+                        "description" : "The name to give the matching directory or file when installed",
+                        "type"        : "string",
+                        "pattern"     : "^[^/\\]+$"
+                    },
                     "filter" : {
                         "description" : "List of files and directories that should be filtered from the install",
                         "$ref"        : "#/definitions/one_or_more_strings"
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -501,7 +501,7 @@ namespace CKAN
                     string outputName = entry.Name;
 
                     // Update our file info with the install location
-                    file_info.destination = TransformOutputName(stanza.file, outputName, installDir);
+                    file_info.destination = TransformOutputName(stanza.file, outputName, installDir, stanza.@as);
                 }
 
                 files.Add(file_info);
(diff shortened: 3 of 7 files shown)


plague006:
This is awesome, definitely looking forward to having this added to our toolset.

dbent:
I've updated the code and schema to prevent directory separators (`/` or `\`) from appearing in the `as` property. I realized it's not really necessary to be able to specify a sub path in `as` since you could always just change the value of `install_to` to include the subdirectories. Since I'm a big fan of [orthogonality](https://en.wikipedia.org/wiki/Orthogonality_%28programming%29) I've updated the code to make that explicit.

Therefore:
- Use `file`/`find`/`find_regexp` to select a file or directory to install.
- Use `install_to` to select the directory in which it should be installed.
- Use `as` to select a name to give the file or directory when it's installed.

techman83:
Ok this looks really neat, with test cases and totally agree with explicit functionality. Will need a release before it's usable.

Issue #1729, opened 2016-05-14
Kerbalism depends on Community Resource, CKAN doesn't seem to think so.


rnewbert:
Thanks for wanting to bring an issue to our attention!
**CKAN Version:** 
1.16.1-0-g2e91715

**Operating System:** 
windows 10

**The issue you are experiencing:** 
Kerbalism mod requires the community resource pack.  attempting to install it with ckan wile the community resource pack is already installed makes ckan throw an error about them being incompatible.  i installed Kerbalism myself, and this caused the issue here, where ckan throws the below message and doesn't let me do anything, the gui show up and menus and buttons respond, but 'add all available updates' and 'install mods' are both grayed out.  further, any interaction other then the file, about, etc menus at the top re-throws the error. 
**How to recreate this issue:** 
install CRP, a mod that requires it (not kerbalism, i recommend USIs mods as thats one of the others i have that requires it), then try to install kerbalism.  install it manually after it tells you they are not compatible.  relaunch ckan.
**CKAN error codes (if applicable):** 
error text: 'CKAN.InconsistentKraken'

error details:
See the end of this message for details on invoking 
just-in-time (JIT) debugging instead of this dialog box.

************** Exception Text **************
The following inconsistencies were found:
CommunityResourcePack conflicts with Kerbalism.   at CKAN.SanityChecker.EnforceConsistency(IEnumerable`1 modules, IEnumerable`1 dlls)
   at CKAN.RegistryManager.Save(Boolean enforce_consistency, Boolean recommmends, Boolean with_versions)
   at CKAN.Main.UpdateRepo()
   at CKAN.Main.OnLoad(EventArgs e)
   at System.Windows.Forms.Form.OnCreateControl()
   at System.Windows.Forms.Control.CreateControl(Boolean fIgnoreVisible)
   at System.Windows.Forms.Control.CreateControl()
   at System.Windows.Forms.Control.WmShowWindow(Message& m)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.Form.WmShowWindow(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

************** Loaded Assemblies **************
mscorlib
    Assembly Version: 4.0.0.0
    Win32 Version: 4.6.1080.0 built by: NETFXREL3STAGE
[...]

PR #1730, opened 2016-05-15
Relationship changes should prompt reinstall

dbent:
@pjf @techman83 @plague006 @KSP-CKAN/wranglers 

This PR makes it so that changes in `provides`, `conflicts`, `depends`, and `recommends` relationships prompt the user to reinstall the affected packages. This should handle [this case](https://github.com/KSP-CKAN/CKAN/issues/1729#issuecomment-219255719).

The basic change is that when we check for equivalent `install` stanzas we also check for equivalent relationship stanzas.

Because updating metadata can and should be able to put the user's registry in an inconsistent state (e.g. user installs Foo and Bar, later Foo conflicts with Bar, on update their repository is now "inconsistent"), I selectively ignore enforcing consistency checks when updating and just tell the user about the inconsistencies. (Annoyingly the GUI does its own thing in `Core/KSP.cs`).

I've also changed the code that checks old metadata against the new metadata to use the *installed* module as the old metadata, not the *available* module. This means if a user decides *not* to reinstall immediately, they'll be prompted to reinstall until the installed metadata matches the repository metadata. This can be done either by selecting yes at the prompt or by manually upgrading the package. Previously, if a user selected no they would not be prompted to reinstall in the future.

### DIFF
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -106,14 +106,14 @@ namespace CKAN.CmdLine
             }
             catch (ModuleNotFoundKraken ex)
             {
-                user.RaiseMessage("Module {0} required, but not listed in index, or not available for your version of KSP", ex.module);
+                user.RaiseMessage("Module {0} required but it is not listed in the index, or not available for your version of KSP.", ex.module);
                 user.RaiseMessage("If you're lucky, you can do a `ckan update` and try again.");
-                user.RaiseMessage("Try `ckan install --no-recommends` to skip installation of recommended modules");
+                user.RaiseMessage("Try `ckan install --no-recommends` to skip installation of recommended modules.");
                 return Exit.ERROR;
             }
             catch (BadMetadataKraken ex)
             {
-                user.RaiseMessage("Bad metadata detected for module {0}", ex.module);
+                user.RaiseMessage("Bad metadata detected for module {0}.", ex.module);
                 user.RaiseMessage(ex.Message);
                 return Exit.ERROR;
             }
--- a/Core/KSP.cs
+++ b/Core/KSP.cs
@@ -418,7 +418,7 @@ namespace CKAN
                     
                 tx.Complete();
             }
-            RegistryManager.Save();
+            RegistryManager.Save(enforce_consistency: false);
         }
 
         #endregion
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -868,7 +868,7 @@ namespace CKAN
         /// </summary>
         /// <param name="add">Add.</param>
         /// <param name="remove">Remove.</param>
-        public void AddRemove(IEnumerable<CkanModule> add = null, IEnumerable<string> remove = null)
+        public void AddRemove(IEnumerable<CkanModule> add = null, IEnumerable<string> remove = null, bool enforceConsistency = true)
         {
 
             // TODO: We should do a consistency check up-front, rather than relying
@@ -887,7 +887,7 @@ namespace CKAN
                     Install(module);
                 }
 
-                registry_manager.Save();
+                registry_manager.Save(enforceConsistency);
 
                 tx.Complete();
             }
@@ -898,7 +898,7 @@ namespace CKAN
         /// Will *re-install* with warning even if an upgrade is not available.
         /// Throws ModuleNotFoundKraken if module is not installed, or not available.
         /// </summary>
-        public void Upgrade(IEnumerable<string> identifiers, IDownloader netAsyncDownloader)
+        public void Upgrade(IEnumerable<string> identifiers, IDownloader netAsyncDownloader, bool enforceConsistency = true)
         {
             var options = new RelationshipResolverOptions();
 
@@ -909,7 +909,7 @@ namespace CKAN
             var resolver = new RelationshipResolver(identifiers.ToList(), options, registry_manager.registry, ksp.Version());
             List<CkanModule> upgrades = resolver.ModList();
 
-            Upgrade(upgrades, netAsyncDownloader);
+            Upgrade(upgrades, netAsyncDownloader, enforceConsistency);
         }
 
         /// <summary>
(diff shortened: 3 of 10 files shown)


politas:
Oooh, this sounds excellent! Excellent enough that I'm going to download and test it!

plague006:
All done within my test branch (https://github.com/plague006/CKAN-meta/archive/test.zip). I used `InterstellarFuelSwitch-Core` as a test since it's a small download. After installing it I pushed `ModuleManager` as a new dependency to IFS-Core https://github.com/plague006/CKAN-meta/commit/fd6980f27e2331b9afb5159f23df7005fac6645c

After hitting refresh I was tossed back to "Manage mods" with the status message at the bottom indicating everything was fine (http://puu.sh/oSKAD/a7cc2fe240.png). Looking at the "Updating repositories" tab the following message is listed:

````
- InterstellarFuelSwitch-Core requires ModuleManager but nothing provides it

Repository successfully updated
````

Back in "Manage mods" no mod can be installed nor uninstalled ("Apply Changes" remains greyed out). I'll note that in the registry in the `installed_modules` section IFS-Core has the MM dependency added.

dbent:
:disappointed: *GUI*. I assume it works fine with the CLI?

plague006:
Ran into a different problem with CLI. Only tested it once so far but when I tried to re-install IFS-core it failed because of the broken dependency. http://puu.sh/oSMmG/adc72469fe.png

dbent:
Updated with changes that should actually handle resolving relationships when you're prompted to reinstall mods and get rid of that scary error message and replace it with a informational message.

plague006:
Will test anew.

plague006:
I tested a case where the dependency can't be met, here's the output: http://puu.sh/oSOV6/a8e0b1000e.png

dbent:
I suppose the best way to handle that case is to try to upgrade each mod individually, catch depends and conflicts exceptions and print them to the user. Probably have something that does that in the next half hour or so.

dbent:
> Probably have something that does that in the next half hour or so.

28 minutes later.

plague006:
Installed `X` then added a dependency and it re-installed `X` and installed the dependency flawlessly. Added a dependency that couldn't be met and got a clear error message. Installed `Y` then added a dependency. `X` failed to update as before but `Y` re-installed and installed dependency as expected.

The effects on the GUI are.. better. Because the inconsistent metadata from mod `X` wasn't written into the registry the GUI allows installs and removals but fails on refresh.

plague006:
Added a few tweaks to the error messages I stumbled upon while testing this. A little more consistent, and a little less jargon-y ("provides" probably isn't very clear to end-users). Feel free to force-push it for cleanliness.

politas:
Compiling that latest version, I get a warning message 
    Net/Repo.cs(301,51): warning CS0168: The variable `ik' is declared but never used

I'm thinking that is a search tag you were using that should be removed?

dbent:
> Added a few tweaks to the error messages I stumbled upon while testing this. A little more consistent, and a little less jargon-y ("provides" probably isn't very clear to end-users). Feel free to force-push it for cleanliness.

Cool thanks.

> I'm thinking that is a search tag you were using that should be removed?

Just the name of an exception variable I never ended up using, not harmful, but I'll remove it.

It'll probably take me another day or two to look into the GUI issues, but this should be good to go after that.

dbent:
@plague006 Can you give me an example of the GUI screwing up with this? From what I've tested things seem to work fine.

Only issue is that it seems the `GUIUser` class which is supposed to manager user interactions in the GUI always returns Yes for every Yes/No dialog, but if that's the way it's always been...

plague006:
The metadata was unable to be updated. (Hitting the Refresh button would lead to a crash)

It's hard to keep track of what I was testing at any particular time or when it was fixed since there's no commit history. I'll give it another go at some point today or tomorrow.

plague006:
CLI continues to be solid. In the case of GUI new dependencies install but aren't marked as installed until CKAN is restarted. http://puu.sh/pfRcj/2ffb777d42.png

dbent:
Thanks @plague006. Pretty sure the GUI is the bane of my CKAN existence. I'll look into when I have a chance.

dbent:
94a987b should fix the issue @plague006 saw.

dbent:
Mono 3.10 build failing because `mozroots --import --ask-remove` is failing because http://mxr.mozilla.org/seamonkey/source/security/nss/lib/ckfw/builtins/certdata.txt?raw=1 redirects to https://hardhat.cdn.mozilla.net/en-US/outages.html

pjf:
Looking at the code, and trying to figure out the order in which things happen. In particular, we have switches that allow for the registry to be saved in an *inconsistent* state, where not all relationship dependencies are satisfied. We didn't have that before.

How do we get into an inconsistent state? I would understand that:

- When a mod's metadata has changed, and the user chooses not to reinstall, we keep the old metadata. If the registry was previously consistent, it will remain consistent.
- When a mod's metadata has changed, and the user chooses to reinstall, we *only* do that if all relationship constraints can be satisfied. The result should have a consistent registry.
- When a mod's metadata has changed and the new metadata doesn't allow for us to reinstall in a consistent fashion, then keeping the old metadata *feels* like the most sensible thing to do, but I don't have a strong argument for that.

Mostly I want to make sure we don't *accidentally* end up with an inconsistent registry and broken installs. (I do have a half-formed idea around an explicit 'broken' package state, but it's still forming.)

dbent:
The problem is in the case of this sequence of events:

- User installs Mod A.
- User installs Mod B.
- Mod A's metadata is changed so that it now conflicts with Mod B.
- User performs an update.

Prior to this PR. What would then happen is:

- CKAN ignores the new conflicts relationship and everything *appears* hunky-dory (when it really isn't).

Now with this PR the following happens:

- CKAN reinstalls Mod A and warns the user of the new conflicts relationship on Mod B.

In both cases the registry is "inconsistent" in that the user has two mods installed that are in conflict with one another, it's just that in the current case the user has no idea about it because CKAN doesn't update the metadata.

In fact the changes were specifically designed to *allow* the registry to exist in an inconsistent state so that users can be warned of new relationship issues, either conflicts that now exist that did not previously or dependencies that are no longer satisfied (for example, if Mod C depends on Mod D, and Mod E is installed which provides Mod D, and that provides is subsequently removed, the user should also be warned). So the possibility of an inconsistent registry is in fact a feature not a bug of this PR.

There's a weak assumption in the existing codebase that a consistent registry will remain consistent for all time so long as CKAN operates correctly, but that's only true *if CKAN ignores relationship changes*. Since this PR no longer ignores relationship changes that assumption can no longer hold.

politas:
@dbent, do you have time to rebase this?

dbent:
Sure, was going to do it sometime soon-ish anyhow.

politas:
Merged

PR #1734, opened 2016-05-15
Fix unit of measure for download speed

### LINKED ISSUE #1732: Kilobytes displayed as kilobits
CKAN v1.16.1-0-g2e91715 (beta)
Windows 10 64-bit

The download speed given by CKAN does not match Task Manager. It appears that it is displaying kiloBytes per second but the unit say kilobits per second.

image

plague006:
Closes #1732

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Bugfixes
 - [Core] In certain cases a `NullReferenceException` could be produced inside error handling code when processing the registry. (#1700 by: keyspace, reviewed: dbent)
 - [GUI] Fix typo in export options. (#1718 by: dandrestor, reviewed: plague006)
+- [GUI] Fix unit of measure for download speed. (#1732 by: plague006, reviewed: )
 
 ### Features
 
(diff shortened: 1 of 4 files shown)


dbent:
[Technically](https://www.youtube.com/watch?v=hou0lU8WMgo), since we're dividing by 2<sup>10</sup>, it's *KB/s* if you're using the [JEDEC binary prefix](https://en.wikipedia.org/wiki/Binary_prefix) standard or *KiB/s* if you're using the [IEC binary prefix](https://en.wikipedia.org/wiki/Binary_prefix) standard. *k* is the SI decimal prefix representing 10<sup>3</sup>.

Normally, I wouldn't bitch but we're using another IEC binary prefix on the same line. :smile:

plague006:
Changed to KB/s and changed all IEC units I could find to JEDEC.

dbent:
LGTM

politas:
Why are we going with JEDEC, rather than converting everything to the ISO/IEC/NIST/IEEE/everyone else in whole world standard?

plague006:
While I'm not saying Windows is the beginning and the end of the conversation, I do think it's safe to say it's the lowest common denominator in terms of the knowledge a typical user has. With that said windows uses [JEDEC Standard 100B.01](https://en.wikipedia.org/wiki/JEDEC_memory_standards#JEDEC_Standard_100B.01)

https://blogs.msdn.microsoft.com/oldnewthing/20090611-00/?p=17933/

Issue #1113, opened 2015-06-13
Add the Unlicense as a valid license


dbent:
The [Unlicense](http://unlicense.org/) is basically a public domain grant and is not currently in our list of [valid licenses](https://github.com/KSP-CKAN/CKAN/blob/master/CKAN.schema#L254).

Came up because of KSP-CKAN/NetKAN#1606.

PR #1737, opened 2016-05-18
Add Unlicense

plague006:
Closes https://github.com/KSP-CKAN/CKAN/issues/1113

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -347,10 +347,11 @@
                 "Perl",
                 "Python-2.0",
                 "QPL-1.0",
+                "Unlicense",
                 "W3C",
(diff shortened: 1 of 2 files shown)


techman83:
Looks like another allowed license to add to the MirrorBot. LGTM, merging with Thanks!

plague006:
I noticed in https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#license that WTFPL needs spec_ver v1.2, is that accurate? If so I'll add an entry for Unlicense.

plague006:
I tested it myself, CKAN absolutely freaks if it gets a license it doesn't recognize.

PR #1742, opened 2016-05-20
Force either `ksp_version` or `ksp_version_max` to be included in .ckans to reduce frequency of promiscuous mods

plague006:
My best guess at how to do this, I have no method to actually test this.

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -256,6 +256,10 @@
             "required": [ "kind" ]
         }
     ],
+    "oneOf": [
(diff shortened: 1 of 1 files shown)


janbrohl:
What about mimicking KSP-AVC's practice to understand ksp_version as targetted version which is mandatory and leaving ksp_version_min and ksp_version_max optional

plague006:
After reconsideration, this is probably better done as a metadata test during Jenkins runs than as a hard change to the schema and spec.

PR #1747, opened 2016-05-22
Added Docker support.

mathuin:
The only change required to the core code was a modification to the root detection routine, to exempt Docker containers from the --asroot requirement.

To build the container:
$ docker build -t ckan .

To use the container to update all mods:
$ docker run --rm -v ${KSPDIR}:/kspdir ckan

To use the container to install MechJeb:
$ docker run --rm -v ${KSPDIR}:/kspdir ckan install MechJeb

Both of the last two lines require that the ${KSPDIR} value be set.

There is a docker-compose.yml supplied which will automatically do this for Linux users.

To use the YAML file to build the container:
$ docker-compose build ckan

To use the YAML file to update all mods:
$ docker-compose run --rm ckan

To use the YAML file to install MechJeb
$ docker-compose run --rm ckan install MechJeb


pjf:
Isn't there some sort of law that everything ends up as a docker container? :)

I've tested this and it works. It looks like a few extraneous commits have got mixed up in the branch, so I'm going to cherry-pick, add a little documentation, and merge.

Thank you very much for helping #DockerAllTheThings! :)

Issue #1749, opened 2016-05-23
Too many public-domain entries


janbrohl:
There are 85 (with duplicates) entries in the index stating the referenced mods are ind the public domain, it does not look like they actually are - see the public domain section in https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#license-specification 

The ids are:
```
TemperatureGaugeKiller
NavBallTextureExport
ScienceFunding
VNG-Parachute-EVALoading
PatentPurchase
ContractConfigurator-SCANsat
StockRT
ReflectionPlugin
Kerbalism
TCShipInfo
rt-config-by-ippo
ExceptionDetector
NavBallsToYou
HarderSolarSystem-4x
FacilityReset
NoMoreGrindRedux
CoolRockets
HoldTheMonoprop
DangerAlerts
StockPlugins
KScale2
StockSCANsat
StockNoContracts
KSPCalendar
LayeredAnimations
ControlLock
HarderSolarSystem-2x
HarderSolarSystem-1x
Kerbalism-Profile-Default
HarderSolarSystem-3x
HarderSolarSystem-generic
```

politas:
On what are you basing the assertion that they are not in the public domain?

plague006:
TemperatureGaugeKiller - to MIT per http://forum.kerbalspaceprogram.com/threads/120731

ContractConfigurator-SCANsat - to WTFPL per https://github.com/SimonTheSourcerer/CCSCANsatMissionPack

VNG-Parachute-EVALoading - to unknown per http://forum.kerbalspaceprogram.com/threads/25305?p=442982&viewfull=1#post442982

NavBallTextureExport - no change per https://bitbucket.org/xEvilReeperx/ksp_navballtexturechanger/src/13d1e0736cd9d67c4ab4dff9e47db2f2f3f9c318/NavBallTextureExport/NBTexExport.cs?at=master&fileviewer=file-view-default

ScienceFunding - no change per https://spacedock.info/mod/554/Science%20Funding

PatentPurchase - no change per https://spacedock.info/mod/692/Patent%20Purchase

StockRT - no change per https://spacedock.info/mod/113/StockRT

ReflectionPlugin - no change per http://forum.kerbalspaceprogram.com/index.php?/topic/63612-105-reflection-plugin-continued-20/

Kerbalism - no change per https://github.com/ShotgunNinja/Kerbalism/blob/master/LICENSE

[...]

PR #1750, opened 2016-05-23
Special comment on public-domain license in Spec.md

janbrohl:
PR for #1749

### DIFF
--- a/Spec.md
+++ b/Spec.md
@@ -166,6 +166,7 @@ The same rules as per the
 * Stripping of trailing zeros is not recognised.
 * (**v1.2**) `WTFPL` is recognised as a valid license.
 * (**v1.18**) `Unlicense` is recognised as a valid license.
+* `public-domain` is taken as to only express the author's intent to put the mod in the public domain, not that it actually is uncopyrighted - see https://en.wikipedia.org/wiki/Public-domain_software
 
 The following license strings are also valid and indicate other licensing not
 described above:


politas:
This is a misapplication of public domain law. Works may revert to public domain in different jurisdictions earlier or later, but if an author places a work in the public domain, it is in the public domain everywhere. This change is unnecessary.

dbent:
> This is a misapplication of public domain law. Works may revert to public domain in different jurisdictions earlier or later, but if an author places a work in the public domain, it is in the public domain everywhere. This change is unnecessary.

Not quite, believe it or not, in some jurisdictions it's basically impossible to relinquish copyright. That's why licenses like [CC0](https://creativecommons.org/about/cc0/) and the [Unlicense](http://unlicense.org/) exist. It's also why prominent public domain works like [SQLite](http://www.sqlite.org/copyright.html) have to have special licenses to deal with the fact.

>  Even though SQLite is in the public domain and does not require a license, some users want to obtain a license anyway. Some reasons for obtaining a license include:
>  - You are using SQLite in a jurisdiction that does not recognize the public domain.
>  - You are using SQLite in a jurisdiction that does not recognize the right of an author to dedicate their work to the public domain.

However, I agree that *intent* is what we're signalling with `public-domain`. If we go by the requirement that a work has to be public domain *everywhere* to be `public-domain` then we'll *never* use `public-domain` because no mod created for KSP would ever fulfill that criteria, at least for the next hundred years or so.

plague006:
> For a one-time fee of US $6000.00 we will add your company name and contact information to a copy of the sample license agreement, sign the completed license, and send you the original signed copy by post and a scan of the signed document by email.

TIL public domain software is hella expensive (to get a license for).

politas:
I am unconvinced that any jurisdictions that do not recognise the public domain, or an author's right to dedicate their work to the public domain actually exist.

janbrohl:
@politas 

There is a reason why licenses like [CC0](https://creativecommons.org/about/cc0/) exist (see [CC on PD](https://creativecommons.org/about/pdm/) )

Also have a look at https://en.wikipedia.org/wiki/Public-domain_software from which I took this quote:
```
In some Jurisdictions, some rights (in particular moral rights) cannot be disclaimed: 
for instance, civil law tradition-based German law's "Urheberrecht" differs here from the Anglo-saxon common law tradition's "copyright" concept.
```

plague006:
The US is the easiest country to find documentation on because of the nature of the internet. So here are a bunch of articles talking about how unclear it is how or even if one can release their copy rights. 

https://www.publicknowledge.org/news-blog/blogs/lets-make-it-easier-to-expand-the-public-domain

https://www.techdirt.com/articles/20150123/15564629797/why-we-still-cant-really-put-anything-public-domain-why-that-needs-to-change.shtml

https://opensource.org/faq#public-domain

To mix it up, this source says you *can* relinquish rights in the US and Europe but doesn't really touch on any other regions: https://cr.yp.to/publicdomain.html

plague006:
Ultimately, abandoning rights and the public domain in general is messy and we're not lawyers so why not have a little caveat/disclaimer included?

politas:
Because it's utterly pointless. Who is ever going to dispute any of this? Who is going to claim copyright protections on game mods they clearly intended to release into the public domain?

janbrohl:
Maybe it really is pointless as the licenses in the index do not have to match those in the archives anyway and are only hints to what the actual license might be.

janbrohl:
Was a pointless attempt to make license info more reliable

PR #1752, opened 2016-05-25
Add download hashes to registry

plague006:
At some point they'll probably be needed client-side, and this gave me an excuse to poke around exactly how the registry is created.

### DIFF
--- a/Core/Types/CkanModule.cs
+++ b/Core/Types/CkanModule.cs
@@ -85,7 +85,7 @@ namespace CKAN
         public Uri spacedock;
     }
 
-    public class DownloadHashsDescriptor
+    public class DownloadHashesDescriptor
     {
         [JsonProperty("sha1")]
         public string sha1;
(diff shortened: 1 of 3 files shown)


techman83:
Can you add a test for this change?

hakan42:
Minor nitpick, could you name it "DownloadHashesDescriptor"? Adding a "e" to the plural?

plague006:
Typo corrected for the class name. Adding a test will take a little longer... Namely I'll have to find a tutorial on what they are and how to write them.

techman83:
@plague006 You should be able to add the Download_Hash data to the test data [here](https://github.com/KSP-CKAN/CKAN/blob/master/Tests/Data/TestData.cs#L274) and then do the asserts [here](https://github.com/KSP-CKAN/CKAN/blob/master/Tests/Core/Types/Module.cs#L34).

plague006:
Thanks @techman83. Haven't had the time to learn up, but can follow instructions and copy+paste like a pro. :D

pjf:
I can't see a reason not to add this, and since I'm bumping the release number for tonight with all the spec changes, this makes a great time to do so.

Tested everything is good, merging now.

PR #1753, opened 2016-05-25
State requirements for metanetkans

### LINKED ISSUE #1751: Spec should list required fields for metanetkans
The spec doesn't specify which fields are required for metanetkans and it'd be nice if it did.

@dbent poking you since you probably have the answer at hand without having to look anything up.

plague006:
Closes  #1751


politas:
LGTM

PR #1769, opened 2016-06-02
JSON Schema in README

janbrohl:
There is no schema for netkan files - updated this.

The schema for ckan files is in the Schema Store - added that info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ We very much welcome contributions, discussions, and especially pull-requests.
 At the core of the CKAN is the **[metadata specification](Spec.md)**,
 which comes with a corresponding [JSON Schema](CKAN.schema).
 
-This repository includes a JSON schema validator that you can use to [validate your files][3].
+This repository includes a validator that you can use to [validate your files][3].
 
 ## CKAN for players
 
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ We very much welcome contributions, discussions, and especially pull-requests.
 ## The CKAN spec
 
 At the core of the CKAN is the **[metadata specification](Spec.md)**,
-which comes with a corresponding [JSON Schema](CKAN.schema).
+which comes with a corresponding [JSON Schema](CKAN.schema) which you can also find in the [Schema Store][8]
 
 This repository includes a validator that you can use to [validate your files][3].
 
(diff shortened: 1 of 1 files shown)


politas:
I don't like the new sentence any more than the old one. Neither seems like an accurate statement. The netkan.exe program is primarily an inflator, not a verifier. It verifies as part of the larger process of expansion. You can use it to verify your creation, but you're testing your file using the full inflation program, not a separate validator.

janbrohl:
The point was just to not link JSON schema validator to mostly netkan.exe making as few changes as possible.

politas:
Is netkan.exe not validating the metadata against the JSON schema? I assume that was the original intention of that sentence.

janbrohl:
There is no .netkan schema (yet) so netkans cannot be validated against it. I dont know if netkan.exe validates the output (.ckan)

plague006:
Netkan does not check against the schema. Neither the input .netkan nor the output .ckan.

janbrohl:
For output it seems like it would be good idea to me to check for schema-validity.

plague006:
> it would be good idea to me

If you would expand on this it would be much easier to understand your rationale.

janbrohl:
Generating invalid ckan files is not very useful (as they might be unreadable for other programs like ckan.exe) so it should be made sure that everything netkan.exe outputs is as valid as possible.

While not as good as actual validity (including semantics), schema-validity can be checked quite easily. Of course netkan.exe should produce valid output anyway but for parts of the input that are just passed through (eg identifier field) at least some checking should be done. A way to do this is schema-validating the output. This would also have the benefit of detecting (some) errors that netkan.exe might produce in special cases.

techman83:
@janbrohl NetKAN does do the basics
```
leon@beast /tmp $ mono netkan.exe test.netkan 
94 [1] FATAL CKAN.NetKAN.Program (null) - spec_version must be specified
leon@beast /tmp $ mono netkan.exe test.netkan 
123 [1] FATAL CKAN.NetKAN.Program (null) - Metadata must have an identifier property
```
And the indexer runs every generated CKAN through the validator before committing to ckan-meta.

ayan4m1:
OK, at this point we are talking about multiple things here: the PR itself is a documentation change, and then there is discussion about an enhancement to netkan regarding validation.

@politas Mind helping decide if the docs are something we want or not?

Either way, please feel free to open or reference an issue that explains the desired validation semantics and functionality - that will let us triage it appropriately and keep PRs moving forward efficiently.

politas:
I think it was more a question of checking the actual functionality. I think the changes to the page are ok, and I've found the Schema Store validator useful even on a .netkan file. 
At some stage, it might be good to have a full .netkan schema, but it would be best to generate it as an extension of the CKAN schema, in order to make sure any changes are kept in sync.

In the interests of cleaning up old PRs, I'm going to merge this.

PR #1771, opened 2016-06-02
Update Spec.md

### LINKED ISSUE #1770: Please clarify install_to in the Spec
At least i did not understand what paths are OK to put in install_to correctly. I thought it would only be these

> Valid values for this entry are GameData, Ships, Ships/SPH(v1.12), Ships/VAB(v1.12), Ships/@thumbs/VAB(v1.16), Ships/@thumbs/SPH(v1.16), Tutorial, Scenarios (v1.14) and GameRoot (which should be used sparingly, if at all)

but i am told some subdirectories are OK, too.  

Please clarify the install_to section because maybe it is  not just me being stupid.

politas:
An attempt to clarify the install_to options.
Closes #1770

### DIFF
--- a/Spec.md
+++ b/Spec.md
@@ -279,11 +279,7 @@ In addition a destination directive *must* be provided:
   and `GameRoot` (which should be used sparingly, if at all).
   Paths will be preserved, but directories will *only*
   be created when installing to `GameData`, `Tutorial`, or `Scenarios`.
-
-(**v1.2**) For `GameData` *only* one *may* specify the path to a specific
-subfolder; for example: `GameData/MyMod/Plugins`. The client *must* check this
-path and abort the install if any attempts to traverse up directories are found
-(eg: `GameData/../Example`).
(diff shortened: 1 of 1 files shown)


plague006:
If you change

```specified under `GameData` *only* (**v1.2**)```
to
```specified only under `GameData` (**v1.2**)```
you maintain the same meaning while also having the spec_version number more clearly attached to `GameData`.

I would also suggest maintaining the line break that previously existed as it now appears as a hard-to-parse blob of text. http://puu.sh/peweI/370e4d780b.png

politas:
I really disliked the line break, since this information should be in the context of the "install_to" bullet point. The separate paragraph broke that connection.

politas:
I really wish there was a way to do second-level bullet points.

dbent:
- One
  - Two
    - Three?

politas:
Ah, like that! Much better.

plague006:
Looks good to me, merging with my thanks.

Issue #1673, opened 2016-04-22
Graphical Glitches Under Mono on OSX


mdfntr:
**CKAN Version:**
```bash
$ mono ckan.exe version
v1.16.1-0-g2e91715 (beta)
```

**Operating System:** OS X 10.11.4 with Mono 4.4.0 (not Homebrew)
```bash
$ uname -rsvp
Darwin 15.4.0 Darwin Kernel Version 15.4.0: Fri Feb 26 22:08:05 PST 2016; root:xnu-3248.40.184~3/RELEASE_X86_64 i386
$ mono --version
Mono JIT compiler version 4.4.0 (mono-4.4.0-branch/a3fabf1 Fri Apr  8 13:48:29 EDT 2016)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           normal
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  x86
	Disabled:      none
	Misc:          softdebug 
	LLVM:          yes(3.6.0svn-mono-master/a173357)
	GC:            sgen
```

**The issue you are experiencing:** Graphical glitches, including misplaced scrollbars and broken tabs.
(Side note: Interestingly, these issues aren't present when running under WINE, but I get certificate errors with WINE right now) See image below.

![screen shot 2016-04-21 at 22 44 12](https://cloud.githubusercontent.com/assets/9371833/14733051/c75da7ca-0814-11e6-8775-93e33970f853.png)

**How to recreate this issue:** Run `mono ckan.exe` and observe.

**CKAN error codes (if applicable):** None

EDIT:
Running under WINE after using winetricks to install `dotnet45` alleviates both the graphical glitches and the certificate errors! Still, though, I would like to know what went wrong with just using plain Mono...

politas:
Can you try rolling back to the Stable Mono release? I'm on Linux rather than MacOSX and have this working fine.

    $ mono --version
    Mono JIT compiler version 4.2.3 (Stable 4.2.3.4/832de4b Wed Mar 16 13:19:08 UTC 2016) 
    Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
    	TLS:           __thread
    	SIGSEGV:       altstack
    	Notifications: epoll
    	Architecture:  amd64
    	Disabled:      none
    	Misc:          softdebug 
    	LLVM:          supported, not enabled.
[...]

PR #1772, opened 2016-06-03
Remove unused field

plague006:
From visual studio:

`Warning	1	The field 'CKAN.Main.UpdateCol' is never used	D:\CKAN\GUI\Main.Designer.cs	1455	44	CKAN-GUI`

And indeed a search for any references to it comes up empty.

### DIFF
--- a/GUI/Main.Designer.cs
+++ b/GUI/Main.Designer.cs
@@ -1452,7 +1452,6 @@ namespace CKAN
         private TextBox LogTextBox;
         private ProgressBar DialogProgressBar;
         private TextBox MessageTextBox;
-        private DataGridViewCheckBoxColumn UpdateCol;
         private TabPage ChangesetTabPage;
         private Button CancelChangesButton;
         private Button ConfirmChangesButton;


politas:
Oh, I'm glad someone has decided to look at this build warning message. Looking into it, now, I'm wondering whether someone created this field in order to address the other build warning: 

    Main.Designer.cs(1497,44): warning CS0108: `CKAN.Main.Update' hides inherited member 
    `System.Windows.Forms.Control.Update()'. Use the new keyword if hiding was intended

It looks to me like there may be an inherited action that's unable to be used because we're using the same identifier.

politas:
I can't see any instance where we're attempting to call that `Update()` action. We seem to have replaced it with `UpdateModsList()`, which does little other thatn over-ride the value of the `repo_updated` parameter as it passes through to `private void _UpdateModsList(bool repo_updated)`

politas:
Which leaves me wondering whether MainRepo.cs:98 

    UpdateModsList(repo_updated: true);

should be replaced with 

    _UpdateModsList(repo_updated: true);

in order to actually force a complete update of the mods list after someone clicks the "Refresh" toolbar button.

politas:
Chasing the rabbot down the hole, I find that the `System.Windows.Forms.Control.Update()` method (I've used the term "action" instead of "method" above, incorrectly)	
"Causes the control to redraw the invalidated regions within its client area."

So this could be what is causing the graphical problems on MacOSX (#1673)

plague006:
Closing in favour of a PR @politas will open to more completely address the source of this issue.

Issue #1766, opened 2016-06-01
Add "Updated" filter type


oldlost:
We all know that for installed mods that get updated we can filter to this and even update those mods with a couple clicks. Great.

It would be nice to have a new filter that show a list of mods that are not installed that have been updated since the last CKAN run. This may allow us to see which mods are now compatible with the current version that were not before, or ones that were recently changed we may want to consider installing.

Thanks for listening and thanks for all the work you do on making this program better.

HebaruSan:
#2494 covers "which mods are now compatible with the current version that were not before."

"recently changed we may want to consider installing" would be more involved. Either the client would have to compare every property from the old and new metadata for every mod (slow!), or we would need to track a last updated timestamp a la #1155.

Issue #1519, opened 2015-11-15
Sort by 'Last Updated'


legacynl:
So with the new update I keep finding myself browsing the KSP forums to check if the mods I'm interested are updated to 1.0.5. This takes a lot of time, and it is easy for certain mods to be buried to the Xth page because other mods have more users.

So I thought it would be neat if the CKAN modlist could be sorted on when a mod was updated. So for example sorting on "Last updated" will give you a list that has the mods that were most recently updated on top. So if this were the case, you would only need to keep an eye on the CKAN-list, and install the mods that you like that pop up at the top.

CharonM72:
I want to second this. This would be super useful, even to see which mods were updated recently and which were not, but especially for new version upgrades.

HebaruSan:
This is a duplicate of #1155.

Issue #1155, opened 2015-06-22
column for date mod updated


anxcon:
alot of times ckan gets a new mod and i don't notice it, be it complete new mod, or recently updated to the current ksp version

if there was a column for date last updated (as in version of mod added or changed, not ckan scan updating full db) i could sort by date :) see whats new

janbrohl:
Currently there are no dates in the underlying ckan-metadata at all (also no helpful ones in the master.zip) so this propably depends on #1116 or on some git-magic (meta-metadata?)

anxcon:
even if the date database is started at current date would be fine, still helps to to sort newest since then :)

pjf:
Lots of people like this, so I'm sticking a bunch of higher priority labels on it.

politas:
When adding this, time zones will need to be considered. It is probably easiest to use UTC Date/time in the metadata, and then look at converting to local time zone in the client.

funkcanna:
This would be very useful.

PyroSA:
This would be awesome addition, I've limited myself to checking the new list after every refresh, as that was the only feasible way of seeing what's fresh. Suspected the lack of backing data would be exist, but even a local date would suffice for my needs.

techman83:
I've given this a lot of thought as we have a much more flexible architecture since #2789 

These are my dot point thoughts:
- Firstly this needs to be discussed and added to the Spec, ie
  - Field assumed to be UTC
  - Full date / time
  - ISO 8601 compliant (2019-11-07T01:29:59+00:00)
- netkan.exe should remain the source of the truth. It should be responsible for populating the `release_date` field (name pending)
  - This would enable us to initially use the system timestamp, but also enhance our transformers to retrieve this information from the relevant APIs where available
- An optional `--release-date` flag on netkan.exe means we can switch this on when we're ready
- Metadata pull request testing on CKAN-meta to fail if `release_date` is changed or removed.
- The indexer can be switched from using an MD5 to an equivalence check via the `Ckan` class. For example, if an existing ckan is found, load it in as a `Ckan` and compare it to the received inflation. The `Ckan` class to have a custom `__eq__` method that compares the objects, but ignores the `release_date`
  - This keeps netkan.exe as the source of the truth
  - Allows us to write to disk exactly as it's received from netkan.exe
  - Keeps our existing logic, we just compare `inflated == existing` instead of the MD5s
  - We can do this in advance to enabling the `release_date` flag
[...]

Issue #1116, opened 2015-06-15
Add "modifed"-key to ckan spec


janbrohl:
As the times in the master.zip are useless (they point to when the file was downloaded) and knowing when a ckan-entry was added or modified is useful eg for finding outdated/old entries, for checking if the last-modified header of the mod-zip fits the entry so mods much newer than their ckan-entry can be identified (like those using the same URL for all versions) without the need for a complete download.
This way a Timeline for changes could also be generated.

I would suggest https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations like "2015-06-15T12:26" with a precision of day, hour or minute precision.

plague006:
I'm not sure how this would be implemented or if it has merit. Pinging @dbent, @Postremus, @pjf since implementing this would probably need netkan, core, and GUI elements.

ayan4m1:
Not sure if I am understanding, but HTTP headers, especially when we aren't in control of the server issuing them, aren't going to be a reliable way of versioning anything, unfortunately.

Modders are free to add an `x_last_modified` property to their netkans. I would prefer to avoid any code or schema changes in response to this request.

politas:
@techman83 has made the case that this is a serious challenge, given the way the netkan robot currently works. I think we should keep the discussion going in #1773. In fact, I'm going to close this as a duplicate.

PR #1773, opened 2016-06-03
Add LastModified to netkan, registry, and GUI

plague006:
**DO NOT MERGE** Until @techman83 gives the all-clear as merging this as-is would likely cause some serious indexer-related headaches.

This introduces a new field into the metadata: `last_modified`. `last_modified` represents the most recent inflation of a .ckan by netkan.exe whether it be initial creation for a new release or updating information on a current release. The time is in UTC to avoid timezone related issues but is presented via the GUI in a user's given local time.

I'd love to have a test for the netkan transform, but don't know how to write one. @dbent input?

@techman83 as we discussed on IRC this will probably require the indexer to "ignore" the last_modified property otherwise we would likely end up inflating each .netkan.

Closes https://github.com/KSP-CKAN/CKAN/issues/1116, closes https://github.com/KSP-CKAN/CKAN/issues/1766, closes https://github.com/KSP-CKAN/CKAN/issues/1519, closes https://github.com/KSP-CKAN/CKAN/issues/1155, closes https://github.com/KSP-CKAN/CKAN/issues/1046

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file.
 - [CLI] `compare` command now checks positive and negative rather than -1/+1 (#1649 by: dbent; reviewed: Daz)
 - [GUI] In windows launch KSP_x64.exe by default rather than KSP.exe. (#1711 by plague006; reviewed: dbent)
 - [Core] Unlicense added to CKAN as an option for mods. (#1737 by plague006; reviewed: techman83)
+- [GUI/NetKAN/Core] Added "Last modified" field and column which allows users to see when a mod was last updated (or had its metadata updated). (#1773 by plague006; reviewed: )
 
 ### Internal
 
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -82,6 +82,10 @@
             "description" : "The content type of the download",
             "type"        : "string"
         },
+        "last_modified" : {
+            "description" : "The UTC time at which metadata was last modified",
+            "type"        : "string"
+        },
         "license" : {
             "description" : "Machine readable license, or array of licenses",
             "$ref"        : "#/definitions/licenses"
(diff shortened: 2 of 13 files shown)


techman83:
@plague006 This looks nice, neat and I really like it. However the indexer is deliberately dumb, it checks for changed files and commits them. This would cause every NetKAN to generate a change on every index. We don't know in advance if the file generated is a new file (as in a new version) or a modified one (metadata update), so doing a comparison against the original file is tricky. We could parse the diff output, but that could break unexpectedly in interesting ways or output the generated ckans to a staging area and do comparison, but that increases the complexity quite a bit.

That isn't to say it's impossible, It requires some pondering. NetKAN is beautiful in its simplicity, but that does come with some gotchas like this.

dbent:
I agree, this is worth doing, but for it to work either the indexer or NetKAN itself will have to get smarter, and like @techman83 I'm not sure what the best solution is.

If the goal is just to let end users know when metadata has changed it may be worth exploring doing this strictly on the client side. Sure it won't be the exact last modified time but last modified *since the client saw it*, but I think that's close enough for government work.

For reference, CKAN uses [NUnit 2.6.x](http://www.nunit.org/index.php?p=docHome&r=2.6.4) as its unit testing framework. The documentation is pretty comprehensive. I structure my tests using the [Arrange-Act-Assert (AAA)](https://msdn.microsoft.com/en-us/library/hh694602.aspx#Anchor_3) pattern. [`GeneratedByTransformerTests.cs`](https://github.com/KSP-CKAN/CKAN/blob/master/Tests/NetKAN/Transformers/GeneratedByTransformerTests.cs) is probably about as simplest as the tests go.

For a test like this you'd probably want to run the transformer and assert that the `last_modified` value is within some acceptable delta of the current time, like one second (the precision of the system clock is going to be higher than the time between when the value is written and when it's read).

It's not really necessary in this case, but generally when testing code that operates on time, you want an interface like `IClock` that can be mocked out in tests to provide specific time values, so that the result of your tests isn't dependent on when they're run, which can be hugely frustrating.

PR #1774, opened 2016-06-04
Update Spec.md

dbent:
@politas


plague006:
Looks good to me, merging with my thanks.

politas:
I would have suggested an example for this, because I don't think it's obvious.

PR #1776, opened 2016-06-05
Column rename "Update" to "UpdateCol" in GUI

politas:
Clears both build warnings that have been happening for some time.

    Warning 1 The field 'CKAN.Main.UpdateCol' is never used D:\CKAN\GUI\Main.Designer.cs 1455 44 CKAN-GUI
    Main.Designer.cs(1497,44): warning CS0108: `CKAN.Main.Update' hides inherited member 
    `System.Windows.Forms.Control.Update()'. Use the new keyword if hiding was intended

Replaces #1772 - See there for details of thought processes
Possible solution for #1673 - `System.Windows.Forms.Control.Update()`'s description is "Causes the control to redraw the invalidated regions within its client area."

This orphaned declaration of UpdateCol may have been part of the fix that was implemented last time the MacOSX redraw glitch came up.

### DIFF
--- a/GUI/Main.Designer.cs
+++ b/GUI/Main.Designer.cs
@@ -62,7 +62,7 @@ namespace CKAN
             this.splitContainer1 = new System.Windows.Forms.SplitContainer();
             this.ModList = new CKAN.MainModListGUI();
             this.Installed = new System.Windows.Forms.DataGridViewCheckBoxColumn();
-            this.Update = new System.Windows.Forms.DataGridViewCheckBoxColumn();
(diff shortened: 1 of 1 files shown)


pjf:
`UpdateCol` dates back to 4a73d89c, which sure enough was renaming all instances of `Update` to `UpdateCol`. I have no idea where we regressed back to `Update`.

Testing now.

pjf:
Seems to be working for me; I've upgraded a bunch of mods and all looks good.

Also, the build warnings are now gone! :D :D 

Merging now

PR #1781, opened 2016-06-19
Cleanse broken KSP version numbers

### LINKED ISSUE #1780: Latest master dies if registry file with correupted KSP versions is loaded.
As of 968848d20640f258575730f9b9edaf30cda75534 (#1645)
if a registry is loaded with
available modules with corrupted KSP strings in the form of `1.1.` then
the client crashes on start-up.

This happens to be the case in my registry file.

Patch incoming with tests now. :)

pjf:
This catches and fixes corrupted version numbers in the form of `1.1.`
that can still persist in registry files (like mine).

Closes #1780, makes #1645 safer.

### DIFF
--- a/Core/Versioning/KspVersion.cs
+++ b/Core/Versioning/KspVersion.cs
@@ -695,6 +695,14 @@ namespace CKAN.Versioning
                     return KspVersion.Any;
                 default:
                     KspVersion result;
+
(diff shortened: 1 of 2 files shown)


politas:
Tested. Prevents the crash

pjf:
Thanks @politas ! Merging! :)

PR #1782, opened 2016-06-19
Fix travis failing in third-party failures

pjf:
These changes:
- Say it's okay for mozroots to fail.
- Avoid hitting github for something they might decline.

The test for these is that travis tests are now passing again, which I need them to do before I can make a release. :)


techman83:
Awesome! Passing tests are the best!

PR #1784, opened 2016-06-19
Workaround string.Format() bug in old Mono versions

dbent:
@pjf


Postremus:
Tested on mono 3.2.8, ckan starts now.
Thanks.

Issue #1793, opened 2016-06-24
Opt In / Opt Out | Required changes to CKAN.


VasVadum:
CKAN has officially become a threat to the community because mod author are beginning to change their mods over to All Rights Reserved in order to stop CKAN from listing.  This will hurt modding in the long run, and cause potentially permanent damage if this ARR trend kicks on just to get back at CKAN for not allowing choice.

You need to change your de-listing policy immediately because I will not continue to support something that is going to ignore this immediate threat.  It has been stated that when CKAN becomes opt in, most of this issue will go away and the people who are angry and wanted to de-list, are likely to no longer want to de-list.

So here is MY personal suggestion which is slightly getting some negative opinions against but I still think it is good for de-listed mods to help us manage everything in one place;
If an author de-lists, you can grey their mod out in the program and still show the home page for users to click and go get the mod manually, also showing a version number on it and all that jazz.  Instead of checkboxes or dashs where you would click to auto install/update a mod, place an X for a delisted mod.  When a mod has an update, turn the "update" column X into a clickable button which brings you to the home page of the mod so that you can find the download URL.  Make sure to make a note however for de-listed mods when a user clicks a home page button in one for the first time or an update button on one for the first time that it pops up an un-closable message that lasts for a minimum of 5 seconds;
`The author has requested that his or her mod not be downloadable via CKAN, so you must do this via his or her page.  Please do not pester the mod author to change their listing status, it was their choice to make and they made it for whatever their reason was and they are even less likely to change their mind if people spam them with requests to list on CKAN.`

This way I don't need to manage mods via browser bookmark folder, and can still find a mod via CKAN even if I can't install it via CKAN.  I think an even better option for this though, is three versions of opt.
**Opt-In:** Your mod is listed and downloadable via CKAN.
**Opt-List:** Your mod is listed in CKAN but not downloadable.  It can still notify of updates too.
**Opt-Out:** Your mod is not listed in CKAN unless it is detected in your game install, and then it is marked red in CKAN with only the information found in the meta-data that the author has provided with his mod.  _(Or don't list at all, I suggest listing if installed because this would allow me to export a list of mods I use easily if I want to share that list with someone, particularly for diagnosing an issue.)_

[...]

Issue #1679, opened 2016-04-23
Vetting metadata through a staging process


Dazpoet:
## Problems ##

When a new version of KSP comes out users go crazy for updated mods, modders are often forced to implement breaking changes because Squad changed something and metadata fails to catch up fast enough. This leads to installation errors which leads to angry users which leads to sad modders who get yelled at.

Our metadata isn't perfect.

## Proposal ##

Implement a 2-stage process for our metadata as a way to ensure better stability and less issues with installations of mods through CKAN.

## Goals ##

- Give our users the best possible metadata
- Give the metadata maintainers breathingroom when new KSP-releases come out
- Become slightly more like Debian
- Give users the alternative of bleeding-edge metadata

## How could this be done ##

Short answer: I have no idea, I don't code.

Long answer: We split CKAN-meta into two distinct branches `stable/master` and `unstable/dev`. Everything we get from our .netkans, webhooks and even manual additions to `CKAN-meta` go into the `unstable/dev` branch for testing. After a yet to be determined vetting process the stable metadata is then merged over into the `stable` branch.

Users who like to live on the edge can help us betatest the unstable branch by adding it as a repository but upon doing so will somehow be prompted with a sign explaining how to report issues on github.

## How could the vetting process look like ##

In a perfect world I'd like to have a bot which [automagically](https://en.wikipedia.org/wiki/Small_matter_of_programming) creates and installs metapackages with the entire non-conflicting modset each time something is added. This would allow us to catch file collisions. A less extreme measure might be to have it atleast run some sort of smart installations trying to install all mods by the same author and/or trying to install mods which filter things that match to some percent (obviously excluding `*.cs`, `_MACOSX` and `Thumbs.db`).

The bot could also run some [random modsets](https://github.com/KSP-CKAN/CKAN/issues/883) every now and then just to see whats up. A more [logical](https://media.licdn.com/mpr/mpr/p/6/005/087/26b/18801d9.jpg) method could be to point the current [statuschecker](http://status.ksp-ckan.org/) to it, which would however not catch file-collisions. I also know that @godarklight and @politas is using some fancy tool to catch things going wrong.

[...]

PR #1795, opened 2016-06-25
Allow foss delisting

politas:
A proposed compromise solution to #1793


Blu3wolf:
Thats no compromise, its negotiating with terrorists.

ThomasKerman:
Thank you for your utterly useful comment. Have a nice day.

passinglurker:
@Blu3wolf careful that's borderline godwin's law talk there.

Any way in the future a system like googles robot.txt included in mod zip files could be used to easily identify a maintainer or group of maintainers intentions to opt out or not.

politas:
Rationale:

CKAN's policy was all about accessibility, and the rights of users. It was based on the assumption that all CKAN was doing was facilitating the installation of mods. Where it went wrong is in not taking into account the fact that in the minds of many CKAN users, CKAN also implicitly defines a support relationship, by linking to the "homepage" forum thread. We have attempted to disabuse users of that idea, but those efforts have been unsuccessful, and are likely to remain so, because people are people.

CKAN should not be in the position of effectively forcing a support relationship on people who are not willing to be part of one.

politas:
@passinglurker You mean an opt-in opt-out method that doesn't involve a request to be added to the Opt-out list page I am creating with this PR?

The only issue I see with that is that embedding it in the the permanent download zip removes any ability to negotiate a change in stance.

passinglurker:
@politas the advantage is the system can be consistently and reliably managed by a bot without haveing to worry about IDing the primary maintainer as long as the bot is pointed at an official download and every time there is an update the bot checks to see if the robot.txt equivalent has changed. 

and either way I don't see how that stops ckan from talking to the maintainers about addressing thier concerns and getting them to change thier minds

also if you are worried about edge cases where the maintainer accepts indexing but doesn't have time to reupload the mod with a different robot.txt equivalent then you can just offer to add them to an exceptions list so the bot can proceed I suppose.

politas:
@passinglurker Ok, I see that could be a useful feature. At the moment, I am focused on what can be done immediately in processes and policy, without requiring time-consuming development work.

passinglurker:
@politas yes I agree what can be done immediately should have priority I did say this was an idea for the future after all

CliftonMarien:
I agree with this idea.

passinglurker:
**I agree with this idea** _in case it wasn't obvious already_

Cphoenix:
I've edited the [adding a mod guide](https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN) to direct people to the opt-out-list.

blowfishpro:
This seems very reasonable to me

dewiniaid:
+1.  Though this it saddens me as a CKAN user to see this going on, I agree it's a necessary measure until a proper fix can be made on both the policy and technical levels.

anxcon:
reasonable

VaPaL:
Agree

BobPalmer:
No surprise either, but I support this as well.  It ensures both parties have appropriate incentives to improve the process.

pjf:
I'm hoping I can nit-pick the wording but (hopefully) not the intent here. Currently the change reads:

> It has become clear that in some cases, it is not possible to preserve mods with permissive licenses and have them remain under permissive licenses. Therefore, CKAN will not attempt to install permissively-licensed mods when the primary maintainer (or a significant group of maintainers if no single primary maintaner is easily identifiable) requests that the mod be taken off CKAN.

This is a little verbose, and also very specific. Can I suggest:

> In some cases it's not possible to install mods via the CKAN and maintain the goodwill of the author(s). The CKAN will not install any mod when the primary maintainer (or a significant group of maintainers when no primary is identifiable) requests the mod be removed from the CKAN.

This provides a consistent policy regardless of license, and provides the ability to request mod removal over a range of issues (eg: broken install, change of maintainer).

BobPalmer:
Thanks pjf.

I think this policy, coupled with a staging repo (where either the authors, CKAN staff, or volunteers can test things out and sort out potential issues) will go a long way to making this a much better tool for all concerned (consumers, publishers, and creators).

Blu3wolf:
concerned creators, perhaps. Consumers, not so much.

BobPalmer:
I expect consumers will be better served by having access to mods again, and with folks working to improve the process than having mods pulled.  Again, thanks everyone for getting us to this point.  @pjf and team - where would be the best place to consolidate discussions about ongoing improvements (the staging area, maintainers, etc.)?

Blu3wolf:
You realise you are talking about having access to mods again after ensuring there need not be any?

pjf:
I'm hoping we can avoid re-discussing existing points, especially after there's already been such extensive conversations in #1793 and elsewhere.

@BobPalmer : With regards to staging areas, #1679 is an existing idea that's had some discussion. It'd be great to review and critique that, I haven't had a full chance to do so myself.

With regards to other discussions, I personally prefer a one-ticket-per-topic, because it means we can close out things when they're done, track changes against issues more effectively, and group them into milestones that are easier to manage. Having said that it's totally appropriate to have a discussion on improvements where new tickets are created as-needed when each idea gains speed.

BobPalmer:
Sounds good.  I'll take a peek.

mgsdk:
Agree

linuxgurugamer:
agree

TeddyDD:
Agree

Cphoenix:
One suggestion for Opt-out-list.md : A "Reason" column could be useful so whomever is reading it would be able to tell if the situation has changed, and also make the distinction for cases like @BobPalmer 's where the opt-out isn't _completely_ opt-out, but rather a "I want to manage my own metadata, but it's okay that my mods are listed".

anxcon:
agreed on the reason column, its information, information isn't bad, and helps new volunteers know if/when a reason is no longer valid

TeddyDD:
"De-list" operation should be specified. Do we remove netkans? Or do we change just change extension to .froze so in case mod author would like to have his mods back on CKAN we have something to start with? Example: https://github.com/KSP-CKAN/NetKAN/pull/4294
What we do with .ckan files? Remove them or just hide by changing extension?

anxcon:
@TeddyDD to clarify, would renaming the extension leave the future open to a "guide" popping up in forums equivalent to "run ckan.exe then type 1234 to see every mod plus the hidden ones" ?

BobPalmer:
It's probably best to remove vs. just freeze tbh.

BobPalmer:
Actually...  maybe offer a choice?  Freeze because they want it down while they sort stuff out, or remove completely because (for whatever reason) they do not wish to come back.

TeddyDD:
@anxcon If you remove netkan you can still get it back using power of git so there is no difference at this point. Freezing netkans:
- makes easier to add mod back to CKAN
- makes easier to advanced ckan users to manually inflate netkan using netkan.exe and keep using mod from ckan. As I said it does not matter if you remove file or rename it. It is possible to get it back anyway. Renaming is just less annoying option.

j-bos:
Also makes it clear that the mod is frozen and not just missing so much less likely to be readded by mistake . . .

Cphoenix:
> Actually... maybe offer a choice? Freeze because they want it down while they sort stuff out, or remove completely because (for whatever reason) they do not wish to come back.

Freezing with the contents in place sounds like a good idea for the former case. When it comes to the latter, @politas has been removing the contents (just leaving the ID it used in the file as the only text, so it's not valid JSON) and then renaming it to .frozen. It removes the contents to there's no chance of something really weird happening and unintentionally having it be relisted (because the JSON deserialization attempt will fail), but it leaves a stub that's easy for anyone primarily using the GitHub browser UI to find the original data should the mod need to be relisted for any reason (author comes back, mod switches hands to a new author who wants to list it, etc.) so there's a starting point at least.

As @TeddyDD said, you can get it back via git anyway. I think the solution @politas has been using over the past few days is the best of both worlds.

TeddyDD:
@Cphoenix could you explain me how to do this? Never heard of that method. I still prefer not removing file content ( I suppose bot ignore everything that has no .netkan extension right?)

Cphoenix:
> @Cphoenix could you explain me how to do this? Never heard of that method. I still prefer to not removing file content ( I suppose bot ignore everything that has no .netkan extension right?)

https://github.com/KSP-CKAN/NetKAN/pull/4271/commits/db97a0423a019c8eb63d115d869f21a2a8e2c96d

Sorry I got that a little backward in my brain (this is what I get for not getting enough sleep :stuck_out_tongue_closed_eyes:  ). He's mostly just been renaming to .frozen , but I actually think this solution might be better. As far as I'm aware, yes, the bot ignores .frozen entirely. 

It's not really removing the data per-se. The old data is easy enough to get back by just reverting the commit that removed it. Just leaving the identifier in there though rather than valid JSON adds another layer of security. If by some fluke a bug is introduced in the bot and it tries to grab a .frozen file, it won't be able to use it, instead of the possibility of it silently failing.

I know I'm being nitpicky; it's a minor concern right now but just a thought.

jrossignol:
You'll want to change my forum name on that list to nightingale. :wink:

politas:
Thanks for that @jrossignol , I was going to chase that up .

politas:
Here's what's been done so far:

CliftonM's .ckan files have been removed (though I kept backups), mainly because I merged his PR that did that, while all other files have been renamed to frozen (including the .netkan files for CliftonM's mods), which acts as a placeholder to signify that the names (and version numbers) have been used, while being ignored by the auto-indexing robot and CKAN client. My understanding is that the requests for removal have been more in protest than a genuine desire for permanent removal, so I took things out of the CKAN client' s awareness without wiping out the useful information they contain.

It is all temporary at this stage. Mods permanently delisted will have their .ckan files removed entirely and their .netkan files stripped of all content so they become name placeholders preventing duplication/confusion.

The intention of this change is that in the long run mods would stay in the listing, to allow autodetection  and downloading of the zip file, but no automated installation would be done, forcing users to open the zip file and manually install the contents. Until the code to handle that case is in place, de-listing is the only way to get the desired effect of not installing mods.

For Ferram's specific issues with FAR (and any other mods with similar concerns such as SVE), my proposed method would be changing the .netkan file to a specific download, disabling auto-indexing, and requiring a PR to be processed for each update, involving approval by the mod author or a designated maintainer.

politas:
The opt-out list may need to have a "delegated approvers" column added at some stage.

pjf:
Merging with thanks, although #1800 will be relevant to many on this thread.

My apologies that merging also automatically closes the ticket. You can continue to discuss policy and process on a closed pull-request, but do consider opening new issues as needed for those that need enduring visibility.

PR #1804, opened 2016-06-28
Remove myself from opt-out

CliftonMarien:
Seeing that this has settled down, I am ready to re-list my mods in CKAN, therefore there is no need to be on this list anymore.


politas:
Thanks, CliftonM

PR #1809, opened 2016-06-29
Ignore ksp instances if we are launching the gui anyway

### LINKED ISSUE #1785: Running `ckan.exe gui` fails if no default ksp install is selected
This may be specific to Linux, or it may not. My ckan.exe is not located in a KSP install

    $ mono ./ckan.exe list
    I don't know where KSP is installed.
    Use 'ckan ksp help' for assistance on setting this.
    $ mono ./ckan.exe gui
    I don't know where KSP is installed.
    Use 'ckan ksp help' for assistance on setting this.

But if I run

    $ mono ./ckan.exe

It will go straight to the "Select KSP install" dialogue.

If I have a ksp install default set, `ckan.exe gui` works

Postremus:
fixes #1785

### DIFF
--- a/Cmdline/Main.cs
+++ b/Cmdline/Main.cs
@@ -141,7 +141,7 @@ This is a bad idea and there is absolutely no good reason to do it. Please run C
                 // Set a KSP directory by its path
                 manager.SetCurrentInstanceByPath(options.KSPdir);
             }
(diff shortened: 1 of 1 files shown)


politas:
Well, that looks like a nice, simple fix! Can you update the comment just below to match the change?

PR #1357, opened 2015-08-09
LockFiles with adjustments

pjf:
This is #1265, but with the addition of tests and extra garbage collection. Specifically we need reviewed:

* a43ec69 (garbage collection)
* b60da6c (tests)

PR #1828, opened 2016-07-13
Add Support for locking the registry

### LINKED ISSUE #706: The CKAN registry should be locked while a client is operating.

politas:
Closes #706 
Working from previous PRs #1357, #1265


Postremus:
-launch gui
-try to start cli without parameters
The cli shows the `CKAN is already running for this instance!` error, but then opens a messagebox with the same text again.

politas:
What happens when you run CLI with no params without having the GUI open?

Postremus:
yes, but showing the error to the user twice (once in cli, once in gui) is not really needed.

politas:
There is a weird case were we are failing to create a lock file issue. When you start a second GUI instance selecting a different KSP install and then switch to the same install as the first instance. The second instance is killed, and all is well unless you then open a different KSP install with the first instance. Under that situation, it doesn't open a new lock file, and if you keep switching KSP installs it keeps not creating lock files. I can't find any other cases where lock files are not handled correctly, and this is going to be a pretty rare situation. But if anyone would like to try as many situations as they can imagine, any further issues might help to pin down the flaw.

politas:
Further investigation shows the second KSP instance is not the culprit. Returning to a KSP Install in which you have already had a lock file created and removed will not create a new lock file.

politas:
Ok, the only remaining issue I can find is the duplicated error messaging, and I'm happy to live with that.

politas:
needs more testing

politas:
Rebased after #1841

Core/KSPManager.cs
```diff
@@ -233,6 +233,13 @@ public void SetCurrentInstance(string name)
                 throw new InvalidKSPInstanceKraken(name);
             }
 
+            //Don't try to Dispose a null CurrentInstance.
+            if (CurrentInstance != null)
+            {
+                // Dispose of the old registry manager, to release the registry.
```
ayan4m1: Possible null dereference

Core/Registry/RegistryManager.cs
```diff
@@ -10,13 +11,17 @@
 
 namespace CKAN
 {
-    public class RegistryManager
+    public class RegistryManager : IDisposable
     {
         private static readonly Dictionary<string, RegistryManager> singleton =
             new Dictionary<string, RegistryManager>();
```
ayan4m1: First off, the formal Singleton pattern is about a class that has a public `getInstance` method which always returns a single instantiation of the pattern it wraps, so nothing about that name is helpful. If I read `static readonly` I already know that it is "initialized once and not scoped to an object reference" so you don't need to name the variable in a way describing it, but instead the purpose it serves. This is just personal opinion but I'd call this `registryCache`.

Core/Registry/RegistryManager.cs
```diff
@@ -10,13 +11,17 @@
...
-
+        
         private static readonly ILog log = LogManager.GetLogger(typeof (RegistryManager));
         private readonly string path;
+        public readonly string lockfile_path;
+        private FileStream lockfile_stream = null;
```
ayan4m1: e.g. lockFileStream

ayan4m1:
Per https://msdn.microsoft.com/en-us/library/ms229043.aspx and https://msdn.microsoft.com/en-us/library/ms229045(v=vs.110).aspx

> DO use camelCasing for parameter names.
> DO NOT use underscores, hyphens, or any other nonalphanumeric characters.

Some of the variables and parameters are in snake_case when you want them in camelCase.

I also added a few other style notes - not trying to be critical of your work at all in case it came off that way.

Cmdline/Main.cs
```diff
@@ -174,6 +176,43 @@ public static int Main(string[] args)
...
+                return Exit.ERROR;
+            }
+            finally
+            {
+                log.Info("Freeing resources");
+
+                // Release the registry lock file if possible.
+                if (manager.CurrentInstance != null)
+                {
+                    manager.CurrentInstance.RegistryManager.Dispose();
```
ayan4m1: If manager.CurrentInstance.RegistryManager is null, this will throw a NullReferenceException. Take a look [here](https://msdn.microsoft.com/en-us/library/ty67wk28.aspx) for an elegant solution to these kinds of issues.

Core/Registry/RegistryManager.cs
```diff
@@ -49,20 +62,122 @@ private RegistryManager(string path, KSP ksp)
...
+
+        /// <summary>
+        /// Tries to lock the registry by creating a lock file.
+        /// </summary>
+        /// <returns><c>true</c>, if lock was gotten, <c>false</c> otherwise.</returns>
+        public bool GetLock()
+        {
+            try
+            {
+                lockfileStream = new FileStream(lockfilePath, FileMode.CreateNew, FileAccess.Write, FileShare.None, 512, FileOptions.DeleteOnClose);
```
ayan4m1: The "[using statement](https://msdn.microsoft.com/en-us/library/yh598w02.aspx)" is a bit safer here and obviates the need for a .Flush() call.

politas: I admit to not knowing how this bit of code works, and I can't figure out how to include both lockFileStream and lockfileWriter in the `using` statement, nor what to do with the `catch` part. That page explains how `using` can handle the `try` and `finally` parts, but has no mention of catching errors. This particular `try` doesn't even have a `finally`. As local variables to the method, shouldn't lockfileStream and lockfileWriter get automatically disposed on return?

politas:
Oh of course. Mono 3.2.8 doesn't support Null Conditional Operators. Rolled that back to the ugly method, now improved to catch all possible nulls

Core/KSP.cs
```diff
@@ -118,6 +122,12 @@ public void Dispose()
                 Cache.Dispose();
         }
 
+
+        ///public void Dispose()
+        ///{
+        ///    this.RegistryManager.Dispose();
```
ayan4m1: If this remains commented out, an additional comment explaining why it was commented out or what it did would be helpful.

politas:
Ah yes, that little bug! Whacked again.

PR #1837, opened 2016-07-24
GUI fixes and solution release/debug fix

Telanor:
Just a small change to make the UI look better for windows 7 users.  Also some of the projects in the solution only had debug builds, so I added release configurations for them

### DIFF
--- a/GUI/Main.Designer.cs
+++ b/GUI/Main.Designer.cs
@@ -980,12 +980,12 @@ namespace CKAN
             // KSPVersionLabel
             // 
             this.KSPVersionLabel.Anchor = ((System.Windows.Forms.AnchorStyles)((System.Windows.Forms.AnchorStyles.Top | System.Windows.Forms.AnchorStyles.Right)));
-            this.KSPVersionLabel.AutoSize = true;
-            this.KSPVersionLabel.Location = new System.Drawing.Point(862, 10);
+            this.KSPVersionLabel.Location = new System.Drawing.Point(788, 10);
             this.KSPVersionLabel.Name = "KSPVersionLabel";
-            this.KSPVersionLabel.Size = new System.Drawing.Size(146, 13);
+            this.KSPVersionLabel.Size = new System.Drawing.Size(227, 13);
             this.KSPVersionLabel.TabIndex = 8;
             this.KSPVersionLabel.Text = "Kerbal Space Program 0.25.0";
+            this.KSPVersionLabel.TextAlign = System.Drawing.ContentAlignment.TopRight;
             // 
             // FilterByNameLabel
             // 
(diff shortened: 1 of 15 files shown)


politas:
I'm guessing the csproj files are used by Visual Studio for building? Our primary build method is make on linux, which would explain the lack, I suppose.

GUI/Main.Designer.cs
```diff
@@ -981,12 +980,12 @@ private void InitializeComponent()
             // KSPVersionLabel
             // 
             this.KSPVersionLabel.Anchor = ((System.Windows.Forms.AnchorStyles)((System.Windows.Forms.AnchorStyles.Top | System.Windows.Forms.AnchorStyles.Right)));
-            this.KSPVersionLabel.AutoSize = true;
-            this.KSPVersionLabel.Location = new System.Drawing.Point(862, 10);
+            this.KSPVersionLabel.Location = new System.Drawing.Point(788, 10);
             this.KSPVersionLabel.Name = "KSPVersionLabel";
-            this.KSPVersionLabel.Size = new System.Drawing.Size(146, 13);
+            this.KSPVersionLabel.Size = new System.Drawing.Size(227, 13);
```
politas: This has been bugging me, too! Thanks for tidying that up!

politas:
I can see no visual difference on my Linux system, so this doesn't appear to be harmful to non-Win10 users. Can you post an image showing the difference it makes for you?

Telanor:
Before: 
before

After:
after

Mostly just affected the buttons on various forms

politas:
Ah, I see. I wonder if there's a way to get Mono to use a different theme on Linux? The flat theme looks much better there, actually.

Telanor:
Well the easy way would be to replace all of the FlatStyle.Flat constants with a function that returns a value based on OS, but the VS designer will just replace them all the second you open up a form.

What does it look like on linux if you go into the Program.cs and disable `Application.EnableVisualStyles();`?

politas:
That appears to again use the system default style.

politas:
Now that I notice the differences, I am left wondering why the flat design was chosen rather than the system design. It's clearly a decision that was made, or all the fields would not be labelled as `*.FlatStyle = System.Windows.Forms.FlatStyle.Flat` - it gives us a consistent UI across platforms, and a modern one. Given the cross-platform nature of CKAN, I'm very hesitant to make sweeping UI changes that may significantly affect users on minority platforms, especially given the regular reports of GUI problems on Macs we're already getting. Could I suggest you consider helping out with #1840, instead?

politas:
Have decided to stick with the flat design and not make things confusing. We will possibly be completely overhauling the GUI in the near future following #1840 Cherry-picked the KSPVersionLabel fix with thanks.

Telanor:
That's fine, I just figured I'd do a pull request since I'd made the
changes for myself anyway

politas:
And  we really do appreciate it, even when we don't end up using it!

PR #1841, opened 2016-07-26
add backward+forward navigation buttons to the UI

b-w:
This PR contains a new feature in the form of browser-like backward and forward navigational buttons that are added to the UI.

ckan-navbuttons

Every time you select a mod, that mod gets added to the "browser history". You can navigate backward and forward through the history using the nav buttons. History is not persisted anywhere, so it only lasts until you close the UI.

Basically it works like you would expect from a web browser, only instead of websites you are browsing KSP mods.

**Motivation for change**

The CKAN UI has a nice feature where you can double-click any mod in the Relationships tree to navigate to that mod. However, there's no way of quickly getting back to the mod you just came from.

I was recently checking out the SETI-Rebalance mod, and viewing all the mods it recommended. There's quite a few, and I found it annoying that I had to re-find SETI-Rebalance in the master list every single time after jumping to a recommended mod. I figured a simple "back" button would make this process a lot nicer, so I added it.

**Impact**

This change affects the UI only, and does not affect any existing functions.

**Testing**

I tested on Windows 10, using .NET framework version 4.6.01038 and Visual Studio 2015 version 14.0.25421.03 Update 3.


politas:
Thanks for your submission! I really like all the tests you put in!

b-w:
Glad you liked my changes! I've used CKAN a lot, so I'm happy to be able to contribute something back.

PR #1848, opened 2016-08-01
Create ~/.local/share/applications/ if it doesn't exist (Linux)

DinCahill:
Discovered this on Ubuntu Server 16.04, where I didn't have this directory.

`Directory.CreateDirectory` creates directories recursively, so it should be fine.

### DIFF
--- a/GUI/URLHandlers.cs
+++ b/GUI/URLHandlers.cs
@@ -126,6 +126,7 @@ Do you want to allow CKAN to do this? If you click no you won't see this message
             if (!File.Exists(MimeAppsListPath))
             {
                 log.InfoFormat("{0} does not exist, trying to create it", MimeAppsListPath);
+                Directory.CreateDirectory(ApplicationsPath);
                 File.WriteAllLines(MimeAppsListPath, new string[] { "[Default Applications]" });
             }
 
--- a/GUI/URLHandlers.cs
+++ b/GUI/URLHandlers.cs
@@ -13,7 +13,7 @@ namespace CKAN
         private static readonly ILog log = LogManager.GetLogger(typeof(URLHandlers));
         public const string UrlRegistrationArgument = "registerUrl";
 
-        private static string MimeAppsListPath = ".local/share/applications/mimeapps.list";
+        private static string MimeAppsListPath = "mimeapps.list";
         private static string ApplicationsPath = ".local/share/applications/";
         private const string HandlerFileName = "ckan-handler.desktop";
 
@@ -21,9 +21,18 @@ namespace CKAN
         {
             if (Platform.IsUnix)
             {
-                string home = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
-                MimeAppsListPath = Path.Combine(home, MimeAppsListPath);
-                ApplicationsPath = Path.Combine(home, ApplicationsPath);
+                string XDGDataHome = Environment.GetEnvironmentVariable("XDG_DATA_HOME");
+                if (XDGDataHome != null)
+                {
+                    ApplicationsPath = Path.Combine(XDGDataHome, "applications");
+                }
+                else
+                {
+                    string home = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+                    ApplicationsPath = Path.Combine(home, ApplicationsPath);
+                }
+                Directory.CreateDirectory(ApplicationsPath);
+                MimeAppsListPath = Path.Combine(ApplicationsPath, MimeAppsListPath);
             }
         }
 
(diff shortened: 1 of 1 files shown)


politas:
Is there a new location for the .desktop files?

DinCahill:
I moved the `CreateDirectory` up to the constructor, and removed a test for the folder existing.

From the [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):

> If, when attempting to write a file, the destination directory is non-existant an attempt should be made to create it with permission 0700.

I'm not immediately sure how we should `chmod 0700`, so that still needs doing.

Postremus:
well, in the autoupdater, we literally just do a `chmod +x`.
https://github.com/KSP-CKAN/CKAN/blob/master/Core/Net/AutoUpdate.cs#L188

politas:
Can we test that the directory creation was successful and throw an error if not?

DinCahill:
@Postremus I'll copy that code, then.

GUI/URLHandlers.cs
```diff
@@ -13,17 +13,26 @@ public static class URLHandlers
...
         {
             if (Platform.IsUnix)
             {
-                string home = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
-                MimeAppsListPath = Path.Combine(home, MimeAppsListPath);
-                ApplicationsPath = Path.Combine(home, ApplicationsPath);
+                string XDGDataHome = Environment.GetEnvironmentVariable("XDG_DATA_HOME");
+                if (XDGDataHome != null)
+                {
+                    ApplicationsPath = Path.Combine(XDGDataHome, "applications");
```
ayan4m1: The `applications` subdirectory is deprecated. The preferred path is [specced here](https://standards.freedesktop.org/mime-apps-spec/mime-apps-spec-1.0.1.html)

DinCahill: Ah, thanks.

DinCahill:
So, we need a strategy for determining which `mimeapps.list` location to use. [Link to the spec](https://standards.freedesktop.org/mime-apps-spec/mime-apps-spec-1.0.1.html).

Should we write to the first one we find out of `$XDG_CONFIG_HOME/mimeapps.list` and `$XDG_DATA_HOME/applications/mimeapps.list`? If neither exists, which should we use? I don't know how old your distro would have to be to not support the one in `$XDG_CONFIG_HOME`.

I think `$XDG_DATA_HOME/applications/` (falling back to `~/.local/share/applications/`) is still the correct place for the `.desktop` file.

politas:
If neither location contains a mimeapps.list, we could possibly assume that the distribution is not supporting mime types, so why bother creating mimeapps.list?

DinCahill:
It doesn't mean mimeapps.list isn't supported. It just means that nothing's created any user-level mime-type associations yet. It's possible that a user's home directory could be completely empty, so we need to create these directories.

politas:
Hmm. Reading that spec, it sounds like we probably don't really need to be messing with mimeapps.list at all. We can put our associations in our ckan-handler.desktop file (which we should probably fix to be just ckan.desktop). mimeapps.list is to add or remove associations in the .desktop files. Am I off base?

politas:
Ok, I suppose that could be possible if people have a dedicated machine for running KSP.

DinCahill:
I think the distinction is that `ckan-handler.desktop` defines that it _supports_ `x-scheme-handler/ckan` (for informational purposes), but `mimetypes.list` is used to actually make the association for URL opening.

If I remove (or delete the line from) `mimetypes.list`, I can't open links. If I instead remove `MimeType=x-scheme-handler` from `ckan-handler.desktop`, it still works.

I've just realized that this makes "CKAN Launcher" show up in application launchers, unsurprisingly, and it should work fine. Its category is "Utility" (comes up under Accessories in XFCE4). The icon is defined as "ckan" but I don't have any of the icons installed (they go in `~/.local/share/icons` or something).

politas:
[Link to spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/) for .desktop files. See the "Registering Mime Types" page.

politas:
I actually made my own CKAN.desktop launcher, just so I could take control of where I'm launching it from and which category it goes in.

politas:
It's almost like Ubuntu is not properly following the Freedesktop standards.

DinCahill:
It says that `.desktop` files simply define the mime-types they support. It's not enough for actually opening the URLs.

politas:
Fair enough.

politas:
`$XDG_CONFIG_HOME/mimeapps.list` seems like where it should go to me. That should be either the same as  `$XDG_DATA_HOME/applications/mimeapps.list` or the correct location for the distribution. If `$XDG_CONFIG_HOME` is null, fall back to `~/.local/share/applications/`

GUI/URLHandlers.cs
```diff
@@ -159,13 +168,6 @@ private static void RegisterURLHandler_Linux()
             parser.WriteFile(MimeAppsListPath, data);
 
             var handlerPath = Path.Combine(ApplicationsPath, HandlerFileName);
-            var handlerDirectory = Path.GetDirectoryName(handlerPath);
```
ayan4m1: What is the rationale behind removing this code?

ayan4m1:
Just want to confirm why there was some code removed, otherwise LGTM.

PR #1851, opened 2016-08-04
Code style alignment

ayan4m1:
I ran CodeMaid over all the source and hand-refactored some classes related to registry and game path detection. I don't think there should be much functional change here but the devil is in the details and this is a very widely scoped changeset.

I will try to keep this rebased against master, for now I am opening the PR to get test results.


ayan4m1:
This PR has gotten messy, closing it for now...

PR #1853, opened 2016-08-04
Update CKAN.schema "as" regexp

politas:
The Regexp on install "as" is not being parsed correctly. swapping the escaped \ and the / seems to work, though.
Should enable KSP-CKAN/NetKAN#4487 to build - I think this is the first time we've actually used "as".

Change tested using http://www.jsonschemavalidator.net/

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -206,7 +206,7 @@
                     "as": {
                         "description" : "The name to give the matching directory or file when installed",
                         "type"        : "string",
-                        "pattern"     : "^[^/\\]+$"
+                        "pattern"     : "^[^\\/]+$"
                     },
                     "filter" : {
                         "description" : "List of files and directories that should be filtered from the install",
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -206,7 +206,7 @@
                     "as": {
                         "description" : "The name to give the matching directory or file when installed",
                         "type"        : "string",
-                        "pattern"     : "^[^\\/]+$"
+                        "pattern"     : "^[^\\\\/]+$"
                     },
                     "filter" : {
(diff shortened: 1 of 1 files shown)


DinCahill:
If we've not used "as" before, then I don't know whether the relevant C# does what it's supposed to do, but as far as this PR goes...

Is the regex supposed to mean "any non-empty string which doesnt contain a \ or /"?

To get that, I believe the regex (in JSON and, I think, Python?) needs to be `^[^\\\\/]+$`. Tested in JSON Schema Validator and ckan-validator.py.

ayan4m1:
Yeah, on the C# side, `"\\".Length` is 1. That would be a place to use `@` like this: `var regex = @"[^\\/]+$";`

politas:
This regexp is only used by the build script currently, and it doesn't use C#. I believe it's Python, @techman83?

politas:
Hmm. Yes, `\\\\` does seem to work better than `\\` ( or `\\\`, for that matter)

String | `"^[^\\/]+$"` | `"^[^\\\/]+$"` | `"^[^\\\\/]+$"` 
------ | ---- | ------ | ------
`test` | pass | pass | pass
`te/st` | fail | fail | fail
`te\st` | fail | fail | fail
`te\\st` | pass | pass | fail
`te\\\st` | fail | fail | fail
`te\\\\st` | pass | pass | fail

DinCahill:
The explanation for Python is at the top of [this page](https://docs.python.org/2/library/re.html). I don't know any more about JSON Schema.

techman83:
@politas yes, testing of the CKANs is done via Python.

politas:
Alright, let's see if it works, then.

PR #1854, opened 2016-08-04
KSPVersionLabel sizing - Cherry-picked from #1837 by Telanor

politas:


ayan4m1:
This makes the version string display fine even with a git SHA1 after it, but there is no actual guarantee that the version string won't get long enough to cause a recurrence of the UI issue.

I think it's worth merging for now.

politas:
We can bump it further when they go to five digit build IDs. That should be a ways down the track. Thanks.

PR #1877, opened 2016-08-30
Parse "action" argument to cmdline case-insensitively

### LINKED ISSUE #889: CmdLine modules should be case insensitive, even though internally they're case sensitive
_Saturday Dec 27, 2014 at 12:28 GMT # Sample: Friday Sep 13, 2013 at 22:58 GMT_
_Originally opened as https://github.com/KSP-CKAN/CKAN-support/issues/26_

---

_From @AstralNomad on November 26, 2014 4:4_

one thing i have noticed with the program... the module names are case sensitive.. might be a better idea to make module names case insensitive.. people dont pay attention and then complain because they cant get a mod to load and dont realize its due to case

_Copied from original issue: KSP-CKAN/CKAN#462_

ayan4m1:
This PR should fix #889 and makes a few light style/formatting fixes to the cmdline project. The `Exit` class has had its properties refactored to CamelCase. A few direct casts were converted to use the C# "as" statement, which passes the responsibility of null checking on to the code being called. The null checks should all be there, but review is appreciated.


ayan4m1:
Closing this for now, the style commits can be pulled into another cleanup branch.

PR #1878, opened 2016-08-30
Replace deprecated mozroots command with cert-sync for Travis

politas:
Explicitly adds ca-certificate-mono package, in case it doesn't get included with mono-complete
@techman83, can you check this?
Closes #1787

PR #1879, opened 2016-08-31
Update SharpZip dependency and remove old Newtonsoft.Json

### LINKED ISSUE #1787: mozroots - Couldn't retrieve the file using the supplied information.
**CKAN Version:** 
latest
**Operating System:** 
ubuntu
```
$ mozroots --import --ask-remove 
Mozilla Roots Importer - version 3.2.8.0
Download and import trusted root certificates from Mozilla's MXR.
Copyright 2002, 2003 Motus Technologies. Copyright 2004-2008 Novell. BSD licensed.

Downloading from 'http://mxr.mozilla.org/seamonkey/source/security/nss/lib/ckfw/builtins/certdata.txt?raw=1'...
Couldn't retrieve the file using the supplied information.
```

ayan4m1:
We were relying on a one-off build of an old version of SharpZipLib because it fixed some issues with Linux-generated archives. The same patches are applied to a NuGet package called SharpZipLib.Patched which is maintained by a much larger community than the DLL we currently rely on.

I added the patched nuget dependency and updated all of the projects that needed it to reference it. This resolves issues running the "GH221" test fixture.

The code currently also relies on a version-controlled copy of Newtonsoft.Json 6.0.8, which is coming up on two years old. I removed the binaries from source control and added the latest stable NuGet packaged version.

All of the tests that run in `master` run here with `nunit-console`. Please review and let me know if you have any concerns.
Closes #1787


politas:
If it works, great! At some stage, we'll need to look at upgrading to the newer nunit version, but we've got some tests that use removed methods we'll need to rewrite.

ayan4m1:
I have done the NUnit 2->3 port in one of the branches in my fork - fixing the CI build now!

politas:
I can't build this branch, either.

politas:
Ok, looks like whatever was blocking Mono 3.8.0 from running the nunit code before still is, so better take that out. Maybe add Mono version 4.0.5 as well, though I'm not aware of any distro using that branch. Debian/Ubuntu are certainly jumping straight from 3.2.8 to 4.2.1 (There are times when a conservative package release strategy is seriously problematic, and this is one of them)

Nuget failing to install just for the mono 3.2.8 build is pretty weird. [This](https://github.com/travis-ci/travis-ci/issues/6189) may have some useful info, but otherwise, raise a new issue with Travis people maybe? I can't find anything exactly describing the problem we have.

politas:
@techman83, does this resolve the issues you had with #1878?
Much as I'd like to maintain Mono 3.2.8, we have been saying for some time that we don't support it, and maybe it's time to let it drop.

.travis.yml
```diff
@@ -1,22 +1,32 @@
 language: csharp
 
+sudo: required
```
techman83: I don't see where we're using Sudo, so we should be able to drop this now.

techman83:
Oh gosh, I wish I could find the page that had the major package versions available per Debian/Ubuntu release that I found. The real issue is that 3.2.8 was the default for a _long_ time. Heck Debian Jessie still uses 3.2.8! Ubuntu only just upgraded to 4.2.1 in Ubuntu 16.04.

Now the mono project have excellent repos for most Distros, so we can either continue to target 3.2.8 (which is becoming increasingly difficult) or target 4.2.1 and provide instructions to update mono on Ubuntu/derivatives older than 16.04 etc here http://www.mono-project.com/docs/getting-started/install/linux/

politas:
Don't know why this one hasn't been marked as Merged. I did everything I normally do.

PR #1880, opened 2016-09-01
Fix Windows-only NullReferenceException and add more ModuleInstaller tests

ayan4m1:
If the user attempts to uninstall a mod in Windows, there is a NullReferenceException where AddParentDirectories tries to compare a null to a string. It is also continuing to walk up above the GameData directory, adding my Program Files directory to the list of directories to delete.

The new code normalizes all compared paths and checks for several edge cases, improving its reliability.

I added new tests to ensure the Windows path oddities I saw don't creep back in.

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -890,28 +890,39 @@ namespace CKAN
         /// Takes a collection of directories and adds all parent directories within the GameData structure.
         /// </summary>
         /// <param name="directories">The collection of directory path strings to examine </param>
-        private HashSet<string> AddParentDirectories(HashSet<string> directories )
+        public HashSet<string> AddParentDirectories(HashSet<string> directories)
         {
             var newDirectories = new HashSet<string>();
-            foreach (string directory in directories)
+            if (directories == null)
             {
-                if (directory.Contains(ksp.GameData()))
+                return newDirectories;
+            }
+
+            var normalGameData = KSPPathUtils.NormalizePath(ksp.GameData());
+            foreach (var directory in directories)
+            {
+                if (!directory.Contains(normalGameData))
                 {
-                    var path = directory;
-                    while ((path != ksp.GameData()) && !newDirectories.Contains(path))
-                    {
-                        newDirectories.Add(path);
-                        DirectoryInfo parent = Directory.GetParent(path);
-                        path = parent.FullName;
-                    }
+                    newDirectories.Add(KSPPathUtils.NormalizePath(directory));
                 }
-                else
+
+                var path = KSPPathUtils.NormalizePath(directory);
+                while ((path != normalGameData) && !newDirectories.Contains(path))
                 {
-                    newDirectories.Add(directory);
+                    newDirectories.Add(path);
+                    var parent = Directory.GetParent(path);
+                    if (parent == null)
+                    {
+                        log.Warn("ModuleInstaller tried to recurse above the filesystem root.");
+                        break;
+                    }
+
+                    path = KSPPathUtils.NormalizePath(parent.FullName);
                 }
             }
-                return newDirectories;
+            return newDirectories;
         }
+
         #region AddRemove
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


Core/ModuleInstaller.cs
```diff
@@ -890,28 +890,39 @@ private void Uninstall(string modName)
...
         {
             var newDirectories = new HashSet<string>();
-            foreach (string directory in directories)
+            if (directories == null)
             {
-                if ( directory.Contains(ksp.GameData()))
+                return newDirectories;
+            }
+
+            var normalGameData = KSPPathUtils.NormalizePath(ksp.GameData());
```
politas: ksp.GameData() is already normalized 

        public string GameData()
        {
            return KSPPathUtils.NormalizePath(
                Path.Combine(GameDir(), "GameData")
            );
        }

Core/ModuleInstaller.cs
```diff
@@ -890,28 +890,39 @@ private void Uninstall(string modName)
...
+            foreach (var directory in directories)
+            {
+                if (!directory.Contains(normalGameData))
```
politas: At this point directory has not been normalized

Core/ModuleInstaller.cs
```diff
@@ -890,28 +890,39 @@ private void Uninstall(string modName)
...
+                    newDirectories.Add(KSPPathUtils.NormalizePath(directory));
                 }
-                else
+
+                var path = KSPPathUtils.NormalizePath(directory);
+                while ((path != normalGameData) && !newDirectories.Contains(path))
                 {
-                    newDirectories.Add(directory);
+                    newDirectories.Add(path);
+                    var parent = Directory.GetParent(path);
```
politas: We should probably use 'KSPPathUtils.GetLeadingPathElements(path)` here, so we'd then be checking for an empty string next

Core/ModuleInstaller.cs
```diff
@@ -889,29 +889,41 @@ private void Uninstall(string modName)
...
+
+            var gameData = KSPPathUtils.NormalizePath(ksp.GameData());
+            return directories
+                .Where(dir => !string.IsNullOrWhiteSpace(dir))
+                // normalize before we deduplicate this list
+                .Select(KSPPathUtils.NormalizePath)
+                // remove any duplicate paths
+                .Distinct()
+                // get rid of paths that are above GameData
+                .Where(dir => dir.Length > gameData.Length)
```
politas: This would exclude any directory with a name shorter than 9 characters that was installed to Gameroot?

Core/ModuleInstaller.cs
```diff
@@ -889,29 +889,41 @@ private void Uninstall(string modName)
...
         {
-            var newDirectories = new HashSet<string>();
-            foreach (string directory in directories)
+            if (directories == null || directories.Count == 0)
             {
-                if ( directory.Contains(ksp.GameData()))
+                return new HashSet<string>();
+            }
+
+            var gameData = KSPPathUtils.NormalizePath(ksp.GameData());
```
politas: This becomes KSPPathUtils.NormalizePath(KSPPathUtils.NormalizePath( Path.Combine(GameDir(), "GameData")))

Core/ModuleInstaller.cs
```diff
@@ -889,29 +889,41 @@ private void Uninstall(string modName)
...
+                    var relativeHead = KSPPathUtils.ToRelative(dir, gameData);
+                    var pathArray = relativeHead.Split('/');
+                    var builtPath = string.Empty;
+                    foreach (var path in pathArray)
                     {
-                        newDirectories.Add(path);
-                        DirectoryInfo parent = Directory.GetParent(path);
-                        path = parent.FullName;
+                        builtPath += path + '/';
+                        results.Add(KSPPathUtils.ToAbsolute(builtPath, gameData));
```
politas: Things can be installed to other directories. We should include any directory not under gameData without iterating for parents.

politas:
TestNullPath() is failing. It looks like `_installer.AddParentDirectories(new HashSet<string>() { sysRoot });` is getting past the sysroot test, at which point ToAbsolute throws a kraken due to sysroot already being absolute.

PR #1881, opened 2016-09-01
Add log4net.xml, refactor logging init and log to file by default

ayan4m1:
The logging initialization has been condensed into one class which will either load config from a log4net.xml in the app startup directory, or failing to find one tries to initialize the existing basic logging system. All of the logging entry points have been refactored to call this `Logging.Initialize()` method.

Tests are passing in the Windows/VS environment, review and feedback appreciated.

This would resolve #1452, which unfortunately has some project changes that are risky and needs a rebase.

### DIFF
--- a/Cmdline/Main.cs
+++ b/Cmdline/Main.cs
@@ -48,8 +48,7 @@ namespace CKAN.CmdLine
                 args = guiCommand.ToArray();
             }
 
-            BasicConfigurator.Configure();
-            LogManager.GetRepository().Threshold = Level.Warn;
+            Logging.Initialize();
             log.Debug("CKAN started");
 
             Options cmdline;
--- a/Cmdline/CKAN-cmdline.csproj
+++ b/Cmdline/CKAN-cmdline.csproj
@@ -68,4 +68,9 @@
     <None Include="app.config" />
     <None Include="packages.config" />
   </ItemGroup>
+  <ItemGroup>
+    <Content Include="log4net.xml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
 </Project>
--- a/Cmdline/log4net.xml
+++ b/Cmdline/log4net.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<log4net>
+  <appender name="CkanConsole" type="log4net.Appender.ConsoleAppender">
+    <layout type="log4net.Layout.PatternLayout">
+      <conversionPattern value="%5level (%file:%line) %message%newline" />
+    </layout>
+  </appender>
+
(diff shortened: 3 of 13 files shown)


Cmdline/Main.cs
```diff
@@ -48,8 +48,7 @@ public static int Main(string[] args)
                 args = guiCommand.ToArray();
             }
 
-            BasicConfigurator.Configure();
-            LogManager.GetRepository().Threshold = Level.Warn;
```
politas: Regression of last change?

politas:
So if I'm understanding this, by default this will not change anything, but if a log4net.xml exists, it'll use that, so we can tweak logging to get just the useful logging output for debugging purposes?

politas:
Our Travis builds are failing since this was merged, on the "prove" step. It appears we are turning debug logging on by default somehow?

ayan4m1:
Ah, I see why - because we are logging to the console during builds, there is more output from the "version" CLI command than just the version number. The test scripts are checking the output against a regex to make sure the version string looks correct, which it doesn't because there is more text being output than just the version string.

My thought is that the test is kinda weak but obviously the functionality has changed which may be undesirable. Let me know how you want to proceed.

politas:
Well, the trouble is that when you run any command, you get debugging level of logging, which shouldn't be the default.

```
$ ckan version execute
108 [1] DEBUG CKAN.CmdLine.MainClass (null) - CKAN started
192 [1] DEBUG CKAN.KSPManager (null) - Loading KSP instances from registry
198 [1] DEBUG CKAN.KSPManager (null) - Loading BuildID-bug from /home/politas/Desktop/KSP/KSP-1.1.2/BuildID-bug
200 [1] DEBUG CKAN.KSP (null) - Initialising /home/politas/Desktop/KSP/KSP-1.1.2/BuildID-bug/CKAN
201 [1] DEBUG CKAN.KSP (null) - Initialised /home/politas/Desktop/KSP/KSP-1.1.2/BuildID-bug/CKAN
208 [1] DEBUG CKAN.KSPManager (null) - Added BuildID-bug at /home/politas/Desktop/KSP/KSP-1.1.2/BuildID-bug

```
I'm not sure why that is. I don't understand why it's doing that, and it shouldn't.

It isn't including the INFO level note "Debug logging enabled" that is inserted when the --debug option is detected.

Where is it looking for that xml file? Is it finding that and using it?

politas:
The online copy of the AVC file is confusing NetKAN - http://ksp-avc.cybutek.net/version.php?id=147

ayan4m1:
I wrote the new log init to fallback if no log4net.xml was found in the current working directory, I'm thinking that is why some installs are DEBUG level by default. The log4net.xml isn't an embedded resource but I think it's possible to package it inside the CKAN binary and reference it from there. That way we are directly controlling the desired log level from inside the binary. I'll open a new PR if that sounds good to you.

PR #1885, opened 2016-09-03
Handle more meta changes

ayan4m1:
Rebased #1730 against master in case it is useful, feel free to close otherwise.

### DIFF
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -106,14 +106,14 @@ namespace CKAN.CmdLine
             }
             catch (ModuleNotFoundKraken ex)
             {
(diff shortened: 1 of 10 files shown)


politas:
#1730 is a public branch, so could you PR to that branch instead of to master?

ayan4m1:
Nah, I had to rebase and the history is different. I could unpush and repush the branch entirely, but I didn't want to overwrite someone's work like that. Ultimately I'm happy to do it, just need some guidance on what to do.

politas:
People use a public branch specifically in order to enable a more collaborative method of working on a change, so don't think of it as "overwriting someone's work", but rather as "contributing to their work", just as we're all contributing to the CKAN on the larger scale.

ayan4m1:
Got it, closing this out.

politas:
Merged

Issue #1725, opened 2016-05-10
Crash on cancelling dependant mod selection


PiggyWhiskey:
CKAN 1.16.1
Windows 10

Clicking "Cancel" on the "Choose Mods" window causes NullReferenceException crash

```
************** Exception Text **************
System.NullReferenceException: Object reference not set to an instance of an object.
   at CKAN.Main.<TooManyModsProvide>c__async2.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at CKAN.MainModList.<ComputeChangeSetFromModList>c__async0.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at CKAN.Main.<UpdateChangeSetAndConflicts>c__async1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at CKAN.Main.<ModList_CellValueChanged>c__async0.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.<>c.<ThrowAsync>b__6_0(Object state)

```

plague006:
Unable to reproduce. I click the checkbox for "Real Plume" and click "Cancel" with no issues.

PiggyWhiskey:
1. Make sure Real Plume and Stock Config is uninstalled.
2. Select Real Plume and the Config and Click Apply
3. Click Apply Changes
4. It will recommend "EngineLighting" click Cancel
5. It will suggest a few other part's mods. Click Cancel.
6. You'll be back to "Choose Mods" with "Real Plume" and "Real Plume - Stock Config" unticked.
7. Click Cancel one more time.

```
System.NullReferenceException: Object reference not set to an instance of an object.
[...]

PR #1887, opened 2016-09-05
Add unit test for "sort by max version" GUI crash

ayan4m1:
I have fixed up the unit test for the sorting issue #1725. It now logs a test failure if a NullReferenceException occurs during a mod install/list refresh process after sorting the list GUI by "Max KSP Version."


politas:
Excellent!

PR #1888, opened 2016-09-06
WIP: Implement a "Replaced by" relationship

### LINKED ISSUE #904: Create a replaced_by relationship to allow seamless continuation of deprecated mods
This was 2 very different issues, 1 revolved around how to de-index deprecated mods, and one is about introducing a replaced_by relationship. I've pruned this conversation to be directed toward the latter topic as the former has a relatively well established workflow.
- @plague006

politas:
I welcome suggestions and comments as I work on this. 

Tracking  steps top level
- [x] Spec
- [x] Schema
- [ ] Test cases
- - [ ] Test valid replacement
- - [ ] Test invalid replacement
- [x] Core changes
- - [x] replaced_by added to CkanModule.cs
- - [x] Replace method added to ModuleInstaller
- - [x] Replace Method accepts resolver options and resolves dependencies
- - [x] New "ModuleReplacement" structure
- - [x] HasReplacement registry query - checks availability and KSP compatibility
- [x] CLI changes
- - [x] Show Replaceable mods in list command
- - [x] New "Replace" Command option
- - [x] Defined options for Replace command
- - [x] Add Replace command to main case
- - [x] Command line resolver options passed to ModuleInstaller.Replace 
- [x] ConsoleUI changes
- - [x] Show replaceable mods in modlist
- - [x] Add "Replace" tag option
- - [x] Call ModuleInstaller.Replace for replaced mods when applying changes
- [ ] GUI changes
- - [ ] Replace column in modlist and visual indicator
- - [ ] replaced_by data shown in Metadata panel
- - [ ] Call to ModuleInstaller.Replace for replaced mods
- - [ ] Add a "Replaceable" GUIModFilter
- - [ ] *discuss* Only allow a single changetype in a given changeset (I can't find an issue for this bug, but we've had it for a long time. Selecting updates and installs together causes a crash)
- - [ ] Add Right-click menu option for Replace
- [ ] Netkan changes
- - [ ] validate replaced_by relationship identifier
- [ ] Make sure no TODO1888 comments remain
Closes #904


politas:
Some additional file changes forced by .gitattributes in that commit.

politas:
There were a whole bunch of files with duplicated listings in the csproj file for tests. I removed all the ones that didn't exist. If anyone understands how to make these tests work, please help!!!! I can't get my head around them.

Cmdline/Action/List.cs
```diff
@@ -111,7 +119,7 @@ public int RunCommand(CKAN.KSP ksp, object raw_options)
 
             if (!(options.porcelain) && exportFileType == null)
             {
-                user.RaiseMessage("\r\nLegend: -: Up to date. X: Incompatible. ^: Upgradable. ?: Unknown. *: Broken. ");
+                user.RaiseMessage("\r\nLegend: -: Up to date. X: Incompatible. ^: Upgradable. >: Replaceable\r\n        A: Authodetected. ?: Unknown. *: Broken. ");
```
HebaruSan: Pls take the 'h' out of "Authodetected"

Spec.md
```diff
@@ -479,7 +479,18 @@ only.
...
-##### resources
+##### replaced-by
+
+(**v1.24**) This is a way to mark a specific mod identifier as being
+obsoleted and tell the client what it has been *replaced by*. It contains a
+single mod that should be selected for installation if this mod is
+selected for updating, while this mod is uninstalled. If this mod identifier
+is brought back to life, an epoch change should be applied. A *replaced_by*
+relationship should be added to the final release of the mod being replaced.
+The listed mod should include a "provides" relationship either to this mod,
```
HebaruSan: The stuff about "provides" makes this sound complicated, since now I have to worry about two changes to two different mods. Will `replaced_by` work on its own? What happens if the provides relation is missing?

Cmdline/Action/Replace.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        public IUser User { get; set; }
+
+        public Replace(IUser user)
+        {
+            User = user;
+        }
+
+
+        public int RunCommand(CKAN.KSP ksp, object raw_options)
```
HebaruSan: What's the rationale for making a new command for this? From reading #904, it sounded like this would tie into the existing `upgrade` command as needed. If we're confident enough in the replacement mod to put it in the metadata, why would we want to trouble the user with knowing whether it's an upgrade or a replaced_by?

CKAN.schema
```diff
@@ -144,6 +144,25 @@
             "items"       : { "type" : "string" },
             "uniqueItems" : true
         },
+        "replaced_by" : {
+            "description" : "Optional pointer to mod that should be selected instead and treated as an update to this mod",
+            "type" : "object",
+            "properties" : {
```
HebaruSan: Could this use `"$ref": "#/definitions/relationship"` like depends, recommends, suggests, supports, and conflicts do? That way we wouldn't have to specify the names and descriptions in multiple places.

HebaruSan:
I'm a little concerned about the implications of using replaced_by with multiple arguments. Presumably this is for the case where I release a mod containing Part 1 and Part 2 and then disappear, and separate people take up separate maintenance of Part 1 and Part 2. But what if my mod also included a Part 3 that no one adopts? Do we want to upgrade users to the replacement mods and lose Part 3? Or do we tell the new maintainers that they can't use `replaced_by` until all functionality is accounted for? It seems like we're putting ourselves in the position to be the arbiters of whether a set of replacement mods is "close enough" to the old mod, which potentially is a lot of work to check what's in the old mod and verify that each piece is in one of the replacement mods, and there will be cases where we lose either way.

Regarding the future plans...

> - [ ] Replace column in modlist and visual indicator

I don't see this as needed. Columns should be for data that I might want to see for multiple mods at once, or sort by. I don't see myself wondering, "Which mods have been replaced?" or needing to perform some task based on that. I think the detailed mod info panel is the right place for this.

Also, can we update the Max KSP Version column? If I'm using ClassicMod on 1.2, and it's only available for 1.2, but it's replaced_by NeoMod on 1.3, then I feel like the right Max KSP Version value is 1.3, since I could safely upgrade my game and (if NeoMod works as intended) still have the same functionality.

Spec.md
HebaruSan: Also, does this advice apply to the multiple-replacement case? Presumably if a mod only replaces part of another mod, it would not be correct to say that it "provides" that mod.

Tests/Tests.csproj
```diff
@@ -35,53 +35,40 @@
...
     <None Include="Data\CKAN-meta-badkan.zip" />
     <None Include="Data\CKAN-meta-testkan.tar.gz" />
     <None Include="Data\CKAN-meta-testkan.zip" />
-    <None Include="Data\data\DogeCoinFlag-1.01-corrupt.zip" />
-    <None Include="Data\data\DogeCoinFlag-1.01.zip" />
-    <None Include="Data\data\DogeCoinFlag-extra-files.zip" />
-    <None Include="Data\data\Firespitter-6.3.5.ckan" />
-    <None Include="Data\data\gh221.zip" />
-    <None Include="Data\data\ksp-avc.version" />
     <None Include="Data\data\KSP\KSP-0.25\GameData\README.md" />
```
HebaruSan: This file is also missing due to the absence of the `Data\data` folder

Tests/Tests.csproj
```diff
@@ -181,7 +169,6 @@
   <ItemGroup>
     <Content Include="Data\CKAN.sln" />
     <Content Include="Data\cmdline.t" />
-    <Content Include="Data\data\CKAN.sln" />
     <Content Include="Data\data\KSP\bad-KSP\readme.txt" />
     <Content Include="Data\data\KSP\KSP-0.25\readme.txt" />
     <Content Include="Data\data\KSP\missing-gamedata\readme.txt" />
```
HebaruSan: Can probably get rid of these three `Data\data` files as well?

Cmdline/Action/Replace.cs
politas: I think it is sufficiently different that users should have to make a distinct choice to replace a mod rather than simply upgrading. We don't want users asking us why they have unknown mods in their list suddenly when they aren't dependencies.

CKAN.schema
politas: I specifically didn't use the definitions/relationship, because this is a special case with limited options. There is only one possible replacement, and we do not allow max_version (because it makes no sense)

Spec.md
politas: There is no multiple-replacement case. We are only permitting a singe replacement mod.

Replaced_by works on its own. This is a "should", not a "must".

Tests/Tests.csproj
politas: Happy to, if they aren't used.

Cmdline/Action/List.cs
politas: Will do.

politas:
The reason I chose to mostly duplicate the contents of "#/definitions/relationship" for "replaced_by" in the schema was to limit replaced_by to a single mod. If at some stage there is a compelling argument to extend replaced_by to multiple mods, we could look at that, but the implementation of the concept I have gone with allows only a single replacement.

For the question of keeping track of original mods, I think if we force users to make explicit choices to replace a mod, we don't need to keep track of that any more, whereas if we were making the replacements invisible using the existing upgrade mechanisms, that would be a concern.

The GUI implementation is still a bit up in the air. I definitely want a separate checkbox column for replacements, and I want some really clear indication in the list that a replacement is there. These are fairly rare events, though, so ideally, the column will only appear when it there are replacements available. (And if I can figure out how to do that, I might apply the same logic to the Update column)

HebaruSan:
Thanks, those answers help to clarify a lot! I'll do another review pass soon.

Cmdline/Action/List.cs
```diff
@@ -65,26 +65,49 @@ public int RunCommand(CKAN.KSP ksp, object raw_options)
...
                             log.InfoFormat("Latest {0} is {1}", mod.Key, latest);
 
                             if (latest == null)
                             {
                                 // Not compatible!
                                 bullet = "X";
+                                //Check if mod is replaceable
+                                if ( current.replaced_by != null)
+                                {
+                                    CkanModule replacement = registry.LatestAvailable(current.replaced_by.name, ksp.VersionCriteria());
```
HebaruSan: `LatestAvailable` throws `ModuleNotFoundKraken` if the passed mod is not compatible with the active version, which will be caught at line 118 and trigger an error message. However, that error message will be misleading since it says the original replaced mod is not in the registry.

I think we need `try`/`catch` blocks around these `LatestAvailable` calls.

(Same comment for the identical block below)

Cmdline/Action/List.cs
```diff
@@ -65,26 +65,49 @@ public int RunCommand(CKAN.KSP ksp, object raw_options)
...
+                                        bullet = ">";
+                                        modInfo = string.Format("{0} {1} > {2} {3}", mod.Key, mod.Value, replacement.name, replacement.version);
+                                    }
+                                }
                             }
                             else if (latest.version.IsEqualTo(current_version))
                             {
                                 // Up to date
                                 bullet = "-";
+                                //Check if mod is replaceable
```
HebaruSan: As a general rule, code that's needed in multiple places should be refactored to a function and called from both places rather than copied. Can you do that with these blocks?

Cmdline/Action/List.cs
```diff
@@ -65,26 +65,49 @@ public int RunCommand(CKAN.KSP ksp, object raw_options)
...
+                                if ( current.replaced_by != null)
+                                {
+                                    CkanModule replacement = registry.LatestAvailable(current.replaced_by.name, ksp.VersionCriteria());
+                                    if (replacement != null)
```
HebaruSan: One of the main questions users seek to answer with CKAN is, "Can I upgrade to the next KSP version yet without losing any of my mods?" Once `replaced_by` is out in the wild, it will become part of that process.

Suppose OriginalMod is for 1.2 only, and RevivalMod replaces it on 1.3 only. Users need to know about that relationship while still on 1.2; is there any place where it's shown currently? It looks like it's not included in `ckan list` so far.

Cmdline/Action/Replace.cs
HebaruSan: Good points. But if a user just wants to get the latest updates and doesn't particularly care if they're called XYZ or XYZReplaced, I could see it being an irritation to have to run multiple commands, every time.

What about using an optional flag to create a combination command? Something like:
`ckan upgrade --all --with-replacements`

Cmdline/Main.cs
```diff
@@ -240,6 +240,11 @@ private static int RunAction(Options cmdline, CommonOptions options, string[] ar
                 case "remove":
                     return (new Remove(user)).RunCommand(manager.CurrentInstance, cmdline.options);
 
+                case "replace":
+                    user.RaiseMessage("Replace command detected in Main");
```
HebaruSan: Could this `RaiseMessage` be a `log.debug` instead? The other commands don't print analogous messages.

Core/Registry/IRegistryQuerier.cs
```diff
@@ -155,5 +155,54 @@ public static bool HasUpdate(this IRegistryQuerier querier, string identifier, K
...
+            if (replaceWith == null) return false;
+            // We definitely have a replacement mod which is compatible with our KSP Version Criteria
+            // If we have any mod version constrainsts, check them now.
+            if (replacedBy.min_version != null)
+            {
+                return !new List<string>(querier.InstalledDlls).Contains(identifier) && !replaceWith.version.IsLessThan(replacedBy.min_version);
+            }
+            else if (replacedBy.version != null)
+            {
+                return !new List<string>(querier.InstalledDlls).Contains(identifier) && replaceWith.version.Equals(replacedBy.version);
```
HebaruSan: I think this check can fail unnecessarily.

Suppose my currently installed mod has:
```
    "replaced_by": {
        "name": "MyMod",
        "version": "1.0.0"
    }
```
... but since that was set up, MyMod 1.1.1 has been released and indexed. The `LatestAvailable` call will find 1.1.1, and then reject it on this line because 1.0.0 and 1.1.1 are different versions. But 1.0.0 is still in the index and could be installed! I think we need a different call than `LatestAvailable` to find the CkanModule if `version` is set.

Cmdline/Action/Replace.cs
```diff
@@ -0,0 +1,148 @@
...
+            }
+
+            User.RaiseMessage("\r\nReplacing modules...\r\n");
+            var registry = RegistryManager.Instance(ksp).registry;
+            var to_replace = new List<ModuleReplacement>();
+
+            if (options.replace_all)
+            {
+                log.Debug("Running Replace all");
+                var installed = new Dictionary<string, Version>(registry.Installed());
```
HebaruSan: `Installed` already returns `Dictionary<string, Version>`. Do we really need to make a copy of it? That will presumably cost CPU time. This should work:
```
var installed = registry.Installed();
```

HebaruSan:
I've been grappling with this scenario:

1. I have ModX 1.0.0 installed, but I do not update my mods or CKAN registry regularly, so it's a bit out of date now
2. At some point in the past, ModX 1.0.1 was indexed and then later marked as `replaced_by` ModY 2.0.
3. I finally update my registry
4. CKAN tells me that my mod is upgradeable (but not replaceable)
5. I upgrade ModX to 1.0.1
6. I think I'm done, but I run `ckan list` to double check
7. Now CKAN tells me it's replaceable

If my goal is to get the most up to date mods available (probably the 80%+ case for our users), then I ought to have the option to replace at step 4, since 1.0.1 is just a stopover release and essentially a waste of bandwidth to download. In other words, if the `LatestAvailable` is replaceable, then treat previous versions of the same mod as replaceable as well. (Obviously this argument doesn't apply if I'm one of the hypothetical picky users who care about replacements happening; in that case 1.0.1 is worth downloading. That's why this should only be an _option_.)

If the current  behavior is kept, could we add a check for replaceability to the end of `Cmdline\Upgrade.cs`? This would at least make sure that the user knew a replacement was available as soon as the replaceable version was installed. I.e.:
```
ModX upgraded; a replacement is available: ModY
```

(I haven't finished auditing `ModuleInstaller.cs` yet, so apologies if this is already handled.)

Cmdline/Action/List.cs
politas: The list command as it is doesn't help with that use case at all, since it only lists installed mods and updates that match the current KSP version. To answer that question, we'd need some new functionality based on the "available" command, I think.

Cmdline/Action/List.cs
politas: What happens when you nest try / catch blocks?

Cmdline/Main.cs
politas: Oh, yes, that was supposed to be a `log.debug`.

Core/Registry/IRegistryQuerier.cs
politas: Good point. Alternatively, we could make the spec only include min_version, which probably makes more sense.

Cmdline/Action/List.cs
HebaruSan: It works as you'd expect, the inner `catch` handles exceptions inside the inner `try` block, and the outer `catch` handles exceptions in the outer `try` block. If the inner `catch` doesn't handle an exception in the inner block or or re-throws it, it gets passed to the outer `catch`.

politas:
I think we'd like to keep the command line fairly simple, and I think CLI users are probably going to be happy to have iterative changes. We're talking about pretty unusual cases, after all. It is probably worth checking replaced_by on latest available in the GUI, since it will more easily show both options.

Let me check the logic you are proposing:

```
If (Current_version or latest_version has a replaced_by), 
    and ( 
        (replaced_by_specified_version is compatible) 
        or (replaced_by_latest_version is compatible and not less than min_version)  ),
then (offer the replacement)
```

Cmdline/Action/List.cs
politas: Yep, and I can then do more detailed testing on the replaced_by version criteria

Cmdline/Action/Replace.cs
politas: I'm not sure why it is like that. I copied the code from the Upgrade command, and that bit is unchanged

Cmdline/Action/List.cs
HebaruSan: I like the symmetry between `modInfo` and `newModInfo` now.

Core/ModuleInstaller.cs
```diff
@@ -1100,6 +1100,90 @@ public void Upgrade(IEnumerable<CkanModule> modules, NetAsyncModulesDownloader n
...
+            // adding everything that needs installing (which may involve new mods to
+            // satisfy dependencies). 
+            var to_remove = new List<string>();
+
+            // Let's discover what we need to do with each module!
+            foreach (ModuleReplacement repl in replacements)
+            {
+                string ident = repl.ToReplace.identifier;
+                InstalledModule installed_mod = registry_manager.registry.InstalledModule(ident);
+                CkanModule ReplaceWith = repl.ReplaceWith;
```
HebaruSan: The compiler is telling me that this variable isn't needed anymore.

Cmdline/Action/List.cs
```diff
@@ -111,13 +126,63 @@ public int RunCommand(CKAN.KSP ksp, object raw_options)
...
             {
-                user.RaiseMessage("\r\nLegend: -: Up to date. X: Incompatible. ^: Upgradable. ?: Unknown. *: Broken. ");
+                user.RaiseMessage("\r\nLegend: -: Up to date. X: Incompatible. ^: Upgradable. >: Replaceable\r\n        A: Autodetected. ?: Unknown. *: Broken. ");
                 // Broken mods are in a state that CKAN doesn't understand, and therefore can't handle automatically
             }
 
             return Exit.OK;
         }
 
+        private string ReplacmentModInfo (RelationshipDescriptor replacedBy, IRegistryQuerier registry, Versioning.KspVersionCriteria versionCriteria)
```
HebaruSan: Slight typo, ReplacmentModInfo vs ReplacementModInfo

HebaruSan: ... but the logic of the function looks solid to me. :+1:

Cmdline/Action/Replace.cs
HebaruSan: Fair enough; it would make sense to reserve this for a separate project that addresses all commands at once, if needed.

Cmdline/Action/List.cs
HebaruSan: Could this function call `IRegistryQuerier.GetReplacement`? They look similar. That way the logic would be even more centralized, and if we decide to make it more complex in the future we can just update that one Core function and `CmdLine.List` would get the changes for free.

Cmdline/Action/Replace.cs
```diff
@@ -0,0 +1,142 @@
...
+                        {
+                            log.InfoFormat("{0} is installed, but it or its replacement is not in the registry",
+                                mod.Key);
+                        }
+                    }
+                }
+            }
+            else
+            {
+                foreach (string mod in options.modules)
```
HebaruSan: Currently if I try to replace mods that aren't replaceable...

```
$ mono ckan.exe replace KerbalEngineerRedux

Replacing modules...


Done!
```

... there's no error message, it just says that it finished. Similarly, I can even give it completely invalid mod names:

```
$ mono ckan.exe replace ThisIsNotAValidMod

Replacing modules...


Done!
```

If I'm trying to replace a mod that actually is replaceable but I mistype the name or capitalize it wrong, I might think I have just performed the replacement when in fact I have done nothing. I think it would be good to raise errors here if mods the user specified are invalid or not replaceable.

Cmdline/Action/Replace.cs
```diff
@@ -0,0 +1,142 @@
...
+
+            if (options.modules.Count == 0 && ! options.replace_all)
+            {
+                // What? No mods specified?
+                User.RaiseMessage("Usage: ckan replace Mod [Mod2, ...]");
+                User.RaiseMessage("  or   ckan replace --all");
+                return Exit.BADOPT;
+            }
+
+            User.RaiseMessage("\r\nReplacing modules...\r\n");
```
HebaruSan: Currently this message is printed even if nothing is replaced:

```
$ mono ckan.exe replace --all

Replacing modules...


Done!
```
... which is a little misleading as no modules have actually been replaced.
Let's move this to right before the `GetInstance.Replace` call, and check `.Count>0` first. Then we can tell the user something like, "No replaceable mods found" if that's the case.

Cmdline/Action/Replace.cs
```diff
@@ -0,0 +1,142 @@
...
+                            }
+                        }
+                    }
+                    catch (ModuleNotFoundKraken kraken)
+                    {
+                        User.RaiseMessage("Module {0} not found", kraken.module);
+                    }                                           
+                }
+            }
+            // TODO: These instances all need to go.
```
HebaruSan: Any idea what this comment means? I see it in `Upgrade.cs`, but if we don't know what change it is suggesting, we might as well not carry it forward.

Core/Registry/IRegistryQuerier.cs
```diff
@@ -155,5 +155,65 @@ public static bool HasUpdate(this IRegistryQuerier querier, string identifier, K
...
+            if (installedVersion == null) return null;  // Mod is not installed, so we don't care about replacements
+            //get the identifier from the replaced_by relationship, if it exists
+            RelationshipDescriptor replacedBy = installedVersion.replaced_by;
+            if (installedVersion.replaced_by == null) return null; // No replaced_by relationship
+            // Now we need to see if there is a compatible version of the replacement
+            try
+            {
+                ModuleReplacement replacement = new ModuleReplacement();
+                replacement.ToReplace = installedVersion;
+                if (installedVersion.replaced_by.version != null)
```
HebaruSan: I tried faking up some data in my `registry.json` to test this, and at first it didn't work because I added the `replaced_by` block to `available_modules` instead of `installed_modules`.

This is a point where my knowledge isn't deep enough; do you know whether that will be a problem in practice? Do we update the data in `installed_modules` with updated data from the same versions in `available_modules` if the metadata online changes? Do we need to check both here?

Cmdline/Action/Replace.cs
HebaruSan: I just tried this with a validly replaceable mod, and the output is the same. We should also print a message for each mod that is successfully replaced.

HebaruSan: There should probably also be a Y/n confirmation prompt. When installing, we list the mods to be installed and prompt the user to confirm, even though they have just typed those exact names at the command line:

```
About to install...

 * Astrogator v0.7.8(cached)

Continue? [Y/n] 
```

But for replacement, when I have not even typed the name of the mod to install, it just does it silently.

Cmdline/Action/Replace.cs
```diff
@@ -0,0 +1,142 @@
...
+            ModuleInstaller.GetInstance(ksp, User).Replace(to_replace, new NetAsyncModulesDownloader(User));
+            User.RaiseMessage("\r\nDone!\r\n");
+
+            return Exit.OK;
+        }
+
+        internal static CkanModule LoadCkanFromFile(CKAN.KSP current_instance, string ckan_file)
```
HebaruSan: Let's share this function with `Install.cs` to avoid duplicating code. This file could call the other's, or vice versa, or this could go in a separate file.

Core/ModuleInstaller.cs
```diff
@@ -1100,6 +1100,90 @@ public void Upgrade(IEnumerable<CkanModule> modules, NetAsyncModulesDownloader n
...
+                    else
+                    {
+                        log.InfoFormat("Replacing {0} with {1} {2}", repl.ToReplace.identifier, repl.ReplaceWith.identifier, repl.ReplaceWith.version);
+                    }
+                }
+            }
+
+            AddRemove(
+                modsToInstall,
+                to_remove,
```
HebaruSan: Minor style thing, but it would be nice to name these consistently. I.e., either:
```
modsToInstall
modsToRemove
```
or:
```
to_install
to_remove
```

HebaruSan:
Dependencies of replacement mods aren't being handled. I installed KER and then faked up data to replace it with GPP, which has several dependencies:

```
$ mono ~/Downloads/ckan replace --all

Replacing modules...

Downloading "https://github.com/Galileo88/Galileos-Planet-Pack/releases/download/1.5.3/Galileos.Planet.Pack.1.5.3.zip"
0 kbps - downloading - 0 MB left - 100%              
Unhandled Exception:
GPP depends on Kopernicus but it is not listed in the index, or not available for your version of KSP.
 * GPP depends on Scatterer but it is not listed in the index, or not available for your version of KSP.
 * GPP depends on Scatterer-config but it is not listed in the index, or not available for your version of KSP.
 * GPP depends on GPPTextures but it is not listed in the index, or not available for your version of KSP.  at CKAN.SanityChecker.EnforceConsistency (System.Collections.Generic.IEnumerable`1[T] modules, System.Collections.Generic.IEnumerable`1[T] dlls) [0x0001c] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.Registry.CheckSanity () [0x00036] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.RegistryManager.Save (System.Boolean enforce_consistency, System.Boolean recommmends, System.Boolean with_versions) [0x0001d] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.ModuleInstaller.AddRemove (System.Collections.Generic.IEnumerable`1[T] add, System.Collections.Generic.IEnumerable`1[T] remove, System.Boolean enforceConsistency) [0x00082] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.ModuleInstaller.Replace (System.Collections.Generic.IEnumerable`1[T] replacements, CKAN.NetAsyncModulesDownloader netAsyncDownloader, System.Boolean enforceConsistency) [0x0032b] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.CmdLine.Replace.RunCommand (CKAN.KSP ksp, System.Object raw_options) [0x00376] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.CmdLine.MainClass.RunAction (CKAN.CmdLine.Options cmdline, CKAN.CmdLine.CommonOptions options, System.String[] args, CKAN.IUser user, CKAN.KSPManager manager) [0x002c3] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.CmdLine.MainClass.Main (System.String[] args) [0x003b7] in <e486b017cc5f4354b7969c2542f78c53>:0 
[ERROR] FATAL UNHANDLED EXCEPTION: The following inconsistencies were found:
GPP depends on Kopernicus but it is not listed in the index, or not available for your version of KSP.
 * GPP depends on Scatterer but it is not listed in the index, or not available for your version of KSP.
 * GPP depends on Scatterer-config but it is not listed in the index, or not available for your version of KSP.
 * GPP depends on GPPTextures but it is not listed in the index, or not available for your version of KSP.  at CKAN.SanityChecker.EnforceConsistency (System.Collections.Generic.IEnumerable`1[T] modules, System.Collections.Generic.IEnumerable`1[T] dlls) [0x0001c] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.Registry.CheckSanity () [0x00036] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.RegistryManager.Save (System.Boolean enforce_consistency, System.Boolean recommmends, System.Boolean with_versions) [0x0001d] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.ModuleInstaller.AddRemove (System.Collections.Generic.IEnumerable`1[T] add, System.Collections.Generic.IEnumerable`1[T] remove, System.Boolean enforceConsistency) [0x00082] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.ModuleInstaller.Replace (System.Collections.Generic.IEnumerable`1[T] replacements, CKAN.NetAsyncModulesDownloader netAsyncDownloader, System.Boolean enforceConsistency) [0x0032b] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.CmdLine.Replace.RunCommand (CKAN.KSP ksp, System.Object raw_options) [0x00376] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.CmdLine.MainClass.RunAction (CKAN.CmdLine.Options cmdline, CKAN.CmdLine.CommonOptions options, System.String[] args, CKAN.IUser user, CKAN.KSPManager manager) [0x002c3] in <e486b017cc5f4354b7969c2542f78c53>:0 
  at CKAN.CmdLine.MainClass.Main (System.String[] args) [0x003b7] in <e486b017cc5f4354b7969c2542f78c53>:0 
```

This means that this command would be very likely to corrupt users' installs in practice.

Cmdline/Action/Replace.cs
politas: All those sort of issues are logInfos. By default, the CLI only reports actual actions it takes.

Cmdline/Action/Replace.cs
politas: It is about the `GetInstance(ksp, User)` stuff below. It's ugly as hell, but it's pretty endemic throughout the code due to someone's possibly unnecessary decisions in the data design. I think the original idea was to have different KSP instances/registries all loaded in memory  at the same time, and allow the user to switch quickly between them. At some stage, that became infeasible, but the fundamental structure will take a significant project to remove.

Core/Registry/IRegistryQuerier.cs
politas: The best way is to mock up a registry like I have done with https://github.com/politas/CKAN-meta/raw/TEST-Replaced_by/TEST-meta-master.tar.gz. Hand editing the registry.json file rarely mimics actual situations correctly.

Cmdline/Action/Replace.cs
HebaruSan: No, the CLI reports errors when mods are missing:

```
$ mono ckan install ThisIsNotAValidMod
Module ThisIsNotAValidMod required but it is not listed in the index, or not available for your version of KSP.
```

Cmdline/Action/Replace.cs
politas: Action/Upgrade has another duplicate of this code. I don't know why it's internal

Cmdline/Action/Replace.cs
politas: Upgrade prints a similar message before anything is actually started or tested.

politas:
I'll have to dive deeper to integrate more of the Install code. The Upgrade code might need a going over to re-check dependencies, too.

Cmdline/Action/Replace.cs
HebaruSan: `internal` is the same as public but only within the same project, so these classes could already share those functions as they are since they're all in the CmdLine project.

HebaruSan:
Updating the registry with this change adds this to every mod in `registry.json`:

```
    "replaced_by": null,
```

This takes up a non-negligible amount of disk space:

```
-rw-r--r-- 1 21247092 Nov  9 16:18 before.json
-rw-r--r-- 1 21518092 Nov  9 16:19 after.json
```

Total increase: 271000 bytes. That's 1.3% of the (already bloated) registry file.

Is there a way to tell the serializer to omit this property when it's `null` instead of including it?

Core/Types/CkanModule.cs
```diff
@@ -161,6 +167,9 @@ public class CkanModule : IEquatable<CkanModule>
         [JsonProperty("depends")]
         public List<RelationshipDescriptor> depends;
 
+        [JsonProperty("replaced_by")]
```
HebaruSan: Let's save disk space with:
```csharp
[JsonProperty("replaced_by", NullValueHandling = NullValueHandling.Ignore)]
```

Cmdline/Action/List.cs
politas: Yeah, I based GetReplacement on this. I will have a look to see if I can re-use that code reasonably. I don't see why not.

politas:
I don't understand why this is failing. It works fine locally for me.

HebaruSan:
Something's wrong with the JSON serializer reference:

```
Could not load file or assembly or one of its dependencies. assembly:Newtonsoft.Json
```

I can reproduce this locally if I delete the `_build` folder first to force it to re-download the NuGet packages. You might still have old downloads in that folder that are masking this problem.

politas:
Ah, yep. Ok, now I can try to figure it out.

HebaruSan:
I sent you a pull request on your branch that should fix it.

Cmdline/Action/List.cs
politas: And.... done!

politas:
Rebased and fixed Replace command dialog. It now only says "Replacing Mods" when it is actually going to replace some mods, and it prompts for confirmation before continuing after listing all changes. Next is to get the Dependencies and suggestions handled properly

politas:
@HebaruSan , I think I've addressed all the issues you've raised so far now. Can you add a consoleui implementation for me?

HebaruSan:
Sure; politas/CKAN#6 now has the changes I had for that. I'm not sure that it's 100% robust, but it should at least cover the basics.

politas:
@HebaruSan , I'm finding it very hard to find time to get back in and finish this. If you would like to take it over and bring it to completion, we might actually be able to get this feature included.

HebaruSan:
@politas, OK, let me know how #2671 looks.

politas:
Closing in favour of #2671

PR #1897, opened 2016-09-12
Add Curse to resources

Olympic1:
Adds the Curse Uri to the `"resources"` stanza together with curse licenses.

```
"resources": {
    "curse": "http://kerbal.curseforge.com/projects/220221"
}
```

### DIFF
--- a/GUI/GUIMod.cs
+++ b/GUI/GUIMod.cs
@@ -131,7 +131,7 @@ namespace CKAN
 
             Abstract = mod.@abstract;
             
-            // If we have homepage provided use that, otherwise use the spacedock page or the github repo so that users have somewhere to get more info than just the abstract.
+            // If we have a homepage provided, use that; otherwise use the spacedock page, curse page or the github repo so that users have somewhere to get more info than just the abstract.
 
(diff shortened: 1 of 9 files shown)


politas:
Looks good, and can go into the patch release.

GUI/GUIMod.cs
```diff
@@ -131,7 +131,7 @@ public GUIMod(CkanModule mod, IRegistryQuerier registry, KspVersion current_ksp_
 
             Abstract = mod.@abstract;
             
-            // If we have homepage provided use that, otherwise use the spacedock page or the github repo so that users have somewhere to get more info than just the abstract.
+            // If we have a homepage provided, use that; otherwise use the spacedock page, curse page or the github repo so that users have somewhere to get more info than just the abstract.
 
             Homepage = "N/A";
             if (mod.resources != null)
```
ayan4m1: Now that we have dropped Mono 3.2 compiler tests we can make this one long `x ?? y ?? z` expression

politas: We haven't done that yet. That'll be after 1.20.1

PR #1899, opened 2016-09-15
Use buildID64.txt in preference to buildID.txt if it exists.

### LINKED ISSUE #1898: 64-bit version files (eg. for KSP 1.2) are ignored.
Basically, KspBuildIdVersionProvider.cs uses buildID.txt, ignoring newer buildID64.txt

With 1.2, KSP is distributed as separate 32-bit and 64-bit versions. Worse, if using the Patch to 64-bit, the old buildID.txt from the full 1.1.3 is left, so CKAN sees that and believes it's 1.1.3 still.

politas:
Chances are, that most users will be using the Win64 version. If not, we just need to advise them to delete buildID64.txt
closes #1898


politas:
Just to make it clear, this is a quick hack that I expect will help the majority of 1.2 upgraders. We should probably check both files and use the highest buildID.

politas:
Closing in favour of #1900

PR #1900, opened 2016-09-15
Use maximum of buildID and buildID64 if both are available

### LINKED ISSUE #1898: 64-bit version files (eg. for KSP 1.2) are ignored.

dbent:
@politas Update of #1899 fixes #1898

Modifies the code to check for both buildID.txt and buildID64.txt and use the maximum KSP version of the two. Also writes a warning if they happen to represent different versions.


Olympic1:
This doesn't seem to work with the current pre-releases of 1.2. I have patched my ksp folder with the 64bit patch but ckan still thinks its a 1.1.3 eventhough it should select the highest buildID.

buildID.txt:
> build id = 01289
2016.06.21 at 20:12:36 CEST
Branch: master

buildID64.txt:
> build id = 01532
2016.09.24 at 06:01:43 CEST
Branch: prerelease

CKAN:
ksp dirs

PR #1906, opened 2016-09-25
Make sure we're updating the build mappings on repository update

dbent:
Fix issue brought up in #1900 that @Olympic1 subsequently deleted?

Looks like this [rebase](https://github.com/KSP-CKAN/CKAN/commit/94335381e274cb07fe77ae1fa6183a2513131b8f#diff-a5b6137749fb38bca7e80257e4e5b158L144) accidentally deleted the line that updates the build mapping, thus leading to stale build mappings.

### DIFF
--- a/Core/Net/Repo.cs
+++ b/Core/Net/Repo.cs
@@ -4,7 +4,9 @@ using System.IO;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
+using Autofac;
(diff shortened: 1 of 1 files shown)


Olympic1:
I don't know how my comment got deleted, I'm having some issues lately on github. Luckily you get notified by mail 😄

Olympic1:
Tested and works. Merging with my thanks

PR #1919, opened 2016-10-20
Add regexp second test for vref filespecs

### LINKED ISSUE #1918: Is there regex in vref
Is there a way to include regex inside a vref, because that would be very useful in [this case](https://github.com/KSP-CKAN/NetKAN/pull/4779).

politas:
Closes #1918
Shouldn't impact any existing $vref filespecs, as they all have complete paths.

### DIFF
--- a/Netkan/Services/ModuleService.cs
+++ b/Netkan/Services/ModuleService.cs
@@ -117,12 +117,8 @@ namespace CKAN.NetKAN.Services
                 for (var i = 0; i < files.Count; i++)
                 {
                     Log.DebugFormat("Testing file '{0}' against path '{1}'", files[i].Name,internalFilePath);
-                    if (files[i].Name == internalFilePath) 
-                    {
-                        remoteIndex = i;
-                        break;
-                    }
-                    else if (Regex.IsMatch(files[i].Name, internalFilePath))
+                    // Test for either an exact match or using the filespec as a regexp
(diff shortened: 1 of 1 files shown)


politas:
#rebuild

Netkan/Services/ModuleService.cs
```diff
@@ -116,7 +116,13 @@ private static AvcVersion GetInternalAvc(CkanModule module, ZipFile zipfile, str
...
                 for (var i = 0; i < files.Count; i++)
                 {
-                    if (files[i].Name == internalFilePath)
+                    Log.DebugFormat("Testing file '{0}' against path '{1}'", files[i].Name,internalFilePath);
+                    if (files[i].Name == internalFilePath) 
+                    {
+                        remoteIndex = i;
+                        break;
+                    }
+                    else if (Regex.IsMatch(files[i].Name, internalFilePath))
```
politas: I thought about combining the tests, and it probably makes sense to do it that way. We're not cycling through the list a second time for the other test. 

Hang on, what's the "?" you've got there?

Olympic1: the "?"  declares a nullable type, and means that the type before it may have a null value.

politas:
The schema is for .ckan files, not .netkan files, so this doesn't affect it.

PR #1920, opened 2016-10-20
Fix default logging on fallback when no XML file

### LINKED ISSUE #1913: CKAN generates a lot of DEBUG output now.
**CKAN Version:** tip as of this morning -- last commit 2764a1ce

**Operating System:** Linux

**The issue you are experiencing:** I run ckan, I get ninety thousand lines of DEBUG output

**How to recreate this issue:** Check out master.  `docker-compose build ckan`.  `docker-compose run --rm ckan`

**CKAN error codes (if applicable):** It succeeds.  It even updates the mods!  It just tells me allllll about it. :-)

politas:
Closes #1913, ref #1881 
This should get our builds working again. "prove" works on this build.

### DIFF
--- a/Core/Logging.cs
+++ b/Core/Logging.cs
@@ -2,6 +2,7 @@
 using System.IO;
 using log4net;
 using log4net.Config;
+using log4net.Core;
 
 namespace CKAN
 {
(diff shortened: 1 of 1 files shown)


ayan4m1:
Yep, still bugs me that BasicConfigurator is using the release configuration to determine the default logging level, but this is a nice workaround.

mathuin:
Can confirm that logging is gone when running Docker container with this code.

politas:
The line I've added was in the code that was replaced with the call to the new Logging code, so we're falling back to what we had before, exactly. Ok, I'll merge this.

PR #1925, opened 2016-10-23
Allow user to specify build map URL

ayan4m1:
This started off as a TODO, figured I'd develop it and it also led to making the SettingsDialog unit testable.

### DIFF
--- a/GUI/SettingsDialog.cs
+++ b/GUI/SettingsDialog.cs
@@ -4,6 +4,8 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
+using System.Net;
+using System.Threading.Tasks;
 using System.Windows.Forms;
 using log4net;
 using log4net.Repository.Hierarchy;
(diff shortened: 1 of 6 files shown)


politas:
I see some changes to the cache location, but it doesn't look like you've made that configurable? We've got an outstanding issue to do that. While you're working with this bit of code, can you sneak that in?

ayan4m1:
Sure, but there's no sneaking really. The download cache directory is obtained by calling a method on a KSP instance, which has access to `GameDir()` and `CkanDir()` and outside the scope of the GUI-only Configuration object.

I had a branch going where I was working on substantial refactoring to the `Main` class - part of that should be making one consistent configuration object - right now all of the GUI config is managed by a class that is not available to the core project.

politas:
Ah, ok. So your changes here are just making the GUI properly show the downloads directory as stored in the KSP instance, rather than assuming the /downloads folder?

GUI/SettingsDialog.cs
```diff
@@ -67,47 +133,45 @@ private void RefreshReposListBox()
 
         private void UpdateCacheInfo()
         {
-            m_cacheSize = 0;
-            m_cacheFileCount = 0;
-            var cachePath = Path.Combine(Main.Instance.CurrentInstance.CkanDir(), "downloads");
```
politas: Ok, I'm understanding that change correctly, then.

politas:
I note that you've changed the cache clearing from removing all the files to deleting the directory. That's going to cause issues for all of us who make their downloads folder a symlink to a single location. (Though, admittedly, we're less likely to ever actually clear the cache that way). Is there a particular reason for deleting the directory rather than the files other than "it's faster"?

ayan4m1:
The existing code tries to delete each file once, but silently gobbles all exceptions, so you'd never know if the issue was permissions vs filename not matching. Makes sense about the symlink though, I can probably just add a log line to the old catch block.

politas:
Adding some logging sounds fine. They'll see the simple fact of a failure to clear the cache in the dialog box.

PR #1930, opened 2016-10-26
Update UI state when download cache is cleared

### LINKED ISSUE #1928: Clear cache needs to mark packages as not cached.
When cache is cleared, mark all formerly cached packages as not cached anymore.

**CKAN Version:** 
v1,20.1-0-gd41fe54 (beta) - KSP 1.2.0.1586
**Operating System:** 
Windows 7 64 bit
Clearing cache deletes cached packages, but does not mark packages as not being cached
download packages
clear cache
click on package in package list
cache file cannot be found (because CKAN deleted it and yet still expects it to be there)
Unhandled exception "Could not find file 'f:\program files\steam\SteamApps\common\Kerbal Space Program\CKAN\downloads\E3E430A1-EVAParachutes-0.1.6.zip'."

ayan4m1:
Fixes #1928 by adding a method to GUIMod which allows the IsCached property to be updated, then calling it once when we finish caching after clicking Download and once when the cache is cleared.

### DIFF
--- a/Core/Net/NetFileCache.cs
+++ b/Core/Net/NetFileCache.cs
@@ -86,7 +86,7 @@ namespace CKAN
         /// When our cache dirctory changes, we just clear the list of
         /// files we know about.
         /// </summary>
-        private void OnCacheChanged()
+        public void OnCacheChanged()
         {
             cachedFiles = null;   
         }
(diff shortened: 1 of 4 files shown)


rkagerer:
I will retest when I get a chance.  If you don't hear back in a couple days (or prefer not to wait) feel free to close and I can reopen if necessary.  Thanks for the fix, much appreciated.

politas:
You'll have to build the current codebase to see this fix; we haven't released a new client, yet.

rkagerer:
Ok, thanks, will wait for the release.

PR #1939, opened 2016-11-13
Use OpenFileDialog instead of FolderBrowserDialog in instance selector

### LINKED ISSUE #1017: When selecting Kerbal directory in GUI, allow text input
_Wednesday Feb 18, 2015 at 22:36 GMT # Sample: Friday Sep 13, 2013 at 22:58 GMT_
_Originally opened as https://github.com/KSP-CKAN/CKAN-GUI/issues/55_

---

_From @alexkindel on February 15, 2015 4:2_

It's very frustrating when you can't have text input. I can browse to my Kerbal directory from steam and then copy paste the address, but can't use it because you can't paste in this type of Browse Dialogue on Windows.

Unrelated: I'm not really sure if GUI issues should go here or in the CKAN-GUI issues tracker?

_Copied from original issue: KSP-CKAN/CKAN#777_

ayan4m1:
This would resolve #1017 by changing the selection behavior for new instances. Instead of the `FolderBrowserDialog` we would be asking the user to select a KSP executable as a way of indicating the game root directory.

I'd like input on how safe that is - I believe that the KSP binaries are still suffixed in .exe when run in Mac/Linux environments but I haven't seen one of these installs.

The actual filter is `KSP*.exe` - so users will see 32- and 64-bit versions but can only select one.

There are also style changes in this file, variable names and changing `SetButtonsEnabled` because "Set*/Get*" methods are not C# style.

### DIFF
--- a/GUI/ChooseKSPInstance.cs
+++ b/GUI/ChooseKSPInstance.cs
@@ -7,35 +7,42 @@ namespace CKAN
 {
     public partial class ChooseKSPInstance : Form
     {
-        private FolderBrowserDialog browseKspFolder;
-        private RenameInstanceDialog renameInstanceDialog;
-        private readonly KSPManager manager;
+        private readonly KSPManager _manager;
+        private RenameInstanceDialog _renameInstanceDialog;
+        private readonly OpenFileDialog _instanceDialog = new OpenFileDialog()
+        {
+            AddExtension = false,
+            CheckFileExists = false,
+            CheckPathExists = false,
+            InitialDirectory = Environment.CurrentDirectory,
+            Filter = "KSP binaries (KSP*.exe)|KSP*.exe",
+            Multiselect = false
+        };
+
+        public bool HasSelections => KSPInstancesListView.SelectedItems.Count > 0;
 
         public ChooseKSPInstance()
         {
-            manager = Main.Instance.Manager;
+            _manager = Main.Instance.Manager;
             InitializeComponent();
 
             StartPosition = FormStartPosition.CenterScreen;
 
-            browseKspFolder = new FolderBrowserDialog();
-
-            if (!manager.Instances.Any())
+            if (!_manager.Instances.Any())
             {
-                manager.FindAndRegisterDefaultInstance();
+                _manager.FindAndRegisterDefaultInstance();
             }
 
             UpdateInstancesList();
-
-            SetButtonsEnabled(false);
+            UpdateButtonState();
         }
 
         private void UpdateInstancesList()
         {
-            SetButtonsEnabled(false);
             KSPInstancesListView.Items.Clear();
+            UpdateButtonState();
 
-            foreach (var instance in manager.Instances)
+            foreach (var instance in _manager.Instances)
             {
                 var item = new ListViewItem { Text = instance.Key, Tag = instance.Key };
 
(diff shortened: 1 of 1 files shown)


politas:
Nope. No .exe on Linux installations. Filenames are "KSP.x86" and "KSP.x86_64". Not sure about MacOS installs.

ayan4m1:
`readme.txt` it is...

politas:
There's going to be a lot of readme.txt files around that aren't indicative of a KSP instance, though. Could we just search for KSP*.* instead?

ayan4m1:
Yep, sounds good.

politas:
Ok, that's working great on Linux. I wonder if @huslage or @100ideas could let us know about the executable filenames on MacOS?

politas:
Turns out the MacOS binary is <gameroot>/KSP.app/Contents/MacOS/KSP
Maybe we should use `buildID*.txt` as the target instead. 'buildID.txt' is present in <gameroot> for both Linux and MacOS, and I think Windows should always have either 'buildID.txt' or 'buildID64.txt'?

ayan4m1:
Done - also fixed a little exception if nothing was selected in the instance list when you clicked "Select" to close the dialog.

politas:
Ok, I tried hitting Select with nothing selected, and got an exception:

`Unhandled exception:
  at CKAN.Main..ctor (System.String[] cmdlineArgs, CKAN.GUIUser user, System.Boolean showConsole) [0x00137] in <bb5821c9a2594cc4b6f5aef2f5b47398>:0 
  at (wrapper remoting-invoke-with-check) CKAN.Main:.ctor (string[],CKAN.GUIUser,bool)
  at CKAN.GUI.Main_ (System.String[] args, System.Boolean showConsole) [0x00057] in <bb5821c9a2594cc4b6f5aef2f5b47398>:0 
  at CKAN.CmdLine.MainClass.Gui (CKAN.CmdLine.GuiOptions options, System.String[] args) [0x00008] in <bb5821c9a2594cc4b6f5aef2f5b47398>:0 
  at CKAN.CmdLine.MainClass.Main (System.String[] args) [0x000a1] in <bb5821c9a2594cc4b6f5aef2f5b47398>:0  `

politas:
Didn't we have the Select button greyed out when no instance is selected before?

ayan4m1:
> ForgetButton.Enabled = RenameButton.Enabled = SelectButton.Enabled = SetAsDefaultCheckbox.Enabled = _manager.Instances.Count > 0;

Only gets disabled if there are actually zero rows - there can be rows and none selected.

Very odd exception, and I don't get the same so it's probably OS-related...

politas:
The code you've replaced set button availability based on whether there was a selection, though.

GUI/ChooseKSPInstance.cs
```diff
@@ -76,55 +84,63 @@ private void SelectButton_Click(object sender, EventArgs e)
...
+                _manager.RemoveInstance(instance);
+                UpdateInstancesList();
+            }
         }
 
-        private void SetButtonsEnabled(bool has_instance)
+        private void UpdateButtonState()
         {
-            ForgetButton.Enabled = RenameButton.Enabled = SelectButton.Enabled = SetAsDefaultCheckbox.Enabled = has_instance;
+            ForgetButton.Enabled = RenameButton.Enabled = SelectButton.Enabled = SetAsDefaultCheckbox.Enabled = _manager.Instances.Count > 0;
```
politas: If you use hasSelections here, the buttons are greyed out correctly, and I don't get any exceptions

`ForgetButton.Enabled = RenameButton.Enabled = SelectButton.Enabled = SetAsDefaultCheckbox.Enabled = HasSelections;`

Issue #1941, opened 2016-11-16
Failed to connect; trying to write to file


Blu3wolf:
The error in question is as follows:

Failed to connect to repository. Exception: Trying to write E:/Program Files (x86)/Steam/steamapps/common/Kerbal Space Program/GameData/UmbraSpaceIndustries/FX/bubble.png but it already exists.

Was on StarTram, have just gone through auto update to Alexey. 

W7 SP1 x64

Exception occurred after going through the auto updater. 

Have not tried for reproduction just yet.

Blu3wolf:
restarting CKAN with a clean install seems to result in no error, so its likely some interaction with older mods being installed?

politas:
I think that file got moved from being installed by one mod to being in a dependency, instead.

bdew:
I've got the same error message but for a different file (GameData\REPOSoftTech\RSTKSPUtils\RSTKSPGameEvents.dll).

Why is CKAN trying to write to files before i even told it to do anything?

Dazpoet:
Maybe a solution to problems like these, us changing the metadata and users being caught in the crossfire, could be that we make a habit of doing an epoch-bump when we change things? That way users will be prompted for an update to what is essentially the same mod but without this issue.

politas:
That'll make for a lot of epoch bumps. Maybe we should hide the epochs from the UI unless a config option is turned on.

Dazpoet:
I would wholeheartedly agree to such a thing, I like my versionnumbers tidy and I guess I'm not alone.

PR #1942, opened 2016-11-18
Hide Epochs in modlist by default

politas:
Changes the main Modlist to use versions  of InstalledVersion and LatestVersion that hide the epoch by default. Added checkbox in settings dialog to allow users to turn this setting off. Epochs still appear in the Mod info panel

ref #1941


politas:
@ayan4m1 , I have a question - Are conditional operators sufficiently faster than if ... then statements to justify the code obfuscation?

PR #1959, opened 2016-12-17
Support modifying KSP compatibility on the command line

dbent:
@politas Support modifying KSP compatibility from the CLI.

Also fixes a bug where compatibility would be too permissive due to unitialized `version` field in `KSP`.

### DIFF
--- a/Cmdline/Action/CompatSubCommand.cs
+++ b/Cmdline/Action/CompatSubCommand.cs
@@ -145,10 +145,10 @@ namespace CKAN.CmdLine.Action
             [VerbOption("list", HelpText = "List compatible KSP versions")]
             public CompatListOptions List { get; set; }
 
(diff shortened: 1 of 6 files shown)


politas:
Ah, @dbent, you are a wonderful human being!

Cmdline/Action/CompatSubCommand.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            return exitCode;
+        }
+
+        private class CompatOptions : CommonOptions
+        {
+            [VerbOption("list", HelpText = "List compatible KSP versions")]
+            public CompatListOptions List { get; set; }
+
+            [VerbOption("add", HelpText = "List compatible KSP versions")]
```
politas: Some more helpful help text would be good

dbent:
Done.

Issue #1840, opened 2016-07-25
Move UI from Winforms to GTK#


politas:
It seems we have many problems with our UI, especially on MacOSX. The Mono Winforms implementation is not being actively maintained, and there are many reports of limitations with its cross-platform capabilities and performance.

[Eto.Forms](https://github.com/picoe/Eto) looks to be a good replacement option. The latest release was last December, there are a small number of recently opened issues and recent activity on the GitHub. 

Discussion has settled on GTK# as a better option. My own experiments shows it to be much easier to work with.

ayan4m1:
Eto is a dying project unfortunately, there has been very little activity this year. Unfortunately, [GTK#](https://github.com/mono/gtk-sharp) is in a similar state though it is being used by several large open-source apps and it is directly under the Mono umbrella. All that said, I agree with the idea that an app with Win/Mac/Lin support will need to use a non-Winforms UI library to ensure a high quality app.

politas:
I got the impression that GTK# wasn't doing a great job with OSX?

ayan4m1:
Well, you're right in that it doesn't provide a "native" visual experience, bug-wise I don't have any information, and it should be feature-complete across all three platforms (Mono GTK# explicitly supports OSX)

ayan4m1:
Ultimately to really do things right on each platform, we'd need to refactor the entire Data <-> UI and create a new layer which would allow per-platform UI code to run. Not a bad idea imo, but a lot of work and it's best to have a clean base before starting something like that.

politas:
I'd be happy with a somewhat imperfect UI that works across all platforms. No one's paying for CKAN, after all.

rspeed:
Seconding GTK# as a solution. Not only would provide a much better experience across all non-Windows platforms, GTK+ is a mature and well-supported framework. So unlike using another mono-specific solution, there is little risk of future abandonment.

Notably, there is also a potential to build an app bundle which contains all of CKAN's dependencies. That would eliminate the need to install mono, or to use the command line at all. Though I'm not sure whether or not gtk-mac-bundler works with GTK#.

100ideas:
What about using [MonoMac.Windows.Forms](https://github.com/Clancey/MonoMac.Windows.Form)?

> "Monomac.Windows.Forms is a project to allow users to reuse 90% of winform code designed on windows to produce an equivalent application for a mac with a native Cocoa UI. (Even Designer generated code!)"

politas:
[...]

PR #1961, opened 2016-12-17
display all mod versions

grzegrzk:
Hi,

I have added tab that displays all versions of currently selected mod. It is displayed on ModInfo pane on the right - next to "Contents" tab.

It is displayed purely for informational reasons - user cannot force particular version installation or something like that.

Mod versions compatible with current KSP are highlighted.

About changes in code:
- I have added new UserControl: MainAllModVersions 
- I have refactored Main.Designer.cs & MainModInfo.cs.  Now MainModInfo.cs is separate UserControl instead of being partial of Main class. I was unable to layout controls properly without Visual Designer and when I try to use it on Main form it destroys layout. Now MainModInfo can be used with visual designer so it is easier to work with it.


politas:
I didn't realise this was a possibility! Now I'm wondering if MainModList can similarly be separated out into a separate file to make the different UI elements easier to work with. I may have to get myself a Windows VM so I can load up VisualStudio

grzegrzk:
> I didn't realise this was a possibility! Now I'm wondering if MainModList can similarly be separated out into a separate file to make the different UI elements easier to work with. I may have to get myself a Windows VM so I can load up VisualStudio

I think it should be not a problem to separate each tab of main window into separate UserControl. Unfortunately I do not know any way to automate it so it requires some manual work. The benefit is that if somebody messes with .design.cs file preventing it to be modified in Visual Studio designer in the future the mess will be only in single file so it will be easier to fix :).

politas:
Indeed. Unfortunately, this work can only be done by someone working in Visual Studio. I'd love to know how Visual Studio somehow remembers things that aren't in the designer.cs files, and why hand-coding them causes such problems

We do have a plan to move to a GTK# interface for improved cross-platform support at some stage (#1840), which as far as I can tell, involves completely recreating the UI. And I'm not sure that there is a Windows visual editor tool for GTK#. You seem far more familiar with UI development than most, so if you could weigh in on #1840, I'd appreciate it.

techman83:
@politas try vscode. Ms has released a Linux version.

politas:
>try vscode. Ms has released a Linux version.

As I understand it, that has none of the GUI editing stuff from Visual Studio.

grzegrzk:
@politas 
Thx for merging.

> I'd love to know how Visual Studio somehow remembers things that aren't in the designer.cs files, and why hand-coding them causes such problems

I have been googling for this for a while and it seems that it doesn't remember things in any special place. It just tries to parse .design.cs files and arrange controls in Visual Designer basing on this parsing. It has some limited possibilities to understand some corner cases, though, so if somebody edits .design.cs files by hand it may be confused and mess things up. The only advice that I have found how to solve it is "do not edit .design.cs files by hand". Of course this is very difficult when developers work on diverse platforms.

>We do have a plan to move to a GTK# interface for improved cross-platform support at some stage (#1840), which as far as I can tell, involves completely recreating the UI. And I'm not sure that there is a Windows visual editor tool for GTK#. You seem far more familiar with UI development than most, so if you could weigh in on #1840, I'd appreciate it.

I will try to investigate it a little bit and provide some input but it is quite difficult topic.

politas:
I've worked out what is causing the designer to revert the column name!

    this.UpdateCol.Name = "Update";

That will always revert the column and leave an unused "this.UpdateCol".
https://github.com/KSP-CKAN/CKAN/blob/master/GUI/Main.Designer.cs#L472-L478

PR #1964, opened 2016-12-24
Introducing IGameComparatorExemptions in order to waive checking version compatibility for select modules.

Rusk85:
Added way of waving through certain modules in terms of version compatibility by pronouncing them compatible to each and every KSP version. This was done due to Strategia missing from CKAN and its dependency CustomBarnKit officially not being supported by the latest build of KSP while in fact it is. The way this waving process selects eligible modules is hardcoded but can be easily made more dynamic.

Also you might want to check why the mod "Strategia" is not appearing in CKAN. According to your ckan-Filerepo it should be there. At least its part of your repo ->


Rusk85:
Didn't push the project file which contains the info about the missing type. I'll fix and push anew.

Issue #1967, opened 2016-12-30
When selecting a mod, all other mods go black.  When unselecting the same mod, stays black


linuxgurugamer:
Steps to reproduce:
1. Start CKAN
2. add "tac" to filter by mod name
3.  Select Filter(compatible)
4.  Select 1.2 in compatible ksp versions
5.  Select TAC Life Support (TACLS)

See the images on the following images:

![ckan1](https://cloud.githubusercontent.com/assets/1630166/21568975/c2af4e5c-ce86-11e6-8b5b-302667e8416a.png)
![ckan2](https://cloud.githubusercontent.com/assets/1630166/21568976/c6089acc-ce86-11e6-96fe-efec2e25fb7f.png)
![ckan3](https://cloud.githubusercontent.com/assets/1630166/21568977/c83ed392-ce86-11e6-939a-5a505330101e.png)

linuxgurugamer:
No need to select anything other than the current in the compatible ksp versions

Rohaq:
This is happening to me, and seems to occur when the selected mod conflicts with another mod (see attachment). Can you click away from the TAC mod and confirm it's background has turned red?

I think it's related to this change, which uses SystemColors..ControlText instead of Color.White: https://github.com/KSP-CKAN/CKAN/pull/1926/files?diff=unified - Running the following powershell module shows this as black on my system: https://gallery.technet.microsoft.com/scriptcenter/Get-SystemColors-using-WPF-d7c31a8c

![6urwgxa 1](https://cloud.githubusercontent.com/assets/246865/21571190/b130cc6c-cec4-11e6-9639-f935c7ff8ca6.png)

linuxgurugamer:
No, that does not happen.

But it is a conflict, USI-LS has a conflict with TACLS

politas:
Right, yes, I can reproduce this on Linux when selecting conflicting mods, too.

PR #1926, opened 2016-10-23
Use SystemColors to source various UI colors

ayan4m1:
Should fix #1011, but needs some in-the-field testing because not all of the colors are necessarily mapped well under the common dark themes.

PR #1968, opened 2016-12-30
Fix for black rows in GUI on conflicting mods.

Rohaq:
Fixes issue #1967 by setting row background color to "Color.Empty" - the default color value for this object.

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -125,14 +125,14 @@ namespace CKAN
                 }
                 else
                 {
-                    if (row.DefaultCellStyle.BackColor != SystemColors.ControlText)
+                    if (row.DefaultCellStyle.BackColor != Color.Empty)
                     {
                         foreach (DataGridViewCell cell in row.Cells)
                         {
                             cell.ToolTipText = null;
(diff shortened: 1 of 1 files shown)


politas:
Merged with huge thanks for tracking that down! I probably would have changed it back to Color.White it had before. Color.Empty is a much better choice.

Rohaq:
Made more sense to set it to the default value, since I couldn't see it set elsewhere!

If you do plan to use SystemColors for stuff, it might be worth being explicit in future for controls if it enables support of system theming properly - though I don't know how well that tends to work cross-platform. It would definitely need testing!

politas:
Change #1926 was trying to improve our support for system themes, rather than direct colour choices. From that perspective, Empty is a great choice, I think. Trying to get the Menubars to match the rest of the UI is a headache under Linux!

linuxgurugamer:
Any chance to get a quick release with this fix, it makes the UI almost unusable for fear of a conflict.

Rohaq:
Just add "New Years Edition" to the end ;)

Issue #1992, opened 2017-01-17
Improve ModuleNotFoundKraken error message to specify the missing module and version


MBradbury:
I've been trying to track down whats gone wrong in my setup that is leading CKAN to throw a ModuleNotFoundKraken on startup. To help debug this I've ended up compiling CKAN with a modified ModuleNotFoundKraken that passes the module string and version to the reason argument of Kraken.

Perhaps it would be better to default to setting this to the following every time the exception is created. `string.Format("Module {0} of version {1} is not available", module, version)`

I was getting this exception from CKAN.Registry.AllAvailable which creates a new ModuleNotFoundKraken without setting the reason. The other option might be to set the reason field here.

politas:
Can you branch and PR to the main CKAN master? That sounds like an excellent change!

MBradbury:
Done, I've also created an issue for the problem I was trying to fix.

PR #1993, opened 2017-01-17
Always provide a string error message for ModuleNotFoundKraken

MBradbury:
As per issue #1992.


politas:
This is not the right way to do this, since the message is not always relevant.  Eg [here](https://github.com/KSP-CKAN/CKAN/blob/master/Core/ModuleInstaller.cs#L1022-L1029), and [here](https://github.com/KSP-CKAN/CKAN/blob/master/Core/Registry/Registry.cs#L498-L509)

Issue #2009, opened 2017-02-23
Proposal: replace "Filter" button icon in the GUI from a looking glass to a filter


JoshKeaper:
A looking glass icon is usually associated with search functionality, rather than filtering (I've never thought until recently that I could actually list all mods, including incompatible; I've just overlooked this function because of the "search" icon). Replacing it with an actual filter icon will make things clearer.

Issue #2007, opened 2017-02-16
Search Boxes out of order when pressing tab. 


Marsfan:
Just a minor aesthetic issue, but the search boxes (mod name, author name, description) are out of order when pressing tab. If one is in the mod name box, pressing tabs goes to description, then jumps back to author name. 

CKAN 1.22.1
Windows 10 64bit

politas:
Slight complexity in fixing this due to the weird reversion bug we have every time someone uses the VS interface editor on the main window. It should be a simple fix, and even the reversion bug is only a code compilation warning that doesn't have an end-user impact, so this should be relatively easy to fix.

PR #2018, opened 2017-03-09
WIP -  GUI fixes 

politas:
Finally have the VS editor properly using UpdateCol instead of Update for the comumn name, so we stop getting build warnings.

Also changed the Filter icon from a search magnifier to a new filter image I created. (I refuse to accept the funnel = filter idea) (#2009)
filter

I have made an attempt to fix the tab order (#2007), but that doesn't seem to be working, so I'll have to do some more work on that.


politas:
Nope, this is all broken and messed up. Killing this branch and starting again.

ayan4m1:
Hey, I have been switching jobs the last few months but mean to get back to CKAN as soon as I can. This seems like a target for discussion maybe, I will try and get on IRC this week to discuss further.

PR #2023, opened 2017-03-17
Switch to all C# native mod downloads

### LINKED ISSUE #840: Download speed requires moving mouse
MacOS X 10.10.3, version 1.6.12 - GUI. 

When installing a package, download speed drops to 10kb/s or less as soon as the mouse stops moving. Keeping the mouse continuously in motion results in expected download rates. Also reported to affect linux in the forums but not personally tested.

politas:
Improvements in Mono and the downfall of kerbalstuff appears to have removed the necessity for curl downloads.
Closes #840

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/packages/
 /_build/
 /.vs/
 Screenshots/
--- a/Core/Net/NetAsyncDownloader.cs
+++ b/Core/Net/NetAsyncDownloader.cs
@@ -83,7 +83,14 @@ namespace CKAN
                 downloads.Add(download);
             }
 
-            DownloadNative();
+            // adding chicken bits
+            if (Platform.IsWindows || System.Environment.GetEnvironmentVariable("KSP_CKAN_USE_CURL") != null) {
+                DownloadNative();
+            }
+            else
+            {
+                DownloadCurl();
+            }
 
         }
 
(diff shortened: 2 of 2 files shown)


politas:
If this works, I'll take out the rest of the curlsharp-related code.

politas:
How robust is that from a multi-platform perspective? Can MacOS users set environment variables in a way that persists?

pjf:
> How robust is that from a multi-platform perspective? Can MacOS users set environment variables in a way that persists?

Apparently it is, but it looks like it's a [right pain when launching from spotlight](http://stackoverflow.com/questions/135688/setting-environment-variables-in-os-x). If they're starting CKAN from the shell then it's easy enough to put them in their `~/.profile`, or just on the cmdline with `KSP_CKAN_USE_CURL=1 mono ckan.exe`.

In any case the suggestion of an environment variable is to give users a short-term chance to back out of the change, rather than an enduring one that always has to be set. If lots of users report that after the update they can't download things then we're back to figuring out why curl is slow, or coming up with creative solutions (like making requests with the native implementation always, but then doing a fallback to curl if they throw recoverable errors).

rspeed:
> but it looks like it's a right pain when launching from spotlight

In recent versions of macOS it's just `launchctl setenv [key] [value]`, though you still have to log out.

PR #2027, opened 2017-04-03
Enable checking against a schema

Olympic1:
> From https://github.com/KSP-CKAN/CKAN-core/pull/133

Rebased PR #992

### DIFF
--- a/Core/CKAN.schema
+++ b/Core/CKAN.schema
@@ -1,0 +1,423 @@
+{
+    "$schema" : "http://json-schema.org/draft-04/schema#",
+    "title" : "CKAN JSON Schema",
+    "description" : "Schema for validating CKAN meta files",
+    "type" : "object",
+    "properties" : {
+        "spec_version" : {
+            "description" : "Version of the CKAN spec this document uses",
+            "oneOf" : [
+                {
+                    "type"        : "integer",
+                    "description" : "The integer '1' is special case",
+                    "minimum"     : 1,
+                    "maximum"     : 1
+                },
+                {
+                    "type"        : "string",
+                    "description" : "A vx.x string",
+                    "pattern"     : "^v[0-9]+\\.[0-9]+$"
+                }
+            ]
+        },
+        "name" : {
+            "description" : "Human readable name of the mod",
+            "type"        : "string"
+        },
+        "identifier" : {
+            "description" : "Unique identifier for this mod.",
+            "$ref"        : "#/definitions/identifier"
+        },
+        "kind" : {
+            "description" : "Package type, defaults to package.",
+            "enum"        : [ "package", "metapackage" ]
+        },
+        "abstract" : {
+            "description" : "Short description of the mod",
+            "type"        : "string"
+        },
+        "description" : {
+            "description" : "Longer description of the mod",
+            "type"        : "string"
+        },
+        "comment" : {
+            "description" : "Free-form comment. Not displayed to users.",
+            "type"        : "string"
+        },
+        "author" : {
+            "description" : "Author, or list of authors.",
+            "oneOf" : [
+                { "type" : "string" },
+                {
+                    "type"        : "array",
+                    "items"       : { "type" : "string" },
+                    "uniqueItems" : true
+                }
+            ]
+        },
+        "download" : {
+            "description" : "URL where mod can be downloaded by tools",
+            "type"        : "string",
+            "format"      : "uri"
+        },
+        "download_size" : {
+            "description" : "The size of the download in bytes",
+            "type"        : "integer"
+        },
+        "download_hash" : {
+            "description" : "A object of hashes of the downloaded file",
+            "type"        : "object",
+            "properties" : {
+                "sha1" : {
+                    "description" : "SHA1 hash of the file",
+                    "type"        : "string"
+                },
+                "sha256" : {
+                    "description" : "SHA256 hash of the file",
+                    "type"        : "string"
+                }
+            }
+        },
+        "download_content_type" : {
+            "description" : "The content type of the download",
+            "type"        : "string"
+        },
+        "license" : {
+            "description" : "Machine readable license, or array of licenses",
+            "$ref"        : "#/definitions/licenses"
+        },
+        "version" : {
+            "description" : "Version of the mod. Drop the leading v",
+            "$ref"        : "#/definitions/version"
+        },
+        "release_status" : {
+            "description" : "Optional release status",
+            "enum"        : [ "stable", "testing", "development" ]
+        },
+        "ksp_version" : {
+            "description" : "Optional target KSP version",
+            "$ref"        : "#/definitions/ksp_version"
+        },
+        "ksp_version_min" : {
+            "description" : "Optional minimum KSP version",
+            "$ref"        : "#/definitions/ksp_version"
+        },
+        "ksp_version_max" : {
+            "description" : "Optional maximum KSP version",
+            "$ref"        : "#/definitions/ksp_version"
+        },
+        "ksp_version_strict" : {
+            "description" : "Optional enforcement of strict version checks (defaults to false)",
+            "type"        : "boolean"
+        },
+        "tags" : {
+            "description" : "A series of descriptive keywords to provide quick filtering and organization",
+            "type"        : "array",
+            "items" : {
+                "type" : "string",
+                "pattern" : "^[a-z0-9-]+$"
+            }
+        },
+        "depends" : {
+            "description" : "Optional list of dependencies",
+            "$ref"        : "#/definitions/relationship"
+        },
+        "recommends" : {
+            "description" : "Optional list of recommended mods",
+            "$ref"        : "#/definitions/relationship"
+        },
+        "suggests" : {
+            "description" : "Optional list of recommended, but not essential mods",
+            "$ref"        : "#/definitions/relationship"
+        },
+        "supports" : {
+            "description" : "Optional list of supported mods",
+            "$ref"        : "#/definitions/relationship"
+        },
+        "conflicts" : {
+            "description" : "Optional list of conflicting mods",
+            "$ref"        : "#/definitions/relationship"
+        },
+        "provides" : {
+            "description" : "A list of virtual packages this mod provides",
+            "type"        : "array",
+            "items"       : { "type" : "string" },
+            "uniqueItems" : true
+        },
+        "resources" : {
+            "description" : "Additional resources",
+            "type" : "object",
+            "properties" : {
+                "homepage" : {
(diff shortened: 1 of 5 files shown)


politas:
@dbent , you're the Cake expert. Can you figure out what's happening here? These .csproj files baffle me.

dbent:
Well we still need to consider if we want to do this. Json.NET Schema is not permissively licensed, but is licensed under [AGPL-3.0](https://www.gnu.org/licenses/agpl-3.0.html) which is copyleft. This means the combined work (ckan.exe, netkan.exe, etc.) has to be licensed under the AGPL. Our source code is MIT licensed which is (A)GPL compatible so that doesn't have to change though.

There's a few things going on in this PR with regards to the build, specifically updating versions of libraries. I'll make a separate PR for that.

Olympic1:
I already thought that it had to do something with the lib updates.

politas:
That doesn't take away the MIT license from the CKAN code, does it? This whole question of multiple licenses gets pretty confusing. Would it mean that someone could fork CKAN and claim only the AGPL applies to their new fork?

Actually, aren't we already including some copyleft packages?

dbent:
IANAL, but here's my understanding:

> That doesn't take away the MIT license from the CKAN code, does it?

No MIT aka Expat license, is ["GPL-Compatible"](https://www.gnu.org/licenses/license-list.html#GPLCompatibleLicenses). GPL-Compatible source can be combined with (A)GPL source to create a derived work that doesn't force the GPL-Compatible soure to be released under the (A)GPL. The derived work (ckan.exe, netkan.exe, etc) however must be.

What this means, is that if someone wants to use ckan.exe or netkan.exe as a library (which is totally possible in .NET), *their* work would have to released under the (A)GPL. To avoid that, they'd have to take our MIT source, excise any bits that refer to (A)GPL code and create a custom build derived only from MIT source. (This could be automated through the use of `#ifdef` in the source and a bit smarter build logic to avoid ILRepacking Json.NET Schema). What's more, the AGPL (and its primary difference with respect to the GPL) is that even use of an application over a network is considered distribution and thus triggers its provisions. So a web application using ckan.exe/netkan.exe would have to be distributed under AGPL.

Now, this person might also be able to avoid that by calling ckan.exe/netkan.exe as a separate process, but if the program is ["specifically designed to require"](https://www.gnu.org/copyleft/gpl.html) ckan.exe/netkan.exe then the provisions probably apply *anyway*.

> Would it mean that someone could fork CKAN and claim only the AGPL applies to their new fork?

No, but it probably means that different source files would have to be under different licenses. Anything that uses JSON.NET Schema would have to be relicensed under AGPL (and relicensing requires consent of all contributors to that file, since CKAN has no copyright assignment to a single individual/organization so every author owns the copyright to their own contributions).

Now that I think about, even excising code could be problematic. Since if the entire file is AGPL, then removing code just makes a new derived work that would also have to be AGPL. Can copyright be specified on a line-by-line basis? Hmmm.

> Actually, aren't we already including some copyleft packages?

Nope. I went through each and every library we use, including ones only used for testing and all except one are licensed under permissive (non-copyleft) MIT, Apache, or BSD licenses. The only one that isn't licensed under a permissive license is ICSharpCode.SharpZipLib.Patched but its licensed under the GPL with an explicit linking exception making it similar to the LGPL.

- Autofac:                            MIT (permissive)
- Castle.Core:                        Apache-2.0 (permissive)
- ChinhDo.Transactions.FileManager:   MIT (permissive)
- CommandLineParser:                  MIT (permissive)
- ICSharpCode.SharpZipLib.Patched:    GPL w/ Linking Exception (semi-permissive)
- ini-parser:                         MIT (permissive)
- log4net:                            Apache-2.0 (permissive)
- Moq:                                BSD 3 Clause (permissive)
- Newtonsoft.Json:                    MIT (permissive)
- NUnit:                              MIT (permissive)
- TxFileManager:                      MIT (permissive)
- CurlSharp:                          BSD 3 Clause (permissive)

The whole issue with the AGPL is a minefield of technicalities, which is why I've never bothered to incorporate JSON.NET Schema into NetKAN's validation.

FYI I've made #2028 which updates the build and all our libraries to the latest and greatest. If we want to pursue this, it should be probably be based off that.

dbent:
Also, the AGPL version of JSON.NET Schema is listed as ["Free with limitations (1000 validations per hour)"](http://www.newtonsoft.com/jsonschema). I'm not entirely sure what that means.

Olympic1:
Well, in that case it would be better to find another way to check the spec or just drop it.

techman83:
To note, both the indexer and Jenkins run the resulting output through a schema validator. So what makes it CKAN-meta has been validated.

It would probably be handy to have in the client for those that make custom ckans or provide early signs that a netkan will fail CI, but overall I don't think the need is exceptionally high. A nice to have for sure, but it's not really a big deal in the scheme of things.

dbent:
[NJsonSchema](https://github.com/NJsonSchema/NJsonSchema) is another project that seems active and is permissively (MIT) licensed.

Dazpoet:
Don't we have someone *cough* @pjf *cough* who loves discussing licenses and how they work and not work?

techman83:
@Dazpoet I think it's more @pjf is exceptionally knowledgeable on the topic and has a passion for ensuring it's done correctly. If there is an MIT based plugin lets use it. Also the restriction added by JSON.NET schema may not be applicable:

https://www.gnu.org/licenses/agpl-3.0.en.html
> All other non-permissive additional terms are considered "further restrictions" within the meaning of section 10. If the Program as you received it, or any part of it, contains a notice stating that it is governed by this License along with a term that is a further restriction, you may remove that term.

Also AGPL is quite at the opposite end of the spectrum when compared to MIT, I'm not super across how it would effect things when used as a library though and we'd want to be very sure we're across the implications. It would certainly effect the ability for it's use in a commercial product, so if there is an MIT licensed project available, lets use that. But we're not in a rush for such a feature and we'd need to do some indexer work if NetKAN started throwing it's own schema exceptions.

ayan4m1:
@Olympic1 Can you resolve conflicts/rebase or close this out?

politas:
@Olympic1, can you have a look at NJsonSchema that @dbent mentioned and see if switching to that is possible?

RSuter:
Hi everybody

Im the creator of NJsonSchema.

Switching should be very simple, only problem i see in the schema is the “required” properties which have no corresponding properties. We have an open issue for that.

Other than that just load the schema with JsonSchema4.FromJsonAsync() and call Validate().

Olympic1:
@dbent, @politas, I converted the PR to NJsonSchema, but not fully.

I followed @RSuter's instructions and fixed the errors. Need to look into some more.

Olympic1:
Rebased. Have anyone had the time to test this PR? I want to move this forward.

Core/Types/CkanModule.cs
```diff
@@ -366,6 +376,33 @@ private void DeSerialisationFixes(StreamingContext like_i_could_care)
...
+            var assembly = Assembly.GetExecutingAssembly();
+            using (var stream = assembly.GetManifestResourceStream(SchemaPath))
+            using (var reader = new StreamReader(stream))
+            {
+                var result = reader.ReadToEnd();
+                MetadataSchema = JsonSchema4.FromSampleJson(result);
+            }
+        }
+
+        private static readonly string SchemaPath = "CKAN.CKAN.schema";
```
HebaruSan: Let's make this a `const` instead of `static readonly`, since the value is known at compile time.

Core/Types/CkanModule.cs
```diff
@@ -366,6 +376,33 @@ private void DeSerialisationFixes(StreamingContext like_i_could_care)
...
+        private static readonly JsonSchema4 MetadataSchema;
+
+        private static async Task<IList<string>> ValidateAgainstSchema(string json)
+        {
+            JsonSchema4 schema = await JsonSchema4.FromJsonAsync(json);
+
+            IList<string> errorList = new List<string>();
+            var errors = schema.Validate(MetadataSchema.ToJson());
```
HebaruSan: If I'm reading this correctly, we process the schema from a JSON string to a JsonSchema4 object in the static constructor, and then back from object to a JSON string in the validation function.

Can we store the return value of the ToJson function, so we don't have to pay the cost of that conversion for every CkanModule object? Presumably we can ensure that the CKAN.schema file bundled in the assembly is valid separately whenever we modify it.

Core/Types/CkanModule.cs
```diff
@@ -292,6 +295,13 @@ public CkanModule(string json, IGameComparator comparator)
         {
             _comparator = comparator;
 
+            var errors = ValidateAgainstSchema(json);
+            if (errors.Result.Any())
+            {
+                var message = "Validation against the schema failed.\n" + string.Join("\n  ", errors);
+                throw new BadMetadataKraken(null, message);
```
HebaruSan: I'd like to test the impact of this exception on the GUI. What would be a good way to do that? I tried corrupting my `registry.json` file in various ways, but I think those tests got caught by other checks (e.g., invalid licenses are already covered elsewhere).

Core/CKAN.schema
```diff
@@ -0,0 +1,423 @@
+{
```
HebaruSan: Is it impossible to move the old CKAN.schema file to this new location? It will be inconvenient to have to maintain two files identically going forward.

Core/Types/CkanModule.cs
```diff
@@ -366,6 +376,33 @@ private void DeSerialisationFixes(StreamingContext like_i_could_care)
...
+                MetadataSchema = JsonSchema4.FromSampleJson(result);
+            }
+        }
+
+        private static readonly string SchemaPath = "CKAN.CKAN.schema";
+        private static readonly JsonSchema4 MetadataSchema;
+
+        private static async Task<IList<string>> ValidateAgainstSchema(string json)
+        {
+            JsonSchema4 schema = await JsonSchema4.FromJsonAsync(json);
```
HebaruSan: This variable represents the module we're validating, not the schema, right? Let's change its name to clarify that.

Core/Types/CkanModule.cs
HebaruSan: Actually, are you sure this call is correct? The `schema` object appears to be the module, and the parameter with ToJson is the schema. This looks like we're trying to use the module to validate the schema rather than the other way around. The [documentation](https://github.com/RSuter/NJsonSchema/wiki/JsonSchema4) passes the data to be validated to the schema's Validate function.

Core/CKAN.schema
politas: It could be a symbolic link instead of a separate file, surely?

Olympic1:
Closing as this was done in #2788

PR #2034, opened 2017-04-20
Add spec/schema to implement mod tags

smattiso:
This patch is related to issue #1021

### DIFF
--- a/Spec.md
+++ b/Spec.md
@@ -378,29 +378,29 @@ This field defaults to `false`, including for `spec_version`s less than
 `v1.16`, however CKAN clients prior to `v1.16` would only perform strict
 checking.
 
-##### categories
+##### tags
 
-The `categories` field describes keywords that a user or program may use to
-classify or filter the mod in a list, but which are not required.
-One or more of either of the following fields may be used:
+The `tags` field describes keywords that a user or program may use to
+classify or filter the mod in a list, but which are not required. These
+may include general tags which define how the mod interacts with or alters
+KSP or specific tags defining what has been added or changed from stock
+gameplay. Tags may contain lowercase alphanumeric characters or hyphens.
 
-- `main` : Keywords which define how the mod interacts with or alters KSP
-- `sub` : Keywords which further describe what has been added or changed
-from stock gameplay
+Example tags:
 
-Example categories:
-
-    "categories" : [
-        { "main" : "physics" },
-        { "main" : "parts" },
-        { "sub"  : "oceanic" },
-        { "sub"  : "thermal" },
-        { "sub"  : "science" }
+    "tags" : [
+        "physics",
+        "parts",
+        "oceanic",
+        "thermal",
+        "science",
+        "my-1-custom-tag"
     ]
 
-The first defined keyword in each category field should be considered the
-most important and used to populate a sortable column, though it is intended
-that all included category fields will be indexed and filterable.
+The first defined keyword listed in the tag field which matches one of a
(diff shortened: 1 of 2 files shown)


ayan4m1:
This looks good to me, it is a simple change which can allow for more complex work to be done. Can the NetKAN guys chime in as to whether or not they like this change? @dbent @Olympic1

smattiso:
Due to my own concerns that my proposed data structure does not appear similar to other data structures in the CKAN file, I am patching this spec to include these multiple category definitions in an array. Also, as 'type' and 'description' seem like keywords, it's probably best practice to avoid using those as descriptors. I am not sure if it is possible to do it the way I originally had it, but better safe than sorry.

dbent:
My main concern is that there could be a lot of hand-wringing over categories, what they should be, which mods belong in which categories, et cetera. Ontology can be a blood sport to some people. Also, if we have a strict enumeration in the spec that means every new category would have to be accompanied by a revision of the spec. Especially problematic if we want to actually do runtime schema-checking and enforcement.

I'd much prefer quasi-freeform tags (restricted to lowercase alphanumerics plus hyphen, i.e. `^[a-z-]+$`) with a list of "standard" tags that people are encouraged (but not restricted) to using.

ayan4m1:
Agreed from me - I think the term "tag" rather than "category" is more useful to the end-user, @smattiso are you amenable to making the changes described by @dbent, as I see it mainly consist of changing "category" to "tags"?

smattiso:
Sure thing. I'll have those updates in a jiffy! ...just as soon as I figure out how to spec quasi-freeform tags... 😛

dbent:
@smattiso You'll want the `pattern` property (there are already examples of it in `CKAN.schema`) and use the pattern: `^[a-z0-9-]+$`, which is "one or more characters that are a lower case letter, number, or hyphen".

Acceptable:
* `foo`
* `foo-bar`
* `foo-bar-100`

Unacceptable:
* `Foo`
* `Foo Bar`
* `Foo Bar !!!`

smattiso:
0-9! That's what I forgot. 😮

dbent:
> 0-9! That's what I forgot. 😮 

No problem, I think I left it off when I first listed the pattern in the previous comment.

FYI, for anyone reading this in the future, I intentionally left out uppercase letters (`A-Z`) so we don't have an issue of different casing or whether or not tags should be compared case sensitively or insensitively or any of that other jazz. A tag should have one single, canonical, representation.

dbent:
Changes look good to me, but you'll want to update the `categories` property in the example file in `Spec.md`.

smattiso:
Haven't had my coffee this morning. *fixed

Spec.md
```diff
@@ -372,6 +380,30 @@ This field defaults to `false`, including for `spec_version`s less than
 `v1.16`, however CKAN clients prior to `v1.16` would only perform strict
 checking.
 
+##### tags
+
+The `tags` field describes keywords that a user or program may use to
```
politas: Please add a **(v1.23)** to indicate that this will only be supported by CKAN clients of version v1.23.0 and higher

Olympic1: We always go up by two numbers (14, 16, 18, 20, 22). So shouldn't the next be v1.24?

smattiso: Adding 1.24...

smattiso:
🙄 That's what I get for logging in with my laptop...
That last commit was mine, y'all, FWIW.

PR #2054, opened 2017-05-23
Canonicalize non-raw GitHub URIs

dbent:
Explanation in comments in code.


politas:
I'm not entirely clear on the issue this fixes. Any chance you could add a test case?

PR #2057, opened 2017-05-27
Fixes #2047

### LINKED ISSUE #2047: System.NullReferenceException with mono 5.0
**CKAN Version:** 
1.22.1-0-g174271a

**Operating System:** 
Debian GNU/Linux 8

CKAN crashes every time I press Download or Apply (with new mod selected) with error below.

Ussue seems to be caused by latest upgrade of mono to 5.0.
After I downgraded mono to 4.8 everything is fine again.

Unhandled exception:
  at (wrapper managed-to-native) CurlSharp.NativeMethods:curl_easy_perform (intptr)
  at CurlSharp.CurlEasy.Perform () [0x00007] in <db7f644cef3d44268a3a52ddab219b4b>:0 
  at CKAN.NetAsyncDownloader.CurlWatchThread (System.Int32 index, CurlSharp.CurlEasy easy, System.IO.FileStream stream) [0x00010] in <db7f644cef3d44268a3a52ddab219b4b>:0 
  at CKAN.NetAsyncDownloader+<DownloadCurl>c__AnonStorey1.<>m__1 () [0x00001] in <db7f644cef3d44268a3a52ddab219b4b>:0 
  at System.Threading.ThreadHelper.ThreadStart_Context (System.Object state) [0x00014] in <4dc8ec68b0964e099af86e50301f5f3c>:0 
  at System.Threading.ExecutionContext.RunInternal (System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, System.Object state, System.Boolean preserveSyncCtx) [0x00071] in <4dc8ec68b0964e099af86e50301f5f3c>:0 
  at System.Threading.ExecutionContext.Run (System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, System.Object state, System.Boolean preserveSyncCtx) [0x00000] in <4dc8ec68b0964e099af86e50301f5f3c>:0 
  at System.Threading.ExecutionContext.Run (System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, System.Object state) [0x0002b] in <4dc8ec68b0964e099af86e50301f5f3c>:0 
  at System.Threading.ThreadHelper.ThreadStart () [0x00008] in <4dc8ec68b0964e099af86e50301f5f3c>:0

Cyclic3:
Fixes https://github.com/KSP-CKAN/CKAN/issues/2047. I have run this on my Debian 9 machine, and it seems to fix the problem with curlsharp

### DIFF
--- a/Core/Net/NetAsyncDownloader.cs
+++ b/Core/Net/NetAsyncDownloader.cs
@@ -84,14 +84,7 @@ namespace CKAN
             }
 
             // adding chicken bits
-            if (Platform.IsWindows || System.Environment.GetEnvironmentVariable("KSP_CKAN_USE_CURL") != null) {
-                DownloadNative();
-            }
-            else
-            {
-                DownloadCurl();
-            }
-
+            DownloadNative();
         }
 
         /// <summary>


cedenday:
Looks good.

cedenday:
You could just change the expression of the if statement and add a comment though.

Cyclic3:
the else statement causes a nullreference exception

Cyclic3:
i have never committed to ckan before, so please do tell me if I did break something

politas:
Oh, bugger. I think that` != null` should be` = null`

politas:
What you're doing is removing the chicken bits, by the way.

politas:
I think I can lay some of the blame on @pjf for my stuff-up in #2023. He suggested the test with != null rather than = null.

politas:
Thanks for finding the issue, @Cyclic3!

Cyclic3:
ok, glad to have been able to help!

Cyclic3:
may I ask a question?
Why do we need libcurl if the mono natives will work? It adds an extra dependency, and doesn't seem to make a vast amount of difference to performance. 
Is there something I'm missing here?

dbent:
> may I ask a question?
Why do we need libcurl if the mono natives will work? It adds an extra dependency, and doesn't seem to make a vast amount of difference to performance.
Is there something I'm missing here?

Mono's TLS implementation is [old](http://tirania.org/blog//archive/2015/Aug.html) and doesn't support TLS 1.1, 1.2, elliptic curve cryptography, or perfect forward secrecy. In the past this meant the native Mono implementation couldn't be used to download files over TLS from KerbalStuff. This has gotten better [recently](http://www.mono-project.com/docs/about-mono/releases/4.8.0/#tls-12-support) but still can't assume that every CKAN installation can access every possible TLS URL.

Cyclic3:
Thank you. This also explains some issues I have had with my own mono programs

politas:
If no one comes asking for how to set the chicken bits between now and the next version or two, we'll probably take it out entirely.

Cyclic3:
what are chicken bits?

cedenday:
https://en.wiktionary.org/wiki/chicken_bit

Cyclic3:
thanks

Issue #1935, opened 2016-11-01
System.TimoutException (transaction timeout) for large downloads on a slower connection


buttercookie42:
Thanks for wanting to bring an issue to our attention!

Before you open a ticket, please do a quick search of the existing issues to see if it's already been brought up.

**CKAN Version:** 
1.20.1-0-gd41fe54 (beta)

**Operating System:** 
Windows 7

**The issue you are experiencing:**
After downloading has finished, instead of the mods installing I just get the unhandled exception window showing that a transaction timeout exception (System.TimoutException) has occured. Luckily the mods do get downloaded into the cache, so the next installation attempt succeeds.

**How to recreate this issue:** 
Download any largeish mod (e.g. Stock Visual Enhancements) on a none too fast internet connection (2 Mbps in my case). It seems like the transaction timeouts currently configured in CKAN might not take account of the fact that in those cases downloads can legitimately take more than a few minutes.

ayan4m1:
We set the timeout [here](https://github.com/KSP-CKAN/CKAN/blob/master/Core/CkanTransaction.cs#L16) to be the maximum time limit proscribed by the transaction library itself. The issue is [documented](https://blogs.msdn.microsoft.com/dbrowne/2010/06/03/using-new-transactionscope-considered-harmful/) (see, MS told us to do it that way 🙉) [elsewhere](http://support.objectivity.com/sites/default/files/docs/objy/R11_0_0/html/csharp/html/d6957adb-fa07-45d5-8433-a9ab602afdec.htm) I agree with the idea that a single CKAN transaction could take over 10 minutes and still complete successfully. Ultimately, we can either hammer a huge number in there or we can make it configurable somewhere. Paging @politas to weigh in on this.

politas:
Spacedock has been having some issues recently, so I would suspect the problem is there. Have you had any issues with mods on Curse or GitHub?

buttercookie42:
@politas
Isn't Stock Visual Enhancements already downloaded from Github? At least all the relevant links are pointing to it being hosted on Github now.

[...]

PR #2073, opened 2017-06-14
Move downloads outside of gui transaction

archer884:
#1935 mentions a timeout exception impacting only the GUI on large and/or slow downloads (taking longer than ten minutes). I believe this can be solved by moving the download process outside of the transaction begun in the GUI (there are additional transactions started inside Core which are not impacted by download times).

Of course, this is a terrible idea if changes to the cache are intended to be covered by the transaction, but I have not dug into the cache code deeply enough to know if this is the case. Thoughts?

To minimize the changes required (and thereby maximize my odds of not breaking anything), I have created the new download workflow *in addition to* the old one rather than replacing the old one. This is largely because the old one is in use in lots of different places throughout the GUI code--for additions and updates as well as bulk installs. This PR pretty much just overhauls bulk installs, since they exhibited the problem for me when I tried to update for 1.3 the other day.

I have just used a build based on this to update my KSP install, so it works. On my machine. We all know what that's worth, surely. :)

First PR here. If I have goofed on anything, I apologize.

### DIFF
--- a/Core/Types/CkanModule.cs
+++ b/Core/Types/CkanModule.cs
@@ -288,11 +288,6 @@ namespace CKAN
         {
             _comparator = comparator;
 
-            if (!validate_json_against_schema(json))
-            {
-                throw new BadMetadataKraken(null, "Validation against spec failed");
-            }
-
             try
             {
                 // Use the json string to populate our object
@@ -362,54 +357,6 @@ namespace CKAN
                 throw new InvalidModuleAttributesException("ksp_version mixed with ksp_version_(min|max)", this);
             }
 
-            if (license == null)
-            {
-                license = new List<License> { new License("unknown") };
-            }
-
-            if (@abstract == null)
-            {
-                @abstract = "";
-            }
-
-            if (name == null)
-            {
-                name = "";
-            }
-        }
-
-        private static bool validate_json_against_schema(string json)
-        {
-
-            log.Debug("In-client JSON schema validation unimplemented.");
-            return true;
-            // due to Newtonsoft Json not supporting v4 of the standard, we can't actually do this :(
-
-            //            if (metadata_schema == null)
-            //            {
-            //                string schema = "";
-            //
-            //                try
-            //                {
-            //                    schema = File.ReadAllText(metadata_schema_path);
-            //                }
-            //                catch (Exception)
-            //                {
-            //                    if (!metadata_schema_missing_warning_fired)
-            //                    {
-            //                        User.Error("Couldn't open metadata schema at \"{0}\", will not validate metadata files",
-            //                            metadata_schema_path);
-            //                        metadata_schema_missing_warning_fired = true;
-            //                    }
-            //
-            //                    return true;
-            //                }
-            //
-            //                metadata_schema = JsonSchema.Parse(schema);
-            //            }
-            //
-            //            JObject obj = JObject.Parse(json);
-            //            return obj.IsValid(metadata_schema);
             license = license ?? new List<License> { License.UnknownLicense };
             @abstract = @abstract ?? string.Empty;
             name = name ?? string.Empty;
--- a/Core/Types/CkanModule.cs
+++ b/Core/Types/CkanModule.cs
@@ -410,6 +410,9 @@ namespace CKAN
             //
             //            JObject obj = JObject.Parse(json);
             //            return obj.IsValid(metadata_schema);
+            license = license ?? new List<License> { License.UnknownLicense };
+            @abstract = @abstract ?? string.Empty;
+            name = name ?? string.Empty;
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


archer884:
Cleaning out PR list.

politas:
The concept seems sound to me. I don't have a good understanding of that section of the code.

archer884:
@politas It's been awhile since I looked, but, from memory...

Basically, a transaction is begun as soon as install is initiated, and during that transaction (which has a 10 minute maximum timeout) we have to download and install all the mods. This usually isn't a problem, but several things can then cause the transaction to time out: lots of mods, a single mod with a large download, or a single download from a slow-ass server.

This patch causes the download process to take place before the transaction is actually started. As noted above, this risks causing changes to the download cache not to be covered by the transaction, but I don't know if the download cache is meant to be covered; I kind of just assumed it's not.

politas:
The only issue I can see is an incomplete download getting cached and poisoning further attempts?

archer884:
@politas So the download cache is transacted?

politas:
I believe we add any successful downloads to the cache regardless of the transaction success, and it looks like Core/Net/NetAsyncModulesDownloader method ModuleDownloadsComplete only adds completed downloads to the cache, so taking the downloads out of the transaction should be ok. I note that you tested this successfully; what platform do you use?

archer884:
That's good to know about the downloads. My system is Windows 10.

Core/Types/CkanModule.cs
```diff
@@ -362,54 +356,9 @@ private void DeSerialisationFixes(StreamingContext like_i_could_care)
...
-            //
-            //                    return true;
-            //                }
-            //
-            //                metadata_schema = JsonSchema.Parse(schema);
-            //            }
-            //
-            //            JObject obj = JObject.Parse(json);
-            //            return obj.IsValid(metadata_schema);
+            license = license ?? new List<License> { License.UnknownLicense };
```
politas: If I understand this correctly, this sets licence to "unknown" if it doesn't match anything in the licence list? We would generally prefer a failure (this code is mainly intended for netkan)

archer884: @politas The old code would create a new instance of an unknown license in that case, on line 367. The new code does the same thing but uses the static `License.UnknownLicense` instead of a new instance. I probably shouldn't have been doing arbitrary little cleanup-y things like that in a branch about timeouts, but the effect (creating the unknown license) is not changed.

politas: Ok. The little code cleanups are making it hard to review, given my poor understanding of the transaction code!

archer884:
@politas I have categorized the work into different commits based on what it actually does. I hope that will make this easier to review. In theory, it should also make it easier to cherry-pick only the parts of the work you want at the moment, if you prefer to go that route.

GUI/MainInstall.cs
```diff
@@ -96,6 +97,31 @@ public partial class Main
                 installCanceled = true;
             };
 
+            GUI.user.RaiseMessage("About to install...\r\n");
```
politas: This message is appearing when the only actions are removing mods

archer884: Well, that sucks. I'll poke that when I get home and see if I can understand what's going on.

politas: Whoops, I deleted this comment from the commit in order to move it to the PR:
"This message is appearing when the only actions are removing mods, and the client seems to get stuck after that."

politas: I haven't tested updates, but I suspect they may have the same problem.

archer884: I haven't made any updates; I just rearranged the pull request to be easier to review. I didn't get to leave my day job yesterday until dark thirty. :(

archer884:
@politas If you look at the `EnsureCache` method on line 259 of `ModuleInstaller.cs`, you'll see where where it accepts a downloader that defaults to null. That method was throwing a null exception in the event that there was nothing to download (and, hence, no downloader was being created).

It now just aborts if there's nothing to download.

GUI/MainInstall.cs
politas: Sorry, I meant "I haven't tested running a changeset that only contains updated to installed mods, but I suspect it would suffer the same problem"

politas:
@archer884 , is there a new commit you need to push up to GitHub for that fix?

archer884:
@politas No, the change is already in here. What I did is that I took this opportunity to rebase on master, pull in all the latest changes from there, and then update my previous commit with the corrected code.

politas:
That seems to have fixed up that issue. Thanks for the reorg, makes it much clearer!

politas:
Now we just need to work out why Jenkins is complaining. @techman83, can you have a look if you've got a minute? Might be more @dbent's sphere of expertise, though.

politas:
Ah, beautiful. Merging!

PR #2103, opened 2017-08-08
Fix forum thread link

HebaruSan:
As per http://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1223-mahia/&do=findComment&comment=3143297


politas:
We've fixed the link in the About dialog, there just hasn't been a new release since then.

Issue #1994, opened 2017-01-17
Detect and correct an empty available_modules in the registry when mods are installed


MBradbury:
**CKAN Version:** 
master (https://github.com/KSP-CKAN/CKAN/commit/683fa13bb8f8a6895f4a6d562316eb006ada6f5a)

**Operating System:** 
Windows 10 x64

**The issue you are experiencing:** 

Somehow my registry.json ended up containing '"available_modules": {},' which led to #1992 while trying to diagnose this. When loading CKAN I get the error message shown at the end. MechJeb isn't at fault, it is simply the first mod trying to be found.

I managed to resolve my issue by moving the registry, restarting CKAN and then copying the available_modules from the new registry to the old registry. Ideally CKAN should be able to detect this and refetch available_modules.

```
CKAN.ModuleNotFoundKraken: Module MechJeb2 of version  is not available
   at CKAN.Registry.AllAvailable(String module) in S:\Code\CKAN\Core\Registry\Registry.cs:line 522
   at CKAN.MainAllModVersions.set_SelectedModule(GUIMod value) in S:\Code\CKAN\GUI\MainAllModVersions.cs:line 28
   at CKAN.MainModInfo.set_SelectedModule(GUIMod value) in S:\Code\CKAN\GUI\MainModInfo.cs:line 56
   at CKAN.Main.ModList_SelectedIndexChanged(Object sender, EventArgs e) in S:\Code\CKAN\GUI\Main.cs:line 490
   at System.Windows.Forms.DataGridView.OnSelectionChanged(EventArgs e)
   at System.Windows.Forms.DataGridView.FlushSelectionChanged()
   at System.Windows.Forms.DataGridView.ClearSelection(Int32 columnIndexException, Int32 rowIndexException, Boolean selectExceptionElement)
   at System.Windows.Forms.DataGridView.SetAndSelectCurrentCellAddress(Int32 columnIndex, Int32 rowIndex, Boolean setAnchorCellAddress, Boolean validateCurrentCell, Boolean throughMouseClick, Boolean clearSelection, Boolean forceCurrentCellSelection)
   at System.Windows.Forms.DataGridView.MakeFirstDisplayedCellCurrentCell(Boolean includeNewRow)
[...]

PR #2115, opened 2017-09-27
Handle cases where there are no available_modules but there are installed_modules

ayan4m1:
If the registry.json does not contain available_modules, but there are installed modules, we should re-create the registry. Relates to #1994 #1992 and supercedes #1993.

### DIFF
--- a/Core/Registry/RegistryManager.cs
+++ b/Core/Registry/RegistryManager.cs
@@ -200,10 +200,14 @@ namespace CKAN
                     )
             };
 
-            
-
-            string json = File.ReadAllText(path);
+            var json = File.ReadAllText(path);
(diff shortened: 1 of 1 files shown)


politas:
Can you explain the state of the registry.json file that will trigger this? I tried both deleting and emptying the "available_modules" of my registry.json, leaving the "installed_modules" section intact, and when I refreshed with this change, it lost track of my installed modules both times, discovering ModuleManager as an AD (which causes huuuge problems).

ayan4m1:
Yeah, I don't know too much about the registry.json state except what has been reported in these issues - available_modules is an empty array. I am not confident in my understanding of how AD is currently handled, so let me do some digging there and get back to you.

PR #2116, opened 2017-09-27
Expand .gitignore to handle packages in subdirectories and build output

ayan4m1:


politas:
doesn't ignoring `/_build/` also ignore all subdirectories? All packages and build output should be in there.

ayan4m1:
It does, binaries/release materials should generally not be committed to version control. My change to the .gitignore does not touch the already-existing `_build` rule.

Issue #2062, opened 2017-06-01
Added download info to Metadata page


linuxgurugamer:
Add a line which shows where the mod is going to be downloaded from to the Metadata page.

Right now it's a mystery, would be helpful for debugging or knowing if some location is having a download problem

Chase22:
@linuxgurugamer Maybe close the issue?

linuxgurugamer:
Yes, since it was now released :-)

HebaruSan:
Oh, you wanted this in the download prompts? I thought it was about the mod info tab.
:+1:

linuxgurugamer:
Either

PR #2119, opened 2017-09-28
Add package download URL as clickable link in metadata section (#2062)

ayan4m1:


politas:
We don't actually want that to be a clickable link. Users should use the Download button on the Contents tab to put it in the cache with the adjusted filename.

Issue #1940, opened 2016-11-14
PathErrorKraken when trying to make an absolute path absolute


raptor-xxl:
Thanks for wanting to bring an issue to our attention!

Before you open a ticket, please do a quick search of the existing issues to see if it's already been brought up.

**CKAN Version:** v1.20.1

**Operating System:** Win 8.1 x64

**The issue you are experiencing:** Unhandled exception

**How to recreate this issue:** Try to upgrade Dang it! Continued mod

**CKAN error codes (if applicable):** 

````
CKAN.PathErrorKraken: c:/program files (x86)/Steam/SteamApps/common/Kerbal Space Program/GameData/DangIt/ModuleManager is already absolute
   at CKAN.KSPPathUtils.ToAbsolute(String path, String root)
   at CKAN.ModuleInstaller.<AddParentDirectories>c__AnonStorey0.<>m__0(String dir)
   at System.Linq.Enumerable.<SelectManyIterator>d__1`2.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Collections.Generic.HashSet`1.UnionWith(IEnumerable`1 other)
   at System.Collections.Generic.HashSet`1..ctor(IEnumerable`1 collection, IEqualityComparer`1 comparer)
   at CKAN.ModuleInstaller.AddParentDirectories(HashSet`1 directories)
   at CKAN.ModuleInstaller.Uninstall(String modName)
   at CKAN.ModuleInstaller.AddRemove(IEnumerable`1 add, IEnumerable`1 remove, Boolean enforceConsistency)
   at CKAN.Main.WasSuccessful(Action action)
   at CKAN.Main.InstallMods(Object sender, DoWorkEventArgs e)
   at System.ComponentModel.BackgroundWorker.OnDoWork(DoWorkEventArgs e)
   at System.ComponentModel.BackgroundWorker.WorkerThreadStart(Object argument)
````

politas:
We've been getting a few reports of this issue. I think we need to go over the logic for the parent directories deletion code.

stmobo:
I've been having the same issue as the OP (trying to update DangItContinued), and I've been looking at the code to maybe try and help debug. I think I've figured out what the problem is.

In short: the code is stumbling on conflicting capitalization of paths. Even though paths are supposed to be case-insensitive (in NTFS, anyways) the code is using normal, case-sensitive, string comparisons.

[...]

PR #2122, opened 2017-09-29
Perform directory root comparison in case-insensitive way

ayan4m1:
This should resolve most cases of #1940 without sacrificing case-sensitivity for the actual path creation/manipulation. The issue arises when ToAbsolute is called on an already-absolute path that did not have the same casing as the mod path. The StartsWith() should return false if the two directories are both absolute but have different casing.

### DIFF
--- a/Core/KSPPathUtils.cs
+++ b/Core/KSPPathUtils.cs
@@ -266,7 +266,10 @@ namespace CKAN
 
             if (Path.IsPathRooted(path))
             {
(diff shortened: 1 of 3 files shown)


politas:
Sounds good. I haven't really got a handle on what's happening, though. Can you add a comment explaining the purpose? Even better would be a test case.

ayan4m1:
Working on a good test case now.

PR #2125, opened 2017-10-04
Add test to cover missing directory entries

HebaruSan:
This test passes on HEAD and fails pre #2120 (FileNotFoundKraken is thrown, tested via cherry-picking).


politas:
We love tests!!!!!

PR #2131, opened 2017-10-13
Loop to find max KSP version instead of assuming ordering

HebaruSan:
As per https://github.com/KSP-CKAN/NetKAN/issues/5945, the "Max KSP version" column is inaccurate for 3-4% of indexed mods:
- if the wrong version is chosen as "latest"
- if the compatible scope of the versions isn't coherent over time

This can discourage users from trusting this column, which is a problem because it should be a main tool for a common task: Determining whether it's safe to update the game.

This code change updates the column to show the actual maximum KSP version across all versions of each mod. The simplest possible strategy is used: loop over the versions and compare. The version ordering problems, accumulated bad metadata, and a few other related issues should still be addressed, so I don't consider this to be a complete fix for that issue, but now at least the column will be reliable.

Additional change to call out for transparency: The "min_version+" format is dropped and displayed the same as the unbounded case. If there's a use case for this, it would be best addressed by showing a true range or list rather than making the code and display confusing for maximums.

### DIFF
--- a/Core/Registry/AvailableModule.cs
+++ b/Core/Registry/AvailableModule.cs
@@ -132,14 +132,14 @@ namespace CKAN
         public KspVersion LatestCompatibleKSP()
         {
             KspVersion best = null;
-            foreach (var pair in module_version) {
+            foreach (var pair in module_version)
+            {
                 KspVersion v = pair.Value.LatestCompatibleKSP();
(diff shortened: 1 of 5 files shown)


politas:
Ooh, very nice! This has bugged me for a very long time! #1327 was closed without actually fixing the problem entirely.

PR #2139, opened 2017-10-14
Check for stale lock files (fixes #2075)

### LINKED ISSUE #2075: Locking pid validation fails
At this moment ckan locks itself by creating a "registery.lock" file. However this feature is relatively annoying on Linux, since shutting down your computer with programs running (this is how I shutdown my computer). Will still leave the lock file behind. Possibly since a "sigkill" is considered as a crash?

I noticed that ckan writes its PID to the lock file. However it doesn't really check it nicely.
image
If I kill ckan wrongly now (`ctr+c`). This happends:
image
Not sure if that % sign is messing it up. But ckan fails to properly validate if their is a instance running on pid `11749`.

I resolved this issue by doing this in my launcher file:
`rm "/games/SteamLibrary/steamapps/common/Kerbal Space Program/CKAN/registry.locked" | mono "/games/SteamLibrary/steamapps/common/Kerbal\ Space\ Program/ckan.exe"`. But this is possibly not the way you guys intended ckan to be launched :wink:.

HebaruSan:
This code change checks whether the registry lock file is stale before attempting to acquire the lock. This eliminates the lock file popup when it's not needed as per #2075, but keeps it when CKAN is actually still running.

I think there are race conditions if the user tries to start multiple copies of CKAN simultaneously; do we care about that?

### DIFF
--- a/Core/Registry/RegistryManager.cs
+++ b/Core/Registry/RegistryManager.cs
@@ -15,7 +15,7 @@ namespace CKAN
     {
         private static readonly Dictionary<string, RegistryManager> registryCache =
             new Dictionary<string, RegistryManager>();
-        
+
         private static readonly ILog log = LogManager.GetLogger(typeof (RegistryManager));
         private readonly string path;
         public readonly string lockfilePath;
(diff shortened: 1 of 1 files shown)


politas:
The big question would have to be, does this work on Windows, too?

HebaruSan:
Good point. I believe it would, but obviously what's needed is real testing. I'll take it for a spin on a Windows VM (which presumably will also involve installing KSP).

politas:
I'm about to do so. You don't need a whole KSP, just the BuildID.txt and a GameData folder

politas:
Seems to work fine on Windows, though it's much harder to get a real registry.locked file there.

HebaruSan:
Yeah, wow, killing it with taskmgr cleans up the file, and it wouldn't even let me make a copy first. I'm vaguely curious as to how you managed it.

politas:
A good setup for testing is a buildID.txt, a readme.txt, GameData and Ships directories.

buildID.txt just needs a valid build ID like so:

    build id = 01804

(That's 1.3.0)

readme.txt is not needed other than to check the failover for bad buildIDs, and should have:

    Version 1.3.0

(obviously, it's best to make the two files match as far as the version)

I just created a dummy registry.locked file with a four digit number (almost guaranteed not to exist)

PR #2147, opened 2017-10-16
Update de-indexing.md

Blu3wolf:
updates policy file to reflect current de-facto policy position
removes statement that CKAN will not honour mod author requests for de-indexing

As per #1793 corrects disparity between current policy and current policy document.

### DIFF
--- a/policy/de-indexing.md
+++ b/policy/de-indexing.md
@@ -3,13 +3,15 @@
 From time to time mod authors ask their works be removed from the CKAN. Whether or not we should do this is a matter of licensing and public good. Our policy is based on two simple principles:
 
 1. **The CKAN, like the legal system, takes licenses very seriously.**
-1. **The CKAN acts in the interests of its users.**
+1. **The CKAN acts in the interests of mod authors.**
 
-Restrictive licenses, including the default "All Rights Reserved", allow the content author absolute control over what happens to their work. **If an All Rights Reserved or other restrictive license author asks for their mod to be de-indexed, we do so promptly and without question.**
+Restrictive licenses, including the default "All Rights Reserved", allow the content author absolute control over what happens to their work. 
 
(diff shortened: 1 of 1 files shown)


HebaruSan:
It still says "we cannot in good faith de-index the mod and still act in the interests of our users".

This document should probably be re-written from scratch rather than patched. It goes into a lot of background about licenses that is no longer relevant if the new rule is to de-index on request.

Blu3wolf:
It does still say that, yes. This reflects an intentional decision on my part rather than an omission or oversight. 

I did make the suggestion it be rewritten, and you may recall that that suggestion was met with the issue being closed #1793

I do not think the discussion about licensing is not relevant, but I do think a rewrite may be in order. The policy document here espouses values which are no longer part of the CKAN.

Blu3wolf:
Perhaps removing lines 15 and 16 would make it clearer?

HebaruSan:
What does that sentence mean now, though? Haven't we decided that we **can** in good faith de-index a mod regardless of license? Does the license play any role in that nowadays?

I think this document should start with a discussion of the conflicting goals and priorities of CKAN. On the one hand, convenience for users. On the other hand, recognizing the potential impact on modders. And then move into how to reconcile these things when a problem arises. Something about obtaining consent as a best practice would be good as well.

It's probably also time to remove comments like "(Hopefully) temporary important information" from the wiki:
https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN#before-you-start

(I would jump in to making some of these edits myself, but I'm about to step out the door for a week without internet, and I don't want to make the spouse any more impatient.)

politas:
Sorry, @Blu3wolf, but this passive aggressive PR is not going to be merged. Delisting permissively-licensed mods is a workaround until we can implement a system to allow listing with manual installation. There is no contradiction.

Blu3wolf:
Well, its a bit mooted, but no, we have decided nothing of the sort. 

That having been said, that discussion was had, and it also failed to generate a new policy document - just a new policy decider.

PR #1929, opened 2016-10-26
Allow uninstall of incompatible mods in GUI

ayan4m1:
Fixes #1842, needs tests but I still can't find a viable way of testing the Main class which is where the ApplyToolButton is declared as a private member.

Issue #1274, opened 2015-07-10
CKAN doesn't like mods that have many things fill one requirement.


FireFaced:
So, I was installing Kofeyh's CKAN file ( http://pastebin.com/raw.php?i=JTsdniZf ) to get all his mods installed, but this error was thrown:

```
 *************\* Exception Text **************
 CKAN.TooManyModsProvideKraken: Too many mods provide RealFuels-Engine-Configs:
 RFStockalike v2.1.7
 - RealismOverhaul v10.1.0
   at CKAN.RelationshipResolver.ResolveStanza(IEnumerable`1 stanza, SelectionReason reason, RelationshipResolverOptions options, Boolean soft_resolve)
   at CKAN.RelationshipResolver.Resolve(CkanModule module, RelationshipResolverOptions >options)
   at CKAN.RelationshipResolver.ResolveStanza(IEnumerable`1 stanza, SelectionReason reason, RelationshipResolverOptions options, Boolean soft_resolve)
   at CKAN.RelationshipResolver.Resolve(CkanModule module, RelationshipResolverOptions >options)
   at CKAN.RelationshipResolver.AddModulesToInstall(IEnumerable`1 modules)
   at CKAN.ModuleInstaller.InstallList(List`1 modules, RelationshipResolverOptions options, IDownloader downloader)
   at CKAN.Main.WasSuccessful(Action action)
   at CKAN.Main.InstallList(HashSet`1 toInstall, RelationshipResolverOptions options, IDownloader downloader)
   at CKAN.Main.InstallMods(Object sender, DoWorkEventArgs e)
   at System.ComponentModel.BackgroundWorker.WorkerThreadStart(Object argument)
```

FireFaced:
Markdown is being weird.

FireFaced:
Anyone?

Wetmelon:
I also get this exception with my own CKAN file for Real Solar System, except it's for the "RSSTextures".

Real Solar System has a "RSSTextures (virtual)" depends, with 3 "provided by" options.

Real Fuels has "RealFuels-Engine-Configs (virtual)" with either RealFuels: Stockalike RF Configs, or RealismOverhaul providing options.

Evidently, this error occurs when multiple options satisfy the virtual dependency.

Wetmelon:
Okay, I was able to get it to work by eliminating the Kraken and using soft_resolve, effectively making it iterate through the options until it gets to the correct, non-conflicting one.

[...]

PR #2150, opened 2017-10-17
Revert "Merge #1929 Allow uninstall of incompatible mods in GUI"

politas:
This reverts commit 85d72b64c3e2d1db1bbc5e608a52eafb82c15c5e, reversing
changes made to bcf87c20409b292f85abcb0ddeef622542ed585d.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
-## v1.24.0-PRE
+## v1.22.6-PRE
 
 ### Bugfixes
-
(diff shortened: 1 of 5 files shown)


politas:
@ayan4m1, There's a bug that crept in with your #1929 PR. When a mod is selected that triggers a dependency choice, the dialog pops up, the user selects an option, and the selected option appears to be checked for install in the modlist, but it doesn't appear to be fully chosen. Applying changes throws a "TooManyModsProvideKraken" (#1274) and selecting other changes causes the Provides selection box to reappear, over and over. The selected provider does appear in the changelist, though. It's very odd.

PR #2156, opened 2017-10-20
Change Icon on FilterTool Button to something filter-y

### LINKED ISSUE #2009: Proposal: replace "Filter" button icon in the GUI from a looking glass to a filter

politas:
Replace the magnifying glass with something better  (I refuse to accept the funnel = filter idea)
filter
Closes #2009


HebaruSan:
I'd like to help review, but there are conflicts now. Can you rebase?

politas:
Rebased

HebaruSan:
Can the old old search.png file be removed? I don't think it's being used anymore.

politas:
Good point

politas:
I'm tempted to leave it in, since it may come in useful in the future, and it's only 1493 bytes.

HebaruSan:
I suppose it's not important, and the new icon definitely works fine. I'll take a stab at following the full merge process, including change log updates...

politas:
We've been putting "Merge #<pr> Blah blah blah" as the comment on the Merge commit, to make the merge commits stand out in the history. Other than that, <3<3<3!!!!

HebaruSan:
That's a good idea; I've added it to the wiki's merging notes to help me remember for next time. Thanks!

PR #1971, opened 2017-01-04
WIP: Handle InconsistentKraken from RelationshipResolver in GUI

politas:
Finally tracked down the cause of the weirdness. Now to fix it!
Closes #1864

PR #2163, opened 2017-10-26
Simplify IUser

HebaruSan:
Currently `IUser` has 6 public events, which makes it awkward to implement this interface. These events are only used inside `NullUser`, so at first blush, they belong inside `NullUser` as an implementation detail. However, all `NullUser` does with them is wire up the interface's functions to `NullUser`'s own equivalent `virtual` functions. This is an unnecessary and cumbersome step of indirection with no advantages over simply calling those functions directly. (Technically this structure could allow third-party classes to inject listeners to various events in other places, but no code currently does this, so the need/advantage is debatable.)

This PR replaces this structure with a much simpler one and does some additional minor clean-up:

- Remove the events from the interface, making it much simpler to implement in new classes, such as a new UI
- Remove the events from `NullUser`
  - Make `NullUser` simply call the virtual functions from inside the interface functions
- Remove the delegate definitions now that they're no longer used
- Remove `WindowWidth` (not used even before this)
- Rearrange the remaining members of `IUser` into a general-purpose group and a progress/updates group
- (My editor auto-deletes trailing whitespace)

After these changes, `Core\User.cs` is much easier to understand, the `IUser` interface is much simpler to implement in new code, and existing code still works.

### DIFF
--- a/Cmdline/ConsoleUser.cs
+++ b/Cmdline/ConsoleUser.cs
@@ -246,10 +246,6 @@ namespace CKAN.CmdLine
         protected override void ReportDownloadsComplete(Uri[] urls, string[] filenames, Exception[] errors)
         {
         }
-        public override int WindowWidth
-        {
-            get { return Console.WindowWidth; }
-        }
 
     }
 }
--- a/Cmdline/ConsoleUser.cs
+++ b/Cmdline/ConsoleUser.cs
@@ -243,9 +243,6 @@ namespace CKAN.CmdLine
                 Console.Write("\r\n{0}", format);
             }
         }
-        protected override void ReportDownloadsComplete(Uri[] urls, string[] filenames, Exception[] errors)
-        {
-        }
 
     }
 }
--- a/Core/Net/NetAsyncDownloader.cs
+++ b/Core/Net/NetAsyncDownloader.cs
@@ -238,7 +238,7 @@ namespace CKAN
                 );
             }
         }
-        
+
         public void DownloadAndWait(ICollection<KeyValuePair<Uri, long>> urls)
         {
             // Start the download!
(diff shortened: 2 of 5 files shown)


HebaruSan:
It looks like `ReportDownloadsComplete` / `RaiseDownloadsCompleted` / `DownloadsComplete` also do nothing. They're called from one place, but that call turns out to be a no-op, so that's removed as well in the latest commit.

politas:
My suspicion would be that as events ahave been removed from IUser's main sections, the editors have left them in the NullUser part by accident, not necessarily realising they were there. This PR seems to be removing a bunch of code, but I can't see any impact. Is it possible some of it is used in obscure commandline cases? I mean, the IDE/build process should flag that, I suppose.

HebaruSan:
"Obscure command line cases" is the story behind `DisplaySelectionDialog` / `RaiseSelectionDialog` / `AskUserForSelection`; the command line code uses this to talk to itself in some places, and it has no other current function, in GUI or Core. I wanted to keep this PR's scope narrow, though, so I left that alone.

I did text searches before removing the events, and then compiled the code after. I believe that would catch anything that's trying to use them. `NullUser`'s virtual function interface is the de facto interface for `IUser`, and the rest of it is seems to be out of sight, out of mind.

politas:
NullUser is about allowing tests to run without interrupting the build process, I gather.

politas:
Great work, by the way, and very much appreciated! It's very easy for codebases like this to get filled up with unused cruft that no one dares touch.

HebaruSan:
Yes, `NullUser` does that, but it also has a strange dual role as a base class for UI-specific child classes. Currently `GUIUser` "is a" `NullUser`, as are `Netkan.ConsoleUser` and `Cmdline.ConsoleUser`, which means that all real errors/notifications the users sees in practice are routed through a class that is named after the idea of _discarding_ them. That's another thing that I decided to leave alone for now.

politas:
That seems odd. Well, if you want to tackle those issues as future projects, it would be most welcome. I find myself flailing in the wind trying to make sense of this code a lot of the time. See #1971 for a good example of that.

PR #2169, opened 2017-11-01
Establish a .gitattributes file

### LINKED ISSUE #409: Establish a .gitattributes file
This can handle making sure we're using spaces instead of tabs, etc etc.

Thanks to @tips48 for spotting that we can do this!

References:
- http://git-scm.com/docs/gitattributes
- https://github.com/Danimoth/gitattributes

My own personal `.gitattributes` file right now is just:

```
*.cs filter=tabspace
```

Olympic1:
Adds a `.gitattributes` file to handle line endings on the repo and makes sure we're using spaces instead of tabs.

"Recreated" files on the new `.gitattributes` file. (the 3 other files in this PR)

Closes #409


HebaruSan:
After pulling this, three files are showing up as permanently having changes in the working directory even though no actual changes are present according to a diff:

- Tests/app.config
- assets/README.md
- nuget.config

This is preventing me from rebasing a feature branch. Checking them out again has no effect---they still show up as changed! I assume this has to do with line endings (I'm on Linux).

Could this .gitattributes file be less aggressive in whatever it's doing?

PR #2172, opened 2017-11-03
Minor spelling and formatting clean-up

HebaruSan:
Lots of very minor changes to comments, strings, and code formatting.

- Spelling/grammar
- Spaces around operators
- Sentence-case capitalization
- One f-bomb made forum-friendly


Netkan/Transformers/ForcedVTransformer.cs
```diff
@@ -21,7 +21,7 @@ public Metadata Transform(Metadata metadata)
             JToken forceV;
             if (json.TryGetValue("x_netkan_force_v", out forceV) && (bool)forceV)
             {
-                Log.InfoFormat("Executing Forced-V transformation with {0}", metadata.Kref);
+                Log.InfoFormat("Executing forced-V transformation with {0}", metadata.Kref);
```
politas: Should make the "V" lower-case too, for consistency

GUI/MainInstall.cs
```diff
@@ -99,8 +99,8 @@ public partial class Main
 
             GUI.user.RaiseMessage("About to install...\r\n");
 
-            // FIXME: here we should heat up the cache with any mods we're going to install, because 
-            // when this transaction types out it fucks up everything else.
+            // FIXME: here we should heat up the cache with any mods we're going to install, because
+            // when this transaction types out it messes up everything else.
```
politas: Thumbs down for the bowdlerisation

HebaruSan: I'll admit, I had to look up that reference; thanks for an interesting Wikipedia visit.

Responses:
1. Comments in the CKAN source code are not classic works of Shakespeare
2. The meaning of the comment is either the same (unless "to mess up" is an Americanism, unbeknownst to me), or clearer since it's not actually related to reproduction
3. The code of conduct expresses an intent to "‬help‭ ‬us‭ ‬create‭ ‬safe‭ ‬and‭ ‬positive‭ ‬experiences‭ ‬for‭ ‬everyone," and I would suggest that "everyone" includes people who find words I couldn't say on daytime TV off-putting

Are you completely opposed to any change here, or is a compromise possible?

politas: Nah, I'm cool with it. My Australian nature sees no problem with the language use, but I accept that some people may find it off-putting. This was more a comment than a request for change.

HebaruSan: Ah, in that case I'll count my blessings that you didn't call me a c*nt, and go about my business! :+1:

PR #2170, opened 2017-11-01
Add include_only fields

Olympic1:
This PR adds two fields which practically work as negative filter/filter_regexp.
Also includes a test and changes to the schema to make sure there are only filter or include_only fields but never both.
Also added the new "filters" to the spec.

Closes #228
Closes #1577

PR #2173, opened 2017-11-03
Fix schema validation errors

HebaruSan:
NetKAN pull requests and the bot are currently failing at the schema validation step, and the output suggests it's something from #2170. This command can be used to reproduce the issue locally:

```
python ./bin/ckan-validate.py /path/to/existing/valid/file.ckan
```

I tried running `CKAN.schema` through the validator at https://www.jsonschemavalidator.net/ , and it flagged two places where `"type"` was used that should have been `"$ref"`. After fixing this, the validation script runs successfully.

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -231,26 +231,26 @@
                     },
                     "include_only" : {
                         "description" : "List of files and directories that should not be excluded from the install",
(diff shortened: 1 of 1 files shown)


Olympic1:
oops, my mistake

PR #2181, opened 2017-11-10
Remove unnecessary using directives

HebaruSan:
My editor got [an upgrade](https://github.com/atom/ide-csharp/issues/7#issuecomment-343563804) today, which to my surprise included the ability to detect when a `using` directive in a C# file isn't needed. To celebrate, here's a pull request!

Most of these are because nothing from that assembly is being used.
Some of them (`using CKAN;` mainly) are because the file's code is already in that namespace.

In theory this might speed up compile times (since the compiler doesn't have to load as many references to parse a file), but that would probably be imperceptible in practice.

### DIFF
--- a/Cmdline/Action/Available.cs
+++ b/Cmdline/Action/Available.cs
@@ -1,11 +1,4 @@
-﻿using System;
 using System.Collections.Generic;
-using System.IO;
-//using System.Linq;
-//using System.Text.RegularExpressions;
-//using CKAN.Exporters;
-//using CKAN.Types;
-//using log4net;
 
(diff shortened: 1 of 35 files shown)


HebaruSan:
#rebuild

politas:
Wow, that's a lot of unnecessary Usings!

politas:
Build before merging this:
```
real	0m30.344s
user	0m19.884s
sys	0m4.204s
```

Build after merging this:
```
real	0m25.354s
user	0m19.508s
sys	0m3.756s
```

A definite difference!

PR #2183, opened 2017-11-11
Fix null-reference exception on failed mod upgrade

### LINKED ISSUE #2162: Unhandled exception when updating mod over a manually-updated mod
### CKAN Version if known (or site URL)
v1.22.6 (Windows 10 64-bit)

### Please describe the expected behaviour
CKAN aborts install with error message and returns to mod-selection window.

### Please describe the actual behaviour
CKAN aborts install with error message and throws an Unhandled Exception "Object reference not set to an instance of an object."

Hitting "Continue" appears to resume normal operation of CKAN, but only parts of the GUI are responsive; most of it is unresponsive after this and the app has to be manually killed using Task Manager.

### What steps can be taken to reproduce the issue? 
1. Install a mod (in my case, KDEX v1.04)
2. Manually install an update of the mod
3. Attempt to update the mod in CKAN

**Error message:**
```
About to install...


Downloading "https://github.com/mwerle/kdex/releases/download/v1.10/kdex-1.10.zip"


Oh no!

It looks like you're trying to install a mod which is already installed,
or which conflicts with another mod which is already installed.

As a safety feature, the CKAN will *never* overwrite or alter a file
that it did not install itself.

If you wish to install KDEX v1.10 via the CKAN,
then please manually uninstall the mod which owns:

GameData/masTerTorch/kdex.version

and try again.

Your GameData has been returned to its original state.




Error!
```

**Exception call stack:**
image

```

   at CKAN.Main._MarkModForUpdate(String identifier)
   at CKAN.Main.PostInstallMods(Object sender, RunWorkerCompletedEventArgs e)
   at System.ComponentModel.BackgroundWorker.OnRunWorkerCompleted(RunWorkerCompletedEventArgs e)


************** Loaded Assemblies **************
mscorlib
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2110.0 built by: NET47REL1LAST
    CodeBase: file:///C:/Windows/Microsoft.NET/Framework64/v4.0.30319/mscorlib.dll
ckan
    Assembly Version: 1.22.0.0
    Win32 Version: 1.22.6
    CodeBase: file:///C:/Games/ckan.exe
System.Core
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2102.0 built by: NET47REL1LAST
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Core/v4.0_4.0.0.0__b77a5c561934e089/System.Core.dll
System
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2110.0 built by: NET47REL1LAST
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System/v4.0_4.0.0.0__b77a5c561934e089/System.dll
System.Configuration
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2046.0 built by: NET47REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Configuration/v4.0_4.0.0.0__b03f5f7f11d50a3a/System.Configuration.dll
System.Xml
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2102.0 built by: NET47REL1LAST
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Xml/v4.0_4.0.0.0__b77a5c561934e089/System.Xml.dll
System.Windows.Forms
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2110.0 built by: NET47REL1LAST
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Windows.Forms/v4.0_4.0.0.0__b77a5c561934e089/System.Windows.Forms.dll
System.Drawing
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2046.0 built by: NET47REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Drawing/v4.0_4.0.0.0__b03f5f7f11d50a3a/System.Drawing.dll
System.Transactions
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2046.0 built by: NET47REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_64/System.Transactions/v4.0_4.0.0.0__b77a5c561934e089/System.Transactions.dll
System.Numerics
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2046.0 built by: NET47REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Numerics/v4.0_4.0.0.0__b77a5c561934e089/System.Numerics.dll
System.Runtime.Serialization
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2112.0 built by: NET47REL1LAST
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Runtime.Serialization/v4.0_4.0.0.0__b77a5c561934e089/System.Runtime.Serialization.dll
System.Data
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2102.0 built by: NET47REL1LAST
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_64/System.Data/v4.0_4.0.0.0__b77a5c561934e089/System.Data.dll
Anonymously Hosted DynamicMethods Assembly
    Assembly Version: 0.0.0.0
    Win32 Version: 4.7.2110.0 built by: NET47REL1LAST
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_64/mscorlib/v4.0_4.0.0.0__b77a5c561934e089/mscorlib.dll
Microsoft.GeneratedCode
    Assembly Version: 1.0.0.0
    Win32 Version: 4.7.2102.0 built by: NET47REL1LAST
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Xml/v4.0_4.0.0.0__b77a5c561934e089/System.Xml.dll
Microsoft.CSharp
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2046.0
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/Microsoft.CSharp/v4.0_4.0.0.0__b03f5f7f11d50a3a/Microsoft.CSharp.dll
System.Dynamic
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2046.0
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Dynamic/v4.0_4.0.0.0__b03f5f7f11d50a3a/System.Dynamic.dll
System.EnterpriseServices
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2046.0 built by: NET47REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_64/System.EnterpriseServices/v4.0_4.0.0.0__b03f5f7f11d50a3a/System.EnterpriseServices.dll
Accessibility
    Assembly Version: 4.0.0.0
    Win32 Version: 4.7.2046.0 built by: NET47REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/Accessibility/v4.0_4.0.0.0__b03f5f7f11d50a3a/Accessibility.dll

************** JIT Debugging **************
To enable just-in-time (JIT) debugging, the .config file for this
application or computer (machine.config) must have the
jitDebugging value set in the system.windows.forms section.
The application must also be compiled with debugging
enabled.

For example:

<configuration>
    <system.windows.forms jitDebugging="true" />
</configuration>

When JIT debugging is enabled, any unhandled exception
will be sent to the JIT debugger registered on the computer
rather than be handled by this dialog box.


```

mwerle:
Fixes #2162 

THIS IS A HACK!  I can't work out why the SortedDictionary "module_version" reverses it's sort-order after a module update has failed; someone more familiar with the codebase may be able to work out the actual bug.

### DIFF
--- a/Core/Registry/AvailableModule.cs
+++ b/Core/Registry/AvailableModule.cs
@@ -75,6 +75,7 @@ namespace CKAN
         public CkanModule Latest(KspVersionCriteria ksp_version = null, RelationshipDescriptor relationship=null)
         {
             var available_versions = new List<Version>(module_version.Keys);
+            available_versions.Sort(new RecentVersionComparer());
             CkanModule module;
             log.DebugFormat("Our dictionary has {0} keys", module_version.Keys.Count);
             log.DebugFormat("Choosing between {0} available versions", available_versions.Count);
--- a/GUI/MainModList.cs
+++ b/GUI/MainModList.cs
@@ -237,14 +237,16 @@ namespace CKAN
 
         public void _MarkModForUpdate(string identifier)
         {
-            foreach (DataGridViewRow row in ModList.Rows)
+            if (!mainModList.full_list_of_mod_rows.ContainsKey(identifier))
             {
-                var mod = (GUIMod) row.Tag;
-                if (mod.Identifier == identifier)
-                {
-                    (row.Cells[1] as DataGridViewCheckBoxCell).Value = true;
-                    break;
-                }
+                return;
+            }
+            DataGridViewRow row = mainModList.full_list_of_mod_rows[identifier];
+
+            var mod = (GUIMod)row.Tag;
+            if (mod.Identifier == identifier)
+            {
+                mod.SetUpgradeChecked(row, true);
             }
         }
     }


HebaruSan:
`AvailableModule.Latest` is called for _every_ mod in `Registry.Available` and `Registry.LatestAvailableWithProvides`, which are in turn called in many other places such as the relationship resolver. Sorting the version list in `Latest` could cause a severe performance degradation across the board, as the registry would start re-sorting tons of lists at various places in miscellaneous operations throughout the code.

I don't think this is safe to merge as-is.

mwerle:
As I said it's a hack; having said that, IMHO it's better than a NullRef. On my (admittedly fast) PC I didn't notice a performance impact; overall I think the the dataset is still too small for it to realistically matter at this stage

Also I used the same sorter as the SortedDictionary uses. Again, I have no idea why the dictionary sort order ends up reversing itself; according to the docs it shouldn't be possible.

Anyway, I just put this out there as I can't get any further with this and it might give someone more familiar with the codebase and C# a starting point.

mwerle:
Ok, found where the dictionaries get their orders reversed; it's when the FileExistsKraken gets re-thrown in ModuleInstaller.cs:419, and re-caught in MainInstall.cs:259.

Before the throw it's in the order it should be, after the throw it's reversed.  How or why the throw affects the sort-order of the dictionaries in the RegistryManager is anybodies guess; this is under Windows/DotNet, btw, haven't tested under Mono.

HebaruSan:
Can we start with simply checking whether the check box cell reference is null before using it?

_MarkModForUpdate currently makes a hard assumption that the identifier it receives **will** be upgradeable, which we've seen is not a justified assumption, and it throws an NRE if it's not. Even if we consider this a serious enough error condition to warrant an exception (I don't think I do), we should be detecting it and throwing a more meaningful exception instead (something with a name like ModNotUpgradeableKraken, perhaps).

Let's keep your rewrite of that function to better resemble _MarkModForInstall :+1:. Then in SetUpgradeChecked, if update_cell is null, just return. And remove the sort call. That should prevent the crash without adversely affecting any other code flows.

I haven't seen your investigation results, so I have to admit I'm skeptical that there's a problem with list sorting. How did you come to that conclusion?

HebaruSan:
I have a possible alternate fix in the works. Can you try your test with the exe in this zip and let me know what happens?

-  [ckan-checkbox-cleanup.zip](https://github.com/KSP-CKAN/CKAN/files/1464328/ckan-checkbox-cleanup.zip)

It's possible to eliminate the possibility of this error happening in the first place with a little refactoring. I'll put this change into a separate pull request if it looks OK.

mwerle:
> Then in SetUpgradeChecked, if update_cell is null, just return. And remove the sort call. That should prevent the crash without adversely affecting any other code flows.

If we just return from SetUpgradeChecked on error then following the failed upgrade you won't be able to upgrade any mod as none will be marked as upgradeable.  It was my initial attempt to fix this.

So not really a fix, but better than a crash.

> I haven't seen your investigation results, so I have to admit I'm skeptical that there's a problem with list sorting. How did you come to that conclusion?

By looking at the list in question in a debugger after every step through the code. Following the "throw", the list is sorted in increasing version order instead of decreasing; this is true for all mods in the "available_mods" registry, not just the mod I was trying to upgrade.

There's some more details in the bug report #2162.

HebaruSan:
What's interesting about `AvailableModule` is that the sorting convention is set _only_ by a parameter to the constructor of the default value of `module_version`, not the type itself, and that object isn't really protected in any way. So if `module_version` was _replaced_ by a `SortedDictionary` from outside, say by a JSON deserializer (it _is_ a `[JsonProperty]!`), it would fall back to the default sort order built in to the `Version` type, which is ascending, opposite to what `AvailableModule` requires and assumes.

If we eliminate `RecentVersionComparer` and make `AvailableModule` work with an ascending-order `SortedDictionary`, that would eliminate the risk of that happening. I'll add that to #2184...

mwerle:
> What's interesting about AvailableModule is that the sorting convention is set only by a parameter to the constructor of the default value of module_version, not the type itself, and that object isn't really protected in any way. So if module_version was replaced by a SortedDictionary from outside, say by a JSON deserializer (it is a [JsonProperty]!),

Interestingly enough when the registry is loaded from the JSON, it's loaded into the correct (ie, descending) order, despite the values in the JSON being in ascending order.  It's only when the exception gets thrown that the order gets changed (well, perhaps the underlying object gets replaced; I didn't take note of the memory addresses..)

We still (I think?) need the "RecentVersionComparer" though as it correctly sorts numerical values rather than by ASCII-order, regardless of whether we keep the dictionary in ascending or descending order.



Anyway, I didn't see your new binary this morning - only saw the notification email after I had already gone out, so I'll have a look at your suggested changes and test them out.

Looks like we might be honing in on a "proper" fix :)

HebaruSan:
The [documentation](https://msdn.microsoft.com/en-us/library/f7fta44c(v=vs.110).aspx) for `SortedDictionary` says

> If type TKey implements the `System.IComparable<T>` generic interface, the default comparer uses that implementation.

... and our key type `Version` does implement `IComparable<Version>`, so I think we still get the sophisticated sort without `RecentVersionComparer`.

mwerle:
Ok, just tested your pull request, and it mostly works (I added review comments to the bits which don't).

Overall a much better fix than my hack :)

HebaruSan:
Cool, I'll be glad to address your review comments. I don't see them yet, however; is there a "share your comments" button that you can click?

mwerle:
Heh, forgot to click on the "Submit Review" button (I've used GitHub some, but am by no means an expert).

I also just worked out it's possible to send a pull request against your pull request, so I (think) I've just done that too.

But I guess for all intents and purposes, this pull request can be closed now.. (apart from the clean-up commit, my "fix" is obsolete). We can continue any discussion in PR #2184

PR #2185, opened 2017-11-14
Cache downloads globally

### LINKED ISSUE #584: Please cache downloads outside of the game directory
Currently almost all of CKAN's data, including downloaded mods, is stored within the `CKAN` directory within a KSP installation. As a consequence, mods previously downloaded and cached in one KSP installation's `CKAN` directory will be re-downloaded when trying to install the same mods into another KSP installation.

Especially with larger packages, such as the high resolution real solar system textures that weight in at something like 2GB, this gets annoying quickly, especially over slow downlinks.

Rather than caching downloads within KSP installations, please cache them in a more global location such as the user's home directory. On linux systems `$XDG_CACHE_DIR/ckan` would seem like a reasonable place.

### LINKED ISSUE #960: Feature Request: Multiple KSP Instances -> share downloads
I have more than one KSP instance, at least one Main copy, a testing copy for bug-hunting, and soonish one to test new mods for compatibility without feeding my saves to the Kraken.

Most mods are shared between them - even in the same versions.
I'd like ckan to store the downloaded archives in a centralized location (config setting for this? env i.e. $CKAN-DOWNLOADS) so I don't have to download a mod 3 times, when one time is enough, also saves HD storage.

Short term workaround could be to have ckan look in the other known installs for the required download.

Olympic1:
This is the beginning of a global cache for downloads. There is a test for read/write access, and an option for moving a cache. Currently, there is no way to go from the local cache in the client into the global cache.

- [x] Test cases
- [x] Core changes
- - [x] Get/Set cache path
- - [x] Get/Set cache path in registry 
- - [x] Move cache folder to path
- - [x] Rewrite `KSP.DownloadCacheDir()` to use new path
- - [x] Save cachepath in registry
- [ ] CLI changes
- - [x] Add `Set Cache` command
- - [ ] Add `Move` command to move an old cache to the new one
- [ ] GUI changes
- - [x] Add option to set cache folder
- - [ ] Add option to move old cache folder to new one
- [x] Store cachepath in config file

Closes #584, closes #960


Core/Net/NetFileCache.cs
```diff
@@ -11,43 +11,54 @@
...
     {
         private FileSystemWatcher watcher;
         private string[] cachedFiles;
         private string cachePath;
         private static readonly TxFileManager tx_file = new TxFileManager();
-        private static readonly ILog log = LogManager.GetLogger(typeof (NetFileCache));
+        private static readonly ILog log = LogManager.GetLogger(typeof(NetFileCache));
 
-        public NetFileCache(string _cachePath)
+        public NetFileCache(string _cachePath = null)
```
HebaruSan: This parameter is still being passed from KSP.cs:61, so this change would not yet have an effect in the full app. The `KSP.DownloadCacheDir()` function probably needs to be retired, with some kind of migration process for old downloads.

Core/Net/NetFileCache.cs
```diff
@@ -215,6 +272,88 @@ public string GetCachedZip(Uri url, bool test_data = false)
...
+            // Create the new path if it doesn't exists
+            if (!Directory.Exists(newPath))
+            {
+                Directory.CreateDirectory(newPath);
+            }
+
+            // Check that we have list permission
+            try
+            {
+                Directory.GetFiles(newPath);
```
HebaruSan: This will load the entire list of files in the target directory into memory, which could be a performance hit if there are already a bunch of files there (e.g., if merging multiple separate instance caches). Is there another check we could do instead? What if we just created the test file?

Core/Net/NetFileCache.cs
```diff
@@ -215,6 +272,88 @@ public string GetCachedZip(Uri url, bool test_data = false)
...
+            }
+
+            // TODO: Start a new transaction
+
+            // Get a list of all the files in the old cache and move them over
+            var files = Directory.GetFiles(cachePath);
+
+            foreach (string file in files)
+            {
+                string newFilePath = Path.Combine(newPath, Path.GetFileName(file));
```
HebaruSan: If we're combining multiple old caches, there's a good chance there will be collisions, as eliminating redundant downloads and disk space is one of the reasons this project was requested. We should check whether `newFilePath` already exists and do something intelligent if it does. Maybe print an error, maybe just delete it if we're moving and the file contents are the same.

politas:
I made a start on this one, too!

Olympic1:
Core functions are in place this ill mostly be bugfixes now.

@politas, I wil add your CmdLine code later

Core/Net/NetFileCache.cs
Olympic1: Rewrote the `KSP.DownloadCacheDir()` and cache gets moved when new path was specified.

Core/Net/NetFileCache.cs
```diff
@@ -215,6 +272,88 @@ public string GetCachedZip(Uri url, bool test_data = false)
...
+
+                if (move)
+                {
+                    tx_file.Move(file, newFilePath);
```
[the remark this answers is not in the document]
Olympic1: Will do some research for checking disk space later this week.

Olympic1:
Also could someone figure where [MoveCacheFailsForNoAccess()](https://github.com/KSP-CKAN/CKAN/pull/2185/files#diff-a8b9da1321e2c360be89f05579a1914bR128) fails exactly. I don't have a linux setup

Genjokoan:
I am new to KSP and looked for how to do this, which led me here.  I can do testing for you on Mac OS and Windows 10.  My Mac kung fu is okay, windows not so much though it is my main KSP machine.

Olympic1:
Thanks that you want to test this out, this needs some testing on Mac OS. Here is a compiled version of `ckan.exe` and `netkan.exe`. Let me know if you come across any issues.

[ckan.zip](https://github.com/KSP-CKAN/CKAN/files/1520598/ckan.zip)

Genjokoan:
Issues, yes.  I have a Mac Mini (Late 2012) but play on a PC.  I installed your ckan.exe and getting it to run involved a little trial and a lot of error.  Even with the wiki install-on-a-mac I think someone with purely point-and-click mindset would give up before getting the app running.
Now, running.  I have my monitor set for 1440p and the difference between the Mac and PC version is remarkable.  The Mac version's graphics were inconsistent.  I can provide screen shots if you want them.
In editing the settings, I would set the cache directory and it would not take.  Set it again, no joy.  Quit, restart, set, no joy.  I kept at it and after about three restarts, the setting stuck, though when you have the setting box open, the cache directory appears blank.  This trend continued when selecting an install to modify, with the cache first going to the install directory and only writing to the "cache directory" after several restarts.
Changing to a different ksp install *required* a restart or the results were unpredictable.  After the first round of selecting mods to apply, going back for more mods - after clicking on one the app would go to the apply now screen with no way of going back to select more.
I am happy to keep testing.  If I were looking at this as a regular player, I would not use it.  I have not tried this version on my PC yet.  I will do so today.  The "regular" version works okay fine on my PC.
Sorry if this is too chatty.  I have not participated in this community process for decades.  If you want to asses my background before rating any feedback, email me outside of git.
It has been a while but if you need screenshots I can provide them (here or offline?) and just maybe I can record some screen video.

Genjokoan:
On the plus side, it works great on windows 10 pro.  *IF* you start with clean installs and double check the settings before you start accessing the cache.  I started with a test setup that had three installs in various states of mods applied and the results were unpredictable.  The first install I tested worked okay fine.  Switching to other installs, I could not get them to transition to the one, new cache.  My three new installs are working well after adding/removing mods on each many times.  I will put more effort into testing on a Mac later.

Olympic1:
wtf, how did this happen? I didn't push anything to this branch or repo. Now I lost everything

HebaruSan:
No idea, but that sucks, and here's a possibly usable backup:

https://github.com/HebaruSan/CKAN/tree/feature/shared-cache

HebaruSan:
Apologies for my previous message, I think that was a copy of my own changes on this topic. Here's what I have of yours:

- https://github.com/HebaruSan/CKAN/tree/olympic1-shared-cache

techman83:
If you know the commit IDs, you should be able to cherry pick them from the refs in your local repo. Checkout a new branch from master and cherry pick them all. It's hard to lose stuff, but it's also hard to find stuff sometimes :-/

Genjokoan:
All I have is the exe's you sent me.

politas:
This is what I have (I'm afraid it's fairly early) https://github.com/politas/CKAN/tree/feature/cache-global

politas:
I though GitHub let you restore deleted branches, though?

Olympic1:
Ok, got it back to where I had it and triggered it so it to reopen.

I don't know why it included the last merged PR but it should be fine.

Core/KSP.cs
```diff
@@ -321,9 +321,16 @@ public string CkanDir()
 
         public string DownloadCacheDir()
         {
-            return KSPPathUtils.NormalizePath(
-                Path.Combine(CkanDir(), "downloads")
-            );
+            var registry = RegistryManager.Instance(this).registry;
```
HebaruSan: This causes all of the instances of the registry to be loaded at start. Since a current registry is 20 MB, this would be quite a delay on systems with many instances. It also causes an unworkaroundable exception if you have an instance with a corrupted registry.json (as I do).

Tests/Core/Cache.cs
```diff
@@ -87,6 +94,100 @@ public void StoreRemove()
...
+            Assert.IsFalse(cache.MoveDefaultCache(""));
+            Assert.IsFalse(cache.MoveDefaultCache(" "));
+        }
+
+        [DllImport("libc", EntryPoint = "chmod", SetLastError = true)]
+        private static extern int Sys_chmod(string path, uint mode);
+
+        private const uint S_IRUSR = 0000400;
+        private const uint S_IWUSR = 0000200;
+        private const uint S_IXUSR = 0000100;
```
HebaruSan: Try `Convert.ToUInt32("0400", 8)`, etc. C# doesn't have octal numeric literals.

Core/Net/NetFileCache.cs
```diff
@@ -96,6 +111,54 @@ public string GetCachePath()
...
+            // Check which platform we're on, as the path depends on it
+            if (!Platform.IsWindows)
+            {
+                // We are on Linux or OSX, attempt to use the XDG standard for getting the path
+                baseDirectory = Environment.GetEnvironmentVariable("XDG_CACHE_HOME");
+
+                // If the above didn't work, fall back to "~/.local/"
+                if (string.IsNullOrWhiteSpace(baseDirectory))
+                {
+                    baseDirectory = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.Personal), ".local", "share");
```
HebaruSan: This is the same as `SpecialFolder.LocalApplicationData` on Linux as far as I can tell. Opportunity to consolidate with below code here.

PR #2192, opened 2017-11-26
Only check first three segments of version in ci (fixes #2191)

### LINKED ISSUE #2191: CI not Pushing Builds to S3
The latest build on S3 is from October 20. Noted in #2189 as Netkan Bot hadn't picked up the fixed Curse API.

```
leon@beast /tmp $ (rm netkan.exe || true) && wget -q http://ckan-travis.s3-website-us-east-1.amazonaws.com/netkan.exe && mono netkan.exe --version
rm: cannot remove 'netkan.exe': No such file or directory
v1.22.7-PRE+bd1c5daa6a5a
```

HebaruSan:
See my comment on #2191; the `mono --version` check is now printing 5.4.0.201 instead of just 5.4.0, and our perl regex extracts that whole string. This pull request changes the perl regex to match just the 5.4.0 part, so it should match BUILD_RELEASE_MONO_VERSION again.

This should get builds deploying to S3 again, which should fix the Curse errors on the status page.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ deploy:
       on:
         repo: KSP-CKAN/CKAN
         branch: master
-        condition: $BUILD_CONFIGURATION = Release && $(mono --version | perl -ne'/version (\S+)/ and print $1') = $BUILD_RELEASE_MONO_VERSION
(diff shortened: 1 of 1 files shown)


techman83:
Awesome, basically it's just ensuring we push up only one artifact by the looks. That should get the job, thanks for your awesome work @HebaruSan

PR #2195, opened 2017-11-28
Update version regex for release builds as well

HebaruSan:
Applying the fix from #2192 to release builds in addition to per-merge builds. The same condition was used in both places, so it would have stopped a release from building as well.


Olympic1:
I think this can be merged as @techman83 merged #2192

Issue #2186, opened 2017-11-14
[Feature] Convert AD mods to CKAN control


APAeroDan:
CKAN v1.22.6

Windows 7 sp1

modmanager was previously installed and autodetected with everything working fine
I selected modmanager from CKAN so that CKAN would update it automatically.
That part works fine.  It updated Modmanager, but killed most of my mods due to some automatic dependencies.
I understand why it did it and why autodetect won't overwrite anything from a safety policy standpoint.
its just a good idea.  But can we add a manual override to the autodetect in the instance that it is already installed, but we want it to be updated from CKAN?

I'm sorry if I'm doing this wrong.  I've never used github before.  I usually code by myself.

HebaruSan:
Possible duplicate of #949.

Issue #2180, opened 2017-11-10
[Feature] Right-click context menu on MainModList


politas:
If we can implement a right-click context menu on the main mod list, it could be the easy gateway to loads of extra functionality. 

**@Olympic1's progress:**
- [x] A clean implementation of mod replacements - #2671
- [x] A `Reinstall` option - #2202
- [x] Select specific version of a mod to install - #2364
- [x] A `Download contents` option - #2202
- [ ] A way to override the autodetected mods so CKAN can control them - #2186 / #949
- [x] A way to add an externally downloaded zip to the cache - #2246
- [ ] And if we ever want to add user tagging of mods, this would be the way to manipulate those:
- - [ ] A way to mark mods - #2176
- - [ ] A way to hide mods from all filters - #2082
- - [ ] An `Add to favourites` option - #1677
- - [ ] Set priority for mods - #1375

We could also display specific conflict information in the context menu
It would also be a good idea to add a "Status" column.

APAeroDan:
I think my issue would tie into this nicely. #2186

Olympic1:
I worked on this a bit and have it working now. What functions do you guys want for the next release?

APAeroDan:
I would love an autodetect override.  I've had several instances where I had the mod preinstalled and I wanted ckan to takeover.  Right now, I have to uninstall the mods before it will do that, and then it auto uninstalls any dependant mods.  So to fix, one has to go through 3 uninstall reinstall cycles... Gets old.  If there were any override, then it would bypass the dependency uninstall.

Olympic1:
Added it to the list and will investigate how to override the `AD` marked mods.

politas:
Feel free to do a PR with only partial functionality. Just getting the basic menu in place is the critical thing, then we can add to it. Thank you thank you thank you for working on this! I've been distracted by other things that have stopped me being able to dive into it.

politas:
I'd really like to use this for the GUI implementation in #1888

APAeroDan:
I've got some programming experience, but it is in firmware.  I've been
meaning to learn GUI.  What is this mod written in?

Olympic1:
> Feel free to do a PR with only partial functionality.

Ok, I'll work out the last bugs and will then make a PR.

> What is this mod written in?

The GUI is written in .NET's library `Windows Forms (WinForms)`

yalov:
-   A way to mark mods #2176
-   An Add to favourites option #1677
[...]

Issue #2176, opened 2017-11-06
[Feature] to mark mod


yalov:
Often you need not to install mod, but somehow to mark it.

For example, for testing in the future, for look into it later, for  keep in list nearby, for checking about release new version, etc.   
Third status of checkbox at middle-click would do this.

![2017-11-06_233746](https://user-images.githubusercontent.com/12796450/32462863-f82a7434-c34b-11e7-88e4-b5fd319780b2.png)

Issue #2154, opened 2017-10-19
Feature request: reinstall


Real-Gecko:
Would be nice to have "Reinstall" button, so that mod is removed and unpacked again, just in case user messes up with mod files.

politas:
Is removing and installing as a two step process  that much of a headache? We can clutter up the UI with separate buttons for every possibility, I suppose...

Real-Gecko:
Sometimes it happens that other mods depend on mod in question, so uninstalling it will trigger uninstallation of dependants. I agree that button is not quite a good idea, maybe right click dropdown?

politas:
Would you be happy with a Command Line implementation only?

politas:
Another possibility is to start using the Mod Versions panel to give us extra functionality - Install a specific version, reinstall current version, set AD version, download selected version...

Real-Gecko:
Shame to ask, but what is Mod Versions panel?

linuxgurugamer:
The panel on the right, where it shows you information on the mod, the tab is Versions

Real-Gecko:
Yep, I think that'll be pretty cool.

Olympic1:
I worked on a right click menu and could add this option to the menu. But we'll need a process for the `reinstall`. I thought of the following:

We uninstall the mod, so that we have clean install, and then install the mod again. When doing this it will remove its dependants, so could this be possible without removing its dependants?

Real-Gecko:
Well, maybe something like a dirty hack: remove mod folder, unpack mod from cache like it's done while installing the mod?

politas:
Should be able to do it like an upgrade, but to the same version instead of the new version.

solarsootysmudge:
As much as like the comments and suggestions so far. I got to bring up the whole policy of not deleting folders on uninstall thing again. I believe that that effects the above conversation and should be considered further.  

CKAN can't never be trusted to do a full reinstall with every mod without end users making a choice of how to handle config files created by the mod at run time. There are mods that install files **after** the CKAN process is completed. The policy rule has been if CKAN did not install it those extra files. It can't remove them.

[...]

PR #2202, opened 2017-11-30
Add right-click context menu

Olympic1:
As said in #2180, here is a right click context menu. I already added a `Download Contents` and a `Reinstall` option.

The `Reinstall` option is just a simple uninstall -> install process. I'll work on that a bit more to show the process in the `WaitTabPage` and maybe a possibility to uninstall the mod whitout uninstalling its dependants.

I also added a check for the left mouse button to double clicking.

**Possible implementations:**
- [ ] A clean implementation of mod replacements #1888 - Will be implemented by @politas 
- [x] A `Reinstall` option #2154 
- [ ] Select specific version of a mod to install #2077 - I'm having some issues with the right click menu
- [x] A `Download contents` option
- [ ] A way to override the autodetected mods so CKAN can control them #2186  
- [ ] A way to add an externally downloaded zip to the cache #1788 - @HebaruSan can you implement this as you did in your console?
- [ ] And if we ever want to add user tagging of mods, this would be the way to manipulate those:
- - [ ] A way to mark mods #2176 
- - [ ] A way to hide mods from all filters #2082 
- - [ ] An `Add to favourites` option #1677 
- - [ ] Set priority for mods #1375


HebaruSan:
Can Reinstall work for a mod that the user manually uninstalled? That's a large percentage of issue reports that come in, so it would be very helpful if we could simply say "right click ModX and choose Reinstall".

Olympic1:
I haven't actually tested that, but I need to work on reinstall anyway. But now time to go to bed.

GUI/Main.cs
```diff
@@ -1200,6 +1206,73 @@ private void reportAnIssueToolStripMenuItem_Click(object sender, EventArgs e)
...
+
+                // Show the context menu
+                ModListContextMenuStrip.Show(ModList, new Point(e.X, e.Y));
+
+                // Set the menu options
+                var guiMod = (GUIMod)ModList.Rows[rowIndex].Tag;
+
+                downloadContentsToolStripMenuItem.Enabled = !guiMod.IsCached;
+
+                if (guiMod.IsInstalled || guiMod.IsAutodetected)
```
politas: Properly handling AD mods is a whole can of worms, that I think would be better handled in a separate PR. There's a lot of different cases to consider. As this is, this throws a ModNotInstalledKraken for AD mods in ModuleInstaller.UninstallList that isn't being caught in the transaction below and thus crashes the client. Can we just make this `if ( guiMod.IsInstalled )`? Then I can merge this and we can all work on extending the right-click functionality one menu item at a time.

Olympic1:
Sorry for waiting, I had to do 12h weeks. Will make the changes

GUI/Main.cs
```diff
@@ -453,7 +427,7 @@ public void UpdateCKAN()
             SwitchEnabledState();
             ClearLog();
             tabController.RenameTab("WaitTabPage", "Updating CKAN");
-            SetDescription("Upgrading CKAN to " + AutoUpdate.Instance.LatestVersion);
+            SetDescription("Upgrading CKAN to {AutoUpdate.Instance.LatestVersion}");
```
HebaruSan: Missing the dollar sign before the staring quote, I believe this will print the literal curly braces as text.

GUI/Main.cs
```diff
@@ -193,10 +191,10 @@ public Main(string[] cmdlineArgs, GUIUser user, bool showConsole)
             }
 
             configuration = Configuration.LoadOrCreateConfiguration
-                (
-                    Path.Combine(CurrentInstance.GameDir(), "CKAN/GUIConfig.xml"),
-                    Repo.default_ckan_repo.ToString()
-                );
+            (
+                Path.Combine(CurrentInstance.GameDir(), "CKAN/GUIConfig.xml"),
```
HebaruSan: Let's use `CurrentInstance.CkanDir()`.

GUI/Main.cs
Olympic1: How did VS screw that up?

Olympic1:
PR should be good to go now.

PR #2204, opened 2017-12-03
Deploy AutoUpdater.exe to S3 and use it to auto-update (fixes #2140)

### LINKED ISSUE #2140: Updater can't deal with spaces in path
**CKAN Version:** 1.22.4

**Operating System:** Arch Linux with Linux 4.13.4

I opened CKAN and was told there was an update. I opted to install said update and it seems to have worked fine. However, I believe the updater was meant to start CKAN again after the update, but all I got was a message on the terminal:
```
Cannot open assembly '/media/ssd1/steam/SteamApps/common/Kerbal': No such file or directory.
```

This looks to me like it did not quite understand the spaces in the directory name. (The full path would probably be `/media/ssd1/steam/SteamApps/common/Kerbal Space Program/ckan.exe`

HebaruSan:
The auto-updater is, ironically enough, not up to date.

`Core/Net/AutoUpdate.cs` tries to download the latest autoupdater exe file. It does this by looking up the latest release on the CKAN-autoupdate repo. This release was published on June 30, 2015, after which that repo was merged into the main CKAN repo and marked as retired. So a 2-year-old file is used instead of the latest code.

A few days after that release was published, on July 2, 2015, [a fix was made to the auto-updater](https://github.com/KSP-CKAN/CKAN/commit/8b882ceb460cffdd3dd08359c7311988a3ba3aad), but it did not take effect. #2140 reports the exact symptoms that that commit was trying to fix.

This pull request tries to bring the latest auto-update code into active service:

- The build process is updated to copy AutoUpdater.exe into the repack directory
  - Which means the deploy process in .travis.yml will copy it to ckan-travis.s3.amazonaws.com
  - This should mean that the latest auto-updater will be available on the internet to clients
- The autoupdate code in `Core/Net/AutoUpdate.cs` is updated to no longer deal with the retired CKAN-autoupdate repository
  - Instead, it downloads AutoUpdater.exe directly from where the build process puts it on the S3 site, like NetKAN-bot does
  - This should mean that the latest auto-updater will be used instead of the one from June 2015

Fixes #2140.

### DIFF
--- a/Core/Net/AutoUpdate.cs
+++ b/Core/Net/AutoUpdate.cs
@@ -21,10 +21,8 @@ namespace CKAN
 
         private readonly Uri latestCKANReleaseApiUrl = new Uri("https://api.github.com/repos/KSP-CKAN/CKAN/releases/latest");
 
-        private readonly Uri latestUpdaterReleaseApiUrl = new Uri(
-            "https://api.github.com/repos/KSP-CKAN/CKAN-autoupdate/releases/latest");
+        private readonly Uri lastestUpdaterUrl = new Uri("https://ckan-travis.s3.amazonaws.com/AutoUpdater.exe");
 
-        private Tuple<Uri, long> fetchedUpdaterUrl;
         private Tuple<Uri, long> fetchedCkanUrl;
 
         public Version LatestVersion { get; private set; }
@@ -59,7 +57,7 @@ namespace CKAN
         /// </summary>
         public bool IsFetched()
         {
-            return LatestVersion != null && fetchedUpdaterUrl != null &&
+            return LatestVersion != null &&
                 fetchedCkanUrl != null && ReleaseNotes != null;
         }
 
@@ -73,7 +71,6 @@ namespace CKAN
 
             try
             {
-                fetchedUpdaterUrl = RetrieveUrl(MakeRequest(latestUpdaterReleaseApiUrl));
                 fetchedCkanUrl = RetrieveUrl(response);
             }
             catch (Kraken)
@@ -126,7 +123,7 @@ namespace CKAN
             string updaterFilename = System.IO.Path.GetTempPath() + Guid.NewGuid().ToString() + ".exe";
             string ckanFilename = System.IO.Path.GetTempPath() + Guid.NewGuid().ToString() + ".exe";
             Net.DownloadWithProgress(new[]{
-                new Net.DownloadTarget(fetchedUpdaterUrl.Item1, updaterFilename, fetchedUpdaterUrl.Item2),
+                new Net.DownloadTarget(lastestUpdaterUrl, updaterFilename),
                 new Net.DownloadTarget(fetchedCkanUrl.Item1, ckanFilename, fetchedCkanUrl.Item2),
             }, user);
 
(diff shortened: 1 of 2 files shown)


Core/Net/AutoUpdate.cs
```diff
@@ -21,10 +21,8 @@ public class AutoUpdate
 
         private readonly Uri latestCKANReleaseApiUrl = new Uri("https://api.github.com/repos/KSP-CKAN/CKAN/releases/latest");
 
-        private readonly Uri latestUpdaterReleaseApiUrl = new Uri(
-            "https://api.github.com/repos/KSP-CKAN/CKAN-autoupdate/releases/latest");
+        private readonly Uri lastestUpdaterUrl = new Uri("https://ckan-travis.s3.amazonaws.com/AutoUpdater.exe");
```
politas: I would much prefer it if we could make this URL in our own ksp-ckan.org domain. @techman83, can you set up a cname entry (say "travis") pointing to the amazonaws.com address so we don't hard-code an external URL?

techman83: Hmm, we lose https in the process of setting a custom domain. Unless we go to the effort of configuring up CloudFront.

Now, we should be aware; We pay for the transit from AWS and we don't currently have a budget for large amounts of downloading from our infrastructure.

HebaruSan: Hmm, in that case, maybe we should keep this download on github.com.

How can we get a cleanly built version of AutoUpdate.exe? I could put a new release on the CKAN-autoupdate repo now, which would also fix #2140, but I'd have to use my locally built file which may or may not be safe depending on mono versions, etc.

techman83: I'm not super across the Autoupdater stuff, but there are ways to have travis release a build onto GitHub.

https://docs.travis-ci.com/user/deployment/releases/

HebaruSan: I believe that's already how CKAN's releases are done, but I'm thinking that adding another exe file to the main release download list would confuse users. "Do I need to download this AutoUpdate.exe? Will CKAN work without it?" Better to handle something like that in the background, as it is now.

techman83: Possibly. We could setup a specific repo for it though, with a Readme stating its purpose.

politas: That sounds like the optimal situation. Revive CKAN-autoupdate repo and use it to push releases of the autoupdater, with a readme and release notes that make it clear this is not something that users should need to install manually. The release script could handle that without any problem?

HebaruSan: I think the release script can't do cross-repo uploads. This one runs on KSP-CKAN/CKAN and can handle releases in that repo, but not KSP-CKAN/CKAN-autoupdate. At least, I'm not able to find documentation or informal discussion of such.

Which implies either manually created releases, or reviving the _code_ of CKAN-autoupdate as well.

politas: Well, I think I'd rather have the Autoupdater included in the main repo releases than have a manual process

HebaruSan: OK, I'll look into that. Currently the auto-update code in ckan assumes the new exe is the first download in the list, so we would either need to control the ordering or add some smarter filtering so it can tell them apart.

I think I'm going to close this PR because we've pretty much decided against this approach, so any further update would pretty much blow these changes away and replace them. Might as well start fresh.

HebaruSan:
Closing in favor of a fresh start not based on S3.

PR #2205, opened 2017-12-05
Avoid NRE in install-from-ckan (fixes #2171)

### LINKED ISSUE #2171: "Install from .ckan" does not catch Krakens
To recreate, make a .ckan file with a spec_version higher than your client and try to install it. The CKAN client closes with no message shown.

HebaruSan:
Fixes #2171; see that issue for details. To sum up, loading a .ckan file can raise exceptions, and currently they're not handled well. In Cmdline they are caught by the uncaught exception handler, and in GUI the exception handler itself causes a null reference exception.

Now in both cases we properly print out the error using an IUser reference.

### DIFF
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -65,7 +65,16 @@ namespace CKAN.CmdLine
                     }
 
                     // Parse the JSON file.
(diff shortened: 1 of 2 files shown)


politas:
Nice work!

PR #2211, opened 2017-12-09
Install version from file when installing from file (fixes #1356)

### LINKED ISSUE #1356: 'ckan install -c' installs latest version in repository instead of the ckan supplied
Highlighted during metadata testing.

```
+ echo Running ckan install -c TriggerAu-Flags/TriggerAu-Flags-2.7.1.0.ckan
Running ckan install -c TriggerAu-Flags/TriggerAu-Flags-2.7.1.0.ckan
+ mono --debug ckan.exe install -c TriggerAu-Flags/TriggerAu-Flags-2.7.1.0.ckan --headless
About to install...

 * TriggerAu-Flags 2.7.2.0
```

HebaruSan:
The `ckan install -c GreatMod.ckan` command loads the file into the registry, and then installs the latest version of GreatMod. This is fine if the file contains the latest version, but if it is an older version, then the command doesn't actually install what's specified in the file.

This pull request updates the install command to include the version from the file when adding to the install list.

Fixes #1356.

### DIFF
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -67,7 +67,8 @@ namespace CKAN.CmdLine
                     // Parse the JSON file.
                     try
                     {
-                        options.modules.Add(LoadCkanFromFile(ksp, filename).identifier);
+                        CkanModule m = LoadCkanFromFile(ksp, filename);
+                        options.modules.Add($"{m.identifier}={m.version}");
                     }
                     catch (Kraken kraken)
                     {


Cmdline/Action/Install.cs
```diff
@@ -67,7 +67,8 @@ public int RunCommand(CKAN.KSP ksp, object raw_options)
                     // Parse the JSON file.
                     try
                     {
-                        options.modules.Add(LoadCkanFromFile(ksp, filename).identifier);
+                        CkanModule m = LoadCkanFromFile(ksp, filename);
+                        options.modules.Add($"{m.identifier}={m.version}");
```
politas: Is this going to throw a Null ref if the .ckan file is incomplete?

HebaruSan: No; LoadCkanFromFile doesn't return null. If there's any problem with the file, it'll throw some other exception.

techman83:
Oh man, I remember this bug. Found it while working on our Metadata CI! I got very confused initially. Awesome work @HebaruSan

Cmdline/Action/Install.cs
politas: Not an entire null, I mean if the .ckan file is missing a version

HebaruSan: The version field is required, so the JSON deserializer throws an exception inside LoadCkanFromFile if it's missing. I deleted the version from this file for testing:

```
$ mono ckan.exe install -c Astrogator-v0.7.8.ckan  --ksp steam
JSON deserialization error: Required property 'version' not found in JSON. Path '', line 67, position 1.: Required property 'version' not found in JSON. Path '', line 67, position 1.
Usage: ckan install [--with-suggests] [--with-all-suggests] [--no-recommends] [--headless] Mod [Mod2, ...]
```

politas: Cool. Thanks for checking that.

politas:
Ah, drat! That commit was supposed to mark this as merged. Should have reset --hard and re-merged, it seems!

PR #2212, opened 2017-12-09
Provide fresh auto updater in releases

### LINKED ISSUE #2140: Updater can't deal with spaces in path

HebaruSan:
This pull request replaces #2204, which was closed because its approach wasn't appropriate. See that PR for details of the problem being solved.

New approach:

- Add AutoUpdater.exe to the list of downloadable files when creating a new release on GitHub
- Update client to look for this file and use it, falling back to the old URL if it's missing

This change assumes that the list of assets in the release will be in the same order as they are in the `.travis.yml` file. I believe this would be the case because it's an array rather than a dictionary, and the easiest way for Travis to implement it would be a simple `foreach` loop, but I was not able to find documentation or a concrete example to confirm this.

We should mention in the next release notes that users do _not_ need to download AutoUpdater.exe manually.

Partially fixes #2140. The error will still occur when updating _to_ this release, because the old clients will still use the 2-year-old auto updater exe, but updating from this release to the next release will use the latest auto update code. If we publish one last manual release to https://github.com/KSP-CKAN/CKAN-autoupdate/releases , then the current clients will update properly as well.

### DIFF
--- a/Core/Net/AutoUpdate.cs
+++ b/Core/Net/AutoUpdate.cs
@@ -20,10 +20,16 @@ namespace CKAN
 
         private static readonly ILog log = LogManager.GetLogger(typeof(AutoUpdate));
 
+        /// <summary>
+        /// The list of releases containing ckan.exe and AutoUpdater.exe
+        /// </summary>
         private static readonly Uri latestCKANReleaseApiUrl = new Uri("https://api.github.com/repos/KSP-CKAN/CKAN/releases/latest");
 
-        private static readonly Uri latestUpdaterReleaseApiUrl = new Uri(
-            "https://api.github.com/repos/KSP-CKAN/CKAN-autoupdate/releases/latest");
+        /// <summary>
+        /// Old release list that just contains the auto updater,
+        /// used as a fallback when missing from main release
+        /// </summary>
+        private static readonly Uri oldLatestUpdaterReleaseApiUrl = new Uri("https://api.github.com/repos/KSP-CKAN/CKAN-autoupdate/releases/latest");
 
         private Tuple<Uri, long> fetchedUpdaterUrl;
         private Tuple<Uri, long> fetchedCkanUrl;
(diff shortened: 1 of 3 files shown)


Core/Net/AutoUpdate.cs
```diff
@@ -73,8 +79,15 @@ public void FetchLatestReleaseInfo()
 
             try
             {
-                fetchedUpdaterUrl = RetrieveUrl(MakeRequest(latestUpdaterReleaseApiUrl));
-                fetchedCkanUrl = RetrieveUrl(response);
+                fetchedCkanUrl = RetrieveUrl(response, 0);
+                // Check whether the release includes the auto updater
+                if (response.assets.Count >= 2) {
+                    // Second asset is AutoUpdater.exe
```
politas: The CKAN.app is now the second asset, and aren't we going to add in a .deb as well? Can we use a regex on the assets to find the correct one, rather than hard-code the position?

HebaruSan: Yup, this needs to be updated every time we add to that list (which hopefully won't be too many more times). It was 1 before the mac PR got merged, and I'll be incrementing it again now that the deb project is in. Luckily GitHub helpfully notices that these PRs all conflict with one another, which ensures that we don't merge without reconciling them.

HebaruSan:
After the latest commit, it now checks the URLs to find the auto updater rather than assuming a particular index.

Issue #1859, opened 2016-08-06
KSP Installation list loses currently invalid installs when adding a new installation


politas:
Steps to reproduce
Add a KSP installation "install1"
Move KSP installation "install1" to a different location
Add KSP installation "install2"
Return KSP installation "install1" to its former location

install1 will not be selectable on starting CKAN

politas:
See also related #1836

ayan4m1:
This is intentional as far as the source is concerned: https://github.com/KSP-CKAN/CKAN/blob/master/Core/KSPManager.cs#L290

politas:
Ideally, unavailable instances would show up in the KSP Install list greyed out, so they can be removed (fixing #1836), and they won't then be forgotten when another instance is added (fixing this)

ayan4m1:
Ah, makes sense - that is kinda more of an enhancement rather than a ckan.dll bug right?

politas:
I suppose you could call it an enhancement that solves two related bugs.

politas:
@ayan4m1 , while you're working on this bit of code, we've had a feature request on the forums to maintain the "default" KSP version when selecting a non-default install. We would then need to add a button to clear the current default. Does that seem simple enough to implement?

PR #2214, opened 2017-12-14
Remove invalid installs from registry at load

### LINKED ISSUE #896: No way in ckan to 'forget' ksp installs that are moved or deleted outside of ckan
<a href="https://github.com/tjanke42"></img></a> **Issue by [tjanke42](https://github.com/tjanke42)**
_Tuesday Jan 27, 2015 at 06:12 GMT # Sample: Friday Sep 13, 2013 at 22:58 GMT_
_Originally opened as https://github.com/KSP-CKAN/CKAN-support/issues/72_

---

Example:
1. Add two KSP installs to CKAN - e.g. KSP-A and KSP-B.
2. Decide you no longer want KSP-B, delete its folder.
3. List the installs - 'ckan.exe ksp list' 
   - CKAN complains that KSP-B is not a valid install  (error message is "WARN CKAN.KSPManager (null) - KSP-B at <path> is not a valid install", where <path> is the path to the missing install).
   - only install KSP-A is listed
4. (Try to) tell CKAN to forget the missing install:
   - "ckan.exe ksp forget KSP-B"
     - response is "Couldn't find install with name KSP-B, aborting"
5. CKAN will continue to complain about the missing install, but will not allow you to 'forget' it

To resolve my version of this problem, I had to go into the windows registry, HKEY_CURRENT_USERS\Software\CKAN, delete the offending KSPInstanceName & KSPInstancePath keys and update the KSPInstanceCount key.

### LINKED ISSUE #1836: Can't "forget" an install that has been removed on disk
**CKAN Version:** 

v1.18.1-0-g0d2c4f0

**Operating System:** 

MacOS 10.11.4

```
% ckan ksp forget rss133
194 [1] WARN CKAN.KSPManager (null) - rss at /Users/lamont/ksp_rss is not a vaild install
204 [1] WARN CKAN.KSPManager (null) - rss133 at /Users/lamont/ksp_rss_1.3.3 is not a vaild install
Couldn't find install with name "rss133", aborting..
```

ckan first prunes out all the invalid installs, then when you try to 'forget' the install it can't find it in its list to remove...

HebaruSan:
If a KSP instance that CKAN knows about is removed manually, an error message is printed at startup:

> 194 [1] WARN CKAN.KSPManager (null) - rss at /Users/lamont/ksp_rss is not a vaild install

But it's impossible to remove it from the list of known installs, because it's not actually added to the list that is manipulated to make that happen.

This pull request updates the loading code to save the list of (valid) instances to the registry again if there were any that were invalid. To alert the user that this is happening, the message now says "removing" at the end:

> 194 [1] WARN CKAN.KSPManager (null) - rss at /Users/lamont/ksp_rss is not a valid install, removing

Fixes #896.
Fixes #1836.

### DIFF
--- a/Core/KSPManager.cs
+++ b/Core/KSPManager.cs
@@ -279,6 +279,7 @@ namespace CKAN
         {
             log.Debug("Loading KSP instances from registry");
 
+            bool needSave = false;
             instances.Clear();
 
             foreach (Tuple<string, string> instance in Win32Registry.GetInstances())
@@ -293,13 +294,9 @@ namespace CKAN
                 }
                 else
                 {
-                    log.WarnFormat("{0} at {1} is not a vaild install", name, path);
+                    log.WarnFormat("{0} at {1} is not a valid install, removing", name, path);
+                    needSave = true;
                 }
-
-                //var ksp = new KSP(path, User);
-                //instances.Add(name, ksp);
-
-
             }
 
             try
@@ -310,6 +307,12 @@ namespace CKAN
             {
                 log.WarnFormat("Auto-start instance was invalid: {0}", e.Message);
                 AutoStartInstance = null;
+            }
+
+            if (needSave)
+            {
+                // Save the list without the invalid entries
+                Win32Registry.SetRegistryToInstances(instances, AutoStartInstance);
             }
         }
     }


politas:
It already does this under certain situations (#1859) and we have a bug request for that. I don't think we want to just automatically delete currently invalid installs. There could just be an unmounted drive.

It would be better if we loaded the invalid installs into the list (but marked them as inactive) and allowed users to remove them.

HebaruSan:
My thought was that supporting invalid game instances would be high risk with low benefit. We would have to add more downstream validity checks in the KSP class (i.e., run them every time you try to use it rather than just when you try to load it, to avoid installing mods in some random or missing directory), then audit all code that uses a KSP object and add exception handling and error outputs, as well as alternate UI formats like graying out invalid installs in GUI.

This would give the user the opportunity to "put back" an invalid instance if they didn't mean to move/delete it. But they could already do that and just re-add the instance if that's what they want. So in exchange for a minor convenience, we've made a lot of code more complex and failure-prone.

politas:
The only real difficulty at the moment is having orphaned installs in the registry, correct? At the moment we load the whole set of valid KSP installs in from the registry on boot, and then maintain that collection. But why? All our operations are done only on a single instance. We ought to be moving to having only a single install of KSP held in memory, and then read the registry when users ask to change installs, and in the initial load where we load either the specified install, the default, or the only valid one. Then we can flag invalid installs in the "Choose KSP Install" dialog and `ksp list` command without any further downstream issues.

HebaruSan:
Interesting point, I hadn't thought of it that way before. I'll take another look at this with that in mind...

HebaruSan:
Alright, replacement PR coming up...

PR #2215, opened 2017-12-15
Add depending mod to missing dependency exception

HebaruSan:
Currently `ModuleNotFoundKraken` is thrown if a dependency isn't met, but this only specifies the mod that can't be installed, not the one that depends on it. This means that if the depending mod is specified indirectly somehow, the user is not told how to avoid the error.

This pull request adds a new Kraken-inheriting class that specifies both the unavailable mod as well as the depending mod, and throws it from the relationship resolver. To maintain backwards compatibility, this new class inherits from `ModuleNotFoundKraken`. We also catch this Kraken in Cmdline, ConsoleUI, and GUI, and extend the messages to the user accordingly. And an identical error-level log message is moved to verbose-level, because we're already throwing an exception.

Before (using KSP 1.3.1):

```
$ mono ckan.exe install FASA --ksp steam
2989 [1] ERROR CKAN.RelationshipResolver (null) - Dependency on TextureReplacer found but it is not listed in the index, or not available for your version of KSP.
Module TextureReplacer required but it is not listed in the index, or not available for your version of KSP.
```

After:

```
$ mono ckan.exe install FASA --ksp steam
FASA 1:v7.2 requires TextureReplacer but it is not listed in the index, or not available for your version of KSP.
```

### DIFF
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -111,6 +111,13 @@ namespace CKAN.CmdLine
                 var installer = ModuleInstaller.GetInstance(ksp, user);
                 installer.InstallList(options.modules, install_ops);
             }
+            catch (DependencyNotSatisfiedKraken ex)
+            {
+                user.RaiseMessage("{0} requires {1} but it is not listed in the index, or not available for your version of KSP.", ex.parent, ex.module);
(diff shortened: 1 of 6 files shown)


politas:
6 tests failing

Issue #1646, opened 2016-04-01
Improved Availability Status


dbent:
See KSP-CKAN/NetKAN#3511

Even though `USI-Core` is compatiable with KSP 1.1.0 it is marked _unavailable_ because one of its dependencies, `FirespitterCore` is not compatible with KSP 1.1.0. To be consistent this should mean that all mods that are compatible with KSP 1.1.0 but depend on `USI-Core` should also be marked _unavailable_ because one of their 2nd degree dependencies is unavailable. In other words, availability status should be determined by traversing the entire dependency tree, not just direct dependencies.

On a related note, we should have some mechanism by which users can see _why_ a mod is unavailable. So for instance you could do something like:

```
> ckan unavailable
Mods unavailable for KSP 1.1.0

* FirespiterCore (v7.1.5) - Firespitter Core
  Incompatible KSP version: 1.0.5
* USI-Core (0.2.0.0) - USI Core
  Unavailable dependency: FirespitterCore
* USI-NuclearRockets (0.2.0.0) - 'Project Orion' Nuclear Pulse Engine
  Unavailable dependency: USI-Core
```

To keep the list managable the `unavailable` command should only consider the _latest_ version of any unavailable mod (No one really cares why a mod released 2 years ago is unavailable with the latest KSP).

ToxicFrog:
I was just coming here to post this bug after spending ten minutes tracking down why most of the USI mods aren't showing up despite having been indexed by NetKAN. Having mods that are compatible, but uninstallable due to unresolved dependencies, show up as such in the list rather than being hidden completely would have helped.

plague006:
https://github.com/KSP-CKAN/CKAN/issues/1695

PR #2216, opened 2017-12-15
Available ∪ Incompatible = All

### LINKED ISSUE #1695: "Compatible" mods with incompatible dependencies aren't listed in any filters.
**CKAN Version:** 1.6.1-0-g2e91715

**Operating System:**  W10/64

**The issue you are experiencing:** search for "FAR/ferram ..." is empty for KSP1.1.2

**How to recreate this issue:** select all, search name 'ferram'

**CKAN error codes (if applicable):** -

the registry.json is valid, has all FAR versions, up to 3:0.15.6.3
if I switch ckan to a KSP 1.1.0 install, I can find/install the correct 3:0.15.6.2
if I switch to a 1.1.2 install, FAR does not show up anywhere, in no list/search.

it does not autodetect it, either (offers it as suggested install though)

I have not pinned down other mods, but I suspect others may be affected.

clipboard01

HebaruSan:
Previously, `Registry.Available` checked for the availability/compatibility of mods' dependencies, but `Registry.Incompatible` only looked at the KSP versions. This meant that mods with compatible KSP versions but with incompatible dependencies were excluded from both of those lists. Since these two lists of mods are used to generate the full list of mods to show in the All filter of GUI, such mods would simply vanish and be completely invisible to the user.

Now the dependency checking logic is refactored into a private helper function called by both `Registry.Available` and `Registry.Incompatible`. Mods with incompatible dependencies will themselves appear in `Registry.Incompatible`, and GUI's All filter will actually show all mods, as expected.

Fixes #1695.

### DIFF
--- a/Core/Registry/Registry.cs
+++ b/Core/Registry/Registry.cs
@@ -418,43 +418,12 @@ namespace CKAN
             {
                 CkanModule available = LatestAvailable(candidate, ksp_version);
 
-                if (available != null)
+                if (available != null
+                    && allDependenciesCompatible(available, ksp_version, modules_for_current_version))
                 {
-                    // we need to check that we can get everything we depend on
-                    bool failedDepedency = false;
-
-                    if (available.depends != null)
-                    {
-                        foreach (RelationshipDescriptor dependency in available.depends)
-                        {
-                            try
-                            {
-                                if (!LatestAvailableWithProvides(dependency.name, ksp_version, modules_for_current_version).Any())
-                                {
-                                    failedDepedency = true;
-                                    break;
-                                }
-                            }
-                            catch (KeyNotFoundException e)
-                            {
-                                log.ErrorFormat("Cannot find available version with provides for {0} in registry", dependency.name);
-                                throw e;
-                            }
-                            catch (ModuleNotFoundKraken)
-                            {
-                                failedDepedency = true;
-                                break;
-                            }
-                        }
-                    }
-
-                    if (!failedDepedency)
-                    {
-                        compatible.Add(available);
-                    }
+                    compatible.Add(available);
                 }
             }
-
             return compatible;
         }
 
@@ -463,8 +432,13 @@ namespace CKAN
         /// </summary>
         public List<CkanModule> Incompatible(KspVersionCriteria ksp_version)
         {
-            var candidates = new List<string>(available_modules.Keys);
+            var candidates   = new List<string>(available_modules.Keys);
             var incompatible = new List<CkanModule>();
+
+            CkanModule[] modules_for_current_version = available_modules.Values
+                .Select(pair => pair.Latest(ksp_version))
+                .Where(mod => mod != null)
+                .ToArray();
 
             // It's nice to see things in alphabetical order, so sort our keys first.
             candidates.Sort();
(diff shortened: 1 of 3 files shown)


politas:
There's something weird going on, here. I've set up a repo - https://github.com/politas/CKAN-meta/raw/test-deps/TEST-meta-deps.tar.gz with the following:

Sarkomand-v1.0  compatible with KSP 1.3.1, depends on SamBelangerFlags
SamBelangerFlags-1 compatible with KSP 1.3, depends on KSP-Slovakia-flags
KSP-Slovakia-flags-0.0.1, compatible with KSP 1.3.0

Loading this Repo against a KSP 1.3.1 install, SamBelangerFlags loses its dependency to KSP-Slovakia-flags and shows as compatible, rather than everything showing as incompatible as it should.

HebaruSan:
Hmm, in Cmdline and ConsoleUI I see just Sarkomand, which is what I'd expect (dependencies are only checked one layer deep as per #1646, both before and after this change; I am not trying to move any currently-Available mods into Incompatible). I'll look into whatever strange thing GUI is doing...

HebaruSan:
OK, GUI doesn't automatically consider mods in `Registry.Incompatible` to be incompatible. Instead it checks the KSP versions, which explains what you're seeing:

https://github.com/KSP-CKAN/CKAN/blob/master/GUI/GUIMod.cs#L76

My goal here is just to get mods like this to _show up at all_; I am not particularly excited about the prospect of screwing around with GUI's existing filters for this. I'll go back and revise the pull request description to make it a bit more vague about exactly where the previously invisible mods show up.

politas:
Ok, that now has the mod with a first-level dependency showing up as incompatible, but it has still lost the "depends" relationship, so there's nothing to show _why_ it's incompatible. If we combine with #2215, users should at least get a better message about why the "Compatible" mod can't be installed. I'm happy to merge this and raise a new issue for the missing dependency in the relationship tab.

HebaruSan:
Hmm, give me another hour or three. There's probably some additional check in GUI that needs to be updated while it's building the dependencies.

politas:
No prob. I've got to go to work and I'm out at a milonga tonight, so I won't be doing any more testing/merging for at least 36 hours or so.

HebaruSan:
Cool, have fun!

I've confirmed that problem is not related to this PR. In the released version, the dependency graph loader only shows dependencies that are available; if `Registry.LatestAvailable` and `Registry.LatestAvailableWithProvides` both fail here, it just gives up and leaves it out:

https://github.com/KSP-CKAN/CKAN/blob/fd270ebcea8332c01288fe44a6bcfa148b72223f/GUI/MainModInfo.cs#L190-L201

This version-incompatible mod has 2 required dependencies, but one isn't shown in GUI because it's not compatible (in this case, with 1.3.0):

https://github.com/KSP-CKAN/CKAN-meta/blob/7b70dd4d2e8f6979e4db43762f64ec17af24eb69/AV8HarrierWITHTRUEVTOLFlight/AV8HarrierWITHTRUEVTOLFlight-104.105.ckan#L15-L22

image

This should be tracked and fixed separately rather than sneaking it into this PR; somehow I don't think there's an issue for it yet (!). Maybe no one looks at the relationships tab too closely?

PR #2225, opened 2017-12-16
Create and release CKAN.app for Mac OS X

### LINKED ISSUE #1989: need support for a quick way for Mac users to use CKAN (CKAN.app)
The repo for this modification is [here](https://github.com/Maxzhao1999/CKAN.app).
I'd love to merge this into the CKAN release, however I'm new to Github so please help, thanks a lot!

HebaruSan:
Mac support is a bit spotty currently, with users resorting to launching the GUI from Terminal.app, and user-created CKAN.app package running GUI from Terminal.app via AppleScript.

- https://github.com/KSP-CKAN/CKAN/wiki/Installing-CKAN-on-OSX
- https://github.com/Maxzhao1999/CKAN.app
- https://forum.kerbalspaceprogram.com/index.php?/topic/154888-macckanapp/

This pull request adds a CKAN.app download to releases, generated in the style of #2187.

Advantages over the user-created package:

- Contains ckan.exe internally rather than a separate download
  - Avoids problems associated with _finding_ ckan.exe, such as quoting spaces in paths
- Terminal.app not required
- Uses standard icon from assets folder

**Manual separate installation of Mono is still required.**

**TESTING REQUIRED, I DO NOT HAVE A MAC!**

Sample build for testing:

- [CKAN.dmg.zip](https://github.com/KSP-CKAN/CKAN/files/1574349/CKAN.dmg.zip)

Fixes #1989.


politas:
One for @KSP-CKAN/macos-testers to have a look at

jwwolff:
So I tested it and it won't launch(gives the error "the application 'CKAN' can't be opened). My guess is your creation of the .app didn't work correctly.

HebaruSan:
Thanks, @jwwolff !
I think that means the launch script isn't executable, which I probably should have thought of before now.
I've updated the download. Could you please try it again?

jwwolff:
Still doesn't work. Same error.

jwwolff:
(I suggest using a program called platypus if you can't figure out the problem. It might help)
p.s. thanks for doing this it'd be super nice to have a .app for Mac users

HebaruSan:
Well luckily this time the error was simple: the new zip was the same as the old zip. I've updated the download again so at least this time you can test the change I _meant_ to make.

jwwolff:
So good news and bad news. Good news, the app is openable and the executable inside the app does launch ckan when launched manually from terminal but bad news the app when opened just leads to the installing ckan on OSX web page. ([](https://github.com/KSP-CKAN/CKAN/wiki/Installing-CKAN-on-OSX)[](url))

HebaruSan:
OK, that's what it's supposed to do if it can't find mono.

If you open a terminal on your system and run "which mono", does it print the correct location?

ProfFan:
Probably it should find mono in its launch script. Then you can set the environment variables for finding mono.

politas:
Ok, now I'm confused - @jwwolf made a new release as mentioned in [this forum post](https://forum.kerbalspaceprogram.com/index.php?/topic/154888-macckanapp/&do=findComment&comment=3247976) - Is that what we have here, or an is it an improved version?

HebaruSan:
This is different from jwwolff's. jwwolff has created one download manually containing one version of ckan.exe, on a repo not owned by KSP-CKAN. This pull request automates the process and adds the new downloads to the main asset list of the releases, so each new version of ckan.exe would be available for Mac.

@ProfFan , it _is_ trying to find mono in the launch script, unless I'm not understanding what you mean by that. Are you saying there are environment variables that will take care of that for us? That would be excellent, any knowledge you can provide on that would be very much appreciated!

politas:
Sorry, I wasn't clear, what I should have said is, does this create a .app file that has all the features @jwwolff has put in his?

HebaruSan:
Ahh, I see. Let me go through his list from that post...

- [X] doesn't use automator
- [X] fixes space handling in the path
- [X] doesn't require a separate download of ckan.exe
- [X] works like any normal Mac app does

Yes, I think we're at feature parity.

politas:
Now I'm wondering whether we should add a Windows installer as well!

PR #2228, opened 2017-12-18
Check game version compatibility when installing specific version

HebaruSan:
## Problem

`ckan install identifier=version` currently bypasses game version compatibility checks. This means that an incompatible module that you normally couldn't install at all, can be installed just by specifying any version:

```
$ mono ckan.exe install AvisorforValentina
Module AvisorforValentina required but it is not listed in the index, or not available for your version of KSP.
```

```
$ mono ckan.exe install AvisorforValentina=1.01
About to install...

 * A visor for Valentina 1.01

Continue? [Y/n] 
```

This makes the `identifier=version` format unsafe by default, since compatibility checking is just turned off; this could be a problem if I intend to install an older _compatible_ version but make a typo in the version number.

It also isn't documented anywhere that it works this way, so a user might not even realize that he's installing incompatible versions once he discovers this argument format.

## Changes

This pull request checks that the version the user specified is compatible with the game version. It also adds the module version to the error message:

```
$ mono ckan.exe install AvisorforValentina=1.01
Module AvisorforValentina 1.01 required but it is not listed in the index, or not available for your version of KSP.
Or `ckan install --allow-incompatible` to ignore module compatibility.
```

To avoid losing the ability to install incompatible mods if the user so chooses, a new explicit Cmdline option `--allow-incompatible` is added:

```
$ mono ckan.exe install --help
CKAN 1.24.0-PRE+5cf8b5017bc5
Copyright © 2014-2017
 
install - Install a KSP mod
Usage: ckan install [options] modules

  --allow-incompatible    (Default: False) Install modules that are not 
                          compatible with the current game version
```

This works for _both_ the `identifier=version` format and the simpler `identifier` format, so you now don't have to know the version number to install the latest incompatible version of a mod:

```
$ mono ckan.exe install AvisorforValentina=1.01 --allow-incompatible
About to install...

 * A visor for Valentina 1.01

Continue? [Y/n] 
```

```
$ mono ckan.exe install AvisorforValentina --allow-incompatible
About to install...

 * A visor for Valentina 1.01

Continue? [Y/n] 
```

### DIFF
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -120,7 +120,14 @@ namespace CKAN.CmdLine
             }
             catch (ModuleNotFoundKraken ex)
             {
-                user.RaiseMessage("Module {0} required but it is not listed in the index, or not available for your version of KSP.", ex.module);
+                if (ex.version == null)
+                {
+                    user.RaiseMessage("Module {0} required but it is not listed in the index, or not available for your version of KSP.", ex.module);
+                }
+                else
+                {
+                    user.RaiseMessage("Module {0} {1} required but it is not listed in the index, or not available for your version of KSP.", ex.module, ex.version);
+                }
                 user.RaiseMessage("If you're lucky, you can do a `ckan update` and try again.");
                 user.RaiseMessage("Try `ckan install --no-recommends` to skip installation of recommended modules.");
                 return Exit.ERROR;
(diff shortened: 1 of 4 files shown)


techman83:
I believe this is used often to force install a non-compatible version. Though there was some work put towards loose compatibility checking options a while back.
Should we allow and warn instead?

politas:
Yes, I think I'd call this a feature, not a bug.

HebaruSan:
Thanks for the replies. I will take another look at this.

HebaruSan:
OK, I added a few more changes to create an explicit option for allowing incompatible versions. This way the command can still be safe by default, and the back door to get around compatibility checking is explicitly documented.

PR #2233, opened 2017-12-24
Use shared installer code in GUI and fix reinstall problems

HebaruSan:
## Install messages in `ModuleInstaller.InstallList`

- For cached modules, added a space between version and "(cached)"
- For non-cached modules, added indication of download host so the user can make a more informed decision about whether to continue with the downloads

```
$ mono ckan.exe install GPP
1) DistantObject-RealSolarSystem (Distant Object Enhancement Real Solar System config)
2) DistantObject-default (Distant Object Enhancement default config)
2
About to install...

 * Galileo's Planet Pack 1.5.88 (github.com)
 * Distant Object Enhancement default config v1.9.1 (cached)
 * Kopernicus Planetary System Modifier 2:release-1.3.1-3 (github.com)
 * ModularFlightIntegrator 1.2.4.0 (cached)
 * scatterer 2:v0.0320b (cached)
 * scatterer - default config 2:v0.0320b (cached)
 * GPP Textures 3.0.0 (github.com)
 * Strategia 1.6.0 (github.com)
 * Custom Barn Kit 1.1.16.0 (ksp.sarbian.com)
 * Contract Configurator 1.23.3 (github.com)
 * Waypoint Manager 2.7.0 (github.com)
 * Final Frontier 1.3.6-3189 (spacedock.info)
 * Distant Object Enhancement v1.9.1 (cached)
 * OPM Galileo 1.2.4 (github.com)
 * Community Terrain Texture Pack 1:1.0.1 (github.com)
 * Kerbal-Konstructs 1.2.0.1 (spacedock.info)
 * Sigma Replacements: Heads H_v0.1.8 (github.com)
 * Sigma Replacements: Suits S_v0.1.5 (github.com)
 * Sigma Replacements: Navigation N_v0.1.3 (github.com)

Continue? [Y/n] 
```

The same changes are visible in ConsoleUI and GUI.

## Code style/refactoring

In addition, some improvements are made to the organization of the code. This should have no visible impact.

- Duplicated installation code in GUI `MainInstaller` removed
  - Replaced with calls to equivalent logic in Core `ModuleInstaller` (modeled after similar function in `ConsoleUI`)
  - Makes the logic consistent in all UIs and easier to maintain (for example, the above change to messages)
  - `ModuleInstaller.ResolveModules` and `ModuleInstaller.EnsureCache` removed because they were only used by the removed redundant GUI code
- `ModuleInstaller.Upgrade` functions accept `IDownloader` reference instead of `NetAsyncModulesDodwnloader`, to make it easier to create alternate downloaders in the future
- All occurrences of HTTP `User-Agent` header properly capitalized

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -190,12 +190,12 @@ namespace CKAN
             {
                 if (!ksp.Cache.IsCachedZip(module.download))
                 {
-                    User.RaiseMessage(" * {0} {1}", module.name, module.version);
+                    User.RaiseMessage(" * {0} {1} ({2})", module.name, module.version, module.download.Host);
                     downloads.Add(module);
                 }
                 else
                 {
-                    User.RaiseMessage(" * {0} {1}(cached)", module.name, module.version);
+                    User.RaiseMessage(" * {0} {1} (cached)", module.name, module.version);
                 }
             }
 
@@ -253,23 +253,6 @@ namespace CKAN
             }
 
             User.RaiseProgress("Done!\r\n", 100);
-        }
-
-        public ModuleResolution ResolveModules(IEnumerable<CkanModule> modules, RelationshipResolverOptions options)
-        {
-            var resolver = new RelationshipResolver(modules, options, registry_manager.registry, ksp.VersionCriteria());
-            return new ModuleResolution(resolver.ModList(), m => ksp.Cache.IsCachedZip(m.download));
-        }
-
-        public void EnsureCache(List<CkanModule> modules, IDownloader downloader = null)
-        {
-            if (!modules.Any())
-            {
-                return;
-            }
-
-            downloader = downloader ?? new NetAsyncModulesDownloader(User);
-            downloader.DownloadModules(ksp.Cache, modules);
         }
 
         public void InstallList(ModuleResolution modules, RelationshipResolverOptions options)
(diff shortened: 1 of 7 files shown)


politas:
In my Replaced_by branch, I want to move from the "Upgrade" code to my "Replace" code, so that the same technique is used for both, which can then also be used for "Reinstall" and "Switch to version [x]". Ultimately, we may be able to fold the "install" and "remove" into the same method. 

My idea is that you pass a "Change" object, which consists of two CkanModules - ChangeFrom and ChangeTo. If ChangeFrom is null, just install ChangeTo. If ChangeTo is null, just uninstall ChangeFrom. If neither is null, Uninstall ChangeFrom and install ChangeTo.

The really tricky part is cleaning up unneeded dependencies. We don't currently maintain a "reason for installation" in our registry, which I think we need. Any module that is installed to fill a dependency, but now has nothing depending on it, should be flagged to the user for possible removal.

politas:
I fear this has been broken by #2202, so I'm re-running Travis' checks

HebaruSan:
Nice catch; I thought this might conflict with some of _my_ other changes, but I didn't think of that one.
Refactored that function to use the common installer code.

I did notice one problem with reinstall (both versions of it): if the remove step removes a dependent mod, the install step doesn't reinstall it. So for example if I install a bunch of stuff and then reinstall ModuleManager, many things will be missing afterwards. That will probably annoy users. I'll submit an issue for it unless I can think of a simple fix.

HebaruSan:
OK, I think that last commit covers the missing dependent mods problem. It will now uninstall the selected mod and any mods that depend on it, as normal, and then re-install all of them.

The one remaining wrinkle is that it'll also prompt you for suggestions and recommendations from those mods. I think that's better than users having mods removed unexpectedly though.

politas:
Hmm. I installed All Y'all, which installed ModuleManager.

I reinstalled ModuleManager, which removed and then installed both All Y'all and ModuleManager. 

When that finished, the Installed checkbox for ModuleManager was no longer checked.
Refreshed - no change.

If I tick the checkbox, it doesn't come up as a change to apply
If I restart CKAN, it will show up as ticked. Seems to be just a visual glitch in MainModList. The mod is still in Installed_Modules, it's just the checkbox that has lost its info.

HebaruSan:
Good catch! `MainModList.ConstructModList` was using the first occurrence of a mod in the change list to get the status, whereas the _last_ occurrence represents the real status.

PR #2234, opened 2017-12-25
Save and restore panel position

### LINKED ISSUE #2208: Feature: divider position persistance
This has been a thing for me pretty much since CKAN.  Would it be possible to make the divider maintain it's position when placed?

I've uploaded a screenshot of what I'm talking about.
screenshot from 2017-12-05 11-01-19

politas:
Stores the SplitterDistance of the main window in GUIConfig.xml as PanelPosition when closing the window and restores it on opening.

Closes #2208

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -285,6 +285,7 @@ namespace CKAN
         {
             Location = configuration.WindowLoc;
             Size = configuration.WindowSize;
+            splitContainer1.SplitterDistance = configuration.PanelPosition;
 
             if (!configuration.CheckForUpdatesOnLaunchNoNag && AutoUpdate.CanUpdate)
             {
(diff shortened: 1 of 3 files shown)


GUI/Main.cs
```diff
@@ -285,6 +285,8 @@ protected override void OnLoad(EventArgs e)
         {
             Location = configuration.WindowLoc;
             Size = configuration.WindowSize;
+            splitContainer1.SplitterDistance = configuration.PanelPosition;
```
HebaruSan: Our method to save/restore the window size doesn't seem two work for maximized windows on Windows (I wanted to play some games over the winter). If I maximize and then move the splitter to be a reasonable distance from the right edge, closing and relaunching gives me a squished right pane in a smaller window, presumably because it's trying to set the pane width to greater than the actual window:

image

politas: If you then maximise, is the pane width correct?

politas: Ah, got it. It's only storing the Window Size when the Window is not maximized, but not storing the fact that the window is maximized. Linux doesn't actually maximize the window, it just makes it fir the window space, so we don't see the issue there.

HebaruSan:
I love a good bonus fix! (confirmed that maximizing works on Windows)

HebaruSan:
Oops, forgot the CHANGELOG. Give me a few minutes...

PR #2236, opened 2017-12-29
Add and change logging to make INFO readable

HebaruSan:
## Background

Unbeknownst to many, including myself until today, [CKAN has log file capabilities](https://github.com/KSP-CKAN/CKAN/wiki/User-guide#logging). It can use several different levels of verbosity, with DEBUG being way too slow and verbose (dozens of megabytes of excruciating detail about nothing), and WARN saying little to nothing (errors only, essentially). For purposes of this pull request, assume we're always talking about INFO, which represents a happy medium in between.

## Problems

Currently the INFO log isn't as useful as it could be. It contains many details that users would not care about, and leaves out other important information about what's happening. A typical session generates a very long log file, but it's hard to figure out what actually happened from it.

## Changes

This pull request aims to make the INFO log useful and readable for typical usage. Each high level step that CKAN takes ought to be logged, while each low level operation should be relegated to DEBUG unless there's a problem.

Moved from INFO to DEBUG (decreased visibility):

- Per-file install/remove messages
- Resolving relationships for modname
- Resolving dependencies for modname
- Found modname version v (when updating registry)

Moved from DEBUG to INFO (increased visibility):

- CKAN started.
- Initialising KSP/CKAN
- Initialised KSP/CKAN
- Cleaning cache directory
- Loading KSP instances from registry
- URL redirected to URL
- Updating the mod list
- Creating / Saving / Loading CKAN registry
- Stored file to cachename

Added to INFO:

- Installed modname
- Removed modname
- CKAN exiting.

After these changes, my INFO logs are a lot easier to make sense of.

### DIFF
--- a/Cmdline/Main.cs
+++ b/Cmdline/Main.cs
@@ -48,7 +48,7 @@ namespace CKAN.CmdLine
             }
 
             Logging.Initialize();
-            log.Debug("CKAN started");
+            log.Info("CKAN started.");
 
             // If we're starting with no options then invoke the GUI instead.
             if (args.Length == 0)
@@ -83,6 +83,10 @@ namespace CKAN.CmdLine
             {
                 return printMissingInstanceError(new ConsoleUser(false));
             }
+            finally
+            {
+                log.Info("CKAN exiting.");
+            }
 
             Options cmdline;
             try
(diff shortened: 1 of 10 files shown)


politas:
Had to fix some merge conflicts on this one - @HebaruSan, can you check that everything from #2230 came through ok?

HebaruSan:
Yeah, it looks OK to me.

PR #2239, opened 2018-01-03
Save timestamped .ckan files after we save the registry

### LINKED ISSUE #1504: "Commit log" with rollback functionality
When it comes to debugging what mods are causing problems, a "commit log" which lists every change you've made would be a huge help. Having a rollback button would help get the game back into a known-good state where the user can then move forward from. Rollbacks should of course be their own commits to the log.

HebaruSan:
## Motivation

Currently if users install and uninstall lots of mods, they have no record of the changes. This can be a problem if one of the mods they install is not compatible, since they have no easy way to return to an older state or even know what the older states were.

We do create a `CKAN/installed-default.ckan` file containing the list of installed mods, but this is overwritten each time the mod list changes.

## Changes

Now each CKAN folder will have a `history` subfolder created, and when the list of installed mods is changed, a new file with a name like this is created:

- `installed-steam-2018-01-02_22-21-37.ckan`

This contains the same info as the `installed-default.ckan` file, but it will not be overwritten by later changes, because the file name contains the change timestamp. This allows the user to quietly accumulate a history of all their mod lists over time, and if they post a problem on the forum, we can tell them, "Check the CKAN/history folder for your older installs."

Also a few code cleanup changes:

- There was a TODO about changing the "default" part of "installed-default" to be the name of the instance. This is now done.
  - To make that happen, the `KSP` class now has a public `Name` property.
  - The existence of `KSP.Name` makes `ModListScreen.InstallName` obsolete, so it is removed.
- There is now a public function specifically for generating custom .ckan files, and that option in GUI no longer depends on making a copy of `installed-default.ckan` after the fact.

Fixes #1504, unless you think it should include a first-class rollback UI.

### DIFF
--- a/Cmdline/Action/KSP.cs
+++ b/Cmdline/Action/KSP.cs
@@ -246,8 +246,7 @@ namespace CKAN.CmdLine
             try
             {
                 string path = options.path;
-                CKAN.KSP ksp = new CKAN.KSP(path, User);
-                Manager.AddInstance(options.name, ksp);
+                Manager.AddInstance(new CKAN.KSP(path, options.name, User));
                 User.RaiseMessage("Added \"{0}\" with root \"{1}\" to known installs", options.name, options.path);
                 return Exit.OK;
             }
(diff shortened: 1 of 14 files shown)


HebaruSan:
I will need some time to wrestle with the compiler and tests...

PR #2243, opened 2018-01-06
Validate downloaded files against metadata before adding to cache

### LINKED ISSUE #62: Support sha1/md5 fingerprints of downloaded files

### LINKED ISSUE #1927: CKAN sometimes saves a mod YZ as CKAN/downloads/ABCDE123-SomeOtherMod-1.2.3.zip
Thanks for wanting to bring an issue to our attention!

Before you open a ticket, please do a quick search of the existing issues to see if it's already been brought up.

**CKAN Version:** 
Affected are at least:
Alexey           (v1.20.1 /  d41fe54)
Yuri               (v1.20.0 /  2f448b9)
StarTrammier (v1.18.1 /  0d2c4f0)
(did not check any older ones)

**Operating System:** 
Windows 10 Pro, Version 1607 (i.e. the Anniversary Edition), Build 14393.321
KSP for Windows, x64, Version 1.2.0.1586

CKAN has problems with downloading mods correctly.  In this case CKAN seems to mix up downloaded/cached filenames and contents.

- Create a completely new KSP directory by unpacking ksp_win64-1-2.zip.
- Start ckan.exe
- point ckan.exe at the new install, update repository data as needed
- find the "SETI-Rebalance" mod and tick it for install
- klick "Apply"
- Agree to the default settings for recommended mods.
- Select all suggested mods.
- CKAN downloads
- CKAN stops during install (this time) with
  "Bad metadata detected for module SAVE 1.4.1-2315: No files found in GameData/Nereid to install!"

Looking at the KSP_win64\CKAN\downloads\A2511907-SAVE-1.4.1-2315.zip, it contains what appears to be the "SETIprobePart" mod: 
    [A2511907-SAVE-1.4.1-2315.zip](https://github.com/KSP-CKAN/CKAN/files/546547/A2511907-SAVE-1.4.1-2315.zip)

However:
- Close CKAN.exe
- delete the KSP_win64\CKAN\downloads\A2511907-SAVE-1.4.1-2315.zip file
- Start CKAN.exe again, select S.A.V.E, switch the rider on the right top from "Metadata" to "Contents" and click on "Download"
- the contents shown seem right for S.A.V.E:
  [A2511907-SAVE-1.4.1-2315.zip](https://github.com/KSP-CKAN/CKAN/files/546553/A2511907-SAVE-1.4.1-2315.zip)

It seems CKAN gets somewhat confused when a larger number of mods are downloaded.

none

If I can be of any help --- say so.  I do understand debugging, though I have very little .NET experience.

### LINKED ISSUE #2100: Silently downloading HTML as ZIP files causing corrupt zip exception.
### CKAN Version if known (or site URL)
v1.22.3
Windows 7

### Please describe the expected behaviour
1. CKAN attempts to download mod.
2. Download is redirected but still outputs data which in this case is HTML.
3. CKAN saves the data as a ZIP in downloads.
4. CKAN checks the hash and attempts to re-download and on subsequent attempts errors.

### Please describe the actual behaviour
1. CKAN attempts to download mod.
2. Download is redirected but still outputs data which in this case is HTML.
3. CKAN saves the data as a ZIP in downloads.
4. CKAN attempts to install mod but errors due to file being 'corrupt'.

### What steps can be taken to reproduce the issue? 
1. Setup a HTTP server that always returns HTML.
2. Get CKAN to download a mod via that server.
Note: The cause of this in my setup is a proxy blocking ZIP files and returning a rejection page.

HebaruSan:
## Problems

Despite storing the download file size, SHA1, and SHA256 for every download in the registry, CKAN currently performs **no** validation checks on downloaded files before it adds them to the download cache. This means that an HTML file or a ZIP for some other mod can be stored in the cache as if it was valid, and CKAN will attempt to use it later if installation is requested.

Now that KSP-CKAN/xKAN-meta_testing#32 has enabled caching of downloads across NetKAN and CKAN-meta pull requests (rather than just within the same request as I initially assumed), caching matters more. KSP-CKAN/NetKAN#6135 nearly got stuck; the author initially uploaded a ZIP in LZMA format, a proprietary 7zip format which CKAN (and Windows Explorer) cannot read, and that PR errored out. However, the validation log (now overwritten) revealed that CKAN only noticed there was a problem when it tried to _install_ the "(cached)" download! So the code in charge of downloading and adding to the cache considers the file valid, so it won't be re-downloaded, but the code in charge of installing does not, so it can't be installed. Luckily the author uploaded a new version; if he had simply updated the existing version, then a "<span>#</span>rebuild" comment would only have re-run the validation on the already "cached" file, and there would have been no way to fix that PR.

A download cache should provide assurances of data integrity.

## Causes

All caching is managed by `NetFileCache`, which operates at a per-URL level. Files downloaded by other components are passed to `NetFileCache.Store` along with the source URL, and the URL is used to generate the name for the file in the cache so it can be found later.

Notably, this class has no access to any information that could be used to validate the files, and it fails to use even the one check it does have access to, the one that detects the LZMA format (that's how the above Catch-22 came about).

## Changes

A new `NetModuleCache` class is added as a wrapper around `NetFileCache` to safeguard cache consistency at a per-module level. Its public interface is defined in terms of `CkanModule` objects rather than `Uri`s, so it always has the information it needs to check whether a given file is valid. Such checks are now performed whenever a file is added to the cache.

All code that previously used `NetFileCache` is updated to use `NetModuleCache` instead, with one exception: Netkan needs to be able to add things to the cache before their size and hash are known, so it continues to use a raw file cache. In most (all?) cases this was simple as there was already a `CkanModule` floating around ready to be used, and we just had to remove "`.download`".

If `NetFileCache` finds an existing invalid file in the cache (added before this PR), it will delete it. Note that this applies only to problems that are independent of metadata such as the LZMA format issue.

The code that calculates SHA1 and SHA256 is moved from Netkan into Core to be shared by whatever code needs it, and the private duplicate copy in ConsoleUI is removed.

`NetFileCache.GetCachedZip` no longer has two different validation modes controlled by an optional `bool`; it now always checks ZIP files' CRCs.

A new `InvalidModuleFileKraken` class is created to describe problems with downloaded files that CKAN attempted to add to the cache.

Since the cache is now a bit more reliable, ConsoleUI is updated to use `IsMaybeCachedZip` when checking if a module is cached for improved response times.

Several existing tests are updated to populate `download_hash` in metadata representing the locally stored test ZIPs.

New Tests are added:

1. Make sure an LZMA format ZIP is not allowed into the cache
2. Make sure a ZIP with a mismatched file size is not allowed into the cache
3. Make sure a ZIP with the correct file size but incorrect SHA1 and SHA256 is not allowed into the cache

Fixes #62.
Fixes #1927.
Fixes #2100.


techman83:
Yeah, I only learned enough C# to write generate the hashes, thanks for implementing this fully :)

HebaruSan:
Ahh, I see. Well, having the hashes in the existing metadata is tremendously helpful, so thanks for that!

Ruedii:
LZMA is not proprietary.

It is entirely open source.

PKZIP is proprietary.  It just happens to have it's reverse engineered design widely available.

Don't confuse proprietary to not implemented in .Net.

HebaruSan:
@Ruedii, CKAN uses `ICSharpCode.SharpZipLib.Zip`, not .NET's built-in library.

PR #2245, opened 2018-01-08
Show status bar text, add progress bar

HebaruSan:
## Problems

CKAN's GUI has a status bar that is always empty:

image

And some code that tries to set it:

https://github.com/KSP-CKAN/CKAN/blob/378d12818ade7831e27a953096f531cc546a9b11/GUI/MainDialogs.cs#L23-L27

Currently they aren't connected.

## Causes

Two different status bar controls are in use at the moment: the one that is visible, and the one that receives the status messages from code.

- `statusStrip1` is the control that's actually visible. It has no contained controls and none of its properties are set by code during downloads/installs.
- `StatusPanel` is a panel which is technically added to the form but not visible. `StatusLabel` is a label which is added to this and which receives the status messages from the application, but again is not visible.

## Changes

The status bar code is rewritten according to the [documentation for StatusStrip](https://msdn.microsoft.com/en-us/library/system.windows.forms.statusstrip(v=vs.110).aspx#Examples).

- `StatusPanel` is removed
- `StatusLabel` is changed to type `System.Windows.Forms.ToolStripStatusLabel` and added to `statuStrip1` instead
- A new `ToolStripProgressBar` control is added to mirror the behavior of the progress bar on the install screen

Together, these changes make the status bar work as intended:

image

### DIFF
--- a/GUI/Main.Designer.cs
+++ b/GUI/Main.Designer.cs
@@ -78,8 +78,8 @@
             this.reinstallToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.downloadContentsToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.ModInfoTabControl = new CKAN.MainModInfo();
-            this.StatusPanel = new System.Windows.Forms.Panel();
-            this.StatusLabel = new System.Windows.Forms.Label();
+            this.StatusLabel = new System.Windows.Forms.ToolStripStatusLabel();
+            this.StatusProgress = new System.Windows.Forms.ToolStripProgressBar();
             this.MainTabControl = new CKAN.MainTabControl();
             this.ManageModsTabPage = new System.Windows.Forms.TabPage();
             this.FilterByAuthorTextBox = new CKAN.HintTextBox();
@@ -124,7 +124,6 @@
             this.splitContainer1.SuspendLayout();
             ((System.ComponentModel.ISupportInitialize)(this.ModList)).BeginInit();
             this.ModListContextMenuStrip.SuspendLayout();
-            this.StatusPanel.SuspendLayout();
             this.MainTabControl.SuspendLayout();
             this.ManageModsTabPage.SuspendLayout();
             this.ChangesetTabPage.SuspendLayout();
@@ -263,16 +262,17 @@
             this.aboutToolStripMenuItem.Text = "About";
             this.aboutToolStripMenuItem.Click += new System.EventHandler(this.aboutToolStripMenuItem_Click);
             // 
+            //
             // statusStrip1
-            // 
-            this.statusStrip1.ImageScalingSize = new System.Drawing.Size(24, 24);
-            this.statusStrip1.Location = new System.Drawing.Point(0, 1016);
+            //
             this.statusStrip1.Name = "statusStrip1";
             this.statusStrip1.Padding = new System.Windows.Forms.Padding(2, 0, 21, 0);
-            this.statusStrip1.Size = new System.Drawing.Size(1544, 22);
             this.statusStrip1.TabIndex = 1;
-            this.statusStrip1.Text = "statusStrip1";
-            // 
+            this.statusStrip1.Items.AddRange(new System.Windows.Forms.ToolStripItem[]
+            {
+                this.StatusLabel, this.StatusProgress
+            });
+            //
             // menuStrip2
             // 
             this.menuStrip2.Anchor = ((System.Windows.Forms.AnchorStyles)(((System.Windows.Forms.AnchorStyles.Top | System.Windows.Forms.AnchorStyles.Left) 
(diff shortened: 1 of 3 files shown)


GUI/Main.Designer.cs
```diff
@@ -124,7 +124,6 @@ private void InitializeComponent()
             this.splitContainer1.SuspendLayout();
             ((System.ComponentModel.ISupportInitialize)(this.ModList)).BeginInit();
             this.ModListContextMenuStrip.SuspendLayout();
-            this.StatusPanel.SuspendLayout();
```
Olympic1: Add `this.statusStrip1.SuspendLayout();`

GUI/Main.Designer.cs
```diff
@@ -1054,7 +1048,6 @@ private void InitializeComponent()
             this.splitContainer1.ResumeLayout(false);
             ((System.ComponentModel.ISupportInitialize)(this.ModList)).EndInit();
             this.ModListContextMenuStrip.ResumeLayout(false);
-            this.StatusPanel.ResumeLayout(false);
```
Olympic1: Add
```
this.statusStrip1.ResumeLayout(false);
this.statusStrip1.PerformLayout();
```

GUI/Main.Designer.cs
```diff
@@ -581,27 +581,22 @@ private void InitializeComponent()
...
-            // 
+            //
+            // StatusProgress
+            //
+            this.StatusProgress.Enabled = true;
+            this.StatusProgress.Visible = false;
+            this.StatusProgress.Minimum = 0;
+            this.StatusProgress.Maximum = 100;
+            this.StatusProgress.Size = new System.Drawing.Size(300, 20);
+            this.StatusProgress.Alignment = System.Windows.Forms.ToolStripItemAlignment.Right;
```
Olympic1: Add `this.StatusProgress.Name = "StatusProgress";`

GUI/Main.Designer.cs
```diff
@@ -581,27 +581,22 @@ private void InitializeComponent()
...
-            // 
+            //
             // StatusLabel
-            // 
-            this.StatusLabel.Dock = System.Windows.Forms.DockStyle.Fill;
-            this.StatusLabel.Location = new System.Drawing.Point(0, 0);
-            this.StatusLabel.Margin = new System.Windows.Forms.Padding(4, 0, 4, 0);
+            //
             this.StatusLabel.Name = "StatusLabel";
-            this.StatusLabel.Size = new System.Drawing.Size(1050, 29);
```
Olympic1: Leave `Size` in

GUI/Main.Designer.cs
```diff
@@ -263,16 +262,17 @@ private void InitializeComponent()
...
             // 
+            //
             // statusStrip1
-            // 
-            this.statusStrip1.ImageScalingSize = new System.Drawing.Size(24, 24);
-            this.statusStrip1.Location = new System.Drawing.Point(0, 1016);
+            //
             this.statusStrip1.Name = "statusStrip1";
             this.statusStrip1.Padding = new System.Windows.Forms.Padding(2, 0, 21, 0);
-            this.statusStrip1.Size = new System.Drawing.Size(1544, 22);
```
Olympic1: Leave `Size` in

GUI/Main.Designer.cs
```diff
@@ -263,16 +262,17 @@ private void InitializeComponent()
             this.aboutToolStripMenuItem.Text = "About";
             this.aboutToolStripMenuItem.Click += new System.EventHandler(this.aboutToolStripMenuItem_Click);
             // 
+            //
             // statusStrip1
-            // 
-            this.statusStrip1.ImageScalingSize = new System.Drawing.Size(24, 24);
-            this.statusStrip1.Location = new System.Drawing.Point(0, 1016);
```
Olympic1: Leave `Location` in

HebaruSan: Oh, the IDE probably needs that stuff for the graphical editor, doesn't it? OK...

GUI/Main.Designer.cs
HebaruSan: Can't:

```
  Main.Designer.cs(127,31): error CS1061: 'StatusStrip' does not contain a definition for 'SuspsendLayout' and no extension method 'Suspse
ndLayout' accepting a first argument of type 'StatusStrip' could be found (are you missing a using directive or an assembly reference?)
```

HebaruSan: Done

GUI/Main.Designer.cs
HebaruSan: Done

GUI/Main.Designer.cs
HebaruSan: Done

GUI/Main.Designer.cs
HebaruSan: Done

PR #2248, opened 2018-01-13
Write desktop file only if doesn't exist

skad:
Since i'm under a dark theme i have read #1011 

The fix 
```
I have found a much better solution meanwhile. Just start ckan with a bright GTK theme. Example:
env GTK2_RC_FILES=/usr/share/themes/Adwaita/gtk-2.0/gtkrc mono ckan.exe
```
works well but each time the app start my desktop file is rewriten.
But imo the desktop file must be handle by package manager (apt / pacman / other) or manually but not by the app itself.

So this is my humble contribution.

:)

### DIFF
--- a/GUI/URLHandlers.cs
+++ b/GUI/URLHandlers.cs
@@ -167,28 +167,26 @@ Do you want to allow CKAN to do this? If you click no you won't see this message
                 return;
             }
 
-            if (File.Exists(handlerPath))
+            if (!File.Exists(handlerPath))
             {
-                File.Delete(handlerPath);
+                File.WriteAllText(handlerPath, "");
+                data = parser.ReadFile(handlerPath);
+                data.Sections.AddSection("Desktop Entry");
+                data["Desktop Entry"].AddKey("Version", "1.0");
+                data["Desktop Entry"].AddKey("Type", "Application");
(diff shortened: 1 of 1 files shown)


HebaruSan:
Code does what it says, but overwriting the file has advantages as well. If we want to make changes to it, overwriting ensures that those changes will be made for upgrading users.

Maybe we could come up with a more direct solution to the underlying problem. @skad, would you mind sharing before/after screenshots for the visual problems you're having?

skad:
Well this is the dark theme bug where the ui is unreadable.

I will try latter to post screen.

But with the lattest version the fix won't work anymore.
But my patch is still revelent i think.

politas:
@skad, there is an alternative for making your changes to the desktop file persistent, which is to save them as a renamed "CKAN.desktop" file, rather than "CKAN Launcher.desktop". Then you could use the menu editor to hide the original CKAN Launcher  and just see your own.

HebaruSan:
> Well this is the dark theme bug where the ui is unreadable.

@skad, we did several fixes subsequent to this to make dark themes work better, see #2529, #2540, #2541, and #2594.

Are you still having those problems with recent versions of CKAN?

skad:
@HebaruSan thank for the update.
I'm not using ckan under dark theme on linux (i'm not using linux for now :'( ) so i can't answer you.
Sorry

politas:
Given the person who had an issue with this doesn't have the issue any more, and the standard is for apps to re-write their desktop file, I'm closing this PR.

PR #2258, opened 2018-01-17
Leave out children already shown in ancestor node

### LINKED ISSUE #2257: Circular logic flaw
There may be just a wee bit of circular logic going on somewhere in CKAN 1.24.0-PRE

screenshot from 2018-01-17 09-20-13

HebaruSan:
## Problem

Many mod relationships are mutual/circular, so mod X conflicts with Y, and Y also conflicts with X.

In the 1.24 pre-release version of the GUI relationships tab, this means you can expand them in an infinite chain if you want to, back and forth forever. This causes visual clutter and is a bit silly looking. Example from #2257:

image

## Changes

Now we omit the children of nodes that already have their children shown as part of their own ancestry. The user will have seen them already when they expanded the identical ancestor node.

image

This is done by walking the tree via `TreeNode.Parent` up to the root and comparing the `CkanModule` objects located in the `.Tag` property.

Fixes #2257.

### DIFF
--- a/GUI/MainModInfo.cs
+++ b/GUI/MainModInfo.cs
@@ -184,6 +184,22 @@ namespace CKAN
             Cursor.Current = prevCur;
         }
 
+        private bool ImMyOwnGrandpa(TreeNode node)
+        {
+            CkanModule module = node.Tag as CkanModule;
+            if (module != null)
+            {
+                for (TreeNode other = node.Parent; other != null; other = other.Parent)
+                {
+                    if (module == other.Tag)
+                    {
+                        return true;
+                    }
+                }
+            }
+            return false;
+        }
+
         private void UpdateModDependencyGraph(CkanModule module)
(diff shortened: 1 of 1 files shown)


politas:
Would it not make sense to just check whether a mod already appears as an expanded treenode and refuse to expand it (showing an "Already expanded" fake node as the only child)? That would prevent loops of three or more mods as well.

HebaruSan:
This solution handles loops of three or more mods. It checks _all_ of the ancestor nodes.

"already appears as an expanded treenode" sounds like it would be path-dependent; whether or not you see a relationship would depend on what other nodes you've expanded previously. I think it's better to use something consistent.

politas:
Ah, ok, I must have misread your description.

As for the path-dependant expansion; it's how a tool I have at work handles expansion of related objects. It's not important.

Issue #2210, opened 2017-12-08
CKAN's GitHub downloads are breaking the rules


HebaruSan:
# GitHub downloading needs a rewrite

(I debated whether to add this as a comment to #1817, but it seems like too much text and detail for that.)

## Problems

Currently if CKAN downloads many files from GitHub at the same time, they often fail with HTTP status code 403-Forbidden. #1817 contains an example, but these reports are common and I've definitely seen it happen myself several times.

## Background

The GitHub API uses 403 codes for throttling; you get 60 unauthenticated requests per hour, and any beyond that return a 403. I encountered this while working on [an unrelated project](https://hebarusan.github.io/releases-timeline/), and I had to use a GitHub token to allow 5000/hour, passed in the HTTP request headers:

```http
Authorization: token <OAuth token here>
```

Currently CKAN's downloads do **not** go through the GitHub API, so this does not necessarily indicate exactly what's going on with them. However, it establishes that 403-Forbidden is sometimes used for throttling, and it becomes more relevant later in discussion of the API.

[Sample API data for releases](https://api.github.com/repos/HebaruSan/Astrogator/releases/7538924), minus the author and uploader fields since they're long and not relevant to this issue:

```json
{
  "url": "https://api.github.com/repos/HebaruSan/Astrogator/releases/7538924",
  "assets_url": "https://api.github.com/repos/HebaruSan/Astrogator/releases/7538924/assets",
  "upload_url": "https://uploads.github.com/repos/HebaruSan/Astrogator/releases/7538924/assets{?name,label}",
  "html_url": "https://github.com/HebaruSan/Astrogator/releases/tag/v0.7.8",
  "id": 7538924,
  "tag_name": "v0.7.8",
  "target_commitish": "master",
  "name": "Frictionless toilet",
  "draft": false,
  "prerelease": false,
  "created_at": "2017-08-28T06:00:32Z",
  "published_at": "2017-08-28T06:09:07Z",
  "assets": [
    {
      "url": "https://api.github.com/repos/HebaruSan/Astrogator/releases/assets/4682631",
      "id": 4682631,
      "name": "Astrogator.zip",
      "label": null,
      "content_type": "application/zip",
      "state": "uploaded",
      "size": 114372,
      "download_count": 5243,
      "created_at": "2017-08-28T06:08:01Z",
      "updated_at": "2017-08-28T06:08:02Z",
[...]

PR #2263, opened 2018-01-22
Accept header and infrastructure for auth tokens

HebaruSan:
## Problems

CKAN's GitHub downloads fail with an HTTP 403 status fairly frequently. We do not know the exact details that trigger it, but it mostly affects large downloads.

It also turns out that the URLs we're using are supposed to be for users and browsers only, not applications. We should migrate to using the GitHub API.

See #2210 for details.

## Cause

GitHub's normal download links are throttled to prevent abuse, and CKAN's use of them is sometimes affected by this.

GitHub prefers applications to use the GitHub API for programmatic access, which requires support for two new HTTP headers, one that specifies the MIME type of the download (without this, the API returns a JSON description of the file instead of the actual download):

```http
Accept: application/octet-stream
```

And one that contains the user's GitHub API authentication token (this one is technically optional, but downloads will still fail frequently without it, because the limit is 60 per hour):

```http
Authentication: token <OAuth token here>
```

### MIME types

The `Accept` header is now set for all module downloads. The default is `application/octet-stream`, but we will use a multi-value format as well when `download_content_type` is set for a module:

```http
Accept: {CkanModule.download_content_type};q=1.0,application/octet-stream;q=0.9
```

`CkanModule` has a new property `download_content_type` that maps to the same value in CKAN-meta (this was absent from the registry previously but stored to the metadata repo for all modules by netkan). This allows us to get the MIME type of a download.

`NetAsyncDownloader` now uses `Net.DownloadTarget` objects instead of `KeyValuePair<Uri, int>` to describe requested downloads, as does all of its calling code.

`Net.DownloadTarget` now has a `mimeType` property storing the MIME type of the download.

The value from `CkanModule.download_content_type` is passed to `Net.DownloadTarget.mimeType` when it's available.

The value from `Net.DownloadTarget.mimeType` is used to set the `Accept` header when it's available.

### Auth tokens

The Windows/Mono registry now has a new `AuthTokens` key (secret value in screenshot deleted in post-processing):

image

When a download is initiated, its `Uri.Host` property is looked up as a key of this key, and if it's found, then the corresponding value is plugged in to the `Authentication` header.

## Still To Be Done In Future Pull Requests

Partial fix for #2210. Will be complete after:

- Cmdline commands for editing the auth token table
- ConsoleUI screen for editing the auth token table
- GUI window for editing the auth token table
- netkan update to use `asset.url` instead of `asset.browser_download_url`
- Conversion of existing URLs to use the API

Note that you can set up auth tokens without UI components by using the registry editor in Windows or an equivalent tool for Mono. When this is done, this version of the client will be able to access GitHub API download URLs with that auth token.

Many thanks to @dbent for wise feedback and ideas that guided these changes in the comments of #2210.

### DIFF
--- a/Core/Net/AutoUpdate.cs
+++ b/Core/Net/AutoUpdate.cs
@@ -144,10 +144,14 @@ namespace CKAN
             // download updater app and new ckan.exe
             string updaterFilename = System.IO.Path.GetTempPath() + Guid.NewGuid().ToString() + ".exe";
             string ckanFilename = System.IO.Path.GetTempPath() + Guid.NewGuid().ToString() + ".exe";
-            Net.DownloadWithProgress(new[]{
-                new Net.DownloadTarget(fetchedUpdaterUrl.Item1, updaterFilename, fetchedUpdaterUrl.Item2),
-                new Net.DownloadTarget(fetchedCkanUrl.Item1, ckanFilename, fetchedCkanUrl.Item2),
-            }, user);
+            Net.DownloadWithProgress(
+                new[]
+                {
+                    new Net.DownloadTarget(fetchedUpdaterUrl.Item1, updaterFilename, fetchedUpdaterUrl.Item2),
+                    new Net.DownloadTarget(fetchedCkanUrl.Item1,    ckanFilename,    fetchedCkanUrl.Item2),
+                },
+                user
+            );
 
             // run updater
             SetExecutable(updaterFilename);
--- a/Core/Net/Net.cs
+++ b/Core/Net/Net.cs
@@ -17,6 +17,7 @@ namespace CKAN
 
     public class Net
     {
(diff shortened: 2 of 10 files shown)


dbent:
Sorry for the confusion, when I said user settings are stored in the registry, I meant the [*Windows* registry](https://github.com/KSP-CKAN/CKAN/blob/master/Core/Win32Registry.cs), which is emulated by Mono on nix systems. This is where global configuration data, like the location of KSP installs, are stored. That's probably the best place for any user tokens.

Everything else looks good to me.

HebaruSan:
Ahh, that makes sense. As a bonus, that eliminates the need to pass that Dictionary around. Please see latest commit.

dbent:
LGTM 👍

PR #2284, opened 2018-02-14
Fallback to archive.org URLs for failed downloads of FOSS packages

### LINKED ISSUE #1682: Upload FOSS mods to the Internet Archive, allow clients to fallback

HebaruSan:
## Background

SpaceDock broke today, which is fun. VITAS reports that it looks like a denial of service attack.

The NetKAN bot has been uploading all permissively-licensed mods to https://archive.org/details/kspckanmods for quite a while now. Many or most of the SpaceDock downloads that are now failing, could in principle fall back to archive.org URLs. Such a feature could also help us deal with GitHub's download throttling or SpaceDock's certificate expirations.

## Changes

Now if the primary download fails for a permissively licensed mod, we try to find it on archive.org as a fallback.

Internally this involves defining a `fallbackUrl` property in `DownloadTarget` and `NetAsyncDownloaderDownloadPart`, both populated from a new `CkanModule.InternetArchiveDownload` property, which itself checks a new `License.Redistributable` property based on a new copy of the list from NetKAN-bot. Then if a download fails, we check whether it has a fallback URL, and if so and if we haven't already tried it, we try it. If the fallback fails as well, then we continue with the normal steps for a failed download.

Fixes #1682.

### DIFF
--- a/Core/Types/CkanModule.cs
+++ b/Core/Types/CkanModule.cs
@@ -624,6 +624,22 @@ namespace CKAN
         }
 
         /// <summary>
+        /// Return an archive.org URL for this download, or null if it's not there.
+        /// The filenames look a lot like the filenames in Net.Cache, but don't be fooled!
+        /// Here it's the first 8 characters of the SHA1 of the DOWNLOADED FILE, not the URL!
+        /// </summary>
+        public Uri InternetArchiveDownload
+        {
+            get
+            {
+                return license.All(l => l.Redistributable)
+                    ? new Uri(
+                        $"https://archive.org/download/{identifier}-{version}/{download_hash.sha1.Substring(0, 8)}-{identifier}-{version}.zip")
+                    : null;
+            }
+        }
+
+        /// <summary>
         /// Format a byte count into readable file size
         /// </summary>
         /// <param name="bytes">Number of bytes in a file</param>
--- a/Core/Net/AutoUpdate.cs
+++ b/Core/Net/AutoUpdate.cs
@@ -152,8 +152,8 @@ namespace CKAN
             Net.DownloadWithProgress(
                 new[]
                 {
-                    new Net.DownloadTarget(fetchedUpdaterUrl.Item1, updaterFilename, fetchedUpdaterUrl.Item2),
-                    new Net.DownloadTarget(fetchedCkanUrl.Item1,    ckanFilename,    fetchedCkanUrl.Item2),
+                    new Net.DownloadTarget(fetchedUpdaterUrl.Item1, null, updaterFilename, fetchedUpdaterUrl.Item2),
+                    new Net.DownloadTarget(fetchedCkanUrl.Item1,    null, ckanFilename,    fetchedCkanUrl.Item2),
                 },
                 user
             );
--- a/Core/Net/Net.cs
+++ b/Core/Net/Net.cs
@@ -121,19 +121,21 @@ namespace CKAN
 
         public class DownloadTarget
         {
-            public Uri    url      { get; private set; }
-            public string filename { get; private set; }
-            public long   size     { get; private set; }
-            public string mimeType { get; private set; }
+            public Uri    url         { get; private set; }
+            public Uri    fallbackUrl { get; private set; }
+            public string filename    { get; private set; }
+            public long   size        { get; private set; }
+            public string mimeType    { get; private set; }
 
-            public DownloadTarget(Uri url, string filename = null, long size = 0, string mimeType = "")
+            public DownloadTarget(Uri url, Uri fallback = null, string filename = null, long size = 0, string mimeType = "")
             {
-                this.url      = url;
-                this.filename = string.IsNullOrEmpty(filename)
+                this.url         = url;
+                this.fallbackUrl = fallback;
+                this.filename    = string.IsNullOrEmpty(filename)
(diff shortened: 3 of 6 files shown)


HebaruSan:
(The failing test is itself due to SpaceDock being down.)

Core/Types/CkanModule.cs
```diff
@@ -623,6 +623,22 @@ public string DescribeInstallStanzas()
             return string.Join(", ", descriptions);
         }
 
+        /// <summary>
+        /// Return an archive.org URL for this download, or null if it's not there.
+        /// The filenames look a lot like the filenames in Net.Cache, but don't be fooled!
+        /// Here it's the first 8 characters of the SHA1 of the DOWNLOADED FILE, not the URL!
+        /// </summary>
```
techman83: Yeah, the hash of that file will remain consistent. The url may not.
```
Produces a filename based of the first 8 digits in sha1 hash,
the 'identifier' and the 'version' in the metadata if the
download_hash exists. Returns '0' if there is no download hash
or has an content type other than zip/gz/tar/tar.gz.
```
There are some [tests](https://github.com/KSP-CKAN/NetKAN-bot/blob/master/t/App/KSP_CKAN/Metadata/Ckan.t#L71) that ensure the correct filenames are generated.

Core/Types/CkanModule.cs
```diff
@@ -623,6 +623,22 @@ public string DescribeInstallStanzas()
...
+            get
+            {
+                return license.All(l => l.Redistributable)
+                    ? new Uri(
+                        $"https://archive.org/download/{identifier}-{version}/{download_hash.sha1.Substring(0, 8)}-{identifier}-{version}.zip")
+                    : null;
+            }
+        }
+
         /// <summary>
```
techman83: Some tests of the filename generation could be useful, but I wouldn't treat this as a blocker.

Core/Types/CkanModule.cs
HebaruSan: Yup, either way makes enough sense to me. I just wanted to note it explicitly since we have two different 8-digit hexadecimal filename prefixes floating around, and it's not easy to tell that they're (supposed to be) different at a glance.

techman83: Yeah, actually makes sense for URL in the cache. If the url changes, you probably do want to re-download.

PR #2297, opened 2018-02-23
Force-allow TLS 1.2 on .NET 4.5

### LINKED ISSUE #2293: Auto-update and other GitHub connections not working on .NET 4.5 and earlier
Background
----------

**CKAN Version:** 
- 1.24.0-PRE

**KSP Version:** 
1.3.1.1891

**Operating System:** 
Windows 10.1

no, I use CKAN for all of my changes

Problem
-------

Opened it.

It to open without error and connect to the repository

**What happened instead?**
I got the errors below, and no updates.

**Screenshots:**

image

```

CKAN.ModuleNotFoundKraken: Exception of type 'CKAN.ModuleNotFoundKraken' was thrown.
   at CKAN.Registry.AllAvailable(String module)
   at CKAN.MainAllModVersions.set_SelectedModule(GUIMod value)
   at CKAN.Main.ModList_SelectedIndexChanged(Object sender, EventArgs e)
   at System.Windows.Forms.DataGridView.OnSelectionChanged(EventArgs e)
   at System.Windows.Forms.DataGridView.FlushSelectionChanged()
   at System.Windows.Forms.DataGridView.ClearSelection(Int32 columnIndexException, Int32 rowIndexException, Boolean selectExceptionElement)
   at System.Windows.Forms.DataGridView.SetAndSelectCurrentCellAddress(Int32 columnIndex, Int32 rowIndex, Boolean setAnchorCellAddress, Boolean validateCurrentCell, Boolean throughMouseClick, Boolean clearSelection, Boolean forceCurrentCellSelection)
   at System.Windows.Forms.DataGridView.MakeFirstDisplayedCellCurrentCell(Boolean includeNewRow)
   at System.Windows.Forms.DataGridView.OnRowCollectionChanged_PostNotification(Boolean recreateNewRow, Boolean allowSettingCurrentCell, CollectionChangeAction cca, DataGridViewRow dataGridViewRow, Int32 rowIndex)
   at System.Windows.Forms.DataGridViewRowCollection.OnCollectionChanged_PostNotification(CollectionChangeAction cca, Int32 rowIndex, Int32 rowCount, DataGridViewRow dataGridViewRow, Boolean changeIsDeletion, Boolean changeIsInsertion, Boolean recreateNewRow, Point newCurrentCell)
   at System.Windows.Forms.DataGridViewRowCollection.OnCollectionChanged(CollectionChangeEventArgs e, Int32 rowIndex, Int32 rowCount)
   at System.Windows.Forms.DataGridViewRowCollection.AddRange(DataGridViewRow[] dataGridViewRows)
   at CKAN.Main._UpdateFilters()
   at CKAN.Main._UpdateModsList(Boolean repo_updated, List`1 mc)
   at CKAN.Main.CurrentInstanceUpdated()
   at CKAN.Main.OnLoad(EventArgs e)
   at System.Windows.Forms.Form.OnCreateControl()
   at System.Windows.Forms.Control.CreateControl(Boolean fIgnoreVisible)
   at System.Windows.Forms.Control.CreateControl()
   at System.Windows.Forms.Control.WmShowWindow(Message& m)
   at System.Windows.Forms.Form.WmShowWindow(Message& m)
```

HebaruSan:
## Background

- GitHub has sunsetted TLS 1/1.1 as of today. [They gave us one year's warning](https://githubengineering.com/crypto-deprecation-notice/), but apparently we missed it or didn't realize it applied to us (or assumed that .NET would handle it properly for us). This means the only way to access an HTTPS URL on GitHub is now via TLS 1.2.
- .NET 4.5 can support TLS 1.2, but it is not enabled by default. .NET 4.6 enables it by default.
- CKAN targets .NET 4.5. Even if you have a later version installed, the behavior of the .NET components will mimic .NET 4.5.

## Problem

Released versions of CKAN can't access GitHub HTTPS URLs anymore. They cause TLS errors.

CKAN uses GitHub HTTPS URLs not just for mod downloads, but also for registry updates and application self-auto-updates. So you can't even refresh the registry or update CKAN itself.

## Workaround

You can force .NET 4.5 to allow _only_ TLS 1.2 by setting a registry key. This allows current versions of CKAN to work:

- https://github.com/KSP-CKAN/CKAN/issues/2293#issuecomment-367868078

However, this affects all applications rather than just CKAN, which might cause problems if any of them need TLS 1/1.1 to operate. Caution is advised.

Until we release a new version, this is the only option for getting CKAN to work.

## Considered but not done

One way to fix this would be to change `TargetFrameworkVersion` in our `csproj` files from v4.5 to v4.6. We probably don't want to do this because some fraction of people might still be running the old framework and would see errors. If you think this is the better solution, please say so below.

## Changes

Now we add `SecurityProtocolTypes.Tls12` to `System.Net.ServicePointManager.SecurityProtocol` at startup via an OR-equals `|=` operator. This enables TLS 1.2 while ensuring that no other supported protocols are disabled.

Credit to @TruePikachu for proposing this solution.

Fixes #2293.

### DIFF
--- a/Cmdline/Main.cs
+++ b/Cmdline/Main.cs
@@ -4,6 +4,7 @@
 // License: CC-BY 4.0, LGPL, or MIT (your choice)
 
 using System;
+using System.Net;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
@@ -49,6 +50,10 @@ namespace CKAN.CmdLine
 
             Logging.Initialize();
             log.Info("CKAN started.");
+
+            // Force-allow TLS 1.2 for HTTPS URLs, because GitHub requires it.
+            // This is on by default in .NET 4.6, but not in 4.5.
+            ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12;
 
             // If we're starting with no options then invoke the GUI instead.
             if (args.Length == 0)
--- a/Netkan/Program.cs
+++ b/Netkan/Program.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Net;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
@@ -30,6 +31,10 @@ namespace CKAN.NetKAN
             try
             {
                 ProcessArgs(args);
+
+                // Force-allow TLS 1.2 for HTTPS URLs, because GitHub requires it.
+                // This is on by default in .NET 4.6, but not in 4.5.
+                ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12;
 
                 // If we see the --version flag, then display our build info
                 // and exit.
--- a/quotes.txt
+++ b/quotes.txt
@@ -70,3 +70,8 @@ It is truly amazing how many extensions to BDArmory are only compatible with a
 KSP version that BDArmory is not compatible with.
 
     -- politas, NetKAN#4169
+%
+<HebaruSan> It's crazy that an app has to specify which version of TLS to use to get an HTTPS URL
+<HebaruSan> I mean, how the heck do we know? You're WebClient, you figure it out!!
+
+    -- regarding CKAN#2297


HebaruSan:
Hmm, we're compiling against some Mono versions that don't have `SecurityProtocolTypes.Tls12`. But how does that square with our TargetFrameworkVersion of v4.5, which should have it?

politas:
Sounds like we need to move to a .Net Target Framework of 4.6, which doesn't rule out any Windows operating systems other than XP, which was already incompatible with 4.5. Looks like we'd be screwing over Debian users again, since even the current Unstable branch (Sid) only has Mono 4.6. If this is the only way to get us working with GitHub again, though...

sundhaug92:
Since this is likely to cause headache for users, should an emergency release be considered?

politas:
https://github.com/KSP-CKAN/CKAN/releases/tag/1.24.0

sundhaug92:
ty @politas

Ruedii:
This is causing serious problems on Ubuntu.  You need to use some sort of case on this line so it only enables it on platforms that need it.

HebaruSan:
@Ruedii, the way to report problems is to create an issue, not post random vague comments on old closed pull requests.

PR #2301, opened 2018-02-23
Fix crash when trying to view empty auth token list

### LINKED ISSUE #2300: NullReferenceException occurs when i click settings>CKAN setting
Background
----------

**CKAN Version:**  1.24.0 (i downloaded bruce 20 min before)

**KSP Version:** 1.3.1

**Operating System:**  windows 10 (64bit)

Problem
-------

execute ckan.exe after i deleted CKAN folder and reset all
and i click immediately settings > CKAN settings 

**Screenshots:**

image

and if i click continue...
image

and i clicked check for updates

image

HebaruSan:
## Problem

If you have never added any auth tokens (and therefore don't have the registry key `HKEY_CURRENT_USER\Software\CKAN\AuthTokens`), you may receive a null reference exception if you:

- Open the GUI settings window
- Edit the auth tokens in consoleui
- Run `ckan authtoken list`

This can be worked around by adding a fake auth token, which you may have to do from the command line with `ckan authtoken add host token`.

## Cause

`Win32Registry.GetAuthTokenHosts()` returns null because `Microsoft.Win32.Registry.CurrentUser.OpenSubKey` returns null.

## Changes

Now `Win32Registry.GetAuthTokenHosts()` returns a zero-element string array if `Microsoft.Win32.Registry.CurrentUser.OpenSubKey` returns null.

Fixes #2300.

### DIFF
--- a/Core/Win32Registry.cs
+++ b/Core/Win32Registry.cs
@@ -103,7 +103,7 @@ namespace CKAN
         public static IEnumerable<string> GetAuthTokenHosts()
         {
             RegistryKey key = Microsoft.Win32.Registry.CurrentUser.OpenSubKey(authTokenKeyNoPrefix);
-            return key?.GetValueNames();
+            return key?.GetValueNames() ?? new string[0];
         }
 
         /// <summary>


politas:
Strange. It worked fine when I added my authtoken.

PR #2311, opened 2018-02-24
Fix GUI sort by size

HebaruSan:
## Background

#2277 updated the Download column in GUI to use "B", "KB", "MB", and "GB" units depending on the magnitude of the size. Previously this column was always in KB.

## Problem

This broke the sorting of this column. There doesn't seem to be a pattern to it anymore.

image

## Cause

The size sort code depended on parsing the text of the cells, which is no longer useful when the units aren't the same or when the strings contain more than just numeric digits.

## Changes

Now the Download column sorts directly by the raw size data. The largest mods are:

image

Fixes https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1240-bruce/&do=findComment&comment=3303380 .

### DIFF
--- a/GUI/MainModList.cs
+++ b/GUI/MainModList.cs
@@ -24,10 +24,9 @@ namespace CKAN
             {
                 return Sort(rows, CheckboxSorter);
             }
-            // XXX: Same for Integer columns
             else if (this.configuration.SortByColumnIndex == 7)
             {
-                return Sort(rows, IntegerSorter);
+                return Sort(rows, DownloadSizeSorter);
             }
             return Sort(rows, DefaultSorter);
         }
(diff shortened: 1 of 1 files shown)


politas:
I saw that comment and thought this would be the way to do it, but I had not a clue how to start.

HebaruSan:
@politas , I think you merged this into your own fork rather than KSP-CKAN's. Was that intentional?

politas:
Ah, must have missed the push to the main fork. I'll fix it when I get home.

politas:
Ok, some definite weirdness happened. This has now been merged, though I broke the GitHub PR connection in the process.

PR #2333, opened 2018-02-27
Check NetKAN for support issues, don't close if last reply is by OP

HebaruSan:
## Motivation

There's a robot that crawls CKAN issues and closes idle support tickets. This is nice for keeping the issues list tidy.

However, users are just as likely to submit support tickets for NetKAN, and those just accumulate.

Also the definition of "idle support ticket" isn't quite fair to issue authors, since it goes purely by time elapsed since last activity, as long as there's at least one comment. Consider this sequence:

1. You submit an issue
2. We mark it as Support and say "What about workaround X?"
3. You say "No, that didn't work"
4. 7 days elapse
5. The bot closes the issue

In this case, the ball is in the dev team's court, and the user has been responsive and hasn't been helped. The ticket should remain open until the _issue author_ is idle for 7 days.

## Changes

- Now the bot looks at the issues in NetKAN as well
- Issues where the latest comment is by the issue creator will no longer be closed

Tagging @techman83 as an FYI in case the server that runs this bot doesn't auto-update the script from GitHub.

### DIFF
--- a/bin/close-old-support-tickets.pl
+++ b/bin/close-old-support-tickets.pl
@@ -18,48 +18,60 @@ use Date::Parse qw(str2time);
 my $token = $ARGV[0] or die "Usage: $0 [token]\n";
 
 my $github = Net::GitHub->new(
-    RaiseError => 1,
+    RaiseError   => 1,
     access_token => $token,
 );
 
-my $date_cutoff = time() - 7 * 86400;   # 7 days ago
+# 7 days ago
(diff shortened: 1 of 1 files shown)


techman83:
Yeah, this is one of those things that don't update automatically. All looks pretty sane to me. Thanks for your efforts!!

I'm very much in favour of bots doing as much of the heavy lifting as possible!

PR #2334, opened 2018-02-28
Actually skip when last comment is by OP

HebaruSan:
@techman83 , KSP-CKAN/Netkan#5172 shouldn't have been closed; I forgot to make it actually skip to the next issue when the last comment is by the OP. It prints the message _saying_ it's going to skip, but continues processing the same issue as before.


techman83:
Aha, cool. Missed that!

Issue #2272, opened 2018-02-01
OSX 32-bit support is being removed


politas:
Apple are planning to remove legacy 32-bit support soon. Currently, we run the CKAN client in 32-bit mode on OSX because Mono Winforms have not been ported to the Cocoa API 

When https://github.com/mono/mono/issues/6701 is resolved, we will be able to allow 64-bit  operation on OSX

HebaruSan:
That issue says they would "really appreciate a pull request," implying the Mono team has no plans to work on this themselves. It sounds like "No more WinForms on Mac" is a distinct possibility.

Thank goodness for cross platform runtimes making low level architecture-related problems like this a thing of the past! :roll_eyes:

politas:
I think that comment was directed at the original poster of that issue who said they had a working version.

Ruedii:
Adding a GTK# GUI option is the best way to insure future multi-platform support.

Just remember, you might want to ship a skin with the Windows version, because the default skin isn't particularly nice.

politas:
How well does GTK# support OSX, though? Will going through all the effort of writing a GTK# GUI implementation actually be any better for our OSX users?

HebaruSan:
I would kind of like to see an Electron-style cross platform UI, where the details are done in Javascript and CSS, but when I looked for such a framework for Mono/.NET, I didn't find one.

Then again, after WinForms, TxFileManager, and CommandLineParser, I'm skeptical of any strategy that sees introducing yet another third party dependency as a magic solution.

ProfFan:
Is there a possibility that CKAN have a stable command line interface and we can then create some frontend for CKAN?

ProfFan:
@politas Thank you for the info! Actually what I am thinking is wrapping CKAN completely with a GUI in another language, i.e. Electron (like @HebaruSan 's idea). This would make CKAN more like other package managers like `apt` and `pacman`, etc., where the actual package management logic is complete separate from the frontend.

politas:
Sure, that's what i meant with the first sentence. Sorry I wasn't more explicit.

HebaruSan:
I've been trying out Electron-NET today. [Electron-build doesn't support building Mac packages on non-Mac OSes due to some signing thing](https://www.electron.build/multi-platform-build), so that kind of throws a wrench into that option (but Windows and Linux can build each other).

[...]

Issue #1334, opened 2015-07-23
Super-awesome Mac client implementation


pjf:
I'm currently at OSCON, sitting next to @desplesda, who has repeatedly expressed an interest in improving the CKAN for the Mac, including <s>the potential for a Gtk#</s> Mac-specific implementation that wouldn't require X11, and which would be less likely to hit bugs in mono.

To foster encouragement, I'm opening this ticket, along with presenting some screenshots of the existing GUI. However it should be noted that some of the UX exists **to work around bugs in mono**; in particular I know we've hit issues with dialog boxes and additional windows.

(For those following along, my conference ends tomorrow, so I'll be playing CKAN catch-up soon after that.)

pjf:
Here are some sample screenshots, for guidance only. :)

![ckan1](https://cloud.githubusercontent.com/assets/12072/8863739/234cb3ce-3150-11e5-84aa-19818d190224.png)
![ckan2](https://cloud.githubusercontent.com/assets/12072/8863745/36ddabd2-3150-11e5-877c-00a4549076f5.png)
![ckan3](https://cloud.githubusercontent.com/assets/12072/8863761/5bb5138c-3150-11e5-8824-1b9be436b4f9.png)

mgsdk:
I played a bit with https://github.com/mono/xwt at one point, and I think @RichardLake also looked at it. I can't recall if there were any issues (apart from the daunting task of rewriting the GUI).

RichardLake:
I think the issue I had with XWT was the lack of binary releases which leads to having to have a OSX machine to build the Cocoa engine backends. Another library I was looking at is https://github.com/picoe/Eto which doesn't seem to have the same issue with needed a OSX build machine.

Crazor:
I thought about doing this for some time now, too. What I'd like to do is write a native Cocoa gui and call into CKAN. So far I'm not quite sure how to do the CKAN calls. All I found so far is how to call Objective-C stuff from Mono, but not the other way round. I suppose if there is a seperate OS X GUI to maintain, it could as well be as native as possible.
I figure I could always call ckan.exe and parse the output, but that would be kinda awkward...

mgsdk:
This may be of interest of you want to call Mono code from a native application: http://www.mono-project.com/docs/advanced/embedding/

mgsdk:
If anyone manages to get embedded Mono to work let me know. I've tried to get the embedding process working on Arch Linux to no avail.

[...]

Issue #760, opened 2015-01-20
quirky GUI on OSX


basbebe:
Hey!

The GUI has some annoying bugs on OS X (for me). I can't say if this is CKAN or Monos behavior.
I understand that the regular Hotkeys work the windows way (CTRL-\* instead of CMD-*).
But whenever I try to enter something into the search field I have to type really really slow because otherwise letter are being dropped or doubled.
And the Interface doesn't update itself. e.g. "updating": The tabs switches to "Updating repositories" and then switches back to "Manage mods" when it is done. Although the tab "Manage mods" is then being active, the UI still shows the "Updating repositories" part where the list of the mods should be… I then have to switch tabs manually again to reset the view.

alampros:
The GUI is definitely a bit finicky on OSX. I sometimes have to "kick" it by scrolling the mod list or attempting to select text in the log output to get it to repaint correctly. The issue can be seen when a mod update fails (in this case EVE) in this recording:
![txabnkmc5c](https://cloud.githubusercontent.com/assets/297461/5842571/e607d028-a173-11e4-9a3a-130c87ffb810.gif)

calebmennen:
Not sure if this is feasible, but could you just signal mono to redraw the gui after every big change (change of tab, check box, etc)? That might help resolve the issue

savagerose:
Looks like the tab issue is a long standing bug in Mono's Carbon-based WinForms driver: https://bugzilla.novell.com/show_bug.cgi?id=509245

I tried getting it to run in X11 instead to work around the issue but recent versions of the OSX Mono distribution don't have that compiled in.

I'm getting a development environment set up so I'll see if I can find a workaround like the one calebmennen mentioned.

savagerose:
Manually setting visibility on the TabPage / controls in the TabPage works pretty well, but there's a case where the mod list doesn't reappear immediately after using the refresh button. It flashes on screen for a second but then disappears until it redraws again, not sure exactly what's going on there.


The search box behavior appears to be a combination of adding even a single row to a DataGridView being really slow and this key handling bug: http://mono.1490590.n4.nabble.com/Incorrect-missing-and-duplicate-keypress-events-td4658863.html

Unless there's a way to address one of those directly, the best work around I can think of would be to wait until the user finishes typing to update the table.

savagerose:
[...]

PR #2341, opened 2018-03-04
Fix building on macOS

phardy:
This PR tweaks the build script to make it easier to build on macOS.

`/usr/bin/sed` on macOS is a BSD sed, which was returning successfully with an empty string trying to parse the Cake version out of packages.config, thereby breaking the build. Rather than expect the user to install GNU sed, I've changed the expression to something that should work on both and (at least to me) reads more easily.

This branch now builds as-is for me on macOS. I've done minor testing that it's still OK with GNU sed on macOS and changing the build script to use `gsed`. I'm hoping CI will catch any other major problems. :)

### DIFF
--- a/build
+++ b/build
@@ -1,5 +1,6 @@
 #!/bin/sh
 set -e
+set -x
 
 arg0=""
 remainingArgs=""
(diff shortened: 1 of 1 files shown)


HebaruSan:
Still works for me on Ubuntu. :+1: 

You're compiling on MacOSX? I hope that means you'll submit Mac-related fixes... :)

phardy:
Got a build environment running today. Expect some updates to the wiki shortly too. :)

My free time is limited, but the winforms problem bothers me and I'd like to take a stab at a native UI.

HebaruSan:
Cool! Issues you might be fixing by doing that, for future reference: #760, #1334, #2272.

politas:
Builds fine under Windows, too.

PR #2354, opened 2018-03-11
Fix very slow response to checkbox changes

### LINKED ISSUE #2310: Slow when choosing mods to install or uninstall
Not the actual uninstall, but when deselecting the mods which I want to remove, sometimes I have to wait a minute or more before it actually deselcts

### LINKED ISSUE #2319: Version requirements not obeyed for recommends
Background
----------

**CKAN Version:** 
1.24.0

**KSP Version:** 
1.3.1

**Operating System:** 
Windows, Linux

No

**Relevant Metadata:**

https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/StockalikeSolarSystem.netkan

```
	"depends": [
		{ "name": "ModuleManager" },
		{ "name": "Kopernicus", "version": "2:release-1.3.1-3" },
		{ "name": "SigmaDimensions", "version": "0.9.6" },
...
	],
	"recommends": [
		{ "name": "SigmaBinary-core", "version": "1:v1.6.9" },
		{ "name": "SigmaSciDefRenamer" },
		{ "name": "Kronometer", "version": "v1.3.1-1" }
	],
```

Problem
-------

1. Run GUI
2. Check StockalikeSolarSystem
3. Apply Changes

ConsoleUI works the same as GUI (i.e., wrong).

SigmaBinary-core 1:v1.6.9 should be presented as a recommendation.

The correct version is installed if you run `ckan install StockalikeSolarSystem` from the command line:

```
$ ckan install StockalikeSolarSystem
About to install...

 * Stockalike Solar System v0.5.5 (github.com, 11.3 MB)
 * Module Manager 3.0.4 (cached)
 * Kopernicus Planetary System Modifier 2:release-1.3.1-3 (cached)
 * ModularFlightIntegrator 1.2.4.0 (cached)
 * Sigma Dimensions 0.9.6 (github.com, 839.4 KB)
 * SASS-DRP 2.0 (cached)
 * SASS-NH 2.0.1 (cached)
 * SASS-OPM 1.2.4 (cached)
 * SASS-RevSS 0.6 (github.com, 162.5 MB)
 * SASS-RevJ 1 (cached)
 * SASS-Saru 0.1 (cached)
 * SASS-SE 1.0 (cached)
 * SASS-ER 1.2.4 (github.com, 1.1 MB)
 * SASS-StockalikeNeptune 1 (cached)
 * SASS-UL 0.5.4 (cached)
 * Sigma Binary 1:v1.6.9 (github.com, 752.5 KB)
 * Science Def Renamer v0.3.2 (github.com, 7.5 KB)
 * Kronometer v1.3.1-1 (cached)

Continue? [Y/n]  
```

**What happened instead?**
SigmaBinary-core 1:v1.6.11 is presented as a recommendation in GUI.

**Screenshots:**

image

image

HebaruSan:
## Problem

Checking or unchecking a checkbox in the main mod list can be very slow, sometimes taking over a minute! Some slowness was observed previously, but it got worse with v1.24.0.

## Causes

My investigation found three sources of slowness.

### Redundant ZIP validation

This line took 8+ seconds in one test that I observed, while everything else was quick:

https://github.com/KSP-CKAN/CKAN/blob/624839ffaa51de25c607531b23f9efb46d4ee612/GUI/MainChangeset.cs#L49-L54

The culprit is `IsCachedZip`. This function performs validation of the ZIP file in the cache for the given module, including CRC checks. This could easily take several seconds for a large download. As of #2277, `Main.UpdateChangesDialog` was running it for every mod listed in the (un)installation confirmation dialog so it could determine whether to tell the user the mod was already "(cached)".

If the mod isn't in the cache, this call is fast. That's why the problem seemed more severe for uninstallation: if you're uninstalling a mod, that means it's installed, which means it's probably in your cache.

### Double reaction to clicks

Relatedly, all of the post-click work is being done twice. When the user clicks, `ModList_CellValueChanged` calls `GUIMod.SetInstallChecked`, which sets `DataGridViewCheckBoxCell.Value`, which calls `ModList_CellValueChanged` again! It stops after two iterations because `GUIMod.IsInstallChecked` is set and checked, but each time through, `ModList_CellValueChanged` recalculates all the dependencies and conflicts. So the baseline response time for clicking a checkbox is twice as long as it should be.

### Slow lookups in `Registry.LatestAvailableWithProvides`

This function is responsible for the core of a lot of CKAN's relationship logic. It takes as its parameters a module identifier and two constraints, a game version and a relationship specification. Its job is to return every `CkanModule` that could reasonably satisfy the given requirement (but only the latest one per identifier).

Unfortunately it works by searching the entire registry for modules compatible with the given version, then exhaustively checking every one of them for a matching `"provides"` property. A typical call to this function takes 400 milliseconds, and we need a lot of them for many common operations. For example, if you click to install USI-LS, you'll see around 3 seconds of delay while the registry is searched and searched again for each of its 4 dependencies.

## Changes

Fixes #2310.

### Less ZIP validation

Now we use `IsMaybeCachedZip` instead. This function just checks whether a file associated with the requested URL exists in the cache, without running the extra validation checks. This makes more sense now that we prevent invalid files from entering the cache as of #2243.

The same change is made for the same reasons to a similar call in `UpdateRecommendedDialog`, affecting recommended and suggested mods, and for `ModuleInstaller`'s installation confirmation prompt.

### Single reaction to clicks

`GUIMod.SetInstallChecked` no longer sets `DataGridViewCheckBoxCell.Value` if it already has the desired value. `SetInstallChecked` is a bit funny in that the source of its input is not clearly defined; it might be called to _react_ to a change in the grid (when all that should happen is that we set `GUIMod.IsInstallChecked`), or it might be called to _initiate_ a change to the grid (when `DataGridViewCheckBoxCell.Value` needs to be changed). The behavior now should be better in the reactive case without harming the initiating case.

### Index the `"provides"` property

Now instead of exhaustively searching the registry in `LatestAvailableWithProvides`, we precalculate an index for it to use when the registry is first initialized. This index contains a mapping from provided identifiers to providing `AvailableModules`. When we need to look up a provided module, the index lets us jump to the specific set of `AvailableModules` that we need to consider, and we simply ask them to provide us their latest modules matching our constraints. We then filter these to ensure that the latest version actually provides our target identifier, in case their `"provides"` relationships have since been modified to no longer do so.

This change reduces the typical run time of `LatestAvailableWithProvides` from 400 milliseconds to essentially 0. There are positive knock-on effects throughout the code base, wherever this function is used:

- Selecting or deselecting mods is instantaneous
- The relationships tab in GUI is noticeably faster
- The install flow is sped up
- Recommendation and suggestion listings are sped up

### Recommendations obey version requirements

Building on these changes, since `LatestAvailableWithProvides` now always obeys the version properties of a relationship, the GUI code for optional dependencies is updated to pass the specific versions found based on those relationships to the installer. This is done by tracking suggested and recommended modules as full `CkanModule` objects instead of string identifiers, which also saves a few redundant lookups here and there.

This is also fixed in ConsoleUI, which now also shows the versions of optional dependencies.

Fixes #2319.

## Other changes

A bunch of the mod list code in `Main.cs` is moved to `MainModList.cs` to make it easier to find when investigating this sort of thing. No functional changes are made to this code.

`GUIUser` is split out from `Main.cs` into its own `GUIUser.cs` file. No functional changes here either.

The GUI change preview screen could give wrong versions for dependencies, because we were using `GUIMod.Version`, which is always the latest compatible version. Now it's fixed by using `CkanModule.version` instead, which is the true version being installed.


linuxgurugamer:
Oh Yay, this was sooo bad.  Thank you for fixing it.

HebaruSan:
Remaining delays are in `Registry.LatestAvailableWithProvides`, 400 milliseconds per call, which has to be called once per mod in the change set and each of their dependencies, each time `RelationshipResolver` runs, which happens twice per click. So a mod with five dependencies would incur a cost of 5 * 2 * 0.4 = 4 seconds just from this one code path.

Real numbers for the example of [USI-LS](https://raw.githubusercontent.com/BobPalmer/CKAN/master/USI-LS.netkan), with 4 dependencies:

- Click at 21:40:41.680
- Finishes processing at 21:40:44.883
- 44.883 - 41.680 = 3.203 total delay after click

Issue #2355, opened 2018-03-12
Unable to build CKAN.dmg on macOS


phardy:
While poking around testing out #2329, I discovered that `macosx/Makefile` is unable to build `CKAN.app` or `CKAN.dmg` on macOS. I know, the irony is fantastic.

The easy issue is that BSD `cp` does not support the `-l` flag for creating hard links. A simple cross-platform fix for that is `s/cp -l/ln/`.

The hard issue is that the Makefile uses `genisoimage` to build an ISO9660 disk image with the Apple extensions. Without getting too far down the rabbithole, [`genisoimage`](https://wiki.debian.org/genisoimage) was created as a fork from mkisofs during a licensing squabble between the cdrtools author and Debian maintainer for cdrtools. It's been unmaintained for several years, and while in theory it'll probably build on FreeBSD, I'm not even sure if I've found the right repository for it.

The Debian folk recommend the `xorriso` package as a drop-in replacement for genisoimage and mkisofs, and it's something that's readily available of macOS through assorted package manages. It doesn't support the `-apple` flag, but does generate an image that I'm able to open and manipulate as expected.

So, I'd like to suggest switching to xorriso for building the macOS packages. Admittedly it's a minor change, and has a little overhead for the folks maintaining build agents (sorry!), but would make life easier for people actively working on `CKAN.app` on a supported platform.

Longer term, it would be amazing to generate a rich dmg using a tool like [appdmg](https://github.com/LinusU/node-appdmg), but my brief research this evening has only found ones that rely on Darwin's `hdiutil`, so aren't portable.

PR #2356, opened 2018-03-12
Allow building CKAN.app on macOS

phardy:
Implements the changes I suggested in #2355 for building the app bundle and disk image on macOS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ addons:
             - make
             - sed
             - libplist-utils
-            - genisoimage
+            - xorriso
             # Stuff for building .deb files
             - gzip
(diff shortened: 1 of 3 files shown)


HebaruSan:
Nice! I think we'd need to change this instance of `genisoimage` to `xorriso` as well so Travis could handle the build:

https://github.com/KSP-CKAN/CKAN/blob/624839ffaa51de25c607531b23f9efb46d4ee612/.travis.yml#L18-L33

Can you add that?

phardy:
Oh, thanks. I had no idea that existed.

HebaruSan:
Thanks for the contribution!

PR #2364, opened 2018-03-16
Install old versions of mods on double click from version list

### LINKED ISSUE #2077: Select which version of a mod to install
**CKAN Version:** 1.22.3

**Operating System:** Win7 64

**The issue you are experiencing:** The UI does not offer a way of installing an older compatible version of a mod while the internal backend is already capable of doing so.

**How to recreate this issue:** Try to install any mod's version that's not the dark green one in the version list.

I've a day ago or so updated AnalogControlContinued because well, why not, it's only the update to make it 1.3 compatible even though I'm still playing 1.2.2.
Turned out that it was a big mistake because something went seriously wrong and it was spewing NREs all over the place.
So I thought "ok no big deal, I'll just revert to the version before".

Well, turns out the UI of CKAN doesn't let you do that, or if it does, it's less obvious than Blender's UI.
I reverted it over the console by now which worked but it'd be really nice if the UI could get updated to allow selecting which version will be installed.

Easiest idea I'd have for this right now would be with a double click in the respective version in the mod's version list.

HebaruSan:
## Background

Currently GUI can only install the latest compatible version of a given mod. So while there may be versions 1.0, 1.1, 1.2, and 1.3 of a mod in the registry, you can only install 1.3.

It may be desirable to install older versions of a mod in some cases. For example, I need to do this to investigate a bug report involving upgrades; if I can only install the latest versions, then I would have to install a bunch of mods and then wait for one of them to upgrade, and repeating such a test would involve more waiting. (Yes, there's a way to do it in Cmdline, but it's nice to have in GUI as well.)

There's a Versions tab that lists all versions of the mod and indicates which are compatible with the current game version. It's not interactive at the moment.

image

## Changes

Now if you double click one of the green versions, CKAN prompts you to confirm that you want to install it:

image

If you click yes, then off to the install flow you go:

image

And when it finishes, the mod appears as installed and upgradeable in the mod list, as expected:

image

From a code perspective, the "import from .ckan file" code is refactored to a standalone function in MainInstall.cs, which is now shared with the newly added double click handler for the versions list view. To get the `CkanModule` from the list view, we now set them into the `.Tag` properties of the list view items.

Fixes #2077.

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -941,51 +941,7 @@ namespace CKAN
 
                 menuStrip1.Enabled = false;
 
-                RelationshipResolverOptions install_ops = RelationshipResolver.DefaultOpts();
-                install_ops.with_recommends = false;
-
-                try
-                {
-                    // Resolve the provides relationships in the dependencies
-                    List<ModChange> fullChangeSet = new List<ModChange>(
-                        await mainModList.ComputeChangeSetFromModList(
-                            registry_manager.registry,
-                            new HashSet<ModChange>()
-                            {
-                                new ModChange(
-                                    new GUIMod(
-                                        module,
-                                        registry_manager.registry,
-                                        CurrentInstance.VersionCriteria()
-                                    ),
-                                    GUIModChangeType.Install,
-                                    null
-                                )
-                            },
-                            ModuleInstaller.GetInstance(CurrentInstance, GUI.user),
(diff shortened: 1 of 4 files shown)


Olympic1:
This is the thing I was working on after the right-click menu, but couldn't it it to work. Glad you figured it out.

This will need a rebuild.

PR #2366, opened 2018-03-17
Clean up url handler registration

### LINKED ISSUE #2362: CKAN.URLHandlers error every time when opening CKAN on Windows
I've received the following error every time while opening any version of CKAN I've tried on Windows 10 64-bit. It doesn't seem to affect any of the functions of the tool, but I thought I'd share it anyway, since searching for the problem does not yield a solution.

```
8468 [1] ERROR CKAN.URLHandlers (null) - There was an error while registering the URL handler for ckan:// - Cannot delete a subkey tree because the subkey does not exist.
8485 [1] ERROR CKAN.URLHandlers (null) -    at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at Microsoft.Win32.RegistryKey.DeleteSubKeyTree(String subkey, Boolean throwOnMissingSubKey)
   at CKAN.URLHandlers.RegisterURLHandler_Win32()
   at CKAN.URLHandlers.RegisterURLHandler(Configuration config, IUser user)
```

HebaruSan:
## Background

When it starts up, the CKAN GUI adds support for `ckan://identifier` URLs on Windows and Linux.

On Windows, it does this by adding this registry key:

image

CKAN might need to escalate privileges to make these changes. It detects this after the fact by catching an `UnauthorizedAccessException` when it attempts to set the registry key. If this exception is thrown, ckan.exe re-launches itself as administrator with the `registerUrl` command line flag, which prompts the user whether to allow the increased privileges and then sets the registry key.

## Problem

The user might see this error when launching CKAN:

```
8468 [1] ERROR CKAN.URLHandlers (null) - There was an error while registering the URL handler for ckan:// - Cannot delete a subkey tree because the subkey does not exist.
8485 [1] ERROR CKAN.URLHandlers (null) -    at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at Microsoft.Win32.RegistryKey.DeleteSubKeyTree(String subkey, Boolean throwOnMissingSubKey)
   at CKAN.URLHandlers.RegisterURLHandler_Win32()
   at CKAN.URLHandlers.RegisterURLHandler(Configuration config, IUser user)
```

If this is fixed, other errors can appear:

```
12186 [1] ERROR CKAN.URLHandlers (null) - There was an error while registering the URL handler for ckan:// - Cannot write to the registry key.
12200 [1] ERROR CKAN.URLHandlers (null) -    at System.ThrowHelper.ThrowUnauthorizedAccessException(ExceptionResource resource)
   at Microsoft.Win32.RegistryKey.EnsureWriteable()
   at Microsoft.Win32.RegistryKey.DeleteSubKeyTree(String subkey, Boolean throwOnMissingSubKey)
   at CKAN.URLHandlers.RegisterURLHandler_Win32()
   at CKAN.URLHandlers.RegisterURLHandler(Configuration config, IUser user)
```

## Causes

The `DeleteSubKeyTree` call doesn't throw `UnauthorizedAccessException` when it can't do its job as we would hope. So this call should have its exceptions caught and ignored just like the lines above it.

But if `UnauthorizedAccessException` _is_ thrown, it's re-thrown after it is caught and handled:

https://github.com/KSP-CKAN/CKAN/blob/9ee358455b338c1164d8605dfd19559ee661919d/GUI/URLHandlers.cs#L51-L67

This is redundant, because no other code is going to be able to do anything more to help the situation than has already been done, and if the privilege-escalated sub-process is able to set the key, then the user has no reason to care about this exception.

## Changes

Now `DeleteSubKeyTree` is called within the `try` block above its current location, and we don't re-throw any `UnauthorizedAccessException`s. This prevents the error message from bothering the user.

Fixes #2362.

### DIFF
--- a/GUI/URLHandlers.cs
+++ b/GUI/URLHandlers.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Diagnostics;
 using System.IO;
+using System.Reflection;
+using Microsoft.Win32;
 using IniParser;
 using IniParser.Exceptions;
 using IniParser.Model;
@@ -11,11 +13,11 @@ namespace CKAN
     public static class URLHandlers
     {
         private static readonly ILog log = LogManager.GetLogger(typeof(URLHandlers));
-        public const string UrlRegistrationArgument = "registerUrl";
+        public  const  string UrlRegistrationArgument = "registerUrl";
 
         private static string MimeAppsListPath = ".local/share/applications/mimeapps.list";
         private static string ApplicationsPath = ".local/share/applications/";
-        private const string HandlerFileName = "ckan-handler.desktop";
+        private const  string HandlerFileName  = "ckan-handler.desktop";
 
         static URLHandlers()
         {
@@ -35,7 +37,7 @@ namespace CKAN
                 {
                     RegisterURLHandler_Linux();
                 }
-                else if(Platform.IsWindows)
+                else if (Platform.IsWindows)
                 {
                     try
                     {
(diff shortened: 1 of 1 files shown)


politas:
So, does this leave the GUI running with elevated privileges? If so, that could be part of why so many people end up running CKAN as administrator (It "worked fine" the first time they ran it!)

HebaruSan:
No, the special `registerUrl` parameter makes it _only_ do the URL registration stuff. The GUI itself is only initialized if that argument is absent:

https://github.com/KSP-CKAN/CKAN/blob/a6619fef5cd12ed771606dd0ee8a841514c6b1dd/GUI/Program.cs#L27-L37

politas:
Ah, thanks, I figured that would be in there somewhere.

PR #2371, opened 2018-03-18
Allow installation to the Missions folder

### LINKED ISSUE #2370: [Request] Install Missions
The new DLC introduced the Mission Builder. This allows players to create missions and share them with the community.

When you created a mission, KSP exports it as a zip file to `KSP_ROOT/Missions/_Exports`. This zip can be shared with the community. When downloaded you have to place the zip in `KSP_ROOT/Missions`. When you select **Play Missions** in KSP, it moves all the zip files to `KSP_ROOT/Missions/_Imported` and extracts the contents to `KSP_ROOT/Missions`.

Now I tested this a bit and this can also be used by CKAN. We can just go directly to the last step and extract the contents of the missions zip file to `KSP_ROOT/Missions`.

**What to do to make this work:**
- [x] Allow CKAN to install to the `Missions` folder
- [x] Add `Missions` folder to spec
- [x] Add `Missions` folder to the schema

Olympic1:
Fixes #2370


Tests/Data/MissionTest-1.0.ckan
```diff
@@ -0,0 +1,18 @@
+{
+    "spec_version"   : "v1.24",
```
Olympic1: We'll have to change this to 1.26 after its release

politas: Actually, there is absolutely no reason we can't make it v1.25. We aren't doing odd-numbered "development" releases.

PR #2386, opened 2018-03-24
Encode spaces in URL output

HebaruSan:
## Problem

Currently in [the output for NetKAN pull request validation](https://ci.ksp-ckan.org/job/NetKAN/7091/console), the download URLs are printed with spaces unescaped:

image

This trips up the Jenkins site's attempt to make the URLs clickable, which could be handy if someone investigating an issue wants to inspect the downloaded files locally. Only the part leading up to the first space is detected, which isn't the full URL that's being downloaded.

## Changes

Now we replace spaces in this output with `%20` as per standard URL encoding.

I looked into the proper functions for encoding URLs in C#, and none of them seemed appropriate for this. The problem is that `Uri.ToString()` outputs a _partially_ valid URL (correctly formatted but may contain spaces), which none of the various encoding functions expect. Some of them would encode the `://` part as `%3A%2F%2F`, which we definitely don't want, and the rest would miss the spaces. So I just went with `.Replace`.

This change affects **only** these specific output statements; no changes are made to the internals of URL handling for downloads, metadata, etc.

## Trivia

You may be familiar with the practice of encoding spaces in URLs as `+`. TIL that this is only valid in the keys and values of form fields, but _not_ in the base URL to which those keys and values are sent (or anywhere in non-form URLs)! This is why I went with `%20` in this case.

### DIFF
--- a/Core/Net/NetAsyncDownloader.cs
+++ b/Core/Net/NetAsyncDownloader.cs
@@ -123,7 +123,9 @@ namespace CKAN
         {
             for (int i = 0; i < downloads.Count; i++)
             {
-                User.RaiseMessage("Downloading \"{0}\"", downloads[i].url);
+                // Encode spaces to avoid confusing URL parsers
+                User.RaiseMessage("Downloading \"{0}\"",
+                    downloads[i].url.ToString().Replace(" ", "%20"));
 
                 // We need a new variable for our closure/lambda, hence index = i.
                 int index = i;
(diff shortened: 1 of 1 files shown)


politas:
What does url.value return?

HebaruSan:
> What does url.value return?

A compiler error, because there is no such property (I tried both `value` and `Value):

```
  Net\NetAsyncDownloader.cs(167,38): error CS1061: 'Uri' does not contain a definition for 'value' and no extension method 'value' accepting a fi
rst argument of type 'Uri' could be found (are you missing a using directive or an assembly reference?) [C:\Users\Paul\github\CKAN\Core\CKAN-core
.csproj]

  Net\NetAsyncDownloader.cs(167,38): error CS1061: 'Uri' does not contain a definition for 'Value' and no extension method 'Value' accepting a fi
rst argument of type 'Uri' could be found (are you missing a using directive or an assembly reference?) [C:\Users\Paul\github\CKAN\Core\CKAN-core
.csproj]
```

politas:
Ah, because they are System.uri rather than System.Security.Policy.url

PR #2392, opened 2018-03-27
Sort by "update"-column on clicking "add available updates"

DasSkelett:
Hey, I just thought of a little enhancement, and decided to create the PR myself.

So now, after clicking the "Add all available updates" button, the modlist automatically sorts by "update" column and focuses/selects and scrolls to first row.

I'm not that experienced at programming and furthermore this is my first commit on CKAN and just copy-pasted together code which seemed to do the right things, after trying to get an overview of the code for some hours :D

Well, it worked after building, so yeah.

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -477,16 +477,20 @@ namespace CKAN
                 }
             }
 
-            
+            // set new sort column
             var new_sort_column = ModList.Columns[1];
             var current_sort_column = ModList.Columns[configuration.SortByColumnIndex];
-
+            
             // Reset the glyph.
             current_sort_column.HeaderCell.SortGlyphDirection = SortOrder.None;
             configuration.SortByColumnIndex = new_sort_column.Index;
             UpdateFilters(this);
 
             ModList.Refresh();
+
+            // Selects the top/bottom row and scrolls the list to it.
+            DataGridViewCell cell = ModList.Rows[0].Cells[2];
+            ModList.CurrentCell = cell;
         }
 
         public void UpdateModContentsTree(CkanModule module, bool force = false)
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -477,6 +477,15 @@ namespace CKAN
                 }
             }
 
+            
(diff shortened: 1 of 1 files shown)


politas:
This seems like something that should be an option. I'm sure some users would find it extremely annoying, I'd be happy for it to be the default behaviour, though. I realise that this would bump up the complexity of the coding significantly, so @DasSkelett, if you'd like some help with that, just say so and I'm sure someone would be happy to make a PR to your branch to handle the option-setting.

DasSkelett:
Thanks for your thoughts.
I started to add an checkbox in the settings GUI, I think I can handle it myself, but it wouldn't be bad if someone reviews the code after that.

So should it be opt-in or opt-out?

politas:
Make it opt-out.

Dazpoet:
It failed and travis posted something about tcp errors so I restarted the job and it passed... sounds like a network issue more than a codeissue.

DasSkelett:
Ah thanks :D

politas:
Could you move the "Hide Epochs" button into the new "More Settings" group box? It doesn't really belong in Updates.

DasSkelett:
Moved it, now looking like this:
ckansettings

One more question:
Why are most controls of the `AutoUpdateGroupBox` depending on `AutoUpdate.CanUpdate`?
Can't we just block the `InstallUpdateButton`, if we can't overwrite the exe?
So instead:
```
private void InstallUpdateButton_Click(object sender, EventArgs e)
{
    if(AutoUpdate.CanUpdate)
    {
        Hide();
        Main.Instance.UpdateCKAN();
    }
    else
   {
        GUI.user.RaiseError("Can't autoupdate. Please check https://github.com/KSP-CKAN/CKAN/ for help!")
   }
}
```
This would solve the annoying Designer Error in VisualStudio and the Settings window would look nicer if `AutoUpdate` isn't available.

politas:
DasSkellet, can you raise a new PR for the change in controls dependency? I'm not sure what designer error you are getting, as I don't use VisualStudio most of the time.

Issue #2391, opened 2018-03-26
Fail to update repository after CKAN is sitting open for a while


linuxgurugamer:
I dont the how long it takes, but this is repeatable:
Windows 10, CKAN 1.24.0

Open CKAN, everything works fine, refreshes work etc.
Leave it open, come back several hours later, and hit refresh, get the following message:

```
Updating repositories...
Connection to https://github.com/KSP-CKAN/CKAN-meta/archive/master.tar.gz could not be established.
Repository update failed!
```

I'm guessing that a cookie or session is timing out, but CKAN isn't restarting from the beginning

HebaruSan:
About how many hours to wait before the problem starts? I launched it about an hour ago, and so far it's OK.

Olympic1:
Tested this too, opened ckan before I left for work. 8h later it couldn't update the repos.

linuxgurugamer:
I can't say how long, I'd have to do long tests.  I'll try to keep an eye on it.
But, it seems that once it fails, it doesn't seem to try to restart the connection

DasSkelett:
Can confirm, had the same problem few days ago.

I can't tell the exact time either, might have been 3-4 hours.

HebaruSan:
Mine's still fine after 11+ hours. :(

Dazpoet:
Had this happen to me once aswell, it disappeared (and didn't come back) after a restart of CKAN. Was probably going for 2-3 hours before it happened.

linuxgurugamer:
@HebaruSan it has to be untouched for a long time.

and, it happens to me all the time

HebaruSan:
@linuxgurugamer, I wasn't using it at all except for clicking Refresh every few hours. Still fine at 24h. Based on that, I'm going to guess that some other factor triggers this, not time (or not _just_ time).

linuxgurugamer:
a few hours isn't enough.  
Open a CKAN, do a refresh, then leave it for 8 hours.

HebaruSan:
I left it for 24 hours.

HamiltonKsos:
I get this immediately on opening CKAN v 1.24.1
Using: Cinnamon Mint 3.6.7. 
New install of mono and CKAN. 

Does not happen to Windows version.

> Connection to https://github.com/KSP-CKAN/CKAN-meta/archive/master.tar.gz could not be establisied.

blackliner:
[...]

PR #2399, opened 2018-04-08
Clean-up and debuggability

### LINKED ISSUE #2376: Build Warning async method
We are getting this build warning from GUI/Main.cs

> Build succeeded.
>
>Warnings:
>
>/mnt/raid/Repositories/monodevelop/CKAN/CKAN.sln (Build) ->
>(Build target) ->
>/mnt/raid/Repositories/monodevelop/CKAN/GUI/CKAN-GUI.csproj (default targets) ->
>/usr/lib/mono/xbuild/14.0/bin/Microsoft.CSharp.targets (CoreCompile target) ->
>
>	Main.cs(725,28): warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.
>
>	 1 Warning(s)
>	 0 Error(s)

We should either re-write this to not be async or use an await, or add a '#pragma warning disable 1998` and matching enable around the method.

### LINKED ISSUE #2396: Ckan Mac version show a cross after "recommended mods"
Background
----------

**CKAN Version:** 
1.24.0

https://github.com/KSP-CKAN/CKAN/files/1756778/CKAN.zip (Using this version since the original .DMG version didn't work)

**KSP Version:** 
1.4.1

**Operating System:** 
macOS 10.13.3 (17D102)

No

Problem
-------

Install KAS, KIS, Trajectories, Dmagic orbital science, Kerbal Alarm clock (basically all mods that has "recommended" page

Run normally

**What happened instead?**
It shows a big red cross after showing the dependency/recommended window

**Screenshots:**

screen shot 2018-03-31 at 01 25 01

HebaruSan:
The past few weeks have seen a steady trickle of obscure, difficult to investigate issues. This pull request attempts to extend CKAN's debugging capabilities in specific areas to make it possible to make progress on those issues. It also addresses a few very small or simple problems that came up.

## Debian dependency on webrequest

A user reported in a random comment on unrelated issue #1817 that the .deb package should depend on the `webrequest` package. This is now added to `debian/control.in`.

## Remove unnecessary `async` keyword

`installFromckanToolStripMenuItem_Click` is marked as `async`, but it doesn't need to be, which causes a warning to be raised at compile time. This is now removed.

Fixes #2376.

## Include webexception info in failure message for repo updates

#2391 contains multiple attestations that refreshing the registry sometimes fails, possibly related to how long CKAN has been running. A`WebException` is thrown when this happens, but its details are discarded.

Since the cause of this is unclear, we now add the exception to the error output. This is not a fix but may allow us to discover how to create a fix.

## Suppress exceptions in OnPaint and skip Mono 3 workaround in Mono 5

A user reports in #2396 that the CKAN window turns into a big red X sometimes on MacOSX. This is apparently something WinForms does when you throw an exception in a control's `OnPaint` method, which we have here:

https://github.com/KSP-CKAN/CKAN/blob/a495f435fff9a8e2c55ca5fdca0cb92d9a03c0e8/GUI/MainModList.cs#L427-L444

That was originally done to work around a bug in Mono 3 that was fixed in Mono 4. The idea was that the code would use the `IsMonoFour` check to avoid running on versions where the bug was fixed. However, now that we are mostly on Mono 5, `IsMonoFour` is no longer an adequate check; since Mono 5 is not Mono 4, this Mono 3 workaround has come back to life and is running on Mono 5.

Now all of that code is wrapped in a `try`/`catch` block that discards all the exceptions, and `IsMonoFour` is replaced with `IsMonoFourOrLater` to reflect the real intent of this code, preventing the Mono 3 workaround from running on Mono 5.

May or may not fix #2396.

## More debug messages around lock acquisition and registry creation

KSP-CKAN/NetKAN#6459 describes an issue in which CKAN "has stopped working" immediately on launch, even if you just run `ckan scan` from the command line. When we tried capturing the `--debug` output, the last message was

https://github.com/KSP-CKAN/CKAN/blob/a495f435fff9a8e2c55ca5fdca0cb92d9a03c0e8/Core/Registry/RegistryManager.cs#L241

In a working install, the next message would be one of these:

https://github.com/KSP-CKAN/CKAN/blob/a495f435fff9a8e2c55ca5fdca0cb92d9a03c0e8/Core/Registry/RegistryManager.cs#L291

https://github.com/KSP-CKAN/CKAN/blob/a495f435fff9a8e2c55ca5fdca0cb92d9a03c0e8/Core/Registry/RegistryManager.cs#L317

So something between those statements is somehow causing a crash. There's a lot of code in that gap. This pull request adds more `DebugFormat` statements to log all those various steps so we can narrow down what's happening.

Not a fix for KSP-CKAN/NetKAN#6459, but should help us to figure out what's going on with that issue.

### DIFF
--- a/Core/Net/Repo.cs
+++ b/Core/Net/Repo.cs
@@ -23,9 +23,6 @@ namespace CKAN
         private static readonly ILog log = LogManager.GetLogger(typeof (Repo));
         private static TxFileManager file_transaction = new TxFileManager();
 
-        // Forward to keep existing code compiling, will be removed soon.
-        public static readonly Uri default_ckan_repo = CKAN.Repository.default_ckan_repo_uri;
-
         /// <summary>
         /// Download and update the local CKAN meta-info.
         /// Optionally takes a URL to the zipfile repo to download.
@@ -92,9 +89,9 @@ namespace CKAN
             {
                 repo_file = Net.Download(repo);
             }
-            catch (System.Net.WebException)
+            catch (System.Net.WebException ex)
             {
-                user.RaiseMessage("Connection to {0} could not be established.", repo);
+                user.RaiseMessage("Failed to download {0}: {1}", repo, ex.ToString());
                 return null;
             }
 
@@ -386,7 +383,7 @@ Do you wish to reinstall now?", sb)))
             // Use our default repo, unless we've been told otherwise.
             if (repo == null)
             {
-                repo = default_ckan_repo;
+                repo = CKAN.Repository.default_ckan_repo_uri;
             }
 
             List<CkanModule> newAvail = UpdateRegistry(repo, ksp, user);
--- a/Core/Platform.cs
+++ b/Core/Platform.cs
@@ -20,7 +20,8 @@ namespace CKAN
         {
             IsMac = IsRunningOnMac();
             IsMono = IsOnMono();
-            IsMonoFour = IsOnMonoFour();//Depends on IsMono
+            // Depends on IsMono
+            IsMonoFourOrLater = IsOnMonoFourOrLater();
         }
 
         // From https://github.com/mono/monodevelop/blob/master/main/src/core/Mono.Texteditor/Mono.TextEditor/Platform.cs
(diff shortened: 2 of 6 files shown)


HebaruSan:
@politas, this would be a decent time to release a 1.24.1 version, in my opinion. The most-encountered bugs in 1.24.0 are fixed, and we don't have very many larger (=higher-risk) changes yet.

I'd be willing to try my hand at preparing such a release if you agree it should happen but would like someone else to take care of it. Just say the word.

politas:
Next release has to be 1.25.0, as we have a spec change as part of the changes. I was thinking it was about time.

politas:
I have free time tomorrow, so I'll push it out then, if no one complains.

PR #2407, opened 2018-04-11
Move AutoUpdate.CanUpdate check to resolve VisualStudio Designer Error

DasSkelett:
# Problem:
VisualStudio is very touchy to manual changes in the Designer, especially in `InitializeComponent()`.
At the moment, it throws an error every time you want to open the Design View of the Settings Form, because the `AutoUpdateGroupBox` and its members depend on `AutoUpdate.CanUpdate` via an if-statement, which checks if the exe is rewriteable.

# My Proposal:
Move the `AutoUpdate.CanUpdate`-check to `InstallUpdateButton_Click` instead of blocking some GUI labels and buttons. Throw an error (should it throw a kraken?), if it's false.
This would also make the SettingsWindow don't look as holey in this case.

### DIFF
--- a/GUI/SettingsDialog.cs
+++ b/GUI/SettingsDialog.cs
@@ -399,8 +399,16 @@ namespace CKAN
 
         private void InstallUpdateButton_Click(object sender, EventArgs e)
         {
-            Hide();
-            Main.Instance.UpdateCKAN();
+            if (AutoUpdate.CanUpdate)
+            {
+                Hide();
+                Main.Instance.UpdateCKAN();
+            }
+            else
+            {
+                GUI.user.RaiseError("Can't autoupdate. Please check https://github.com/KSP-CKAN/CKAN/ for help!");
+            }
+            
         }
 
         private void CheckUpdateOnLaunchCheckbox_CheckedChanged(object sender, EventArgs e)
--- a/GUI/SettingsDialog.Designer.cs
+++ b/GUI/SettingsDialog.Designer.cs
@@ -43,7 +43,6 @@
             this.ClearCKANCacheButton = new System.Windows.Forms.Button();
             this.CKANCacheLabel = new System.Windows.Forms.Label();
             this.AutoUpdateGroupBox = new System.Windows.Forms.GroupBox();
-            this.HideEpochsCheckbox = new System.Windows.Forms.CheckBox();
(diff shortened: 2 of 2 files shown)


DasSkelett:
It was added in #2187 (https://github.com/KSP-CKAN/CKAN/pull/2187#issuecomment-348795156 / https://github.com/KSP-CKAN/CKAN/pull/2187/commits/bd006337eb8a861a21cf3b2196818c3d2a1e2427#diff-77661e48afbbc3ba9cde9b81106f683e).

Olympic1:
Wil test tonight after work but looks good at first glance.

politas:
Oooh, I just made my executable read-only and yes, that is rather ugly. Thanks for this PR, DasSkelett!

GUI/SettingsDialog.cs
```diff
@@ -399,8 +399,16 @@ private void CheckForUpdatesButton_Click(object sender, EventArgs e)
...
-            Hide();
-            Main.Instance.UpdateCKAN();
+            if (AutoUpdate.CanUpdate)
+            {
+                Hide();
+                Main.Instance.UpdateCKAN();
+            }
+            else
+            {
+                GUI.user.RaiseError("Can't autoupdate. Please check https://github.com/KSP-CKAN/CKAN/ for help!");
```
HebaruSan: Would someone opening this URL be able to figure out what was going on and/or what to do about it?

politas: Ooh, that's a good point. Just merged this, unfortunately. Pointing to the latest release would probably be better, no?

DasSkelett: Oups, you were a bit faster than me.
Actually, I wanted to ask if this would work first, before detailing the error message...
Well, should I open a new PR?

politas: Make it point to https://github.com/KSP-CKAN/CKAN/releases/latest I think. @HebaruSan , does that seem like a useful link to you?

DasSkelett: Yes, maybe something like:
"Error during update.
Can't update automatically, because ckan.exe is read-only or we are not allowed to overwrite it. Please update manually via https://github.com/KSP-CKAN/CKAN/releases/latest ." ?

HebaruSan: That message sounds good to me!

PR #2408, opened 2018-04-13
Better AutoUpdate.CanUpdate Error Message

DasSkelett:
Revised error message if `AutoUpdate` is blocked with more details and better link.

@HebaruSan ; @politas 
#2407


politas:
Looks pretty good to me.

Issue #2412, opened 2018-04-18
Mod information sidebar is compressed


AndersDotExe:
CKAN version 1.25.0

KSP version 1.4.2.2110, Steam 64-bit

OS: Win10 64-bit

No changes have been made to my GameData folder apart from a few mods I had to download manually

Issue: mod information sidebar, [screencap here](https://imgur.com/a/sXgwA), happens every time CKAN starts, replaced CKAN install with new version ~10 minutes ago, has been a persistent issue for 1-2 months, a bit after KSP 1.4

![screenshot](https://i.imgur.com/vS4dpEO.png)

HebaruSan:
Does it reset to that state after you drag the splitter?

AndersDotExe:
Ah, I tried dragging the splitter and that seems to have fixed it. I did look for a splitter before, just couldn't find it. It does not seem to be resetting to the compressed state either. Thanks!

PR #2413, opened 2018-04-18
Splitter and tabstrip visual improvements

HebaruSan:
## Problem

GUI has a vertical splitter between the mod list and the mod info pane and a horizontal splitter within one of the tabs of the mod info pane, but some users have found this to be non-obvious and ask whether information has been removed from the display (see #2412 and KSP-CKAN/NetKAN#6288).

The tab style chosen for the mod info pane is also a bit ugly and gray.

screenshot

splitter low splitter high

## Changes

Now each pane of these splitters has a minimum size:

- The mod info pane always has enough width for all of its tabs
- The top part of the metadata tab always has enough height for both of its fields
- The bottom part of the metadata tab always has enough height for its fields
- The vertical splitter itself is widened slightly to try to make it more obvious
  (I tried to enlarge the horizontal splitter as well, but setting the property had no effect.)

And the tabs are changed to a more conventional style, which also gives the tab pages a white background.

Minimum sizes of the panes illustrated:

image

### DIFF
--- a/GUI/Main.Designer.cs
+++ b/GUI/Main.Designer.cs
@@ -453,12 +453,14 @@
             // 
             this.splitContainer1.Panel1.Controls.Add(this.ModList);
             // 
+            this.splitContainer1.Panel1MinSize = 200;
             // splitContainer1.Panel2
             // 
             this.splitContainer1.Panel2.Controls.Add(this.ModInfoTabControl);
+            this.splitContainer1.Panel2MinSize = 300;
             this.splitContainer1.Size = new System.Drawing.Size(1522, 836);
             this.splitContainer1.SplitterDistance = 1156;
-            this.splitContainer1.SplitterWidth = 6;
+            this.splitContainer1.SplitterWidth = 10;
             this.splitContainer1.TabIndex = 7;
             // 
             // ModList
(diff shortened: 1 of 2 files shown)


DasSkelett:
Can you push all labels in the lower part of metadata tab some Pixels down?
I think the Version text is sticking a bit to close to the splitter.

politas:
Have you fixed it so the main splitter still works when tabs other than the mod details is shown? (An issue I noticed the other day but haven't had a chance to look at in detail, yet)

HebaruSan:
I haven't consciously made any changes along those lines, but testing it out I don't see a problem with it either. I might not be understanding the issue though.

politas:
Seems to work fine in this branch, so all good. I do note that the abstract is using the "Input Box" background colour, while the rest of the metadata is using the "Windows" background. Can we get them using the same style?
image

HebaruSan:
@politas, that was to fix that control being _gray on white_ on Windows, which happened because it's a `RichTextBox` with `ReadOnly` set to true, and .NET and Mono apparently don't use the same colors for tabs. I've changed it to a `Label` and removed the color manipulation, so it should match the main background on both platforms now.

politas:
Cool. The .Net/Mono promise of multi-platform UI development is not quite fulfilled, eh?

PR #2418, opened 2018-04-22
Hide StatusProgress-bar if user cancels installation

### LINKED ISSUE #2417: Canceling on "recommended mods" tab doesn't stop the progress bar
First of all, I think I can fix this myself, I just wanted to inform there's a bug. I'll open a PR then.
If I need help, I let you know.

Background
----------

**CKAN Version:** 
v1.25.1
(Should be in regular 1.25 too)

Problem
-------

Choosed a mod (with recommends) to install and applied the changes, than cancelled on the "Choose recommended mods"-tab page.

the progress bar stops running and the status label says sth. like "User cancelled Installation process!".

**What happened instead?**
The progress bar (`StatusProgress`) in the lower right corner keeps cycling as well as the `StatusLabel` just says "Error!" which is a bit vague and also incorrect, as there is no real error, just a cancellation.

**Screenshots:**
progressbar

DasSkelett:
**Problem:**
If you manually cancel the Installation process, the lower right Progress bar keeps moving.

**Solution:**
Reset and hides `StatusProgress` bar if user cancels installation.
Give a better `StatusMessage` too (as "Error!" is not only verry vague, it's incorrect too in this case.)

Fixes #2417

### DIFF
--- a/GUI/MainInstall.cs
+++ b/GUI/MainInstall.cs
@@ -392,18 +392,11 @@ namespace CKAN
                 RetryCurrentActionButton.Visible = false;
                 UpdateChangesDialog(null, installWorker);
             }
-            else if(!result.Key && installCanceled){
+            else if(installCanceled)
+            {
                 // User cancelled the installation
-                AddStatusMessage("Manually cancelled installation process!");
-                SetDescription("The user cancelled the installation manually!");
+                FailWaitDialog("Installation cancelled by user!", "User cancelled the installation manually!", "Install failed!");
                 UpdateChangesDialog(result.Value, installWorker);
-                RetryCurrentActionButton.Visible = true;
-                Util.Invoke(statusStrip1, () =>
-                {
-                    StatusProgress.Value = 0;
-                    StatusProgress.Style = ProgressBarStyle.Continuous;
-                    StatusProgress.Visible = false;
-                });
             }
             else
             {
--- a/GUI/MainInstall.cs
+++ b/GUI/MainInstall.cs
@@ -392,6 +392,19 @@ namespace CKAN
                 RetryCurrentActionButton.Visible = false;
                 UpdateChangesDialog(null, installWorker);
             }
+            else if(!result.Key && installCanceled){
+                // User cancelled the installation
+                AddStatusMessage("Manually cancelled installation process!");
+                SetDescription("The user cancelled the installation manually!");
+                UpdateChangesDialog(result.Value, installWorker);
+                RetryCurrentActionButton.Visible = true;
+                Util.Invoke(statusStrip1, () =>
+                {
+                    StatusProgress.Value = 0;
+                    StatusProgress.Style = ProgressBarStyle.Continuous;
+                    StatusProgress.Visible = false;
+                });
+            }
             else
             {
                 // There was an error
--- a/GUI/MainInstall.cs
+++ b/GUI/MainInstall.cs
@@ -330,11 +330,13 @@ namespace CKAN
 
             // If we're going to install something anyway, then don't list it in the
             // recommended list, since they can't de-select it anyway.
-            foreach (var kvp in selectable)
+            // The ToList dance is because we need to modify the dictionary while iterating over it,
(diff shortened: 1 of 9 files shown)


DasSkelett:
Interestingly, the GUI switches to the Modlist tab afterwards, and I have no ideo, why, as it shouldn't do it on errors though (but I don't know how to simulate an error to test this).

GUI/MainInstall.cs
```diff
@@ -390,6 +390,19 @@ private void PostInstallMods(object sender, RunWorkerCompletedEventArgs e)
                 RetryCurrentActionButton.Visible = false;
                 UpdateChangesDialog(null, installWorker);
             }
+            else if(!result.Key && installCanceled){
```
HebaruSan: We don't need to check `result.Key` here, because we're already in the `else` block of `if (result.Key)`.

Also, can you put the curly brace on a new line? This file follows that convention currently.

GUI/MainInstall.cs
```diff
@@ -390,6 +390,19 @@ private void PostInstallMods(object sender, RunWorkerCompletedEventArgs e)
...
+                // User cancelled the installation
+                AddStatusMessage("Manually cancelled installation process!");
+                SetDescription("The user cancelled the installation manually!");
+                UpdateChangesDialog(result.Value, installWorker);
+                RetryCurrentActionButton.Visible = true;
+                Util.Invoke(statusStrip1, () =>
+                {
+                    StatusProgress.Value = 0;
```
HebaruSan: It would be nice if we didn't have to directly manipulate specific properties of GUI objects here. Some sort of helper function to encapsulate this at a higher level.

`MainWait.cs` already has a partial API along those lines with `ShowWaitDialog` and `HideWaitDialog`. Maybe it should also have `FailWaitDialog`, which could then be called by the cancel and error blocks here? Looks like the parameters should be the messages to display.

DasSkelett: `HideWaitDialog` already gets passed a bool `success`, but it's not used yet.
Either we keep this and differentiate `success` or we create an explicit function.

HebaruSan: I thought the idea was to avoid hiding it, though, so the user can look at the errors.

DasSkelett: Well, as you mention it...
It would be a bit odd to have a funtion called HideWaitDialog to not hide the WaitDialog in case of an error...

DasSkelett:
So, I created a new `FailWaitDialog()` with error messages as Parameters wich executes all GUI calls.
Also I deleted the `HideWaitDialog()` on `CancelCurrentActionButton_Click()` so the Status log stays open.

And because I have bad reflexes and I clicked cancel for testing too slow a lot of times, I realized, that after clicking cancel we don't check if the installation still concluded successfully, resulting in those now false error messages and the GUI not recognizing the newly installed mods.

DasSkelett:
Need to sleep a night about this as this is some high level sh*t which definitely exceeds my capabilities by now.

politas:
>this is some high level sh*t which definitely exceeds my capabilities

Boy, I know what _that_ feels like!

DasSkelett:
Okay this was some strange rebase...

DasSkelett:
So I added the check if the installation completed despite "canceling", Abit hacky, but working (I think)
Not ready-to-merge yet, because for unknown causes the retry button stopped working and the GUI laggs shortly after canceling...

politas:
What build fail? :grinning:

politas:
@DasSkelett , is this ready to merge, or are you still needing to work on it some more?

DasSkelett:
Hey, back again, sorry for the long wait.
I'll try to get behind what I did back then. As I just tested, if I cancel AFTER the install completed, it still says error, the "apply changes" button is still falsely active (not thaaaat bad, because if you click it it says already installed, but would be nicer if it would be deactivated) and the retry button moves to a buggy position.
I think I should open a new PR without all those rebase errors? @politas @HebaruSan

PR #1866, opened 2016-08-12
Fix issue where CKAN would crash after installing mods if you are sorting by KSP version.

Antialiased:
This was related to mods with ksp_version 'any' in JSON, which is handled differently to mods with no ksp_version field, and the UI code was not knowledgeable about the difference

PR #2420, opened 2018-04-22
Replace max KSP version empty string with "any" in GUI

DasSkelett:
If there's no max KSP version defined in a mod's metadata, CKAN GUI displays it as an empty string.
See maxV1

It would be prettier and more informative if there is some text, so I changed the empty string to ~~"N/D"~~ "any".
This applies to the lower right metadata box and also the column in the modlist, as you can see here:
maxV2

### DIFF
--- a/GUI/GUIMod.cs
+++ b/GUI/GUIMod.cs
@@ -115,7 +115,7 @@ namespace CKAN
             // KSP.
             if (latest_available_for_any_ksp != null)
             {
-                KSPCompatibility = KSPCompatibilityLong = registry.LatestCompatibleKSP(mod.identifier)?.ToString() ?? "N/D";
+                KSPCompatibility = KSPCompatibilityLong = registry.LatestCompatibleKSP(mod.identifier)?.ToString() ?? "any";
 
                 // If the mod we have installed is *not* the mod we have installed, or we don't know
                 // what we have installed, indicate that an upgrade would be needed.
--- a/GUI/GUIMod.cs
+++ b/GUI/GUIMod.cs
@@ -115,7 +115,7 @@ namespace CKAN
             // KSP.
             if (latest_available_for_any_ksp != null)
             {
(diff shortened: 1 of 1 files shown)


DasSkelett:
Would N/A be better?

HebaruSan:
Unlimited is clearer, but kind of long.
What about something like "Any"?

politas:
Didn't we used to have "Any" and we stopped? Ah, yes #1866 and related #1882 and #1887

politas:
Yes, I think in between we changed the display and sorting values for this column to different values.

politas:
I do think "any" would be preferable to "N/D" (which I admit I am not instantly parsing, myself). Not defined?

DasSkelett:
Yes, N/D stands for "Not Defined", but if we already have trouble understanding it, it's probably not a good idea. So, "any" now?

politas:
If you can push a new commit with that change in the next 12 hours, I'll include it in the 1.25.1 release I'm going to put out in the morning.

DasSkelett:
Done!

linuxgurugamer:
Might I suggest that you make sure this works EXACTLY as the current KSP-AVC works (which I'm now maintaining)?  Some bugs were fixed which might affect this

linuxgurugamer:
Not saying this is wrong, just want to review it together

DasSkelett:
This shouldn't affect AVC at all, since it just changes just the representation of an empty maxVersion string  in the GUI.

politas:
>Might I suggest that you make sure this works EXACTLY as the current KSP-AVC works (which I'm now maintaining)? Some bugs were fixed which might affect this

That's a completely different bit of code (assuming you're referring to the parsing of max KSP from .version files)

yalov:
empty to "any"? 
Then maybe 9.99.99, 99.99.99, 1.99.99 also to any?

https://github.com/KSP-CKAN/NetKAN/issues/5356

PR #2424, opened 2018-04-24
Treat installed DLC as compatible dependency

HebaruSan:
## Problem

Modules that depend on `MakingHistory-DLC` are treated as incompatible in mod listings, even if you have the DLC.

You are still able to _install_ such modules via CmdLine, even though they are displayed as incompatible.

Relatedly, the MakingHistory-DLC entry in the relationships tab always shows up as red and "(not indexed").

image

## Cause

DLC checks were only added to functions that check whether an identifier is _installed_. To determine whether a module is _compatible_, we look up all of its dependencies via `LatestAvailableWithProvides` regardless of whether they're already installed, which returns a module, and there are no modules for Dlc objects, so this function doesn't find anything for `MakingHistory-DLC`.

The relationships tab in GUI also checks dependencies itself, and also wasn't updated.

## Changes

Now the helper function that checks dependencies treats installed DLC identifiers as compatible.

The relationships tab also now checks for installed DLC and shows it as compatible if it's installed.

image

Fixes https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1250-wallops/&do=findComment&comment=3362539

Tests are added:

- When no DLC is installed, make sure we consider a mod depending on MakingHistory-DLC as incompatible
- When MakingHistory-DLC is installed, make sure we consider a mod depending on MakingHistory-DLC as compatible
- When MakingHistory-DLC 1.1.0 is installed, make sure we consider a mod depending on MakingHistory-DLC 1.1.0 as compatible
- When MakingHistory-DLC 1.0.0 is installed, make sure we consider a mod depending on MakingHistory-DLC 1.1.0 as incompatible

### DIFF
--- a/Core/Registry/Registry.cs
+++ b/Core/Registry/Registry.cs
@@ -498,7 +498,8 @@ namespace CKAN
                 {
                     try
                     {
-                        if (!LatestAvailableWithProvides(dependency.name, ksp_version).Any())
+                        if (!dependency.MatchesAny(null, InstalledDlls.ToHashSet(), InstalledDlc)
+                            && !LatestAvailableWithProvides(dependency.name, ksp_version).Any())
                         {
                             return false;
                         }
--- a/GUI/MainModInfo.cs
+++ b/GUI/MainModInfo.cs
@@ -302,6 +302,13 @@ namespace CKAN
             }
             catch (ModuleNotFoundKraken)
             {
+                // Maybe it's a DLC?
+                ModuleVersion installedVersion = registry.InstalledVersion(identifier, false);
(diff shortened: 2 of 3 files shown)


HebaruSan:
DO NOT MERGE, several unintended changes were committed because I forgot to add just the relevant chunk rater than the whole file...

HebaruSan:
This should probably have some tests as well. I'll look into that, please stay on the line for the next available representative.

politas:
Sounds like a good candidate for a 1.25.1 release.

PR #2426, opened 2018-04-25
Ignore splitter exceptions

HebaruSan:
## Problem

In #2413, I set the `Panel1MinSize` and `Panel2MinSize` properties on our splitters, because they look like a simple way to ensure that important content is visible. Of course with .NET, even something this simple has to be arduous:

image

This happens at start-up if our saved splitter size is no longer allowed in the current window size. You can get to such a state easily by resizing the window with no problems, but then if you close and restart this exception is thrown.

## Changes

We may not be able to convince Microsoft how dumb it is to throw this exception, but now we do the next best thing and ignore it. The min sizes of the mod info pane are also reduced to about 75% of their previous values to allow them to fit a bit better into small windows.

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -280,7 +280,15 @@ namespace CKAN
             Location = configuration.WindowLoc;
             Size = configuration.WindowSize;
             WindowState = configuration.IsWindowMaximised ? FormWindowState.Maximized : FormWindowState.Normal;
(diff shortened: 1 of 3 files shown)


HebaruSan:
@politas, we should not ship a 1.25.1 without this fix (or a better solution to the same problem). This exception locks the user out of the application, so the only way to work around it is to edit the config file where the saved splitter position is stored.

Issue #2428, opened 2018-04-26
Self-conflicts treated as actual conflicts at upgrade


DasSkelett:
Background
----------

**CKAN Version:** 
v1.25.1

**KSP Version:** 
v1.4.2.2110

**Operating System:** 
Windows10 64bit

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** 
No

Problem
-------

**What steps did you take in CKAN?**
a) Marked all mods to update via `Add available updates`-button and clicked apply
b) Marked all mods manually (not via button)

**What did you expect to happen?**
a) Mods to update show in changset tab
b) `Apply changes`-button gets activated

**What happened instead?**
a) Mods not showing in changeset tab, clicking apply does nothing
b) `Apply changes`-button doesn't get activated

**Screenshots:**
a) Selecting all via button:
![nochanges1](https://user-images.githubusercontent.com/28812678/39316805-55b552c2-497a-11e8-8080-c3b4e780d3f3.JPG)
then clicking apply (no changes listed):
![nochanges2](https://user-images.githubusercontent.com/28812678/39316817-5e363ab0-497a-11e8-9428-bfb774ed3391.JPG)

b) Selecting manually (button won't get activated):
![nochanges3](https://user-images.githubusercontent.com/28812678/39316828-66fac62a-497a-11e8-96a4-8a980fbc8c03.JPG)

Thoughts
----------
This might either be because scatterer has its new version listed as conflict:
![scattererconflict](https://user-images.githubusercontent.com/28812678/39316951-b9f7c5b2-497a-11e8-9d92-503660d056d6.JPG)
~~or the only change since v1.25.0 affecting anything in that area I know would be #2425~~
Scratch that, I can't imagine how refreshing the row could cause this.

DasSkelett:
So uninstalling and reinstalling worked perfectly.

HebaruSan:
Confirmed that the weird self-conflict is a factor. I installed `v0.0329_for_1.4.2` and clicked Refresh and Add available updates, then saw the empty changeset problem as reported. Then I restarted after manually editing my `registry.json` to remove the conflict of Scatterer with Scatterer, and now it's not empty:

![image](https://user-images.githubusercontent.com/1559108/39320669-5b78f9da-4974-11e8-840a-11be590c8097.png)

(Not sure why the sunflare and config packages aren't listed there, since they're also checked for upgrade.)

[...]

PR #2429, opened 2018-04-26
Don't Force Apply button active when no update selected

DasSkelett:
Problem:
----------
#2428 showed, that `MarkAllUpdatesToolButton_Click()` always sets `ApplyToolButton.Enabled` to true, even without changeset. This is a very edge case of a bug which should get fixed soon (the changeset shouldn't be empty at all), but nevertheless, it would be better if the button doesn't get activated manually.

Solution:
---------
So I removed this `ApplyToolButton.Enabled` call, as it is unnecessary in addition, because `ApplyToolButton` gets always updated in `ChangeSetUpdated()` which again is called by `UpdateChangeSetAndConflicts()`, which then again is called by `ModList_CellValueChanged()`, in turn automatically called on any checkbox-changes...
_Ouf_
In short, `ApplyToolButton.Enabled` is called everywhere, anytime.

Furthermore I rearranged (and hopefully) improved the code for marking mods for updates.
Feedback welcome!

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -466,16 +466,7 @@ namespace CKAN
 
         private void MarkAllUpdatesToolButton_Click(object sender, EventArgs e)
         {
-            foreach (DataGridViewRow row in ModList.Rows)
-            {
-                var mod = (GUIMod)row.Tag;
-                if (mod.HasUpdate && row.Cells[1] is DataGridViewCheckBoxCell)
-                {
-                    MarkModForUpdate(mod.Identifier);
-                    mod.SetUpgradeChecked(row, true);
-                    ApplyToolButton.Enabled = true;
-                }
-            }
+            MarkModsForUpdate();
 
             // only sort by Update column if checkbox in settings checked
             if (Main.Instance.configuration.AutoSortByUpdate)
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -466,7 +466,14 @@ namespace CKAN
 
         private void MarkAllUpdatesToolButton_Click(object sender, EventArgs e)
         {
-            MarkModsForUpdate();
+            foreach (DataGridViewRow row in ModList.Rows)
+            {
+                var mod = (GUIMod)row.Tag;
+                if (mod.HasUpdate)
+                {
+                    MarkModForUpdate(mod.Identifier);
+                }
+            }
 
             // only sort by Update column if checkbox in settings checked
             if (Main.Instance.configuration.AutoSortByUpdate)
@@ -487,8 +494,6 @@ namespace CKAN
             }
 
             ModList.Refresh();
-
-
         }
 
         public void UpdateModContentsTree(CkanModule module, bool force = false)
--- a/GUI/MainModList.cs
+++ b/GUI/MainModList.cs
@@ -219,21 +219,16 @@ namespace CKAN
             }
         }
 
-        public void MarkModsForUpdate()
+        public void MarkModForUpdate(string identifier)
         {
-            Util.Invoke(this, () => _MarkModsForUpdate());
+            Util.Invoke(this, () => _MarkModForUpdate(identifier));
         }
         
-        public void _MarkModsForUpdate()
+        public void _MarkModForUpdate(string identifier)
         {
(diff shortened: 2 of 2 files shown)


politas:
I can see the logic behind putting the loop in MarkAllUpdatesToolButton_Click and having the single mark for update code in MainModList. It makes the MarkModForUpdate method more reusable, and puts the logic of "Mark all available updates" where it arguably should be (attached to the button that the user clicks to do that).

On the other hand, it's looping inside a loop, which is inefficient. Do we need to loop across the entire gridview to find a specific identifier in ModList? Is there a way to directly access the correct row? Some kind of LINQ query?

I think it you want to put the "Mark all available updates" looping code into MainModList (where it may arguably belong as an action taken on the ModList) it would be ideal to have it as a separate method. Simply putting the loop from MarkAllUpdatesToolButton_Click into MainModList maintains the nested loop, so that would be redundant. Your new single loop method is much more efficient. If we were calling MarkModForUpdate from anywhere else, it would be an easy call. It doesn't seem rational to leave the old code there if it isn't used.

Is there a way to rescue MarkModForUpdate by removing its loop? 

Let me be clear, I'm entirely cool with that answer being "no", I'm just spilling out what's going on in my brain. There are many times with this code where I find myself thinking that the AwesomeCodersWhoCameBefore must have had Reasons for the choices they made. Those Reasons aren't necessarily right, though. :smiley:

politas:
So before your changes, it was split up . You've basically merged the two elements (Find all mods which have updates) fromMarkAllUpdatesToolButton_Click and (Change the update checkbox) from MarkModForUpdate. If we change MarkModForUpdate to get the relevant row using `DataGridViewRow row = mainModList.full_list_of_mod_rows[identifier];` then we get rid of the inefficiency and keep the concept of a method doing a single conceptual task. 

I'm assuming that refreshing the modlist will cause the status of the Apply button to change?

DasSkelett:
So like this now?
I dont't know what `row.Cells[1] is DataGridViewCheckBoxCell` is for, this just checks if the first cell of the row is a checkbox?

politas:
It's testing to see if the second column (UpdateCol) is a checkbox. You might want to check what makes the status of that column change. I would have thought that mod.HasUpdate includes the same checks, so it may be redundant.

SetUpgradeChecked has a comment,  `//Contract.Requires<ArgumentException>(row.Cells[1] is DataGridViewCheckBoxCell);` So checking for that might be to avoid a type mismatch exception? I don't know why the Contract.Requires is commented out.

DasSkelett:
I got it:
When  we initiate all the cells in `ConstructModList()`, the cell `updating` (= `row.Cells[1]`) gets either declared as
* `DataGridViewCheckboxCell` and casted to `DataGridViewCell` if the mod is installed, or
* `DataGridViewTextBoxCell` if the mod is not installed:
```
var updating = mod.IsInstallable() && mod.HasUpdate
    ? (DataGridViewCell) new DataGridViewCheckBoxCell() {
        Value = myChange == null ? false
            : myChange.ChangeType == GUIModChangeType.Update ? true
            : false
     } : new DataGridViewTextBoxCell() {
         Value    = "-"
     };
```

But because `GUIMos.HasUpdate` = `registry.HasUpdate()` which checks if the mod is installed, the type check should be redundant:
```
/// <summary>
/// Is the mod installed and does it have a newer version compatible with version
/// We can't update AD mods
/// </summary>
public static bool HasUpdate(this IRegistryQuerier querier, string identifier, KspVersionCriteria version)
{
    CkanModule newest_version;
    try
    {
        newest_version = querier.LatestAvailable(identifier, version);
    }
    catch (ModuleNotFoundKraken)
    {
        return false;
    }
    if (newest_version == null) return false;
    return !new List<string>(querier.InstalledDlls).Contains(identifier)
        && querier.IsInstalled(identifier, false)
        && newest_version.version.IsGreaterThan(querier.InstalledVersion(identifier));
}
```

politas:
Cool. This should make for a good little speedup.

DasSkelett:
Ah, forgot to remove the comments...
We'll, may the next one tumbling over those lines worry his brain...

PR #2430, opened 2018-04-27
Ignore conflicts between versions of same mod

### LINKED ISSUE #2428: Self-conflicts treated as actual conflicts at upgrade

HebaruSan:
## Problem

If a mod conflicts with itself, it can no longer be upgraded as of #2339; see #2428 for details of a specific example involving Scatterer.

## Cause

When we upgrade, we ask the `SanityChecker` to look for missing dependencies and conflicts in a scenario in which the new versions are installed without removing the old versions. Since the old and new modules are both there, we check whether they conflict with one another, and a self-conflict is treated as a real conflict.

Prior to #2339, this was avoided by ignoring conflicts if the identifiers were the same, but when we made conflicts version-specific, we only ignored that conflict if the identifier _and version_ are the same.

## Changes

Now `SanityChecker` will ignore any conflict between two mods with the same identifier, as it did historically. Other conflicts will still care about versions, so #2339 is not voided.

A test is added to catch regressions.

Fixes #2428.

### DIFF
--- a/Core/Relationships/SanityChecker.cs
+++ b/Core/Relationships/SanityChecker.cs
@@ -132,18 +132,15 @@ namespace CKAN
             IDictionary<string, UnmanagedModuleVersion> dlc
         )
         {
-            modules = modules?.AsCollection();
-
             var confl = new List<KeyValuePair<CkanModule, RelationshipDescriptor>>();
-
             if (modules != null)
             {
-                foreach (var m in modules.Where(m => m.conflicts != null))
+                foreach (CkanModule m in modules.Where(m => m.conflicts != null))
                 {
-                    // Remove self from the list, so we're only comparing to OTHER modules
-                    var others = modules.Where(other => !ReferenceEquals(other, m)).AsCollection();
-
-                    foreach (var dep in m.conflicts)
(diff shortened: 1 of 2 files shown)


HebaruSan:
Tests in progress...

PR #2432, opened 2018-04-27
Replace max KSP "9.9.9" with "any"

yalov:
https://github.com/KSP-CKAN/CKAN/pull/2420
https://github.com/KSP-CKAN/NetKAN/issues/5356

CKAN GUI replace:

```
0.90    // unchanged
1.4     → 1.4.9
1.4.10  → 1.4.9
1.4.90  → 1.4.9
1.4.99  → 1.4.9
1.9.9   → any
9.99.99 → any
```

### DIFF
--- a/GUI/GUIMod.cs
+++ b/GUI/GUIMod.cs
@@ -115,23 +115,18 @@ namespace CKAN
             // KSP.
             if (latest_available_for_any_ksp != null)
             {
-                string str = registry.LatestCompatibleKSP(mod.identifier)?.ToString();
-                int major = registry.LatestCompatibleKSP(mod.identifier).Major;
-                int minor = registry.LatestCompatibleKSP(mod.identifier).Minor;
-                int patch = registry.LatestCompatibleKSP(mod.identifier).Patch;
+                KspVersion ver = registry.LatestCompatibleKSP(mod.identifier);
+                string str = ver?.ToString();
+                int major = ver.Major;
+                int minor = ver.Minor;
+                int patch = ver.Patch;
 
 
                 if (str == null
-                    || (major + 1) % 10 == 0 
-                    || (major == 1 && (minor + 1) % 10 == 0)
+                    || (major + 1) % 10 == 0                                  
+                    || (major == 1 && minor > 90 && (minor + 1) % 10 == 0)    // 1.99.x, not 1.9.x
                     )
                     KSPCompatibility = "any";
-
-                else if (major == 1
-                    && minor != -1 
-                    && (patch > 8 || patch == -1)
-                    )
-                    KSPCompatibility = major + "." + minor + ".9";
 
                 else
                     KSPCompatibility = str;
--- a/GUI/GUIMod.cs
+++ b/GUI/GUIMod.cs
@@ -115,7 +115,27 @@ namespace CKAN
             // KSP.
             if (latest_available_for_any_ksp != null)
             {
-                KSPCompatibility = KSPCompatibilityLong = registry.LatestCompatibleKSP(mod.identifier)?.ToString() ?? "any";
+                string str = registry.LatestCompatibleKSP(mod.identifier)?.ToString();
+                int major = registry.LatestCompatibleKSP(mod.identifier).Major;
+                int minor = registry.LatestCompatibleKSP(mod.identifier).Minor;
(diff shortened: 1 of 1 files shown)


HebaruSan:
I still don't understand why we wouldn't want to display accurately the values from the metadata.

DasSkelett:
Seems like a good approach to me, but should we really consider a.b.9 as the highest possible a.b version number?
It might be unlikely, but technically it is _possible_ that there is a double-digit version number in the future (a.b.cc). There already have been two-digit (0.90; 0.25; ...) Releases.

GUI/GUIMod.cs
```diff
@@ -115,7 +115,27 @@ public GUIMod(CkanModule mod, IRegistryQuerier registry, KspVersionCriteria curr
             // KSP.
             if (latest_available_for_any_ksp != null)
             {
-                KSPCompatibility = KSPCompatibilityLong = registry.LatestCompatibleKSP(mod.identifier)?.ToString() ?? "any";
+                string str = registry.LatestCompatibleKSP(mod.identifier)?.ToString();
+                int major = registry.LatestCompatibleKSP(mod.identifier).Major;
```
HebaruSan: Please don't call `LatestCompatibleKSP` more than once. It loops over all versions of the mod checking their compatibility, so the return value should go into a local variable.

GUI/GUIMod.cs
```diff
@@ -115,7 +115,27 @@ public GUIMod(CkanModule mod, IRegistryQuerier registry, KspVersionCriteria curr
...
+                int minor = registry.LatestCompatibleKSP(mod.identifier).Minor;
+                int patch = registry.LatestCompatibleKSP(mod.identifier).Patch;
+
+
+                if (str == null
+                    || (major + 1) % 10 == 0 
+                    || (major == 1 && (minor + 1) % 10 == 0)
```
HebaruSan: Treating 1.9 as "any" is going to get us in trouble if Squad ever releases a 1.10 or 2.0 version.

yalov:
`patch` only twice reached .5 so .9 should be ok for the patch.

Now I start thinking about minor... it will reach 1.10.0 one day... so maybe 1.9.9 better leave unchanged

@HebaruSan ok, will fix that

GUI/GUIMod.cs
```diff
@@ -115,7 +115,27 @@ public GUIMod(CkanModule mod, IRegistryQuerier registry, KspVersionCriteria curr
...
+                    )
+                    KSPCompatibility = "any";
+
+                else if (major == 1
+                    && minor != -1 
+                    && (patch > 8 || patch == -1)
+                    )
+                    KSPCompatibility = major + "." + minor + ".9";
```
HebaruSan: This would cause problems if Squad released a 1.4.10 version.

HebaruSan:
Regardless of whether we can convince ourselves now that .5 or .9 are safe, it raises the question again of why do this in the first place? It's hard coding a series of dubious assumptions about not just how many patches there will be but how Squad chooses to _represent_ those versions in their numbering scheme. If any of those assumptions is violated in the future, then this code will need to be patched again with the new assumptions.

There's also value in being able to inspect the true values in the GUI, for example if you want to find these mods with weird versions for testing purposes. All things considered, I think we should leave this code as it is.

yalov:
remove patch changing, but still don't wan't have 99.99.99's

yalov:
> if Squad ever releases a 1.10 or 2.0 version

ok... they really can switch to "chrome" numeration. 
Maybe you are right.

politas:
We have a table of actual ksp versions; maybe we can use that to programattically limit max ksp compatibility to the most recent version

So, if a mod says max version is 1.3.99, but we know that the highest patch number in the 1.3 series is 2, then show max KSP as 1.3.2.
I would still say replace anything with a major version above 2 with "any"

yalov:
where is this table?

maybe find real_max_major, real_max_minor, and real_max patch (now they are 1, 4, 5 respectively), and use them to determine what should be converted to "any", and should be used x.y.9 or x.y.99

DasSkelett:
https://github.com/KSP-CKAN/CKAN/blob/master/Core/builds.json

HebaruSan:
And you can access that file via:

```csharp
List<KspVersion> versions = new KspBuildMap(new Win32Registry()).KnownVersions;
```

PR #2437, opened 2018-04-29
"any" and "x.y.max" in the KSP Max column (GUI)

yalov:
https://github.com/KSP-CKAN/CKAN/pull/2432
add dictionary, should be ok. 

```
// don't like that, but it is the right way
0.90.0 → 0.90.0
0.90   → 0.90.9     

// there is v1.2.9, so 1.2.max is 1.2.99
1.2.9   → 1.2.9  
1.2.45  → 1.2.45
1.2.99  → 1.2.99
1.2.999 → 1.2.99

1.4     → 1.4.9   // until there isn't 1.4.9, then 1.4 → 1.4.99
1.4.3   → 1.4.3
1.4.90  → 1.4.9  // until there isn't 1.4.9, then 1.4.90 → 1.4.99
1.4.99  → 1.4.9

1.9 or 1.9.x   → any   // until there isn't 1.9.0, then 1.99.x → any
2.0.0 → any               // discussable
9.99.99 → any
```

### DIFF
--- a/GUI/GUIMod.cs
+++ b/GUI/GUIMod.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Windows.Forms;
 using System.Linq;
 using CKAN.Versioning;
+using CKAN.GameVersionProviders;
 
 namespace CKAN
 {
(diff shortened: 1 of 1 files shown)


yalov:
your question is about how it works at scan, you are trying to say what 1.2 or 1.2.99 really not compatible with 1.2.9, because 1.2.9 is 1.3.0.rc.

I am not trying to solve that, it is just about column gui. People who cares already set 1.2.8, but if someone set 1.2.999 it is works for 1.2.9 and column value should correspond that

PR #2442, opened 2018-05-08
Show conflict messages in status bar

HebaruSan:
## Problems

If you try to install two mods that conflict with one another, they'll be highlighted red, but there are no hints as to the cause if they're not on-screen.

If you manually install a mod, and then attempt to install or upgrade a mod in CKAN that conflicts with it, the mod is not highlighted red and there are no other clues as to what's going on, see KSP-CKAN/NetKAN#6533.

Some effort was made toward creating a popup for this in #2237, but it was not completed. Part of the challenge is that a popup can be disruptive and is not something we would want to happen every time the user clicks a checkbox.

## Causes

Conflicts previously were used to mark rows red, abort the creation of a change set, and leave the Apply button disabled, but specific descriptive messages weren't shown.

The code that highlights conflicts relies on `RelationshipResolver.ConflictList`, which only checks conflicts between `CkanModule`s and ignores DLLs and DLCs.

## Changes

The changes in this pull request are somewhat high risk and should be subjected to close scrutiny. It's possible that I missed an important test case, so please try any obscure things you can think of and report problems!

### Status bar shows conflict messages

Now if you select two conflicting mods for install, a short description of the conflict is shown in the status bar to explain why the rows are highlighted:

image

If more than one conflict is present, only one will be shown, because we stop looking once we find the first. If the displayed problem is fixed, then the other conflict will be shown. Hopefully this will be enough to help users figure out what to do.

Previously, clicking any row would clear the previous status bar message. This doesn't make sense anymore now that the status bar contains useful information about the current conflict. This is now removed, and instead the message is cleared if we are able to successfully calculate a change set without problems.

### Conflicts with DLLs included in `RelationshipResolver.ConflictList`

`SanityChecker` is rewritten to throw a new `BadRelationshipsKraken` exception instead of `InconsistentKraken`. This new exception inherits from the old one for backwards compatibility. Where `InconsistentKraken` provided only a list of opaque error strings, the new exception provides machine-readable descriptions of the problems to allow calling code to react to them.

`RelationshipResolver` now always runs `SanityChecker`. (`SanityChecker`'s exception is only passed along if not turned off in the resolver options, as before.) If any conflicts with DLLs and DLCs are found, these are added to the data structure that provides `RelationshipResolver.ConflictList` (by way of the new exception described above). This allows the conflicting row to be highlighted. A description of the conflict will be shown in the status bar as described above:

image

### Small things

The ", cannot install both" suffix is removed from the messages for conflicts, for consistency with other places that already don't have this.

In the course of investigation, I found several places where we check `variable is null`. The `is` operator in C# dynamically checks _types_, not values, and from what I can tell it's better to check `variable == null`. This pull request makes that change.

The `var` keyword obscures code because it's harder to tell what type an object is and what we're allowed to call on it. This is replaced with types in several places.

The status bar tooltip is now set to the text it contains, in case some message is very long and some runtime displays such tooltips (I was not able to get these to display on Mono).

Some conflict-checking code in `CkanModule` was duplicative of logic in `RelationshipDescriptor`. Now the duplication is removed by having one function call the other.

A blank comment line was missing from the auto generated code in `Main.Designer.cs`. This is now restored, which hopefully will help prevent parsing problems in Visual Studio.

### DIFF
--- a/Core/Extensions/EnumerableExtensions.cs
+++ b/Core/Extensions/EnumerableExtensions.cs
@@ -8,7 +8,7 @@ namespace CKAN.Extensions
     {
         public static ICollection<T> AsCollection<T>(this IEnumerable<T> source)
         {
-            if (source is null)
+            if (source == null)
                 throw new ArgumentNullException(nameof(source));
 
             return source is ICollection<T> collection ? collection : source.ToArray();
@@ -16,7 +16,7 @@ namespace CKAN.Extensions
 
         public static HashSet<T> ToHashSet<T>(this IEnumerable<T> source)
         {
-            if (source is null)
+            if (source == null)
                 throw new ArgumentNullException(nameof(source));
 
             return new HashSet<T>(source);
--- a/Core/Relationships/RelationshipResolver.cs
+++ b/Core/Relationships/RelationshipResolver.cs
@@ -201,13 +201,13 @@ namespace CKAN
             // adding them to the list. This *must* be pre-populated with all
             // user-specified modules, as they may be supplying things that provide
             // virtual packages.
-            foreach (var module in ckan_modules)
+            foreach (CkanModule module in ckan_modules)
             {
                 log.DebugFormat("Preparing to resolve relationships for {0} {1}", module.identifier, module.version);
 
                 //Need to check against installed mods and those to install.
                 var mods = modlist.Values.Concat(installed_modules).Where(listed_mod => listed_mod.ConflictsWith(module));
-                foreach (var listed_mod in mods)
+                foreach (CkanModule listed_mod in mods)
                 {
                     if (options.procede_with_inconsistencies)
                     {
@@ -216,8 +216,8 @@ namespace CKAN
                     }
                     else
                     {
-                        throw new InconsistentKraken(string.Format("{0} conflicts with {1}, can't install both.", module,
-                            listed_mod));
+                        throw new InconsistentKraken(
+                            $"{module} conflicts with {listed_mod}");
                     }
                 }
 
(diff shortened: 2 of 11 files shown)


dbent:
I'll see if I can get to this in the next couple of days.

dbent:
Code looks good, although I haven't actually tested it yet. A few general remarks:

> In the course of investigation, I found several places where we check `variable is null`. The `is` operator in C# dynamically checks _types_, not values, and from what I can tell it's better to check `variable == null`. 

That used to be the case, C# 7.0 introduced a functional-style [pattern matching syntax](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#pattern-matching). The compiler is smart enough to recognize that an `is` check against a constant (like `null`) is just a equality check. `obj is null` wouldn't even be valid in C# versions prior to 7.0.

For confirmation you can check the IL produced by these two pieces of code:
```C#
object o = null;
if (o == null)
	Console.WriteLine();
```

```C#
object o = null;
if (o is null)
	Console.WriteLine();
```

In the first case you get:
```
IL_0000:  nop         
IL_0001:  ldnull      
IL_0002:  stloc.0     // o
IL_0003:  ldloc.0     // o
IL_0004:  ldnull      
IL_0005:  ceq         
IL_0007:  stloc.1     
IL_0008:  ldloc.1     
IL_0009:  brfalse.s   IL_0011
IL_000B:  call        System.Console.WriteLine
IL_0010:  nop         
IL_0011:  ret         
```

In the second case you get:
```
IL_0000:  nop         
IL_0001:  ldnull      
IL_0002:  stloc.0     // o
IL_0003:  ldloc.0     // o
IL_0004:  ldnull      
IL_0005:  ceq         
IL_0007:  stloc.1     
IL_0008:  ldloc.1     
IL_0009:  brfalse.s   IL_0011
IL_000B:  call        System.Console.WriteLine
IL_0010:  nop         
IL_0011:  ret    
```

They're identical. In both cases the two operands are simply loaded onto the stack (lines 3-4) and then compared using the `ceq` instruction (line 5), no dynamic type checking is performed.

> The `var` keyword obscures code because it's harder to tell what type an object is and what we're allowed to call on it. This is replaced with types in several places.

I generally disagree, one of the benefits of static typing is that we can leverage tools like the IDE to aid the programmer. We know what type an object is because the compiler, and by extension the IDE, knows. We know what we're allowed to call on it because the compiler, and by extension the IDE, knows.

In general, the only time I wouldn't use `var` is if somehow the specific type of a reference is critically important and wouldn't be caught statically. For example, let's say you perform some huge set of LINQ statements that produce an `IEnumerable<T>`. Iterating over this would lead to multiple enumeration and cause all the LINQ statements to reexecute multiple times. Therefore to indicate that the enumerable should be preemptively enumerated for performance reasons you may declare the variable storing the result as `List<T>`, forcing a `.ToList()` at the end and preventing multiple enumeration. But even then, I'd opt to use the most basic type or interface that captures my intent (which is simply to prevent multiple enumeration), in which case I'd use `ICollection<T>`.

politas:
Thanks, @dbent, I was hoping for exactly that sort of discussion!

In testing, I note that when the modlist is refreshed (removing all selections) thanks to changing KSP compatibility settings, the status bar message indicating conflicts does not disappear.

PR #2444, opened 2018-05-13
Improve handling of missing game version

### LINKED ISSUE #2443: Windows error after KSP reinstall: Object reference not set to an instance of an object
Background
----------

**CKAN Version:** 
1.25.1

**KSP Version:** 
1.4.3.2152

**Operating System:** 
Windows 10

KSP has been recently reinstalled

Problem
-------

(After reinstalling KSP)

1. Open CKAN (running from Download folder, not KSP folder)
2. CKAN notifies me that the KSP version has been updated, and raises the Compatible Ksp Versions window
3. Click Save

I expect my preferences to be saved

**What happened instead?**
An error is thrown: Object reference not set to an instance of an object

**Workaround**
Moving ckan.exe to KSP's root folder seems to have resolved the issue

N/A, but see Windows exception message below

```
   at CKAN.KSP.SaveCompatibleVersions()
   at CKAN.CompatibleKspVersionsDialog.SaveButton_Click(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnClick(EventArgs e)
```

HebaruSan:
## Problem

If you:

1. Have exactly one game instance
2. Set at least one additional compatible version in the compatible versions dialog
3. Remove `buildID.txt` and `readme.txt`, _without_ removing the KSP folder or the CKAN folder
4. Launch CKAN

... then on load, you get the compatible versions popup warning you about a version change (if in GUI), and clicking Save causes an exception:

```
   at CKAN.KSP.SaveCompatibleVersions()
   at CKAN.CompatibleKspVersionsDialog.SaveButton_Click(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnClick(EventArgs e)
```

(One way to make this happen is apparently to uninstall in Steam, then re-install to a different Steam library folder. See #2443.)

The problem is in Core, so the other UIs are also affected, but the details vary.

## Causes

`KSP.Version()` is allowed to be null in most places (and it happens when the buildID.txt and readme.txt files are missing), but the code to load and save compatible versions doesn't like it. (This is somewhat ironic as that code would be one of the better ways of handling this situation, since you could use it to set the correct game version manually.)

For saving, `ToString()` is called on the null reference.

For loading, if `compatible_ksp_versions.json` has a null value for the `VersionOfKspWhenWritten` property, it is passed to `KspVersion.Parse`, which throws exceptions on null. (That property can't be null currently, but it will be possible once we allow the save to proceed with a null.)

To be considered invalid, a game instance must lack a GameData folder. If it merely lacks the files that provide the game version, this is treated as fine; you can attempt to open and use such an instance even though we know it's missing at least some data that should be there.

## Changes

Now the null conditional operator ("`?.`") is used to avoid a null reference exception at save. This allows a null value to be written to the file.

Now we use `KspVersion.TryParse` instead of `Parse` to allow `VersionOfKspWhenCompatibleVersionsWereStored` to be set to null at load.

Finally, a game instance without a version is now treated as invalid. This means you can't open them from the Select KSP Install window, and they won't be auto-opened at launch. If your only known instance loses its version info (say, by uninstalling in Steam), then it'll automatically open the Select KSP Install window instead.

Fixes #2443.

### DIFF
--- a/Core/KSP.cs
+++ b/Core/KSP.cs
@@ -69,7 +69,7 @@ namespace CKAN
             }
         }
 
-        public bool Valid { get { return IsKspDir(gameDir); } }
+        public bool Valid { get { return IsKspDir(gameDir) && Version() != null; } }
 
         /// <summary>
         /// Create the CKAN directory and any supporting files.
@@ -124,7 +124,7 @@ namespace CKAN
         {
             CompatibleKspVersionsDto compatibleKspVersionsDto = new CompatibleKspVersionsDto();
 
-            compatibleKspVersionsDto.VersionOfKspWhenWritten = Version().ToString();
+            compatibleKspVersionsDto.VersionOfKspWhenWritten = Version()?.ToString();
             compatibleKspVersionsDto.CompatibleKspVersions = _compatibleVersions.Select(v => v.ToString()).ToList();
 
             String json = JsonConvert.SerializeObject(compatibleKspVersionsDto);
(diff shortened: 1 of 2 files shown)


politas:
`1) Failed : Tests.Core.KSP.SaveCompatibleVersions_MissingVersionData_NoCrash
  Expected: No Exception to be thrown`

New test is failing?

HebaruSan:
Oh yeah, that test no longer applies after the latest changes. Replaced with one confirming that Valid is false.

PR #2451, opened 2018-05-21
Remove "v" in "installed version" and "latest version" columns

yalov:
```
v3.4.5-whatever → 3.4.5-whatever
V.3.4.5         → 3.4.5
Version_3.4.5   → 3.4.5
V21             → 21
vessel_5.1      → vessel_5.1
```

cons:	
```
V1deo → 1deo
```

Not sure what is `2` in `2:v3.4.5`, and why it in `LatestVersion` and not in the GUI column, I mean where `2:` is removed from `LatestVersion`, but `2:` is leaved there.
```
2:v3.4.5 → 2:3.4.5
```

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -1162,6 +1162,21 @@ namespace CKAN
             }
         }
 
+
+
+        /// <summary>
+        /// Remove prepending v V. Version_ etc
+        /// </summary>
+        public static string StripV(string version)
+        {
+            Match match = Regex.Match(version, @"^(?<num>\d\:)?[vV]+(ersion)?[_.]*(?<ver>\d.*)$");
+
+            if (match.Success)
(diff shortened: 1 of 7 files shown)


yalov:
https://github.com/semver/semver/issues/235

> In my opinion the leading 'v' belongs to a tag (e.g. on a git repo) but does not belong to the version string itself

yalov:
> We have a setting to hide them in the mod list
then maybe also hide from update or install mods process window?

check out [x]Science SCAN Version Tab - not sure what happens there, but looks like really vV5.13 < v5.14, epoch is needed.
SEP uses v-or-not-v by random, that why 2.1.5 in the bottom.
QuickStart changed author, so v is disappear and epoch is neaded.

All this wouldn't be there if v is excluded from SCAN in the core... in the some  'Version' core-object ?

politas:
https://github.com/KSP-CKAN/CKAN/issues/963 is relevant here.

PR #2463, opened 2018-06-20
Add description to ModInfoTab

### LINKED ISSUE #2458: Suggestion: Show "description" for a mod in the GUI
Background
----------

**CKAN Version:** v1.25.1

Suggestion
-------

Show both the abstract and description for a mod. To be fancy, use the abstract as a summary which, if clicked on, opens the description.

There's currently not much point adding a description instead of an abstract in NETKAN as it's never seen in the GUI (which, I would hesitate to guess, over 99% of users use instead of the command-line version).

politas:
Adds a Mod's "description" beneath the "abstract" in the top part of the ModInfoPanel

Closes #2458


HebaruSan:
Looks good. It would be nice to shrink the gap between the abstract and the description to fit the content, as in a flow layout, but after extensive experimentation I was only able to make it look _worse_ (WinForms strikes again!).

Working on the merge now...

PR #2486, opened 2018-08-06
Fix GUIMod crash when module doesn't have a compatible game version

### LINKED ISSUE #2481: NullReferenceException at launch in GUIMod constructor
**NOTE:** CKAN 1.25.1 runs just fine.  This is only an issue on 1.25.2.

Background
----------

**CKAN Version:** 1.25.2

**KSP Version:** 1.4.5.2243

**Operating System:** macOS X 10.13.6 with Mono: Mono JIT compiler version 5.12.0.226 (2018-02/9824e260f56 Mon Apr 30 18:29:27 EDT 2018)

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** NO

Problem
-------

**What steps did you take in CKAN?** Started it from command-line as I always do: mono ckan.exe

**What did you expect to happen?** CKAN to run and not crash.

**What happened instead?** CKAN Crashed with a stack trace (previous version, 1.25.1 runs just fine)

**Screenshots:**
screen shot 2018-08-05 at 11 44 38 pm

HebaruSan:
## Problems

As of #2437, if you have an installed module that has since been completely de-indexed (e.g., MiniAVC: KSP-CKAN/CKAN-meta#561, https://github.com/KSP-CKAN/CKAN-meta/commit/6cd629492ab5c35e02137b93ecb20b9d8f3dcb1a), this exception happens at GUI startup:

```
  at CKAN.GUIMod..ctor (CKAN.CkanModule mod, CKAN.IRegistryQuerier registry, CKAN.Versioning.KspVersionCriteria current_ksp_version, System.Boolean incompatible) [0x003d6] in <60e037f139274e47b1d6c0206eefe140>:0 
  at CKAN.Main+<>c__DisplayClass229_0.<_UpdateModsList>b__2 (CKAN.InstalledModule m) [0x00006] in <60e037f139274e47b1d6c0206eefe140>:0 
  at System.Linq.Enumerable+SelectEnumerableIterator`2[TSource,TResult].MoveNext () [0x00048] in <cc3b329d40bd4675ae0e985e302972af>:0 
  at System.Linq.Enumerable+WhereEnumerableIterator`1[TSource].MoveNext () [0x0004e] in <cc3b329d40bd4675ae0e985e302972af>:0 
  at CKAN.Main._UpdateModsList (System.Boolean repo_updated, System.Collections.Generic.List`1[T] mc) [0x000fe] in <60e037f139274e47b1d6c0206eefe140>:0 
  at CKAN.Main+<>c__DisplayClass228_0.<UpdateModsList>b__0 () [0x0001b] in <60e037f139274e47b1d6c0206eefe140>:0 
  at CKAN.Util.Invoke[T] (T obj, System.Action action) [0x0002d] in <60e037f139274e47b1d6c0206eefe140>:0 
  at CKAN.Main.UpdateModsList (System.Boolean repo_updated, System.Collections.Generic.List`1[T] mc) [0x0001c] in <60e037f139274e47b1d6c0206eefe140>:0 
  at CKAN.Main.CurrentInstanceUpdated () [0x00062] in <60e037f139274e47b1d6c0206eefe140>:0 
  at CKAN.Main.OnLoad (System.EventArgs e) [0x002c2] in <60e037f139274e47b1d6c0206eefe140>:0 
  at System.Windows.Forms.Form.OnLoadInternal (System.EventArgs e) [0x00023] in <beba0b11547e4333a33d4d8cdf7b7c51>:0 
Waiting on threads to park on joinable thread list timed out.
```

Unrelatedly, if you had the old standalone MiniAVC module installed, it was not shown in the mod list.

## Causes

This happens when one of your installed modules is totally de-indexed. I only understood this after @cculianu shared his `registry.json`, so this PR's description previously said the cause was unknown:

> <details>
> <summary>Old obsolete investigation notes</summary>
> The exact cause is a bit mysterious. Only @cculianu has reported seeing this so far. A testing component revealed that the exception happens while the GUI is attempting to generate a mod list row for MiniAVC 1.0.3.0:
> 
> ```
> System.Exception: Failed to construct a GUIMod for MiniAVC 1.0.3.0 (Object reference not set to an instance of an object)
> ```
> 
> The same testing component established that the exception is thrown inside this block from #2437:
> 
> https://github.com/KSP-CKAN/CKAN/blob/ab714c224b9b446ac4a6d9c9b209b73f14bfa752/GUI/GUIMod.cs#L135-L173
> 
> However...
> 
> - MiniAVC isn't indexed in CKAN! It's only distributed as a bundled DLL inside other mods.
> - If CKAN installed MiniAVC.dll as part of another mod, then it would not be added to the mods list but rather counted as an installed file and ignored, so it must have gotten there some other way; yet the issue report says no manual changes were made to GameData.
> - 1.0.3.0 is a valid MiniAVC release version (from 2014), but there is no known pathway by which CKAN could have gotten that version string without metadata for it in the registry!
> 
> Hitting a dead end in trying to make this crash happen on my PC, I turned to auditing the code.
> </details>
> 

The crash happens here; since MiniAVC isn't in the registry, the `LatestCompatibleKSP` call returns null, and shortly after that the code attempts to access properties of that null reference:

https://github.com/KSP-CKAN/CKAN/blob/ab714c224b9b446ac4a6d9c9b209b73f14bfa752/GUI/GUIMod.cs#L152-L154

## Changes

- #2437's version formatting logic is moved from being embedded in `GUIMod`'s constructor to `KspVersion.ToYalovString`. This eliminates the possibility of the version object being null inside the function. Instead, null is now checked before the function is called, and we fall back to a version string of "Unknown" in that case. It also allows us to use this logic in other places as well as test it (not included in this PR).
- The `VersionsMax` dictionary is now static and only generated once rather than being redundantly re-generated for every module.
- A bunch of commented code is removed.
- `VersionMaxWasGenerated` is removed and the generation code it governed is moved to a static constructor which accomplishes the same thing more reliably with less overhead.
- `GUIMod`'s comparison operations now check the `Identifier` rather than the `Name`. This allows the old MiniAVC module to appear in the mod list, where it previously was hidden because its `Name` was the same as KSP-AVC's.

Fixes #2481.

### DIFF
--- a/Core/Versioning/KspVersion.cs
+++ b/Core/Versioning/KspVersion.cs
@@ -2,7 +2,9 @@
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Collections.Generic;
 using Newtonsoft.Json;
+using CKAN.GameVersionProviders;
 
 namespace CKAN.Versioning
 {
@@ -228,6 +230,65 @@ namespace CKAN.Versioning
             return _string;
         }
 
+        private static Dictionary<string, KspVersion> VersionsMax = new Dictionary<string, KspVersion>();
+
+        /// <summary>
+        /// Generate version mapping table once for all instances to share
+        /// </summary>
+        static KspVersion()
+        {
+            // Should be sorted
+            List<KspVersion> versions = new KspBuildMap(new Win32Registry()).KnownVersions;
+            VersionsMax[""] = versions.Last();
+            foreach (var v in versions)
+            {
+                // Add or replace
+                VersionsMax[$"{v.Major}"          ] = v;
+                VersionsMax[$"{v.Major}.{v.Minor}"] = v;
+            }
+        }
+
+        /// <summary>
+        /// Get a string to represent a game version.
+        /// </summary>
+        /// <returns>
+        /// String representing max game version.
+        /// Partly clamped to real versions, partly rounded up to imaginary versions.
+        /// </returns>
+        public string ToYalovString()
+        {
+            KspVersion value;
+
+            if (!IsMajorDefined
+                // 2.0.0
+                || _major > VersionsMax[""].Major
+                // 1.99.99
+                || (_major == VersionsMax[""].Major && VersionsMax.TryGetValue($"{_major}", out value) && _minor >= UptoNines(value.Minor)))
+            {
+                return "any";
+            }
+            else if (IsMinorDefined
+                && VersionsMax.TryGetValue($"{_major}.{_minor}", out value)
+                && (!IsPatchDefined || _patch >= UptoNines(value.Patch)))
+            {
+                return $"{_major}.{_minor}.{UptoNines(value.Patch)}";
+            }
+            else
+            {
+                return ToString();
+            }
+        }
+
+        /// <returns>
+        ///   0 - 9   //  9  - 99    //  99 - 999
+        ///   1 - 9   //  10 - 99    // 100 - 999
+        ///   8 - 9   //  98 - 99
+        /// </returns>
+        private static int UptoNines(int num)
+        {
+            return (int)Math.Pow(10, Math.Floor(Math.Log10(num + 1)) + 1) - 1;
+        }
+
         /// <summary>
         /// Converts the value of the current <see cref="KspVersion"/> to its equivalent
         /// <see cref="KspVersionRange"/>.
@@ -239,7 +300,7 @@ namespace CKAN.Versioning
         /// </para>
         /// <para>
         /// For example, the version "1.0.0.0" would be equivalent to the range ["1.0.0.0", "1.0.0.0"], while the
-        /// version "1.0" would be equivalent to the range ["1.0.0.0", "1.1.0.0"). Where '[' and ']' represent 
+        /// version "1.0" would be equivalent to the range ["1.0.0.0", "1.1.0.0"). Where '[' and ']' represent
         /// inclusive bounds and '(' and ')' represent exclusive bounds.
         /// </para>
         /// </returns>
@@ -247,7 +308,7 @@ namespace CKAN.Versioning
         {
             KspVersionBound lower;
             KspVersionBound upper;
-        
+
             if (IsBuildDefined)
             {
                 lower = new KspVersionBound(this, inclusive: true);
--- a/GUI/GUIMod.cs
+++ b/GUI/GUIMod.cs
@@ -268,7 +268,7 @@ namespace CKAN
 
         private bool Equals(GUIMod other)
         {
-            return Equals(Name, other.Name);
+            return Equals(Identifier, other.Identifier);
         }
 
         public override bool Equals(object obj)
@@ -281,7 +281,7 @@ namespace CKAN
 
         public override int GetHashCode()
         {
-            return (Name != null ? Name.GetHashCode() : 0);
+            return Identifier?.GetHashCode() ?? 0;
         }
 
     }
--- a/GUI/GUIMod.cs
+++ b/GUI/GUIMod.cs
@@ -42,9 +42,6 @@ namespace CKAN
         public bool IsCKAN { get; private set; }
         public string Abbrevation { get; private set; }
 
-        private bool VersionMaxWasGenerated = false;
-        private Dictionary<string, KspVersion> VersionsMax;
-
         /// <summary>
         /// Return whether this mod is installable.
         /// Used for determining whether to show a checkbox in the leftmost column.
@@ -66,19 +63,6 @@ namespace CKAN
         public string Version
         {
             get { return IsInstalled ? InstalledVersion : LatestVersion; }
-        }
-
-        /// <returns>
-        ///   0 - 9   //  9  - 99    //  99 - 999    
-        ///   1 - 9   //  10 - 99    // 100 - 999
-        ///   8 - 9   //  98 - 99    // 
-        /// </returns>
-        private int UptoNines(int num)
-        {
-            //if (num == 0)
-            //    return 0;
-            //else
-                return (int)Math.Pow(10, Math.Floor(Math.Log10(num + 1)) + 1) - 1;
         }
 
         public GUIMod(CkanModule mod, IRegistryQuerier registry, KspVersionCriteria current_ksp_version, bool incompatible = false)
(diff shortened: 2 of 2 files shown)


cculianu:
@HebaruSan -- I tried the provided ckan from the latest .zip you posted and.. it WORKS!  Hurray!

I am not sure why MiniAVC is there.  I have a pretty ancient KSP install from a long time ago.  All my mods I installed with CKAN, though.  No manually installed mods.

Here is what my mod list screen looks like:

screen shot 2018-08-07 at 4 34 48 am

I do seem to have a MiniAVC folder with MiniAVC in there from September 2016.. Should I delete that?

HebaruSan:
@cculianu, that's great news! Thanks for trying that out.

What if you change the filter from "Installed" to "All" and type "MiniAVC" in "Filter by mod name"? Does MiniAVC show up in the list in CKAN? Maybe it was indexed a long time ago and you have metadata from back then...?

cculianu:
Nope.  Still not there.

screen shot 2018-08-07 at 12 21 39 pm


Here is my entire CKAN folder (sans downloads/ folder).  Maybe this will help you forensically? : 

[CKAN.zip](https://github.com/KSP-CKAN/CKAN/files/2265798/CKAN.zip)

HebaruSan:
Wow, it does think you have MiniAVC installed as a full module! And the `install_time` tells us that you installed it in September 2016, so it must have been indexed back then:

```json
	"installed_modules": {
		"MiniAVC": {
			"install_time": "2016-09-26T17:41:31.496489+03:00",
			"source_module": {
				"abstract": "This plugin will notify you when updates are available for certain mods",
				"author": [
					"cybutek"
				],
				"download": "http://ksp.cybutek.net/miniavc/MiniAVC-1.0.3.0.zip",
				"download_size": 34408,
				"download_hash": {
					"sha1": "8B956D21495ACD0911EE4C181E049A0C664D3BB4",
					"sha256": "092AAFEB0FBC8EB002366136FE3AED4F496BADD7940861A8C42028D4F2822DDA"
				},
				"identifier": "MiniAVC",
				"ksp_version_strict": false,
				"license": [
					"GPL-3.0"
				],
				"name": "KSP AVC",
				"release_status": "stable",
				"resources": {
					"repository": "https://github.com/CYBUTEK/KSPAddonVersionChecker",
					"homepage": "http://forum.kerbalspaceprogram.com/threads/79745"
				},
				"version": "1.0.3.0",
				"spec_version": "1"
			},
			"installed_files": {
				"GameData/MiniAVC": {},
				"GameData/MiniAVC/MiniAVC.dll": {
					"sha1_sum": "F6-5D-DB-C2-41-CF-AE-83-3D-CC-39-F2-EA-FC-E3-EE-51-B1-EE-2F"
				}
			}
		},
```

It should be possible to find this in the metadata repositories, I'll take another look...

HebaruSan:
OK, KSP-CKAN/NetKAN#1455 has the background info. There was an effort to support MiniAVC as a standalone mod so other mods could recommend it and users could choose whether to install it. But the architecture of MiniAVC doesn't work well for this: it checks the versions of mods in its DLL's folder and subfolders, so if we put the DLL in its own folder it won't check anything, but if we put it in GameData it will check too much. (RoverDude also thought that de-bundling MiniAVC was too much interference with mod authors' wishes.)

KSP-CKAN/CKAN-meta#561 added a .ckan file as a stopgap while they worked on how a .netkan would work.

Here's the bot updating it somehow: https://github.com/KSP-CKAN/CKAN-meta/commit/f26d34f655c74c88e7ef70257211209926fae1c8

A year later KSP-CKAN/NetKAN#3934 removed the filtering out of bundled copies of MiniAVC.

Then a year after that the .ckan file was deleted: https://github.com/KSP-CKAN/CKAN-meta/commit/6cd629492ab5c35e02137b93ecb20b9d8f3dcb1a

Long story short, the symptom we're looking for is: Install a mod, then it gets de-indexed.

cculianu:
I'm actually a computer programmer by trade.  I am not familiar enough with NetKAN or any of what even MiniAVC does (version control/checker?).

I understood about 95% of that.  But one question remains: Should I delete that MiniAVC?  What happens when a module is ancient?  Does it still get loaded?  Will it do Bad Things? :)

I'm really glad I was able to help you fix this and that you fixed it so fast!

It's possible other people in the wild that have been using KSP since the early days (and CKAN) somehow ended up with this bug and/or in this situation.

I did nothing outside the ordinary except use CKAN to install mods (apparently including MiniAVC)... so.. there ya go.

Now.. should I delete MiniAVC from GameData/?

HebaruSan:
Indeed, thanks for your help!

At this point I would not delete it manually. You would end up with an installed module entry in your registry pointing to files that no longer exist, which may cause errors. The ideal endgame is that you would uninstall it from within CKAN itself, which would then update all the tracking data structures properly. But you can't do that currently because it doesn't show up in the mod list, so that's the next thing I plan to investigate.

cculianu:
Ok, well.. until then. MiniAVC that hopefully does no evil will continue to live, like a virus, in my GameData folder.

Do let me know if I can help any further, ha ha.

Pretty awesome how fast you fixed it. Great project.  I know C# well too.. maybe someday when my plate is freer I can also contribute to this project (although sometimes too many cooks spoil the pot).

Really good useful software, this CKAN.  Thumbs up on all counts.  👍

linuxgurugamer:
Note that CKAN is perfectly happy if you DELETE files.  When it updates or deletes, it is fine with files being missing.

HebaruSan:
Update, it's hiding MiniAVC because `GUIMod`'s comparison operations rely on `GUIMod.Name`:

https://github.com/KSP-CKAN/CKAN/blob/37c8477acccd8cebe204da4905cd67760e6bdb31/GUI/GUIMod.cs#L324-L340

... and the old MiniAVC module's `name` property is "KSP AVC", which is shared with the actual KSP-AVC:

https://github.com/KSP-CKAN/NetKAN/blob/56fc0358d7ba4ad514c748299e51aafaee1a49b0/NetKAN/KSP-AVC.netkan#L6

So GUI thinks MiniAVC is already in the list when it's not. This should be identifier-based instead. This could be causing assorted other problems if there are other cases of mods inappropriately sharing the same name.

HebaruSan:
I'm guessing LGG requested your review to confirm that the test build was working (good idea). Or I somehow misclicked it and don't remember. Either way, thanks!

cculianu:
Hmm. Should I try building it and seeing if old defunct mini avc shows up now? Or did your other change make it not appear despite this change?

(Actually if you can post a .zip again that’s easier as I’m not set up to build on this Mac and booting up windows for this is slightly painful as my windows system needs updates and is crufty)

HebaruSan:
Sure, here's a new test build:

- [ckan.zip](https://github.com/KSP-CKAN/CKAN/files/2267918/ckan.zip)

You should see a new row labeled "KSP AVC" (due to the shared name problem mentioned above), and unchecking its checkbox should allow you to uninstall. After that, the standard release of CKAN v1.25.2 should be able to start without the error.

cculianu:
I love you! It's there!  I am deleting it now!  YOU ARE THE BEST! 💃 

screen shot 2018-08-08 at 12 52 43 am

cculianu:
Indeed! I've been running the demo build you sent me for weeks now without a single issue.  I'm super happy!!

PR #2494, opened 2018-08-08
Replace New in repository filter with Newly compatible filter

### LINKED ISSUE #1923: A filter should show mods that are new, and those that are compatible with my version for the first time.
I want to look at all of the mods available for my version of KSP that were not available last time I refreshed the package list. I think this should replace the new in repository filter. When I use the new in repository filter, it's because I've already looked through all the mods in the compatible list, and after refreshing, I see that there are more mods in the compatible list.

Knowing whether a mod has ever been in the repo or not isn't very useful, since what really matters is if it's updated for my version of KSP.

HebaruSan:
## Motivation

GUI's "New in repository" filter shows mods that had no versions at all in the index before the last refresh. It's pretty rare for a mod to be newly indexed, so the usefulness of this is limited. Users would prefer a filter that shows newly indexed mods plus previously-incompatible mods that just had a compatible version indexed.

## Changes

"New in repository" is replaced with "Newly compatible". It now shows mods that were incompatible before the last refresh and compatible after. This is accomplished by limiting the old and new module lists that we compare to only those that are compatible.

| Compatible before? | Compatible after? | New? |
| --- | --- | --- |
| No | No | No |
| Yes | No | No |
| Yes | Yes | No |
| No | Yes | Yes |

Newly-indexed *incompatible* mods will also appear in this filter, in case anyone is interested in them.

Fixes #1923.

### DIFF
--- a/GUI/Main.Designer.cs
+++ b/GUI/Main.Designer.cs
@@ -397,7 +397,7 @@
             // 
             this.FilterNewButton.Name = "FilterNewButton";
             this.FilterNewButton.Size = new System.Drawing.Size(307, 30);
-            this.FilterNewButton.Text = "New in repository";
+            this.FilterNewButton.Text = "Newly compatible";
             this.FilterNewButton.Click += new System.EventHandler(this.FilterNewButton_Click);
             // 
             // FilterNotInstalledButton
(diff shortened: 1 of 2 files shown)


politas:
> Newly-indexed incompatible mods will also appear in this filter, in case anyone is interested in them.

Is it just me, or is that going to end up with someone complaining that it makes no sense?

HebaruSan:
That's possible. It'll be easy enough to change it if there are convincing complaints. My preference is to keep functionality where possible, and the existing filter shows them.

PR #2518, opened 2018-08-29
Add download count column to GUI

### LINKED ISSUE #2415: Sort by "popularity" column
While #481 was about making mods author see their stats, this issue is about 

> Being able to sort by popularity and type would be a nice median way for new users to locate mods they aren't familiar with.

HebaruSan:
## Motivation

KSP-CKAN/NetKAN-bot#67 created a script that updates a [download_counts.json file in CKAN-meta](https://github.com/KSP-CKAN/CKAN-meta/blob/master/download_counts.json) containing the download counts of mods. This data is available now (and already being downloaded whenever anyone clicks Refresh!), but it's not being used yet.

## Changes

Now when you refresh the registry, we extract the `download_counts.json` file and parse it, then store the values it contains in a dictionary at `Registry.download_counts`.

A new Downloads column is added to GUI displaying this value for each mod. The user can sort this column to identify major mods that they haven't tried before.

~~Congrats to KIS for winning the popularity contest:~~

Correction: Congrats to EVE for winning the popularity contest (now that KSP-CKAN/NetKAN-bot#73 is counting older branches):

image

Fixes #2415.

### DIFF
--- a/Core/Net/Repo.cs
+++ b/Core/Net/Repo.cs
@@ -12,6 +12,7 @@ using ICSharpCode.SharpZipLib.GZip;
 using ICSharpCode.SharpZipLib.Tar;
 using ICSharpCode.SharpZipLib.Zip;
 using log4net;
+using Newtonsoft.Json;
 
 namespace CKAN
 {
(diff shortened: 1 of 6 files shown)


yalov:
EVE, KIS, etc would be "popular" for every counting system.

But when Universal Storage II overcome Universal Storage, or SSPEredux overcome SSPE, in few years? I think they already popular.

For a new mod to be qualified as popular (with all_downloads) is not about quality, but about how long a mod is bouncing around.

if mod1 and mod2 is used by equal amount of people, but mod1 has 20 updates since 1.2, and mod2 is new (2 update since 1.4), then I don't see them as "mod1 10 times more popular than mod2"

politas:
With all these extra columns being added, I'm starting to feel a need for the ability to hide columns.

PR #2536, opened 2018-10-10
Configurable cache size limit

### LINKED ISSUE #2438: [Enhancement suggestion] Purge old downloads from cache
I think it would be great when a mod updates to a newer version of ksp(a version the player has currently installed) ckan should delete the older version from ..\Kerbal Space Program\CKAN\downloads folder, replacing it with the version that is installed alongside with the currently installed version of ksp. 

Issue: I see at least two different versions of many mods in my ckan downloads folder, and all ive been doing is updating mods to new versions when I see them available. it looks like ckan doesn't touch the older versions of mods once a newer version is installed alongside an updated version of ksp. 

Since Ckan can point to multiple ksp installs, a decision should be made if: Can I use this with multiple KSP installs at the same time(lets say I have 1.4.3 and 1.3.2 installed in different directories)? Can the settings be set based off what ksp version im currently using(we obviously don't want 1.4+ mods installing to a 1.3.0-2 just like that, or mods simply not appearing at all for a 1.3 install when you also have 1.4+ available )

Thank you!! I'm glad I could help!

Background
----------

**CKAN Version: 1.25.1** 

**KSP Version: 1.4.3** 

**Operating System: Windows 10 x64**

HebaruSan:
## Motivation

CKAN's download cache currently grows without limit. We write to it, but we don't remove files unless the user runs `ckan clean` or clicks the Clear button in the cache section of the GUI settings. This means that in order to avoid filling up a disk, the user might have to manually monitor the cache size and clear it periodically, or delete individual files. It would be nice to have something more automated.

## Changes

This pull request builds upon #2535 to create a configurable cache size limit. The default is to allow unlimited cache size, as it is today. (This pull request has been rebased to show only its own changes, where formerly #2535's commit was included.)

### Core

`Win32Registry.CacheSizeLimit` now stores the maximum number of bytes allowed in the cache, or `null` if unlimited.

`ModuleInstaller` enforces this limit after it finishes installing mods. Note that this means that we may temporarily exceed the limit during installation if required to fit the mods being installed. This also means that a user can set a limit of 0 bytes if they wish all downloads to be downloaded every time and never cached, and that will work.

The algorithm to purge cache files will delete, in order:

1. Files that do not correspond to any download in the registry (mods that you downloaded before their URLs changed, such as old KerbalStuff files)
2. Files that are in the registry but are not compatible with any of your game instances (mods that you downloaded when you were using old game versions)
3. Files that are compatible with at least one of your game instances

Within each subgroup, files are deleted in age order, oldest first. Deletion stops once the total cache size falls under the limit. Both the main configurable cache folder and the legacy per-instance caches are checked and purged together according to the above rules, and we compare the sum of their sizes to the limit to determine compliance.

### CmdLine

The `ckan cache` command has two new subcommands:

- `ckan cache showlimit` - Print the cache size limit in MB, or "Unlimited" if unlimited
- `ckan cache setlimit <new limit>` - Set the cache size limit to the argument in MB, or clear it to allow unlimited caching if no argument is given

These commands do not enforce the limit. That will happen after mods are installed.

### GUI

The cache section in the settings is expanded, the buttons are pushed down, and a limit field is added in the middle:

image

Changing this value does **not** immediately enforce the new limit (hence the size being over the limit in the above screenshot). That only happens after you install some mods. I don't want to auto-delete files while the user is still editing the value in the field, which may be smaller than their intended value.

The Clear button is replaced with a Purge dropdown button. When clicked, two options appear, one of which enforces the configured limit, the other of which deletes all files in the cache. If the cache is empty, the dropdown is disabled. If the cache is under the current limit, then the dropdown option to enforce the limit is disabled.

Fixes #2438.

### DIFF
--- a/Cmdline/Action/Cache.cs
+++ b/Cmdline/Action/Cache.cs
@@ -25,6 +25,12 @@ namespace CKAN.CmdLine
             [VerbOption("reset", HelpText = "Set the download cache path to the default")]
             public CommonOptions ResetOptions { get; set; }
 
+            [VerbOption("showlimit", HelpText = "Show the cache size limit")]
+            public CommonOptions ShowLimitOptions { get; set; }
+
+            [VerbOption("setlimit", HelpText = "Set the cache size limit")]
+            public SetLimitOptions SetLimitOptions { get; set; }
+
             [HelpVerbOption]
             public string GetUsage(string verb)
             {
@@ -45,11 +51,15 @@ namespace CKAN.CmdLine
                         case "set":
                             ht.AddPreOptionsLine($"Usage: ckan cache {verb} [options] path");
                             break;
+                        case "setlimit":
+                            ht.AddPreOptionsLine($"Usage: ckan cache {verb} [options] megabytes");
+                            break;
 
                         // Now the commands with only --flag type options
                         case "list":
                         case "clear":
                         case "reset":
+                        case "showlimit":
                         default:
                             ht.AddPreOptionsLine($"Usage: ckan cache {verb} [options]");
                             break;
@@ -63,6 +73,12 @@ namespace CKAN.CmdLine
         {
             [ValueOption(0)]
             public string Path { get; set; }
+        }
+
+        private class SetLimitOptions : CommonOptions
+        {
+            [ValueOption(0)]
+            public long Megabytes { get; set; } = -1;
         }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


politas:
This is great, but can we add a manual "purge" button to bring the cache down to the limit set? If the UI is getting too cluttered, we could just replace the "Clear" button, and users could set the cache max size to 0 if they want to completely empty it.

HebaruSan:
Replaced the Clear button with a Purge dropdown menu button with both of those choices.

Issue #2525, opened 2018-09-23
CKAN theme usage


7ranceaddic7:
Background
----------
**CKAN Version:** 
1.25.3

**KSP Version:** 
n/a

**Operating System:** 
Ubuntu 18.04, Linux 4.15

**Gnome Theme**
Arc-Dark

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** 
n/a

Problem
-------
Main Window, Menu Bar, Action Bar: White background, pale font.

Any ideas how to get the background color to theme standard, #383c4a>
![screenshot from 2018-09-23 13-56-52](https://user-images.githubusercontent.com/4062616/45932931-17509c00-bf39-11e8-91f9-294c5fe51c66.png)

StripedMonkey:
This is also an issue with Adwita-dark. Certain UI Elements just don't seem to follow any sort of theming. In older versions of Ckan I believe I just used `GTK_THEME=` and set Ckan to a light theme to let it even be usable however it doesn't appear that Ckan follows this anymore Using the GTK2 method doesn't work either.

Orochimarufan:
The GTK2 way worked for me:
$ env GTK2_RC_FILES=/usr/share/themes/Breeze/gtk-2.0/gtkrc ckan
(substitute Breeze for your light theme of choice)

Still would be nice if it properly worked w/ dark themes, but most mono apps seem to be broken in this regard

LastStarDust:
Same problem here with Adapta-Nokto theme. The Orochimarufan workaround has no effect in my case (tested with Adwaita light theme)

HebaruSan:
Can someone please post installation/setup instructions for Arc-Dark, Adwita-dark, or Adapta-Nokto? The dark theme I use is available via `apt install numix-gtk-theme`, then using the old (and [now removed!?](https://forum.antergos.com/topic/9515/gnome-tweaks-no-more-global-dark-theme/3)) dark theme toggle. I don't see APT packages by these other names.

HebaruSan:
- The menu bar has no color properties set
- The toolbar has `BackColor` set to `SystemColors.Control`
- The mod list has `BackgroundColor` set to `SystemColors.Window`

It looks like Mono's ToolStrip (and MenuStrip which inherits from it) may ignore its BackColor property; it always passes SystemColors.Control to ToolStripRenderEventArgs.BackColor instead of its own BackColor:

[...]

PR #2540, opened 2018-10-15
Fix text colors for dark themes

HebaruSan:
## Problems

In a dark theme, the status log and parts of the mod info pane feature dark text on a dark background:

image

image

Looks fine in a normal theme:

image

## Causes

1. A text box in .NET defaults to `WindowText` (black) on `Window` (white), but if you set its `ReadOnly=true`, as we do in this case, the background is changed to `Control` (dark gray) without changing the foreground, to give it a disabled appearance
2. We set the `ForeColor` of the mod info pane labels to `ControlDarkDark`; the meaning of this value is:
   [Gets a Color structure that is the dark shadow color of a 3-D element.](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.systemcolors)
   This color is not meant to be used for text foreground colors, and using it assumes a light background color.

## Changes

Now these places look better.

1. The log text box's `ForeColor` is now `ControlText`, to match its background.
   image
2. The labels in the mod info pane now have a `ForeColor` of `GrayText`, which the spec defines as "Gets a Color structure that is the color of dimmed text." This is meant to be used as a text color, and it will provide visual distinction from the normal text color:
image
   Standard themes still look OK:
   image

Somewhat related to #2525 but not a fix for the specific things in that issue.

### DIFF
--- a/GUI/Main.Designer.cs
+++ b/GUI/Main.Designer.cs
@@ -865,6 +865,7 @@
             this.LogTextBox.Multiline = true;
             this.LogTextBox.Name = "LogTextBox";
             this.LogTextBox.ReadOnly = true;
+            this.LogTextBox.ForeColor = System.Drawing.SystemColors.ControlText;
             this.LogTextBox.ScrollBars = System.Windows.Forms.ScrollBars.Vertical;
             this.LogTextBox.Size = new System.Drawing.Size(1505, 851);
             this.LogTextBox.TabIndex = 8;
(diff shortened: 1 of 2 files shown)


Olympic1:
Looks good. I personally have trouble reading the links on the dark theme, could it be possibe to give it a lighter color using `LinkColor`?

I find `System.Drawing.SystemColors.Highlight` a good one.

HebaruSan:
I'm not 100% sure `Highlight` is an improvement; at least in my theme, it's still pretty hard to read:

image

That's intended to be used as a background color for a selected element, and I'd rather not risk making things worse for some users by using it in a "clever" way. (The fact that `SystemColors` has no explicit `LinkColor` member says a lot about how long ago some of this stuff must have been designed!)

PR #2541, opened 2018-10-19
Fix toolbar background colors for dark themes

### LINKED ISSUE #2525: CKAN theme usage

HebaruSan:
## Problem

CKAN's toolbars are hideous with a dark theme:

screenshot

## Cause

See mono/mono#11287; Mono is trying to use `SystemColors.ButtonFace` for the left edge of a toolstrip, fading to a hard coded light color with a linear gradient. But Mono doesn't load `ButtonFace` from the system settings, instead leaving it with a compile time default.

## Upstream fix

mono/mono#11287 is trying to fix this in Mono by switching from `ButtonFace` to `SystemColors.Control`, which is set by Mono's theme loader. However, there are some downsides:

- Even with the fix, the default toolstrip renderer still uses an ugly gradient to a hard coded light color at the right edge
- That fix won't reach users until Mono puts out a new release and users upgrade to it (assuming it gets merged)

upstream

## Changes

Luckily this part of Mono is 90% customization / overrides / defaults / etc., so there are plenty of hooks here for us to make it work however we want. When Mono wants to draw a control, it first checks whether that control has a `.Renderer` property, and if it does, it delegates the drawing to that object. Now we populate that property with our own renderer that just draws the background color normally. This looks better than the renderer built into Mono, even after the other fix, and will work for all users regardless of their version of Mono.

image

Fixes #2525.

### DIFF
--- a/GUI/CKAN-GUI.csproj
+++ b/GUI/CKAN-GUI.csproj
@@ -114,6 +114,7 @@
     <Compile Include="FormCompatibility.cs">
       <SubType>Form</SubType>
     </Compile>
+    <Compile Include="FlatToolStripRenderer.cs" />
     <Compile Include="GUIMod.cs" />
     <Compile Include="GUIUser.cs" />
     <Compile Include="HintTextBox.cs">
--- a/GUI/FlatToolStripRenderer.cs
+++ b/GUI/FlatToolStripRenderer.cs
@@ -1,0 +1,25 @@
+using System;
+using System.Drawing;
+using System.Windows.Forms;
+
+namespace CKAN
+{
+    /// <summary>
+    /// Custom toolstrip renderer that just fills the background with the BackColor.
+    ///
+    /// The default Mono one uses a gradient between a hardcoded light color and a system color that Mono doesn't load correctly.
+    /// </summary>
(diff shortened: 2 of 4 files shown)


Olympic1:
Looks good. Could it be possible to also fix the checboxes?

HebaruSan:
The Mono code for this is extremely obfuscated, but as far as I can tell so far, no, the visual style of those checkboxes is inaccessible to modification by applications. The core drawing code gets the backcolor from a reference to a full checkbox control, which doesn't exist for the grid, so a wrapper layer around that creates a temporary checkbox control from scratch using defaults that we can't change.

(It's kind of impressive how little is accomplished here with so much code. At least 8 Mono source files are involved in providing this degree of non-flexibility.)

PR #2543, opened 2018-10-21
Clean up mod list spacebar handling in GUI

### LINKED ISSUE #2534: Inconsistent checkbox behavior when pressing spacebar
Background
----------

**CKAN Version:** v1.25.3

**Operating System:** Windows (8.1, but looks like others as well)

Problem
-------
When selecting a mod from list, 'Installed' checkbox can be toggled by pressing spacebar. But if 'Installed' checkbox was clicked by mouse, then pressing spacebar toggles it 2 times (first when pressing down, second when releasing).
If there is also 'Update' checkbox available, then:
-If mod was selected by clicking outside 'Update' checkbox, pressing spacebar works as described above
-If mod was selected by clicking 'Update' checkbox, pressing spacebar toggles 'Installed' checkbox and releasing it toggles 'Update' checkbox

clicks
(on the gif I held spacebar for a fraction of second so there is a pause between pressing and releasing)

Probably caused by conflict between whatever toggles 'Installed' checkbox of selected mod when pressing spacebar and OS's toggling of selected checkbox using tab controls (note the dotted border).

HebaruSan:
## Problem

If you focus the mod list and press space in GUI, some surprising behaviors can occur depending on exactly where you click:

animation

- If you click the Install checkbox, then the checkbox will toggle when you press space and then again when you let go
- If you click the Upgrade checkbox, then pressing space can toggle both checkboxes

## Causes

CKAN has some special code to toggle checkboxes when you press space.

- The grid itself handles checkbox toggling in `KeyDown`, but we were doing it in `KeyPress`, which happens later
- Our code fired regardless of which cell was focused, and if that cell had a checkbox, then multiple toggles would happen at different times
- Our code only ever toggled the Install checkbox, so if you click an upgradable row and press space, it would try to remove the mod rather than upgrade it

## Changes

Now the behavior is more consistent and less surprising.

- Now we handle space in `KeyDown` for consistency with the built in behavior
- If one of the checkbox cells has the focus, we let the grid handle it
- If there's an Update checkbox, we toggle it preferentially, otherwise we toggle the Install checkbox

Fixes #2534.

### DIFF
--- a/GUI/MainModList.cs
+++ b/GUI/MainModList.cs
@@ -289,26 +289,39 @@ namespace CKAN
         /// </summary>
         private void ModList_KeyDown(object sender, KeyEventArgs e)
         {
-            DataGridViewCell cell = null;
             switch (e.KeyCode)
             {
                 case Keys.Home:
                     // First row.
-                    cell = ModList.Rows[0].Cells[2];
+                    ModList.CurrentCell = ModList.Rows[0].Cells[2];
+                    e.Handled = true;
                     break;
 
                 case Keys.End:
                     // Last row.
-                    cell = ModList.Rows[ModList.Rows.Count - 1].Cells[2];
+                    ModList.CurrentCell = ModList.Rows[ModList.Rows.Count - 1].Cells[2];
+                    e.Handled = true;
                     break;
-            }
 
-            if (cell != null)
-            {
-                e.Handled = true;
-
-                // Selects the top/bottom row and scrolls the list to it.
-                ModList.CurrentCell = cell;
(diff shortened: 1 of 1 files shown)


7ranceaddic7:
Speaking of surprising behaviour, I stumbled an edge case where spacebar crashes CKAN.

Steps to Replicate:
Launch CKAN.
Select desires instance.
DO NOTHING - Don't click anything.  Don't Arrow or Tab anywhere.  NOTHING.
Hit Spacebar.
CTD

Does this fix handle that behavior?  If not, I can open new issue.

PR #2555, opened 2018-10-25
Bump spec version

ChucklesTheBeard:
This value hasn't been updated, though the following updates are shown elsewhere in the file:
24 - include_only, include_only_regexp
25 - install_to / missions


HebaruSan:
Thanks for catching that!

PR #2556, opened 2018-10-27
Show mod info for change set, recs/sugs, and providers

### LINKED ISSUE #1868: Show the sidebar with Metadata (at least) for changeset stages
I usually don't check the recommended and suggested mods that comes with what I want to install. Often don't check the dependencies either.

It would be good if I could click on what comes up and get the same Metadata sidebar with info and link to homepage and such so I can see what they are before deciding if I want to skip them or not.

Currently I have to actually cancel the installation, go back to the mod list, remember what they were, search them up, and so on. Not very user friendly :disappointed: 

**CKAN Version:**  1.18.1

**Operating System:** Windows 10

HebaruSan:
## Motivation

GUI has several places where mods are listed but all the user can access is their name and abstract.

- Change set
- Recommendations / suggestions
- Providing modules

If you want to see the long description, license, home page, relationships, etc., you have to cancel whatever you're doing and then look up the mods separately on the main mod list.

## Changes

Now the mod info pane is available in these lists if you click a row. This allows the user to get complete info about any mod about which they're being asked to make decisions. The mod info pane will disappear if you clear the list's selection.

Main mod list:

image

Provides:

image

Change set:

image

Recommendations:

image

To accomplish this, the structure of the main form is rearranged so the vertical splitter now contains the main tab control and the mod info:

Pre:

- MainTabControl
  - ManageModsTabPage
    - splitContainer1
      - ModList
      - MainModInfo

Post:

- splitContainer1
  - MainTabControl
    - ManageModsTabPage
      - Modlist
  - MainModInfo

Then when the selection changes in any of those lists, we get the `CkanModule` for it, create a `GUIMod`, and pass it to the mod info pane.

Fixes #1868.

### DIFF
--- a/GUI/Main.Designer.cs
+++ b/GUI/Main.Designer.cs
@@ -446,26 +446,29 @@
             | System.Windows.Forms.AnchorStyles.Left) 
             | System.Windows.Forms.AnchorStyles.Right)));
             this.splitContainer1.FixedPanel = System.Windows.Forms.FixedPanel.Panel2;
-            this.splitContainer1.Location = new System.Drawing.Point(0, 111);
+            this.splitContainer1.Location = new System.Drawing.Point(0, 35);
             this.splitContainer1.Margin = new System.Windows.Forms.Padding(4, 5, 4, 5);
             this.splitContainer1.Name = "splitContainer1";
-            // 
+            this.splitContainer1.Dock = System.Windows.Forms.DockStyle.Fill;
+            //
             // splitContainer1.Panel1
-            // 
-            this.splitContainer1.Panel1.Controls.Add(this.ModList);
-            // 
+            //
+            this.splitContainer1.Panel1.Controls.Add(this.MainTabControl);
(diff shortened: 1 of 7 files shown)


politas:
This is a great idea, but I find that the mod info pane is oddly higher than the mod list. Could you add a spacer to move the mod tabs down a bit. Perhaps throw a CKAN logo in it?

HebaruSan:
Uh oh, UI design? I was afraid that might happen... :stuck_out_tongue_winking_eye: 
I'll see what I can do about that.

HebaruSan:
What if we put the name/description pane above the tabstrip? Then it can serve as a sort of label for all the tabs, as well as displacing them downwards to deal with the visual offset situation:

image

HebaruSan:
Pushed a new commit with the above change.

politas:
Hmm, that's better, but still looks a little weird when you stretch out the window to see more of the modlist on a 4K monitor. I'll see if it grows on me.

Issue #2562, opened 2018-11-01
Spacebar Crashes CKAN


7ranceaddic7:
Background
----------

**CKAN Version:** 
1.25.3

**KSP Version:** 
1.4.5

**Operating System:** 
Linux Jupiter 4.18.0-11-generic # 12-Ubuntu SMP Tue Oct 23 19:22:37 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

Problem
-------
CKAN crashes after hitting spacebar.

**What steps did you take in CKAN?**
Steps to Replicate:
1. Launch CKAN.
2. Highlight desired instance.
3. Click 'Select' and put your hands on your head.
    1. (Seriously, put'em on your head.)

![screenshot from 2018-11-01 15-37-38](https://user-images.githubusercontent.com/4062616/47884642-c90fa200-ddee-11e8-9ba8-dfece4052736.png  "CKAN - KSP Instance Selection Window")

4. CKAN updates repositories and then displays MainWindow, as shown here:
![screenshot from 2018-11-01 15-29-20](https://user-images.githubusercontent.com/4062616/47883654-138f1f80-ddeb-11e8-89a8-ad252bcbaa87.png "CKAN Mod Selection Window")

5. At this point, your hands should still be on your head.  Don't touch your mouse. Don't click anything. Don't Arrow or Tab anywhere. ABSOLUTELY NOTHING.

6. Now, you have permission to hit Spacebar.

7. CTD.  Don't profit.

**What did you expect to happen?**
tick state change on/off of highlighted (this will alway be first) mod listed

**What happened instead?**
CTD

```
System.InvalidOperationException: No editing control defined
  at System.Windows.Forms.DataGridViewCell.InitializeEditingControl (System.Int32 rowIndex, System.Object initialFormattedValue, System.Windows.Forms.DataGridViewCellStyle dataGridViewCellStyle) [0x00015] in <50f22134ee3a4d3ab382d1d406dc2ad0>:0 
  at System.Windows.Forms.DataGridView.RefreshEdit () [0x0002f] in <50f22134ee3a4d3ab382d1d406dc2ad0>:0 
  at (wrapper remoting-invoke-with-check) System.Windows.Forms.DataGridView.RefreshEdit()
[...]

PR #2563, opened 2018-11-02
Cache listings of legacy cache dirs

HebaruSan:
## Problem

In a comment in #2562, @7ranceaddic7 noted a significant performance degradation with the current master HEAD version compared to 1.25.3. I audited the changes since then and identified 3 possible causes. This is the first.

## Cause

`NetFileCache` maintains a `cachedFiles` string array of files currently in the cache. This list is searched to find files matching a given module's download property.

When we consolidated the instance-specific caches in #2535, existing instance-specific cache folders were treated as a fallback if a file was not found in the main cache dir. However, the files in those fallback folders were not added to `cachedFiles` because I wasn't sure how best to do that or if it was necessary. Instead, those files were enumerated from the directories on the fly every time.

On closer inspection, this might cause a big slowdown because GUI checks `IsMaybeCachedZip` for every single mod on load, which implies O(n * m) disk operations, where n is the number of mods in the registry and m is the number of files in the legacy caches. If they were added to `cachedFiles` instead, then those disk accesses could be avoided and replaced with faster memory accesses.

## Changes

Now we consolidate all of the files in all of the caches into the single `cachedFiles` array. It turns out this is much easier and simpler than what I did previously. The cache will work the same as before, except faster because it will no longer have to hit the disk for legacy cache dirs.

And as long as we're touching this code, `cachedFiles` is now a dictionary mapping from URL hashes to cache entries. This eliminates the need to loop through every entry and check whether it starts with our target hash, which should result in a general performance improvement for cache-related tasks.

### DIFF
--- a/Core/Net/NetFileCache.cs
+++ b/Core/Net/NetFileCache.cs
@@ -26,7 +26,8 @@ namespace CKAN
     public class NetFileCache : IDisposable
     {
         private FileSystemWatcher watcher;
-        private string[] cachedFiles;
+        // hash => full file path
+        private Dictionary<string, string> cachedFiles;
         private string cachePath;
         private KSPManager manager;
         private static readonly ILog log = LogManager.GetLogger(typeof (NetFileCache));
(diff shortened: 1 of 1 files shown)


7ranceaddic7:
Additional information:

I just performed an update on a mod (BobsPanicBox).  I have two 1.4.5 installs - one with one without MHE.  While performing the update with CKANv1.25.4 and found the update process performance degraded as well. The CKANv1.25.3 performed the same update quickly.

I can test for actual time, but the performance differential was markedly noticeable.

Olympic1:
@7ranceaddic7 Did you got around to test this fix out?

politas:
Oooh, sounds good. I have noticed a significant performance hit since the One True Cache system was added. After every function that downloads files, it takes ages for it to get back to the modlist. And ok, maybe I don't need 1,270 items in my cache taking up a half a Terabyte...

politas:
Wow, that really sped up the cache delay! Now hardly noticeable!

PR #2571, opened 2018-11-07
Treat AVC min without max as open range

HebaruSan:
## Problem

A .version file containing this:

```json
    "KSP_VERSION":
    {
        "MAJOR":1,
        "MINOR":5,
        "PATCH":0
    },
     "KSP_VERSION_MIN":{
         "MAJOR":1,
         "MINOR":5,
         "PATCH":0
     },
```

Is resulting in this metadata:

```json
    "ksp_version": "1.5.0",
```

The minimum value is essentially ignored.

## Cause

The [KSP-AVC spec](http://ksp.cybutek.net/kspavc/Documents/README.htm) says:

- **KSP_VERSION** - Optional, Required for MIN/MAX
  Version of KSP that the add-on was made to support.
- **KSP_VERSION_MIN** - Optional
  Minimum version of KSP that the add-on supports.
  _Requires KSP_VERSION field to work._
- **KSP_VERSION_MAX** - Optional
  Maximum version of KSP that the add-on supports.
  _Requires KSP_VERSION field to work._

This implies that if you want to specify `KSP_VERSION_MIN`, you *must* specify `KSP_VERSION`. This does not make sense, but it's in the spec, and some modders may feel compelled to follow it strictly as written.

Under that interpretation, if you wanted to specify "all versions after 1.5.0", you would create a file as shown above, but it would not translate into equivalent CKAN metadata.

## Changes

Now if a .version file has `KSP_VERSION_MIN` and `KSP_VERSION` but not `KSP_VERSION_MAX`, we don't set `ksp_version_max`. (And vice versa for `MAX` and `MIN`.)

### DIFF
--- a/Netkan/Transformers/AvcTransformer.cs
+++ b/Netkan/Transformers/AvcTransformer.cs
@@ -110,9 +110,23 @@ namespace CKAN.NetKAN.Transformers
             var existingKspMax = existingKspMaxStr == null ? null : KspVersion.Parse(existingKspMaxStr);
 
             // Get the minimum and maximum KSP versions that are in the AVC file.
-            // Use specific KSP version if min/max don't exist.
-            var avcKspMin = avc.ksp_version_min ?? avc.ksp_version;
-            var avcKspMax = avc.ksp_version_max ?? avc.ksp_version;
+            // http://ksp.cybutek.net/kspavc/Documents/README.htm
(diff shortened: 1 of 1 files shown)


politas:
And because we set the other min/max field to a null value, it doesn't populate that field in the .ckan, is that right?

HebaruSan:
Yup, null will mean that the property isn't set.

PR #2577, opened 2018-11-16
Show recommendations of installed modules in GUI

### LINKED ISSUE #1764: Search/filter for mods recommended/suggested by installed mods
Hi!  Happy CKAN user here.  Just thought I'd offer a suggested feature that could help me out:

So, sometimes I install mods but elect not to install some of their "Recommended" or "Suggested" mods, because I'm not sure, but I'd like to find those mods later and inspect them.  Or, sometimes mod A recommends mod B, and that doesn't seem to trigger a page re: the mods that mod B recommends (unless I wasn't paying attention, which is possible).

I was thinking there are a few things that could be added to the filter menu:
- Mods recommended by installed mods
- Mods suggested by installed mods

Some other lower-priority ideas (maybe in a sub-menu, because it's a lot of options):
- Mods that depend on installed mods
- Mods that recommend installed mods
- Mods that suggest installed mods
  - These three could be useful to e.g. find a new mod B that extends or just plays nicely with an installed mod A, even if installed mod A isn't aware of new mod B.
- Mods that conflict with installed mods
  - As the inverse of the above, this could be useful to find mods that provide (incompatible) alternatives to mods you already have installed.
  - I'm imagining this would be bidirectional, i.e. the union of "uninstalled mods that our installed mods say they conflict with" and "uninstalled mods that say they conflict with our installed mods".

(Not that I'm trying to create a bunch of work here, I just figured that if someone's already mucking about in the filters, it shouldn't be too tricky to add several / all of the above. 😉 )

Thanks for this awesome tool!  133 mods installed at once and rising!  😄

### LINKED ISSUE #2268: Recommendations with virtual dependencies don't appear in GUI
Background
----------

**CKAN Version:** 
1.22.6 and current HEAD

**KSP Version:** 
1.3.1

**Operating System:** 
Windows 7, Windows 10

No

Problem
-------

1. Create a .netkan file with Scatterer as a recommendation
2. Run it through netkan.exe
3. In GUI, File &rarr; Install from .ckan...
4. Select the .ckan file generated in step 2

Scatterer should appear in the list of recommended mods to install

**What happened instead?**
Other recommendations appear, but Scatterer doesn't.

**Screenshots:**
screenshot 2018-01-29 20 58 10

I instrumented the code with various exception handlers in `Main.GetShowableMods`:

```
Exception thrown looking up Scatterer: Exception of type 'CKAN.InconsistentKraken' was thrown.
   at CKAN.SanityChecker.EnforceConsistency(IEnumerable`1 modules, IEnumerable`1 dlls)
   at CKAN.RelationshipResolver..ctor(IEnumerable`1 modules, RelationshipResolverOptions options, IRegistryQuerier registry, KspVersionCriteria kspversion)
   at CKAN.RelationshipResolver..ctor(IEnumerable`1 module_names, RelationshipResolverOptions options, IRegistryQuerier registry, KspVersionCriteria kspversion)
   at CKAN.Main.GetShowableMods(Dictionary`2 mods)
```

```
Kraken for Scatterer: The following inconsistencies were found:
Scatterer has an unmet dependency: Scatterer-sunflare is not installed
 * Scatterer has an unmet dependency: Scatterer-config is not installed
```

That function turns off handling of virtual mods and turns _on_ consistency checks:

https://github.com/KSP-CKAN/CKAN/blob/dfd03d7051e56dede16d4212eaad38d6f652cc51/GUI/MainInstall.cs#L410-L417

Then when it does a consistency check, it discovers that Scatterer has unsatisfied dependencies, since they're virtual, and the `InconsistentKraken` makes it skip to the next recommendation:

https://github.com/KSP-CKAN/CKAN/blob/dfd03d7051e56dede16d4212eaad38d6f652cc51/GUI/MainInstall.cs#L423-L438

Scatterer's metadata isn't actually invalid, so it should be possible for another mod to recommend it.

Thanks to @sigma88 for reporting this via forum PM.

HebaruSan:
## Motivation

When installing a mod, CKAN shows the user other mods that are recommended or suggested by that mod, and the user can choose whether to install them as well.

If the user chooses not to install these mods, but then later has second thoughts, there's no easy way to try again. They would have to remember what the mods were and choose to install them, check the Relationships tab in mod info, or uninstall and re-install the dependent mods.

## Changes

Now there's a new File &rArr; Audit recommendations menu option:

image

If you click this and none of your installed mods have recommendations or suggestions, it'll tell you that:

image

Otherwise, it takes you to the Choose recommended mods tab, listing the mods that are recommended by currently installed mods.

image

Clicking Continue jumps to Choose suggested mods:

image

If you click Cancel at any point, or if you don't select anything to install, then it just stops. Otherwise it takes you to the installer flow. (One possibly confusing point here is that the recommendations / suggestion lists might appear again if any of the mods you chose to install have recommendations or suggestions.) The chosen mods are downloaded and installed:

image

### Code details

The private member variable `Main.toInstall` is removed and replaced with a local variable, to improve encapsulation and because that's how it's already used. To share this object effectively, it's now a parameter to the other functions that use it. Where `RecommendedModsContinueButton_Click` previously manipulated `toInstall` directly, now `ShowSelection` does that work instead.

A lot of the code in `MainInstall.cs` was specifically for handling recommendations and suggestions. This is now split out into `MainRecommendations.cs`, along with a function or two from another file, to make it easier to find and reduce the clutter of `MainInstall.cs`.

`MainRecommendations.cs` then has a bit of new code at the end to gather installed mods' recommendations and suggestions and feed them into the selection prompts, then launch the installer.

`ShowSelection` uses `Monitor.Wait` to wait for the cancel or continue buttons to be clicked and call `Monitor.Pulse`, which requires the foreground GUI thread to be unoccupied, so the menu item click handler has to start a new thread. I used `Task.Factory.StartNew` for this, but I'm not sure that's the best way, please advise.

Fixes #1764.

### Recommendation conflicts with mods we're installing

If you select several mods to install and some have recommendations or suggestions, recommendations and suggestions that conflict with installed mods will be hidden, but ones that conflict with mods that are *planned to install in this change set* will be shown.

Now recommendations and suggestions will only be shown if they don't conflict with installed mods **or** mods planned to be installed in the current change set.

Fixes KSP-CKAN/NetKAN#6709.

### Recommendations with virtual dependencies

If you select a mod to install that has a recommendation with a virtual dependency, that recommendation won't appear in the list, even if it can be satisfied by installing some combination of allowable mods. For example, `ExtrasolarPlanetsBeyondKerbol-stock` recommends `Scatterer`, which depends on `Scatterer-sunflare` and `Scatterer-config`, each of which are provided by several possible mods (assuming you have KSP 1.3; some haven't been updated past that point yet). Installing `ExtrasolarPlanetsBeyondKerbol-stock` won't ask if you want to install `Scatterer`.

Now such recommendations show up as available in GUI.

This requires two main pieces:

1. Allowing the recommendation to appear. This requires us to handle `TooManyModsProvideKraken` in the recommendation logic, which we now do in a new `CanInstall` function.
2. Prompting the user to resolve the virtual dependency when installing. This requires handling `TooManyModsProvideKraken` in `Main.InstallMods`. This may have existed at some point, but if it did it went missing at some point. Now it's definitely present, so you can be asked to choose a mod to satisfy a virtual dependency after you choose a recommendation.

Fixes #2268.

### DIFF
--- a/GUI/CKAN-GUI.csproj
+++ b/GUI/CKAN-GUI.csproj
@@ -168,6 +168,8 @@
     <Compile Include="MainModList.cs">
       <SubType>Form</SubType>
     </Compile>
+    <Compile Include="MainRecommendations.cs">
+    </Compile>
     <Compile Include="MainRepo.cs">
       <SubType>Form</SubType>
     </Compile>
--- a/GUI/Main.Designer.cs
+++ b/GUI/Main.Designer.cs
@@ -40,6 +40,7 @@
             this.toolStripSeparator2 = new System.Windows.Forms.ToolStripSeparator();
             this.importDownloadsToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.toolStripSeparator3 = new System.Windows.Forms.ToolStripSeparator();
+            this.auditRecommendationsMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.ExitToolButton = new System.Windows.Forms.ToolStripMenuItem();
             this.settingsToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.cKANSettingsToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
@@ -177,6 +178,8 @@
             this.toolStripSeparator2,
             this.exportModListToolStripMenuItem,
             this.toolStripSeparator3,
+            this.auditRecommendationsMenuItem,
+            this.toolStripSeparator3,
             this.ExitToolButton});
             this.fileToolStripMenuItem.Name = "fileToolStripMenuItem";
             this.fileToolStripMenuItem.Size = new System.Drawing.Size(50, 29);
@@ -233,6 +236,14 @@
             this.toolStripSeparator3.Name = "toolStripSeparator3";
             this.toolStripSeparator3.Size = new System.Drawing.Size(278, 6);
             // 
+            //
+            // importDownloadsToolStripMenuItem
+            //
+            this.auditRecommendationsMenuItem.Name = "auditRecommendationsMenuItem";
+            this.auditRecommendationsMenuItem.Size = new System.Drawing.Size(281, 30);
+            this.auditRecommendationsMenuItem.Text = "Audit recommendations";
+            this.auditRecommendationsMenuItem.Click += new System.EventHandler(this.auditRecommendationsMenuItem_Click);
+            //
             // ExitToolButton
             // 
             this.ExitToolButton.Name = "ExitToolButton";
@@ -1258,6 +1269,7 @@
         private System.Windows.Forms.ToolStripSeparator toolStripSeparator2;
         private System.Windows.Forms.ToolStripMenuItem importDownloadsToolStripMenuItem;
         private System.Windows.Forms.ToolStripSeparator toolStripSeparator3;
+        private System.Windows.Forms.ToolStripMenuItem auditRecommendationsMenuItem;
         private System.Windows.Forms.ToolStripMenuItem ExitToolButton;
         public System.Windows.Forms.ToolStripMenuItem settingsToolStripMenuItem;
         private System.Windows.Forms.ToolStripMenuItem cKANSettingsToolStripMenuItem;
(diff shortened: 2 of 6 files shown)


politas:
I like this. I always like smaller chunks of code per file where it makes sense.

politas:
We've had a bunch of new features added so far for this version. Any chance we could get some tests addressing these features, to make sure nothing gets broken in the future?

HebaruSan:
I will put together a PR with a bunch of tests in the next couple of days.

PR #2565, opened 2018-11-03
Minimize CKAN to tray and automatic refreshing

Olympic1:
### Core
Added an automatic refreshing of CKAN that happens after N minutes. This option can be set via the settings menu and is stored in the windows registry. This behaviour starts when the GUI is loaded and can be paused via the respective options, or by setting the rate to `0`.

### GUI
The GUI settings window's updates section is halved and a new behaviour section is added next to it to save some space:

settingswindow

* The `Enable tray icon` shows a CKAN icon in the tray. The window is still shown on the taskbar.
* The `Minimize to tray` allows minimizing to the tray. The window is hidden from the taskbar.
* The refresh rate to set an interval in minutes to refresh the modlist automatically.
* The `pause refreshing` pauses the automatic refreshing. This is disabled when `RefreshRate` is set to `0`.

When the GUI is minimized, you can double click the icon to bring back the GUI.  
Also right clicking the icon shows a menu:

traymenu

* The updates option shows the amount of available updates. This is disabled when none are available and is clickable when there are. Clicking the option selects all of the available updates.
* `Refresh` refreshes the modlist
* `Pause` pauses the automatic refreshing. This is disabled when `RefreshRate` is set to `0`.
* `Open CKAN` opens the GUI
* `Open KSP Directory` opens the current KSP directory
* `CKAN Settings` opens the settings menu
* `Quit` closes CKAN

When an update becomes available, it pops up a balloon notifying you that there is an update. Clicking it installs it. (Thanks @HebaruSan)

### Other
* Removed a duplicate `StartPosition` in the settings window
* Redone the tab order in the settings window
* Moved `Refresh on startup` from updates to more settings

Closes #2377.

PR #2587, opened 2018-11-24
Fix tray icon menu position and color on Linux (Gnome)

HebaruSan:
## Problem

#2565 added a tray icon with a context menu. On Linux, it has light text on a white background and it can be displayed mostly off-screen:

screenshot

(The icon is invisible as well. I don't know why that is, so this pull request doesn't attempt to fix it.)

## Cause

Mono?

## Changes

Now the background matches other menus, and the position is always on screen:

image

In addition, the tray icon related code is moved from `Main.cs` to `MainTrayIcon.cs`.

### DIFF
--- a/GUI/CKAN-GUI.csproj
+++ b/GUI/CKAN-GUI.csproj
@@ -176,6 +176,9 @@
     <Compile Include="MainTabControl.cs">
       <SubType>Component</SubType>
     </Compile>
+    <Compile Include="MainTrayIcon.cs">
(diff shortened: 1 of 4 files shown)


politas:
Nope, doesn't show up anywhere, and then refuses to unminimise

politas:
I may well be the only Mate user of CKAN, and I don't care about this feature, as it doesn't have the main element I wanted for a tray applet (handling multiple KSP installs).

PR #2592, opened 2018-12-03
Add legend for relationships tab

HebaruSan:
## Motivation

Someone mentioned recently that it can be inconvenient to read the relationships tab because the meaning of the icons aren't apparent, you have to hover them to see what the relationship is. It's a fair point.

## Changes

Now a legend is added above the relationships tree, similar to the one above the list of versions:

image


yalov:
Maybe instead of the legend, to make them "folders":
```
[-] The mod:
    [-] Depends:
        Mod1
        Mod2
        Mod3
    [-] Supports:
        Mod4
        Mod5
```

PR #2602, opened 2018-12-05
Show correct error messages after canceling (un)installations/upgrades

DasSkelett:
## Problem
If you cancel the installation/uninstall/update process after it's already complete, but before the Cancel button is greyed out you still get the "Error!" message and the Apply changes button is still active.
grafik
(Ignore the location of the Retry button atm, this seems to be another bug)

This is due to the entire GUI PostInstallMods() function only checking for e.Result.Key wich is set to false if the installation is canceled, despite the real result possibly being successful. This leads to those false error messages and  GUI behavior.

## Solution
e.Result.Key is now set to the value of processSuccesful (if you know a better name, let me know!). PostInstallMods now also distinguishes the "installCanceled but positive processSuccesful case.
The corresponding messages and GUI actions are now handled by a new FailWaitDialog() in MainWait.cs.

Furthermore the GUI ModList gets updated and UpdateChangesDialog is called. This solves the Apply changes button falsely being active as well as "installed" checkboxes having the false state.

Also the now empty changes tab is hidden (as after normal, not canceled installation).
ckan_3
grafik

Redo of #2418

### DIFF
--- a/GUI/MainInstall.cs
+++ b/GUI/MainInstall.cs
@@ -147,8 +147,9 @@ namespace CKAN
                 installCanceled = true;
             };
 
-             // checks if all actions were successfull
-             bool resolvedAllProvidedMods = false;
+            // checks if all actions were successfull
+            bool processSuccessful = false;
+            bool resolvedAllProvidedMods = false;
             // uninstall/installs/upgrades until every list is empty
             // if the queue is NOT empty, resolvedAllProvidedMods is set to false until the action is done
             while (!resolvedAllProvidedMods)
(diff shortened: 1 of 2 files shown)


GUI/MainInstall.cs
```diff
@@ -147,30 +147,43 @@ public async void InstallModuleDriver(IRegistryQuerier registry, CkanModule modu
...
                 {
                     e.Result = new KeyValuePair<bool, ModChanges>(false, opts.Key);
-                    if (!installCanceled && toUninstall.Count > 0)
+                    if (toUninstall.Count > 0)
                     {
-                        installer.UninstallList(toUninstall);
+                        resolvedAllProvidedMods = false;
+                        if (!installCanceled) {
+                            installer.UninstallList(toUninstall);
+                            resolvedAllProvidedMods = true;
```
HebaruSan: This variable is specifically to track handling of TooManyModsProvideKraken. It should not be getting set after uninstalls like this.

DasSkelett: Ohhh, and I thought that this was a strange variable name for a simple "Just keep going until everything is done" while loop... 
Yeah so basically I create a new bool for my "everything done?" checks.

politas:
@DasSkelett , can you fix the conflict on this with #2599?

PR #2608, opened 2018-12-08
Remove unnecessary parameter in Configuration methods

DasSkelett:
## defaultRepo
`Configuration.LoadOrCreateConfiguration()` gets passed a `defaultRepo` parameter.
This is a relic of the early days of CKAN it seems, when the repositories were saved in the GUI configuration, but this got moved to the registry in 536637f7fe2fded1878bf66fb9b5aabee012770f.
Since then the parameter sits there unused.

## path
Also `SaveConfiguration()` gets passed a `path` parameter, which is redundant since there is a specific path variable in the `Configuration` class. It's a bit dangerous too, because it could lead to a difference between the real save path and the one set in the configuration.
Now SaveConfiguration() takes the value out of the according object of Configuration, which happened in `Save()` before.

## Add `SpaceAfterMethodCallName` to code formatting guideline
I added `$2.SpaceAfterMethodCallName = False` to the solution file, because MonoDevelop always inserted a space between method calls an round brackets.
Don't know if Visual Studio is okay with this, but I think in the worst case VS auto-removes the line again.
I encountered a strange Github display bug with the indentation in CKAN.sln, hence the 2 additional commits.

### DIFF
--- a/CKAN.sln
+++ b/CKAN.sln
@@ -89,5 +89,6 @@ Global
 		$2.inheritsSet = Mono
 		$2.inheritsScope = text/x-csharp
 		$2.scope = text/x-csharp
+        $2.SpaceAfterMethodCallName = False
 	EndGlobalSection
 EndGlobal
--- a/GUI/Configuration.cs
+++ b/GUI/Configuration.cs
@@ -81,7 +81,7 @@ namespace CKAN
             SaveConfiguration(this, path);
         }
 
-        public static Configuration LoadOrCreateConfiguration(string path, string defaultRepo)
+        public static Configuration LoadOrCreateConfiguration(string path)
         {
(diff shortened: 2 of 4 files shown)


DasSkelett:
As I wrote this PR I was wondering if SaveConfiguration() shouldn't be private instead of public, this would force Save() to be called from outside, which I understood is the goal of having this separate function.

DasSkelett:
That's a great rule of thumb :P

DasSkelett:
Okay, made it private, also LoadConfiguration(). This needed changing two tests to calling LoadOrCreateConfiguration() instead of LoadConfiguration(), but that doesn't affect their purpose.

HebaruSan:
Build fail? What build fail? :wink:

PR #2610, opened 2018-12-09
Allow selection of text in mod info panel

### LINKED ISSUE #2591: Selectable information on the metadata tab
Sometimes I need to copy author or identifier fields values.
Adding copyable mod-name to the metadata tab would also be cool.

DasSkelett:
## Feature request
#2591 asked for the ability to select and copy the text out of some labels in the mod info panel.
This was not possible because you cant copy text out of labels.

## Changes
I changed the **mod name (at the top), the mod author and the mod identifier label** to not-editable textboxes with the same background and text color as the labels had before. Also I removed the per standard in textboxes activated border, so they are seamless.
They look like the labels before (in dark and in white themes, at least with mono on Kubuntu), but now you can copy text.
Some pics:
bright dark

## Important
**I directly manipulated the `MainModInfo.Designer.cs` in MonoDevelop, and as far as I know, Visual Studio doesn't really like this, so someone using VS should verify whether VS complains/changes something back by itself or not.**
I also renamed the comments inside the Designer.cs for this, but you never know.

Closes #2591

### DIFF
--- a/GUI/MainModInfo.Designer.cs
+++ b/GUI/MainModInfo.Designer.cs
@@ -20,17 +20,6 @@
             base.Dispose(disposing);
         }
 
-        /// <summary>
-        /// Create a new TransparentTextBox control that allows the backcolor to be transparent
-        /// </summary>
-        private partial class TransparentTextBox : System.Windows.Forms.TextBox
-        {
-            public TransparentTextBox ()
-            {
-                SetStyle (System.Windows.Forms.ControlStyles.SupportsTransparentBackColor, true);
-            }
-        }
-
         #region Component Designer generated code
 
         /// <summary>
@@ -44,8 +33,8 @@
             this.splitContainer2 = new System.Windows.Forms.SplitContainer();
             this.MetaDataUpperLayoutPanel = new System.Windows.Forms.TableLayoutPanel();
             this.MetadataModuleNameTextBox = new TransparentTextBox();
-            this.MetadataModuleAbstractLabel = new System.Windows.Forms.Label ();
-            this.MetadataModuleDescriptionLabel = new System.Windows.Forms.Label ();
+            this.MetadataModuleAbstractLabel = new System.Windows.Forms.Label();
+            this.MetadataModuleDescriptionTextBox = new TransparentTextBox();
             this.MetaDataLowerLayoutPanel = new System.Windows.Forms.TableLayoutPanel();
             this.IdentifierLabel = new System.Windows.Forms.Label();
             this.MetadataIdentifierTextBox = new TransparentTextBox();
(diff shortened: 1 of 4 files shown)


GUI/MainModInfo.Designer.cs
```diff
@@ -139,17 +139,21 @@ private void InitializeComponent()
...
+            this.MetadataModuleNameTextBox.Font = new System.Drawing.Font("Microsoft Sans Serif", 9.75F, System.Drawing.FontStyle.Bold, System.Drawing.GraphicsUnit.Point, ((byte)(0)));
+            this.MetadataModuleNameTextBox.ForeColor = System.Drawing.SystemColors.GrayText;
+            this.MetadataModuleNameTextBox.Location = new System.Drawing.Point(3, 0);
+            this.MetadataModuleNameTextBox.Name = "MetadataModuleNameTextBox";
+            this.MetadataModuleNameTextBox.Size = new System.Drawing.Size(340, 46);
+            this.MetadataModuleNameTextBox.TabIndex = 0;
+            this.MetadataModuleNameTextBox.Text = "Mod Name";
+            this.MetadataModuleNameTextBox.ReadOnly = true;
+            this.MetadataModuleNameTextBox.TextAlign = System.Windows.Forms.HorizontalAlignment.Center;
+            this.MetadataModuleNameTextBox.BackColor = System.Drawing.SystemColors.Menu;
```
HebaruSan: I think this should be `Control`. This isn't on a menu.

DasSkelett: Okay, this is weird. I first took Control, but that gave me a pure white background, on dark and bright themes.
Now I tested again, and it was the right color ¯\_(ツ)_/¯
Will change it.

DasSkelett:
Oh, on Windows there are two different colors for the normal background and the tab pages...
Gosh, why are there so many?

I'm not sure if it is needed to have the version number / the license or the release status as a textbox. I can't image that someone wants to copy that text. But for the sake of everything being the same, it would make sense.

DasSkelett:
Ah I think I got it. For the ModuleName textbox I should use `SystemColors.Control` and for the lower ones `SystemColors.Menu`.
I'll change it and boot up Windows real quick to verify.

DasSkelett:
Ooooookay, I got it now, really:
Upper panel: Windows+Linux `SystemColors.Control`
Lower panel: Windows `SystemColors.Window` | Linux `SystemColors.Control`

See:
e_linux-bright e_linux-dark e_windows (The mod name textbox always has `.Control`)

I begin to assume this might be a mono implementation bug?

Nevertheless, either we go the complex way and set the background color of the labels during runtime with OS checks, or we just change all labels in the right column to textboxes, so at least they have the same colors. That would solve @yalov's request to.
The third way I see is to change the background color of the tab pages to the same as the upper panel (like it seems to be on Linux already), if that's possible. Theoretically the easiest way.

HebaruSan:
It's not a bug, just different styles of tabstrips. Mono's is how tabs looked in Windows 98.

Can you check what value `MetadataTabPage.BackColor` has on Windows? It seems to be `Control` on Linux in my testing. There's gotta be *some* value somewhere that already has the right color.

DasSkelett:
On Windows it throws the following error:
``` 
Unbehandelte Ausnahme: System.ArgumentException: Das Steuerelement unterstützt keine transparenten Hintergrundfarben.
   bei System.Windows.Forms.Control.set_BackColor(Color value)
   bei System.Windows.Forms.TextBoxBase.set_BackColor(Color value)
   bei CKAN.MainModInfo.<>c__DisplayClass2_0.<.ctor>b__0()
   bei CKAN.Util.Invoke[T](T obj, Action action)
   bei CKAN.MainModInfo..ctor()
   bei CKAN.Main.InitializeComponent()
   bei CKAN.Main..ctor(String[] cmdlineArgs, KSPManager mgr, GUIUser user, Boolean showConsole)
   bei CKAN.GUI.Main_(String[] args, KSPManager manager, Boolean showConsole)
   bei CKAN.CmdLine.MainClass.Gui(KSPManager manager, GuiOptions options, String[] args)
   bei CKAN.CmdLine.MainClass.RunSimpleAction(Options cmdline, CommonOptions options, String[] args, IUser user, KSPManager manager)
   bei CKAN.CmdLine.MainClass.Execute(KSPManager manager, CommonOptions opts, String[] args)
   bei CKAN.CmdLine.MainClass.Main(String[] args)
```
-> The control element does not support transparent backgrounds.
I tried to move it outside the Designer into the constructor of , but same error...
```
public MainModInfo()
{
    InitializeComponent()
    System.Drawing.Color newOne = MetadataTabPage.BackColor;
    Util.Invoke(MetadataModuleAuthorTextBox, () => MetadataModuleAuthorTextBox.BackColor = newOne);
    ...
}
```

DasSkelett:
Yep, that works. Nice!
ckan1 ckan2

GUI/MainModInfo.Designer.cs
```diff
@@ -20,6 +20,17 @@ protected override void Dispose(bool disposing)
             base.Dispose(disposing);
         }
 
+        /// <summary>
+        /// Create a new TransparentTextBox control that allows the backcolor to be transparent
+        /// </summary>
+        private partial class TransparentTextBox : System.Windows.Forms.TextBox
```
HebaruSan: Could you please put this class in a separate file? I think we should limit the .Designer.cs files to things that the VS editor code generation can handle.

DasSkelett: Ouups, forgot about bitchy VS :)

PR #2627, opened 2018-12-28
Add possibility to clone KSP installs and create dummy ones

DasSkelett:
~As I am a responsible developer and always test my changes on multiple systems before release...~
Because I am currently out of reach of my PC (christmas, family and useless stuff like this), I tried to work on CKAN a bit from a device with no KSP install.
But CKAN is rather bitchy about having a correct KSP install, and doesn't even let you in the GUI (which is the correct behavior in my opinion).

So to solve this issue, I added the `ckan ksp fake` command to create a dummy KSP directory, containing a GamaData folder, a `buildID.txt`, a `buildID64.txt` and a `readme.txt`.

This should also be useful for quick tests of mod related issues.

---
Following options are implemented:
- [x] choose the KSP version
- [x] get the build number for the version 
- [x] and the path (currently a folder `_fakeKSP` is created inside the folder from which the command is executed)
- [x] support DLC with different versions (no validation yet)
- [x] automatically add the ksp instance to the known instances? Would make sense, I think.
- [x] add validation
- [x] more logging
- [x] add a clone method to clone an existing KSP instance
- [x] implement the clone method in the cmdline
- [x] add a buildID64
- [x] add tests

Side effects:
* New `CopyDirectory()` function to properly copy a folder with all subfolders. Implemented as a transaction.
* New `FindKnownVersion()` function to find a build number for a ksp version
* New `InBuildMap()` bool to check if a version is in the build map.

Right now the command structure is as follows:
`ckan ksp fake [commonOptions] name path version dlcVersion`
* where name is the name with which the new instance is added to the known instances
* and     path can be absolute or relative
* and     version is in the form that it can be parsed by `KspVersion.Parse()` (i.e. e.g. `1.5.1`)
* and     dlcVersion is the version for the dlc. Not being evaluated right now. To exclude the DLC, type "none".

And:
`ckan ksp clone [commonOptions] instanceNameOrPath newname newpath`
*  where instanceNameOrPath is the name of an existing instance (-> see `ckan ksp list`) or the path to it
* and name is the name for the new instance
* and path is the path for the new instance (relative or absolute)
* This command copies the FULL KSP directory, including all exes, logs, game files and whatsoever.

### DIFF
--- a/Cmdline/Action/KSP.cs
+++ b/Cmdline/Action/KSP.cs
@@ -100,7 +100,7 @@ namespace CKAN.CmdLine
         {
             [ValueOption(0)] public string nameOrPath { get; set; }
             [ValueOption(1)] public string new_name { get; set; }
-            [ValueOption(1)] public string new_path { get; set; }
+            [ValueOption(2)] public string new_path { get; set; }
         }
 
         internal class RenameOptions : CommonOptions
@@ -301,83 +301,61 @@ namespace CKAN.CmdLine
             string new_name = options.new_name;
             string new_path = options.new_path;
 
-            // Try instanceNameOrPath as name and search the registry for it.
-            if (Manager.HasInstance(instanceNameOrPath))
+
+            try
             {
-                CKAN.KSP[] listOfInstances = Manager.Instances.Values.ToArray();
-                foreach (CKAN.KSP instance in listOfInstances)
+                // Try instanceNameOrPath as name and search the registry for it.
+                if (Manager.HasInstance(instanceNameOrPath))
                 {
-                    if (instance.Name == instanceNameOrPath)
+                    CKAN.KSP[] listOfInstances = Manager.Instances.Values.ToArray();
+                    foreach (CKAN.KSP instance in listOfInstances)
                     {
-                        // Found it, now clone it.
-                        try
+                        if (instance.Name == instanceNameOrPath)
                         {
+                            // Found it, now clone it.
                             Manager.CloneInstance(instance, new_name, new_path);
+                            break;
                         }
-                        catch (NotKSPDirKraken kraken)
-                        {
-                            // Two possible reasons:
-                            // First: The instance to clone is not a valid KSP instance.
-                            // Only occurs if user manipulated directory and deleted files/folders
-                            // which CKAN searches for in validity test.
-
-                            // Second: Something went wrong adding the new instance to the registry,
-                            // most likely because the newly created directory is not valid.
-
-                            User.RaiseError(kraken.ToString());
-                            log.Error(kraken);
-                            return Exit.ERROR;
-                        }
-                        catch (IOException e)
-                        {
-                            // The new path is not empty
-                            // The exception contains a message to inform the user.
-
-                            User.RaiseError(e.ToString());
-                            log.Error(e);
-                            return Exit.ERROR;
-                        }
-                        break;
                     }
                 }
-            }
-            // Try to use instanceNameOrPath as a path and create a new KSP object.
-            // If it's valid, go on.
-            else if (new CKAN.KSP(instanceNameOrPath, new_name, User) is CKAN.KSP instance && instance.Valid)
-            {
-                // Same as above.
-                try
+                // Try to use instanceNameOrPath as a path and create a new KSP object.
+                // If it's valid, go on.
+                else if (new CKAN.KSP(instanceNameOrPath, new_name, User) is CKAN.KSP instance && instance.Valid)
                 {
                     Manager.CloneInstance(instance, new_name, new_path);
+
                 }
-                catch (NotKSPDirKraken kraken)
+                // There is no instance with this name or at this path.
+                else
                 {
-                    // Two possible reasons:
-                    // First: The instance to clone is not a valid KSP instance.
-                    // Only occurs if user manipulated directory and deleted files/folders
-                    // which CKAN searches for in validity test.
-
-                    // Second: Something went wrong adding the new instance to the registry,
-                    // most likely because the newly created directory is not valid.
-
-                    User.RaiseError(kraken.ToString());
-                    log.Error(kraken);
-                    return Exit.ERROR;
-                }
-                catch (IOException e)
-                {
-                    // The new path is not empty
-                    // The exception contains a message to inform the user.
-
-                    User.RaiseError(e.ToString());
-                    log.Error(e);
-                    return Exit.ERROR;
+                    throw new NoGameInstanceKraken();
                 }
             }
-            // There is no instance with this name or at this path.
-            else
+            catch (NotKSPDirKraken kraken)
             {
-                NoGameInstanceKraken kraken = new NoGameInstanceKraken();
+                // Two possible reasons:
+                // First: The instance to clone is not a valid KSP instance.
+                // Only occurs if user manipulated directory and deleted files/folders
+                // which CKAN searches for in validity test.
+
+                // Second: Something went wrong adding the new instance to the registry,
+                // most likely because the newly created directory is not valid.
+
+                User.RaiseError(kraken.ToString());
+                log.Error(kraken);
+                return Exit.ERROR;
+            }
+            catch (IOException e)
+            {
+                // The new path is not empty
+                // The exception contains a message to inform the user.
+
+                User.RaiseError(e.ToString());
+                log.Error(e);
+                return Exit.ERROR;
+            }
+            catch (NoGameInstanceKraken kraken)
+            {
                 log.Error(kraken);
                 User.RaiseError(String.Concat(kraken.ToString(), "\n", "No instance with this name or at this path: ", instanceNameOrPath));
                 return Exit.ERROR;
@@ -521,18 +499,18 @@ namespace CKAN.CmdLine
             log.Info("Creating a new fake KSP install");
 
             // Parse all options
-            bool DLC = !(options.dlcVersion.ToLower() == "none");
             string installName = options.name;
             string path = options.path;
             KspVersion version;
-            string dlcVersion = options.dlcVersion;
+            // dlcVersion is null if no user wants no simulated DLC
+            string dlcVersion = options.dlcVersion.ToLower() == "none" ? null : options.dlcVersion;
             try
             {
                 log.Debug("Parsing KSP version");
                 version = KspVersion.Parse(options.version);
                 if (!version.IsBuildDefined)
                 {
-                    version = version.AddBuildToVersion();
+                    version = version.FindKnownVersion();
                 }
                 if (version == null)
                 {
@@ -550,7 +528,7 @@ namespace CKAN.CmdLine
             try
             {
                 // Pass all arguments to CKAN.KSPManager.FakeInstance() and create a new one.
-                Manager.FakeInstance(installName, path, version, DLC, dlcVersion);
+                Manager.FakeInstance(installName, path, version, dlcVersion);
             }
             catch (ArgumentOutOfRangeException e)
             {
--- a/Cmdline/Action/KSP.cs
+++ b/Cmdline/Action/KSP.cs
@@ -1,8 +1,9 @@
 using System;
+using System.IO;
 using System.Linq;
+using CKAN.Versioning;
 using CommandLine;
 using CommandLine.Text;
-using CKAN.Versioning;
 
 namespace CKAN.CmdLine
 {
@@ -13,22 +14,22 @@ namespace CKAN.CmdLine
         internal class KSPSubOptions : VerbCommandOptions
         {
             [VerbOption("list",    HelpText = "List KSP installs")]
-            public CommonOptions ListOptions     { get; set; }
+            public CommonOptions  ListOptions    { get; set; }
 
             [VerbOption("add",     HelpText = "Add a KSP install")]
-            public AddOptions    AddOptions      { get; set; }
+            public AddOptions     AddOptions     { get; set; }
 
             [VerbOption("rename",  HelpText = "Rename a KSP install")]
-            public RenameOptions RenameOptions   { get; set; }
+            public RenameOptions  RenameOptions  { get; set; }
 
             [VerbOption("forget",  HelpText = "Forget a KSP install")]
-            public ForgetOptions ForgetOptions   { get; set; }
+            public ForgetOptions  ForgetOptions  { get; set; }
 
             [VerbOption("default", HelpText = "Set the default KSP install")]
             public DefaultOptions DefaultOptions { get; set; }
 
-            [VerbOption("fake", HelpText = "Fake a KSP install")]
-            public CommonOptions FakeOptions { get; set; }
+            [VerbOption("fake",    HelpText = "Fake a KSP install")]
+            public CommonOptions  FakeOptions    { get; set; }
 
             [HelpVerbOption]
             public string GetUsage(string verb)
@@ -46,7 +47,12 @@ namespace CKAN.CmdLine
                     ht.AddPreOptionsLine("ksp " + verb + " - " + GetDescription(verb));
                     switch (verb)
                     {
-                        // First the commands with two string arguments
+                        // First the commands with three string arguments
+                        case "fake":
+                            ht.AddPreOptionsLine($"Usage: ckan ksp {verb} [options] version name path");
+                            break;
+
+                        // Second the commands with two string arguments
                         case "add":
                             ht.AddPreOptionsLine($"Usage: ckan ksp {verb} [options] name url");
                             break;
@@ -64,7 +70,6 @@ namespace CKAN.CmdLine
 
                         // Now the commands with only --flag type options
                         case "list":
-                        case "fake":
                         default:
                             ht.AddPreOptionsLine($"Usage: ckan ksp {verb} [options]");
                             break;
@@ -95,6 +100,15 @@ namespace CKAN.CmdLine
         internal class DefaultOptions : CommonOptions
         {
             [ValueOption(0)] public string name { get; set; }
+        }
+
+        internal class FakeOptions : CommonOptions
+        {
+            [ValueOption(0)] public string version { get; set; }
+            [ValueOption(1)] public string name { get; set; }
+            [ValueOption(2)] public string path { get; set; }
+            [Option("dlc", DefaultValue = true, HelpText = "Include faked DLC.")]
+            public bool dlc { get; set; }
         }
 
         // This is required by ISubCommand
@@ -158,7 +172,7 @@ namespace CKAN.CmdLine
                             break;
 
                         case "fake":
-                            exitCode = FakeNewKSPInstall();
+                            exitCode = FakeNewKSPInstall((FakeOptions)suboptions);
                             break;
 
                         default:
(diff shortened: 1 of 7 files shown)


HebaruSan:
That could certainly simplify the validation scripts:

https://github.com/KSP-CKAN/xKAN-meta_testing/blob/36ccd470cde1c6a02a8a3fc267e2eddad2b6640c/NetKAN/build.sh#L36-L99

DasSkelett:
Added some features, but can't test them because I get the cast error everytime I run `ksp fake 1.5.1 testName testPath` or similar.

Cmdline/Action/KSP.cs
```diff
@@ -373,21 +387,57 @@ private int SetDefaultInstall(DefaultOptions options)
...
+            // Try to create the folder structure and the txt files
             try
             {
-                string folder = "_fakeKSP/";
-                System.IO.Directory.CreateDirectory(folder);
-                System.IO.Directory.CreateDirectory(folder + "GameData");
-                System.IO.File.WriteAllText(folder + "buildID.txt", "build id = 023352018.10.17");
-                System.IO.File.WriteAllText(folder + "readme.txt", "Version 1.5.1");
+                // Create a KSP root directory, containing a GameData folder, a buildID.txt and a readme.txt
+                string directory = Path.Combine(path, "fakeKSP");
```
HebaruSan: Can we treat the path given by the user as the full path? It might not be suitable to have "fakeKSP" added in all circumstances.

Cmdline/Action/KSP.cs
```diff
@@ -373,21 +387,57 @@ private int SetDefaultInstall(DefaultOptions options)
...
+                Directory.CreateDirectory(directory);
+                Directory.CreateDirectory(Path.Combine(directory, "GameData"));
+                File.WriteAllText(Path.Combine(directory, "buildID.txt"), String.Format("build id = {0}", version.Build));
+                File.WriteAllText(Path.Combine(directory, "readme.txt"), String.Format("Version {0}", version.ToYalovString()));
+
+                // If a installed DLC should be simulated, we create the needed folder structure and the readme.txt
+                if (DLC && version.CompareTo(KspVersion.Parse("1.4.0")) >= 0)
+                {
+                    Directory.CreateDirectory(Path.Combine(directory, "GameData", "SquadExpansion", "MakingHistory"));
+                    File.WriteAllText(Path.Combine(directory, "GameData", "SquadExpansion", "MakingHistory", "readme.txt"), "Version 1.1.0");
```
HebaruSan: Let's make the DLC version a command line option. The validation script hard codes it because it has no way of knowing the right version to use, but your new command could be used in other contexts where a desired version is known.

Cmdline/Action/KSP.cs
```diff
@@ -373,21 +387,57 @@ private int SetDefaultInstall(DefaultOptions options)
...
-                System.IO.Directory.CreateDirectory(folder);
-                System.IO.Directory.CreateDirectory(folder + "GameData");
-                System.IO.File.WriteAllText(folder + "buildID.txt", "build id = 023352018.10.17");
-                System.IO.File.WriteAllText(folder + "readme.txt", "Version 1.5.1");
+                // Create a KSP root directory, containing a GameData folder, a buildID.txt and a readme.txt
+                string directory = Path.Combine(path, "fakeKSP");
+                Directory.CreateDirectory(directory);
+                Directory.CreateDirectory(Path.Combine(directory, "GameData"));
+                File.WriteAllText(Path.Combine(directory, "buildID.txt"), String.Format("build id = {0}", version.Build));
+                File.WriteAllText(Path.Combine(directory, "readme.txt"), String.Format("Version {0}", version.ToYalovString()));
```
HebaruSan: ToYalovString would convert things like 1.4.99 to 1.4.9, and the user might have a reason for wanting to use 1.4.99. Let's use the normal ToString instead.

HebaruSan:
General thought here, but would we ever want this capability in GUI or ConsoleUI? If so, it would make sense to migrate the business logic into Core and reduce the CmdLine code to just calling into it. Not sure about the answer to that myself, this might make sense as a text-only utility thing.

politas:
Ooh, don't suppose you could expand it into a more generic "clone install" command with a "--dummy" option to make the new install a blank one?

DasSkelett:
I can see the reason behind a "clone install", but I think it's a bit counter-intuitive to use a flag behind "clone" to create a new (dummy) instance. In "clone" you want to copy an existing one (which is kinda the definition of the word), in "--dummy"/"fake" you create an entirely new one.
This shouldn't be hidden in a flag, I think having those two as different commands is more intuitive.

DasSkelett:
Do we have a "mapVersionToBuild" function somewhere?
I thought `KspVersion.Parse()` does this, but it doesn't. If patch and/or build is undefined, it just stays undefined (-1). So atm, if a user enters `ckan ksp add smth smwh 1.4 none` he gets `Version 1.4` in the readme.txt and `-1` in the buildID.txt.
This leads to `AddInstall()` to throw an error, because it obviously doesn't know `Version 1.5`.

So is there a function that does this already implemented anywhere, or do I have to write a new one?

DasSkelett:
Thanks. I  thought about autocompletion with a 0 if the patch is missing, so "1.5" is parsed to "1.5.0", because 1.5 is a more natural way to say it.
But then, in the netkan files we interpret "1.5" as a range from "1.5.0" to "1.5.XXX", which means all versions starting with "1.5".
But we can't do that in this case, so an error might be the right answer.

DasSkelett:
@politas So you mean moving the new feature to "core" as @HebaruSan proposed, but with the "clone" function and the "fake" function all in one, and running `ckan KSP clone <instanceName>` or `ckan KSP fake <name> <path> <version> <dlcVersion>` both call the new core function.

Did I get this right?

HebaruSan:
Do we actually want "cloning" to create a *fake* instance? I would have thought the point of it would be to copy a *real* instance, e.g. clone my Steam install so I could put different mods in the copy.

DasSkelett:
So my last commit moved most of the fake-code to `KSPManager`. Should enable writing tests and implementing it in the GUI or ConsoleUI.

I've decided to split `FakeInstance()` and `CloneInstance` in two seperate functions, but `CloneInstance()` calls `FakeInstance()` once it parsed the version and stuff.
Hope that's kinda how you imagined it, but I'm open to critic and suggestions.

A new `public KspVersion AddBuildToVersion()` inside `Versioning.KspVersion` (better names welcome) tries to find the right build number for the version object it is called on out of the build map using `KnownVersions`. Then it returns the new "complete" version object.
I decided to throw an error if the patch is not specified, as @HebaruSan  proposed.

Also addded some logging.

Core/KSPManager.cs
```diff
@@ -151,6 +152,76 @@ public KSP AddInstance(KSP ksp_instance)
...
+                if (KSP.IsKspDir(new_path)) {
+                    throw new BadInstallLocationKraken("There is already a KSP instance at this path. Delete the old one first.");
+                }
+
+                log.DebugFormat("Creating folder structure and text files at {0} for KSP version {1}", Path.GetFullPath(new_path), version.ToString());
+
+                // Create a KSP root directory, containing a GameData folder, a buildID.txt and a readme.txt
+                Directory.CreateDirectory(new_path);
+                Directory.CreateDirectory(Path.Combine(new_path, "GameData"));
+                File.WriteAllText(Path.Combine(new_path, "buildID.txt"), String.Format("build id = {0}", version.Build));
```
HebaruSan: `version.Build` could still be null here, e.g. if I choose a fake version like "1.2.4". You'll probably have to do an explicit `IsBuildDefined` check and skip this step if false.

Core/KSPManager.cs
```diff
@@ -151,6 +152,76 @@ public KSP AddInstance(KSP ksp_instance)
...
+        {
+            if (existing_instance.Valid)
+            {
+                KspVersion version = existing_instance.Version();
+
+                CKAN.DLC.MakingHistoryDlcDetector dlcDetector = new DLC.MakingHistoryDlcDetector();
+                bool DLC = dlcDetector.IsInstalled(existing_instance, out string identifier, out UnmanagedModuleVersion moduleVersion);
+                string dlcVersion = moduleVersion.ToString();
+
+                FakeInstance(new_name, new_path, version, DLC, dlcVersion);
```
HebaruSan: Do we actually want "cloning" to create a fake instance? I would have thought the point of it would be to copy a real instance, e.g. clone my Steam install so I could put different mods in the copy.

DasSkelett: Oops, might have misunderstood that one.
So if we really copy an existing one, what should we copy? The whole directory with all of KSPs files?

Core/KSPManager.cs
DasSkelett: `AddBuildToVersion()` returns a  null object if it can't find the version in the build map, and therefore `FakeNewKSPInstall` throws an exception before `FakeInstace()` is called.
But if we call it from anywhere else in the future, it might make sense to to a check inside this function.

Core/KSPManager.cs
HebaruSan: I would think we would copy just enough to create a pristine working copy, but let's let @politas weigh in.

Core/KSPManager.cs
DasSkelett: Thinking more about this, it would be useful to have a new `public bool IsValid()` in `KspVersion`.
That could be checked in `FakeInstance()` and if `false`, throw an exception.
I don't see a sense in allowing unreal KSP versions in a  dummy instance.

Core/KSPManager.cs
politas: I meant copy the whole directory with all files, so that users can then trial some new mods without messing up their main install. Or create a base install with a minimal set of core mods, and clone that to try different planet packs or strategy replacements, or tech trees or whatever.

DasSkelett: Okay, gotcha.

DasSkelett:
Sooooo, I changed quite a bit again.
We have now `ckan ksp clone [CommonOptions] <instanceName> <newname> <newpath>` to _really_ copy an existing KSP directory, with **all** folders and files inside.
I created a new `Utilities` class because I had no idea where else to add a `CopyDirectory()` function.
Now `ckan ksp fake` also creates a `buildID664.txt`, in the case that this will be relevant some day.
I also wrote a `public bool IsValid()` in ` Versioning.KspVersion` to be able to check if versions are official/real ones.

Core/Utilities.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
+using System.IO;
+
+namespace CKAN
+{
+    public static class Utilities
+    {
+        /// <summary>
+        /// Copies a directory and optionally its supdirectories.
```
politas: Typo - Subdirectories

Core/Versioning/KspVersion.cs
```diff
@@ -444,6 +444,53 @@ public static bool TryParse(string input, out KspVersion result)
...
+        /// </summary>
+        /// <returns>The build for the version. Null if version is not known in the build map.</returns>
+        public KspVersion AddBuildToVersion ()
+        {
+            if (!IsPatchDefined)
+            {
+                throw new ArgumentOutOfRangeException();
+            }
+
+            List<KspVersion> knownVersions = new KspBuildMap(new Win32Registry()).KnownVersions;
```
politas: So add a 'knownVersion.Reverse()' here to in-place reverse the list, and we'll find the last valid build number, most likely faster.

politas:
Will the following command work?
`ckan ksp clone <destdir> --ksp-dir <sourcedir>`

Core/Versioning/KspVersion.cs
DasSkelett: Yep, good idea 👍

DasSkelett:
No, currently it only accepts instance names known in the registry.
But I think I will implement it as ` ckan ksp clone <instanceNameOrPath> <newname> <newpath>`, that should be easier.

DasSkelett:
As I'm going through the code, I found that when testing a KSP directory for validity, we only ask if `GameData` exists (despite the GUI specifically asking for a `buildID.txt`).
https://github.com/KSP-CKAN/CKAN/blob/48d29f1313ea1b5261a8c4014031c1db076ab6a5/Core/KSP.cs#L208-L217
That kinda affects my new `clone` and `fake` in the sense that I can't properly test if everything went right.

In my opinion it would make sense to expand the checks for a `buildID.txt` and a `readme.txt`.
I'm not sure if they were part of a KSP install since the earliest versions, but I think so.

Cmdline/Action/KSP.cs
```diff
@@ -77,6 +96,13 @@ internal class AddOptions : CommonOptions
             [ValueOption(1)] public string path { get; set; }
         }
 
+        internal class CloneOptions : CommonOptions
+        {
+            [ValueOption(0)] public string nameOrPath { get; set; }
+            [ValueOption(1)] public string new_name { get; set; }
+            [ValueOption(1)] public string new_path { get; set; }
```
HebaruSan: One of these should probably be `ValueOption(2)`

DasSkelett: Interesting, what exactly does this number do? Because my tests worked despite the typo...

HebaruSan: It should be the index of that option in the list. Maybe it's got some sanity checking built in (hard to check the code for this dependency because it's been completely changed since the version CKAN uses).

Tests/Core/KSPManager.cs
```diff
@@ -109,6 +109,37 @@ public void AddInstance_ManagerHasInstance()
             }
         }
 
+        [Test]
+        public void CloneInstance_BadInstance_ThrowsNotKSPDirKraken()
```
HebaruSan: This test seems to be failing.

https://travis-ci.org/KSP-CKAN/CKAN/jobs/473907148#L2592

Core/Versioning/KspVersion.cs
```diff
@@ -444,6 +444,54 @@ public static bool TryParse(string input, out KspVersion result)
                 return false;
             }
         }
+
+        /// <summary>
+        /// Searches the build map if it knows the version.
+        /// </summary>
+        /// <returns><c>true</c>, if version is valid, <c>false</c> otherwise.</returns>
+        public bool IsValid()
```
HebaruSan: This notion of "valid" is specific to the operation you're creating and doesn't apply in other cases. E.g., a version of "1.5" is valid when parsing it from a .ckan file, but this function would return false. This logic should probably live someplace more related to creating fake installs.

Core/Versioning/KspVersion.cs
```diff
@@ -444,6 +444,54 @@ public static bool TryParse(string input, out KspVersion result)
...
+            }
+            return false;
+        }
+
+        /// <summary>
+        /// Returns a "complete" KspVersion object, including the build number.
+        /// If a version number has multiple possible builds, it takes the latest one.
+        /// </summary>
+        /// <returns>The build for the version. Null if version is not known in the build map.</returns>
+        public KspVersion AddBuildToVersion ()
```
HebaruSan: This name is misleading; this function doesn't "add" anything. It's more of a "find known version" function.

Tests/Core/KSPManager.cs
DasSkelett: Yep, found the mistake already and is fixed in the next commit.

Cmdline/Action/KSP.cs
DasSkelett: Okay, thanks

Cmdline/Action/KSP.cs
```diff
@@ -248,6 +292,113 @@ private int AddInstall(AddOptions options)
...
+            // Try to use instanceNameOrPath as a path and create a new KSP object.
+            // If it's valid, go on.
+            else if (new CKAN.KSP(instanceNameOrPath, new_name, User) is CKAN.KSP instance && instance.Valid)
+            {
+                // Same as above.
+                try
+                {
+                    Manager.CloneInstance(instance, new_name, new_path);
+                }
+                catch (NotKSPDirKraken kraken)
```
HebaruSan: Is there any way we could consolidate these `catch` blocks with the identical ones above? Any future fixes to these areas would have to be made in multiple places.

Core/KSPManager.cs
```diff
@@ -151,6 +152,103 @@ public KSP AddInstance(KSP ksp_instance)
...
+                Directory.CreateDirectory(Path.Combine(new_path, "GameData"));
+
+                // Don't write the buildID.txts if we have no build, otherwise it would be -1.
+                if (version.IsBuildDefined)
+                {
+                    File.WriteAllText(Path.Combine(new_path, "buildID.txt"), String.Format("build id = {0}", version.Build));
+                    File.WriteAllText(Path.Combine(new_path, "buildID64.txt"), String.Format("build id = {0}", version.Build));
+                }
+
+                File.WriteAllText(Path.Combine(new_path, "readme.txt"), String.Format("Version {0}", version.ToString()));
```
HebaruSan: Will this output the build ID? SQUAD's readme.txt file doesn't contain that.

Core/Versioning/KspVersion.cs
DasSkelett: So in KSPManager?

Core/KSPManager.cs
DasSkelett: Yep, it does, but it shouldn't. Fixing...

Core/Utilities.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    public static class Utilities
+    {
+        /// <summary>
+        /// Copies a directory and optionally its subdirectories.
+        /// </summary>
+        /// <param name="sourceDirPath">Source directory path.</param>
+        /// <param name="destDirPath">Destination directory path.</param>
+        /// <param name="copySubDirs">Copy sub dirs recursively if set to <c>true</c>.</param>
+        public static void CopyDirectory(string sourceDirPath, string destDirPath, bool copySubDirs)
```
HebaruSan: Could/should we do this within a Transaction? Seems like this is as likely to fail as installing a mod would be, and it would be nice to be able to auto clean it up if that happens.

Core/KSPManager.cs
```diff
@@ -151,6 +152,103 @@ public KSP AddInstance(KSP ksp_instance)
...
+            }
+        }
+
+        /// <summary>
+        /// Create a new fake KSP instance
+        /// </summary>
+        /// <param name="new_name">The name for the new instance.</param>
+        /// <param name="new_path">The loaction of the new instance.</param>
+        /// <param name="version">The version of the new instance. Should have a build number.</param>
+        /// <param name="DLC">Whether to fake the DLC too.</param>
```
HebaruSan: Do we need an independent `DLC` parameter, or could we simply check whether `dlcVersion` is null? It doesn't look like passing these parameters in unexpected ways would be useful, e.g., DLC=true and dlcVersion=null, or DLC=false and dlcVersion="1.0.0".

Core/KSPManager.cs
```diff
@@ -151,6 +152,103 @@ public KSP AddInstance(KSP ksp_instance)
...
+                throw new BadInstallLocationKraken("The specified folder already exists and is not empty.");
+            }
+
+            try
+            {
+                log.DebugFormat("Creating folder structure and text files at {0} for KSP version {1}", Path.GetFullPath(new_path), version.ToString());
+
+                // Create a KSP root directory, containing a GameData folder, a buildID.txt/buildID64.txt and a readme.txt
+                Directory.CreateDirectory(new_path);
+                Directory.CreateDirectory(Path.Combine(new_path, "GameData"));
```
HebaruSan: Could this could create the other folders that the validation scripts currently do? This would help with being able to install mods.

https://github.com/KSP-CKAN/xKAN-meta_testing/blob/36ccd470cde1c6a02a8a3fc267e2eddad2b6640c/NetKAN/build.sh#L54-L59

Core/Versioning/KspVersion.cs
DasSkelett: Or change the name to `IsKnown`? This would also clear up the use case.

Cmdline/Action/KSP.cs
DasSkelett: Yep, changed in next commit.

Core/Versioning/KspVersion.cs
HebaruSan: Could you use `IsBuildDefined` instead? I think that's what this check amounts to ultimately.

DasSkelett: `IsBuildDefined` name is already used:
https://github.com/KSP-CKAN/CKAN/blob/0604a12739acefa30f0ee69223ba43dbb52cbe64/Core/Versioning/KspVersion.cs#L79
Also, it doesn't only check if the build number (if specified) is valid, it works for non-build-defined versions to (i.e. 1.5.1 returns true, 1.5.1.2335 returns true, but 1.5.9 returns false, so does 1.5.1.6666).
What about `InBuildMap` instead?

HebaruSan: Sorry, I meant check `IsBuildDefined` instead of `IsValid` in `FakeInstance`. In both cases, the check would be true for a known version and false otherwise, right?

DasSkelett: Something I can't see your new comment on GitHub, but I got it per mail @HebaruSan 

> Sorry, I meant check IsBuildDefined instead of IsValid in FakeInstance. In both cases, the check would be true for a known version and false otherwise, right?

No, IsBuildDefined only checks if there is a build number specified (i.e. not -1), but not if it is the right one. Also IsBuildDefined can't find fake version numbers like 1.4.99.

HebaruSan: I deleted that comment because I realized it was wrong. But email notifications can't be deleted. :(

Core/Utilities.cs
DasSkelett: Does that work the way I did it now?

politas:
The only reason we ask for a BuildID.txt when adding a KSP install is to be able to use the File Selector rather than the directory selector dialogue, because it works way better.

politas:
Cool! Now we just need to add it to the GUI interface.

HebaruSan:
@DasSkelett , FYI, [the validation scripts are now set up to use this](https://github.com/KSP-CKAN/xKAN-meta_testing/pull/44), and working fine!

PR #2635, opened 2019-01-06
Allow overriding menu tip in ConsoleUI

HebaruSan:
## Motivation

The screen tip for the main menu is always "F10 - Menu" in ConsoleUI, but sometimes a more specific string might be helpful.

## Changes

Now the "F10 - Menu" tip is changed on two screens:

- "F10 - Links" on the mod info screen
- "F10 - Sort" on the game instance list screen

As a bit of side cleanup, `ConsoleScreen.LeftHeader` and `ConsoleScreen.CenterHeader` are now changed from `Func<string>` variables to virtual string functions (to be consistent with the newly added `MenuTip`). This is a cleaner way of providing a function that can be overridden in a child class.

### DIFF
--- a/ConsoleUI/AuthTokenListScreen.cs
+++ b/ConsoleUI/AuthTokenListScreen.cs
@@ -15,9 +15,7 @@ namespace CKAN.ConsoleUI {
         /// </summary>
         public AuthTokenScreen() : base()
         {
-            LeftHeader   = () => $"CKAN {Meta.GetVersion()}";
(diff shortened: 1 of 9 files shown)


DasSkelett:
Nice :+1:

HebaruSan:
Note that we reserve the right to change it back to Menu if we ever need to add non-link content to that menu. :)

DasSkelett:
I am willing to make this comprise.

PR #2648, opened 2019-01-10
Derive User classes directly from IUser interface instead of the NullUser class.

DasSkelett:
Problem:
---
Our UI specific `(I)User` classes for interacting with the... user are pretty messy.
Besides to the consoleUI all of them are inheriting from `NullUser` instead of directly extending the `IUser` interface, which kinda misses the point of having an interface.
Also the `NullUser` implements a `Display...()` method for each `IUser` method _on top_ of the existing `Raise...()` methods, just to let the `Raise..()` ones call the `Display...()` ones without any reason to do so.

Solution:
---
Now all `User` classes inherit directly from `IUser`.
`NullUser` is now only used in tests. 
All `Display...()` methods are removed.
Added LOTS of code documentation, because I could have needed them...

And i implemented the selection dialog for the GUI.

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -177,7 +177,15 @@ namespace CKAN
                 }
             }
 
-            bool ok = User.RaiseYesNoDialog("\r\nContinue?");
+            bool ok;
+            if (User.GetType().ToString() == "CKAN.GUIUser")
+            {
+                ok = true;
+            }
+            else
+            {
+                ok = User.RaiseYesNoDialog("\r\nContinue?");
+            }
 
             if (!ok)
             {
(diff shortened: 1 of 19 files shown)


DasSkelett:
Added the GUI dialog, looks like this:
seldiag-windows selDiagUbuntuDark

"  -- Default"  is added to the default selection, if any.

Core/ModuleInstaller.cs
```diff
@@ -177,7 +177,15 @@ public void InstallList(ICollection<CkanModule> modules, RelationshipResolverOpt
                 }
             }
 
-            bool ok = User.RaiseYesNoDialog("\r\nContinue?");
+            bool ok;
+            if (User.GetType().ToString() == "CKAN.GUIUser")
```
HebaruSan: The old method that GUI used for suppressing this prompt was pretty bad, but this might be worse.

What about a new `bool confirmPrompt = true` parameter, which only GUI would pass as false?

DasSkelett: Yeah, I don't like my solution either, but I searched for a way to get around this without adding something to the interface and all the User classes, because it is only needed for the GUI. But maybe I should do it like this.

PR #2656, opened 2019-01-10
Invoke control accesses in `UpdateModsList()`

DasSkelett:
#2617 made `UpdateModsList()` run in a background thread.
So all accesses to `FilterToolButton` and `UpdateToolButton` are made via this new task, and not the UI thread.
This didn't throw any exception on runtime or anything, but in Visual Studio if you debug CKAN.

Now all accesses to any GUI elements in `UpdateModsList()` are done through invocation.


politas:
Nononono, you're not catching me with that again. I reject your title!

PR #2657, opened 2019-01-11
Set focus to mod list after loading

### LINKED ISSUE #2651: Can't scroll the modlist if it's not selected on Linux
Background
----------

**CKAN Version:** 
1.25.4

**Operating System:** 
Kubuntu 18.04.1

Problem
-------

1) Open it up
2) Move your mouse over the mod list but don't select a mod 
3) Scroll up or down

You can scroll through the mod list.

**What happened instead?**
You can't, the mod list doesn't move.
Also happens if you "deselect" the mod list (click anywhere else in the program)

Only affects Linux (and maybe MacOS?), not Windows.

Was initially fixed in #1584, but doesn't seem to work anymore.

HebaruSan:
## Problem

On loading GUI, the focus is somewhere invisible (the splitter or the tab control, from some debugging). It should be on the mod list.

## Cause

Not clear exactly what caused this. Possibilities:

- #2617 moved the mod list loading process into a background process and switched the GUI into the progress bar screen while it runs. The extra switching between tabs may have confused WinForms.
- #2556 made the mod info sidebar show/hide when mods are focused, and also rearranged the hierarchy of controls so the main tabs are inside the splitter. Some of this happens during load, so that may also have confused WinForms.

## Changes

Now we explicitly set the focus to the mod list at the end of `UpdateModsList`, after we switch back to that tab.
(I'd like to find a more generic fix, but none of the solutions I tried worked.)

Fixes #2651.

### DIFF
--- a/GUI/MainModList.cs
+++ b/GUI/MainModList.cs
@@ -258,6 +258,7 @@ namespace CKAN
             tabController.HideTab("WaitTabPage");
             tabController.SetTabLock(false);
             Util.Invoke(this, SwitchEnabledState);
+            Util.Invoke(this, () => Main.Instance.ModList.Focus());
         }
 
         public void MarkModForInstall(string identifier, bool uncheck = false)


politas:
Oh, yes!  That's been so annoying!

PR #2658, opened 2019-01-13
Small text/number formatting changes to mod list

DasSkelett:
* Align the download count number to the right
* Align the size text to the right
* Format the download count number with thousands separators.
* Change size appendages to binary prefixes following [ISO/IEC 80000-13:2008](https://en.wikipedia.org/wiki/ISO/IEC_80000) and  [IEEE 1541-2002](https://en.wikipedia.org/wiki/IEEE_1541-2002) ;)

### DIFF
--- a/Core/Types/CkanModule.cs
+++ b/Core/Types/CkanModule.cs
@@ -737,11 +737,11 @@ namespace CKAN
             if (bytes < K) {
                 return $"{bytes} B";
             } else if (bytes < K * K) {
-                return $"{bytes / K :N1} KB";
+                return $"{bytes / K :N1} KiB";
             } else if (bytes < K * K * K) {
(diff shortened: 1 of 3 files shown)


politas:
You can bet that some people will complain about the KiB/MiB/GiB change.

HebaruSan:
... starting with me. Is that necessary? It seems like pedantry for its own sake.

DasSkelett:
No, not necessary. But I think more correct.
So why don't do it this way?
It's not as if this was a big effort to change.

HebaruSan:
Basically because no one uses those abbreviations, and everyone uses MB and GB.

DasSkelett:
So make the people use them ; )

HebaruSan:
Well, at least 'i' is a small character. :man_shrugging:

PR #2659, opened 2019-01-14
Remove ConfirmPrompt from IUser

HebaruSan:
#2648 added a `IUser.ConfirmPrompt` property. This doesn't belong there because it's just a way for two core functions to avoid raising specific yes/no prompts in GUI (which prior to #2648 was done with a hack of setting `GUIUser.displayYesNo = null`). There is no general concept of "confirm prompt" that needs to be represented at this level.

Now this is removed and replaced with parameters in `ModuleInstaller.InstallList` and `ModuleInstaller.UninstallList`, where they're actually used. GUI passes false, everyone else uses the default of true.

### DIFF
--- a/Cmdline/ConsoleUser.cs
+++ b/Cmdline/ConsoleUser.cs
@@ -31,15 +31,6 @@ namespace CKAN.CmdLine
         public bool Headless { get; }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


DasSkelett:
Ohhh, you meant it as a parameter of the function.... Didn't get that.

PR #2628, opened 2018-12-31
Dispose caches in tests

HebaruSan:
## Problem

As of recently (probably #2613), `./build test` is failing on Mono for me with this:

```
1) Failed : Tests.Core.ModuleInstaller.ModuleManagerInstancesAreDecoupled
  Expected: No Exception to be thrown
  But was:  <System.InvalidOperationException: object_op ---> System.IO.IOException: The configured user limit (128) on the number of inotify instances has been reached.
  at System.IO.CoreFX.FileSystemWatcher.StartRaisingEvents () [0x0007b] in <eeec8e88b1fc462e84ade7bf334d0c97>:0 
  at System.IO.CoreFX.FileSystemWatcher.StartRaisingEventsIfNotDisposed () [0x00019] in <eeec8e88b1fc462e84ade7bf334d0c97>:0 
  at System.IO.CoreFX.FileSystemWatcher.set_EnableRaisingEvents (System.Boolean value) [0x0001d] in <eeec8e88b1fc462e84ade7bf334d0c97>:0 
  at (wrapper remoting-invoke-with-check) System.IO.CoreFX.FileSystemWatcher.set_EnableRaisingEvents(bool)
  at System.IO.CoreFXFileSystemWatcherProxy+<>c.<StartDispatching>b__9_0 (System.IO.CoreFX.FileSystemWatcher internal_fsw, System.IO.FileSystemWatcher fsw) [0x00048] in <eeec8e88b1fc462e84ade7bf334d0c97>:0 
  at System.IO.CoreFXFileSystemWatcherProxy.Operation (System.Action`4[T1,T2,T3,T4] map_op, System.Action`2[T1,T2] object_op, System.Object handle, System.Action`2[T1,T2] cancel_op) [0x00184] in <eeec8e88b1fc462e84ade7bf334d0c97>:0 
   --- End of inner exception stack trace ---
  at System.IO.CoreFXFileSystemWatcherProxy.Operation (System.Action`4[T1,T2,T3,T4] map_op, System.Action`2[T1,T2] object_op, System.Object handle, System.Action`2[T1,T2] cancel_op) [0x001a6] in <eeec8e88b1fc462e84ade7bf334d0c97>:0 
  at System.IO.CoreFXFileSystemWatcherProxy.StartDispatching (System.Object handle) [0x00004] in <eeec8e88b1fc462e84ade7bf334d0c97>:0 
[...]

PR #2660, opened 2019-01-16
Support depends on any_of lists

### LINKED ISSUE #1693: CKAN tests mod version interdependencies incorrectly
Bug report as requested by @plague006 on KSP-CKAN/NetKAN#3841. 

**CKAN Version:** v1.16.1-0-g2e91715 (beta)
**Operating System:** Windows 10 KP
**The issue you are experiencing:** The Custom Asteroids config file .ckans each depend on a minimum version of Custom Asteroids, but have a much broader range of KSP versions. Currently the dependencies are:
- [CustomAsteroids-v1.2.0.ckan](https://github.com/KSP-CKAN/CKAN-meta/blob/master/CustomAsteroids/CustomAsteroids-v1.2.0.ckan): requires KSP 1.0.0-1.0.5
- [CustomAsteroids-v1.3.0.ckan](https://github.com/KSP-CKAN/CKAN-meta/blob/master/CustomAsteroids/CustomAsteroids-v1.3.0.ckan): requires KSP 1.1.0-1.1.2
- [CustomAsteroids-Pops-Stock-Inner-v1.2.0.ckan](https://github.com/KSP-CKAN/CKAN-meta/blob/master/CustomAsteroids-Pops-Stock-Inner/CustomAsteroids-Pops-Stock-Inner-v1.2.0.ckan): requires Custom Asteroids >= v1.0.0
- [CustomAsteroids-Pops-Stock-Inner-v1.3.0.ckan](https://github.com/KSP-CKAN/CKAN-meta/blob/master/CustomAsteroids-Pops-Stock-Inner/CustomAsteroids-Pops-Stock-Inner-v1.3.0.ckan): requires Custom Asteroids >= v1.3.0

The intended behavior is that Pops-Stock-Inner v1.3.0 gets installed if you have Custom Asteroids v1.3.0, and v1.2.0 gets installed if you have any older version.

However, trying to install Custom Asteroids for KSP 1.0.5 (Custom Asteroids v1.2.0), which requires Pops-Stock-Inner as a dependency, causes CKAN to report a conflict between Pops-Stock-Inner v1.3.0 and Custom Asteroids v1.2.0. There is no option to download Pops-Stock-Inner v1.2.0, even though it is compatible with both KSP 1.0.5 and Custom Asteroids v1.2.0 and would satisfy Custom Asteroids's config dependency.

**How to recreate this issue:** (I will shortly be modifying `CustomAsteroids-Pops-*.netkan` to work around this issue, so the original example can't be used. The issue should be reproducible with any mod that depends on a newer version of a currently installed mod.)

### LINKED ISSUE #2522: Mod Relationships do not allow for user to select 1 hard dependency from a list of possible hard dependencies
Background
----------

**CKAN Version:** 1.25.3

**KSP Version:** 1.4.5

**Operating System:** Windows 10 

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** No... although all of this and the above it kind of unnecessary in this situation.

Problem
-------

The issue I have pertains to how CKAN cannot intelligently handle dependencies when there are multiple plugins that can offer a solution for a mod but the the mod developer cannot provide a list to a player to select just 1 from that list to provide the dependency. 

For example (and specific to my situation), I provide a Skybox mod to the player base which requires 1 of a few mods as a dependency in order for it to function: Drop-in Replacement Textures, Sigma Replacements: Skybox, TextureReplacer OR TextureReplacerReplaced. All 4 of these mods provide functionality for my mod to work but I cannot list them all as hard dependencies as there may be conflicts between the mods providing this functionality, a mod may not update to the latest version of KSP and therefore hold back my mod being offered to the player base and installing all 4 does not provide the player with an intelligent solution to this. Only choosing 1 as a hard dependency limits the scope of my mod and may not fit in with the desired mod of choice of the player. Listing all 4 mods as Recommends of Suggests is not a good solution as the player may inadvertently untick all the mods and not install the dependent plugin causing my mod to not function.

What I would suggest is that a procedure to be put into the scope of the relationship handling to allow for mod developers to state a list of dependent mods from within a list/group of mods and that they are only able to pick one from that list. If one of the mods is already installed, then bypass this action as normal.

I have mentioned it previously on the CKAN forum thread and HebaruSan has confirmed that this behavior is not handled at this time by CKAN and discussed options here: https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1253-woomera/&do=findComment&comment=3440100

I appreciate that this would be a fair amount of new logic for the relationships system but may give the system a lot more intelligence with regards to dependency handling.

HebaruSan:
## Background

Sometimes a mod needs some functionality to be provided by a third party dependency, and there are several choices available, but they have different APIs, so the dependent mod has to implement support for each one separately.

For example, TextureReplacer, TextureReplacerReplaced, SigmaReplacements-SkyBox, and DiRT can all be used to replace the stock SkyBox with a custom one, but they're all mutually incompatible and so a dependent mod has to implement support for them separately one by one.

The `provides` property can't handle this (well), since there is no baseline abstract concept that they could sensibly `provide`. Consider if we had all four of the above mentioned mods `provide` something like SkyBoxReplacer; then we would not have a solution for a dependent mod that only implemented compatibility with 2 or 3 of them. We would need one `provides` tag for each possible permutation of supports that a mod could implement, which would be unworkable.

## Changes

**NOTE: This is a high risk change. Core pieces of `RelationshipResolver` are modified, which risks breaking common tasks. Please review with care!**

Now a new `any_of` format for relationships is allowed:

```json
    "depends": [
        {
            "any_of": [
                { "name": "SigmaReplacements-Heads" },
                { "name": "TextureReplacerReplaced" }
            ]
        },
        { "name": "ModuleManager" }
    ],
```

This creates a requirement that can be satisfied by any of the specified modules. If the user installs a module with the above syntax, they are prompted to choose which dependency to use:

```
$ ~/Downloads/KSP/CKAN/_build/ckan.exe install -c ScottManleyHeadPack-1.0.0.ckan 
Too many mods provide SigmaReplacements-Heads OR TextureReplacerReplaced. Please pick from the following:

1) SigmaReplacements-Heads (Sigma Replacements: Heads)
2) TextureReplacerReplaced (TextureReplacerReplaced)
1
About to install...

 * Scott Manley Head Pack 1.0.0 (cached)
 * Sigma Replacements: Heads H_v0.2.4 (cached)

Continue? [Y/n] 


Installing mod "ScottManleyHeadPack 1.0.0"
Installing mod "SigmaReplacements-Heads H_v0.2.4"
Updating registry
Committing filesystem changes
Rescanning GameData
Done!
```

(It works in GUI too.)

The schema and spec are updated to incorporate this syntax.

`RelationshipDescriptor` is now an abstract base class (let me know if an interface would be better somehow) with two child classes, `ModuleRelationshipDescriptor` (formerly known as `RelationshipDescriptor`) and `AnyOfRelationshipDescriptor` which models the new syntax and mainly holds a `List<RelationshipDescriptor>` and multiplexes access to it.

Various tests are updated to instantiate `ModuleRelationshipDescriptor` now that `RelationshipDescriptor` is abstract.

A new `JsonRelationshipConverter` class handles parsing .ckan files, populating lists of `RelationshipDescriptor` with instances of the appropriate child classes based on the contained properties.

Parts of `RelationshipResolver` are refactored to handle the new syntax, mainly by calling virtual members of `RelationshipDescriptor`.

Fixes #2522.

## Reverse dependencies fix

Previously if you installed a mod that depended on another mod, but the latest of version of the dependency conflicts with or depends on a different version of the original mod, then you'd get an error.

Now older versions of the dependency will be checked to see whether they're compatible.

Includes tests to make sure this works.

Fixes #1693.

## Tests fix

I was still getting the error from #2628, I think because the cache objects in KSPManager weren't being disposed. Now this is added to the affected tests.

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -338,12 +338,12 @@
                     {
                         "type": "object",
                         "properties": {
-                            "any-of": {
+                            "any_of": {
                                 "description": "List of modules that can satisfy this relationship",
                                 "$ref": "#/definitions/relationship"
                             }
                         },
-                        "required": [ "any-of" ]
+                        "required": [ "any_of" ]
                     }
                 ]
             }
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -312,26 +312,40 @@
             "description" : "A relationship to a list of mods",
             "type" : "array",
             "items" : {
-                "type" : "object",
-                "properties" : {
-                    "name" : {
-                        "description" : "Identifier of the mod",
-                        "$ref"        : "#/definitions/identifier"
+                "oneOf": [
+                    {
+                        "type" : "object",
+                        "properties" : {
+                            "name" : {
+                                "description" : "Identifier of the mod",
+                                "$ref"        : "#/definitions/identifier"
+                            },
+                            "version" : {
+                                "description" : "Optional version",
+                                "$ref"        : "#/definitions/version"
+                            },
+                            "min_version" : {
+                                "description" : "Optional minimum version",
+                                "$ref"        : "#/definitions/version"
+                            },
+                            "max_version" : {
+                                "description" : "Optional maximum version",
+                                "$ref"        : "#/definitions/version"
+                            }
+                        },
+                        "required" : [ "name" ]
                     },
-                    "version" : {
-                        "description" : "Optional version",
-                        "$ref"        : "#/definitions/version"
-                    },
-                    "min_version" : {
-                        "description" : "Optional minimum version",
-                        "$ref"        : "#/definitions/version"
-                    },
-                    "max_version" : {
-                        "description" : "Optional maximum version",
-                        "$ref"        : "#/definitions/version"
+                    {
+                        "type": "object",
+                        "properties": {
+                            "any-of": {
+                                "description": "List of modules that can satisfy this relationship",
+                                "$ref": "#/definitions/relationship"
+                            }
+                        },
+                        "required": [ "any-of" ]
                     }
-                },
-                "required" : [ "name" ]
+                ]
             }
         },
         "install_to" : {
--- a/Cmdline/Action/Show.cs
+++ b/Cmdline/Action/Show.cs
@@ -155,7 +155,7 @@ namespace CKAN.CmdLine
             }
 
             user.RaiseMessage("- status:\t{0}", module.release_status);
-            user.RaiseMessage("- license:\t{0}", string.Join(", ", module.license)); 
+            user.RaiseMessage("- license:\t{0}", string.Join(", ", module.license));
             #endregion
 
             #region Relationships
@@ -185,7 +185,7 @@ namespace CKAN.CmdLine
                 user.RaiseMessage("\r\nProvides:");
                 foreach (string prov in module.ProvidesList)
                     user.RaiseMessage("- {0}", prov);
-            } 
+            }
             #endregion
 
             user.RaiseMessage("\r\nResources:");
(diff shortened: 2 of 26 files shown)


politas:
I'm testing this PR, and I'm not sure when things changed, but if I get a failure applying a changeset and click retry, The ChangeSet tab reappears, and is selected, but the actual tab contents remain as the Status page, so to re-appy the changeset, I need to switch off and back to the Changeset tab.

PR #2671, opened 2019-01-29
Support replaced_by property

### LINKED ISSUE #904: Create a replaced_by relationship to allow seamless continuation of deprecated mods

HebaruSan:
## Motivation

See #904; sometimes after a mod goes out of maintenance, a successor takes its place. It would be helpful if we could give users a cue that this has happened.

## Changes

This is #1888 rebased onto current master, plus:

- Replace checkbox column
- Replaced by field added to mod info pane
- Replacements handled in change sets and installer
- Replaceable filter

For testing, I set up a fake relationship to replace KSP-AVC with ZeroMiniAVC:

image

image

Checking the box updates the change set to remove the replaced mod and install the replacement:

image

Not done:

- Right click menu option. I didn't think this fit well since we have a checkbox for it anyway.
- Netkan validation. This would be done in xKAN-meta_testing instead, e.g. KSP-CKAN/xKAN-meta_testing#45
- Tests

Fixes #904.
Closes #1888.

## Side fixes

- Reset the Update checkbox when clearing the change set (discovered because we need to do the same with the Replace checkbox)
- Centralize filter definitions in `IsModInFilter` rather than implementing them in multiple places
- Release status in mod info was often wrong because it wasn't updated when null, now it changes to N/A
- The mod list right click menu now looks better in dark themes


DasSkelett:
I don't really like having another checkbox column at the beginning, especially one that will be used very rarely.

But I can't think of a better way right now either.

politas:
Yeah, I was trying to work a way of having an image column to indicate either an update or a replacement  for the existing checkbox, but kept hitting brick walls.

PR #2681, opened 2019-02-10
Releases option for Netkan

HebaruSan:
## Motivation

KSP-CKAN/NetKAN#7003 pointed out that two versions of UniversalStorage2 are missing from CKAN-meta. This happened because multiple versions were uploaded within the same bot pass, so only the most recent release was indexed, and the other was skipped. This has always been a risk, but it was mitigated for a while by the SpaceDock-CKAN web hook. Now with KSP-SpaceDock/SpaceDock#192 it is a bigger problem.

Currently the easiest way I know to fix this is to create fake .netkan files with hardcoded `download` properties, and then run Netkan once on each of them. This could miss things, but it's usually safe and can be checked by diffing against other versions.

It would be nice to be able to generate older .ckan files with less hassle.

## Changes

Now Netkan has a new `--releases` option that accepts a parameter of either `all` or a number (the default is 1). If passed, Netkan will process the specified number of releases of a module, starting with the most recent, and generate a .ckan file for each one when the `$kref` uses one of the following:

- Curse
- GitHub
- SpaceDock
- Jenkins

To make this happen, `ITransformer.Transform` now returns `IEnumerable<Metadata>` instead of just `Metadata`, which means transformers can generate multiple outputs from one input. All existing transformers are updated to `yield return` their previous outputs to generate a single-element sequence. Existing tests are updated to use `.First()` to get the first `Metadata` object in the returned lists.

For the transformers that process the `$kref` values listed above, their source APIs now have some form of `All` or `AllReleases` property that includes everything in the given repository for that module. When `--releases` is passed, we parse it into an `int?` with `null` representing `"all"` and pass this value in a new `releases` parameter of their constructors, and these transformers will generate and return one new `Metadata` object for each release up to the specified number, which will then be subjected to all normal further downstream processing.

```
$ netkan.exe ~/Downloads/KSP/NetKAN/NetKAN/UniversalStorage2.netkan --releases all
$ ls -gG U*
-rw-r--r-- 1 1435 Feb 10 00:53 UniversalStorage2-1.3.1.2.ckan
-rw-r--r-- 1 1435 Feb 10 00:52 UniversalStorage2-1.3.1.3.ckan
-rw-r--r-- 1 1435 Feb 10 00:51 UniversalStorage2-1.3.1.4.ckan
-rw-r--r-- 1 1435 Feb 10 00:50 UniversalStorage2-1.3.1.5.ckan
-rw-r--r-- 1 1435 Feb 10 00:49 UniversalStorage2-1.3.1.6.ckan
-rw-r--r-- 1 1435 Feb 10 00:49 UniversalStorage2-1.3.1.7.ckan
-rw-r--r-- 1 1435 Feb 10 00:48 UniversalStorage2-1.3.1.8.ckan
-rw-r--r-- 1 1435 Feb 10 00:47 UniversalStorage2-1.3.1.9.ckan
-rw-r--r-- 1 1435 Feb 10 00:53 UniversalStorage2-1.4.5.2.ckan
-rw-r--r-- 1 1435 Feb 10 00:51 UniversalStorage2-1.4.5.3.ckan
-rw-r--r-- 1 1435 Feb 10 00:50 UniversalStorage2-1.4.5.4.ckan
-rw-r--r-- 1 1435 Feb 10 00:50 UniversalStorage2-1.5.1.5.ckan
-rw-r--r-- 1 1435 Feb 10 00:49 UniversalStorage2-1.5.1.6.ckan
-rw-r--r-- 1 1435 Feb 10 00:48 UniversalStorage2-1.5.1.7.ckan
-rw-r--r-- 1 1435 Feb 10 00:48 UniversalStorage2-1.5.1.8.ckan
-rw-r--r-- 1 1435 Feb 10 00:47 UniversalStorage2-1.6.0.9.ckan
```

### DIFF
--- a/Netkan/CmdLineOptions.cs
+++ b/Netkan/CmdLineOptions.cs
@@ -28,6 +28,9 @@ namespace CKAN.NetKAN
         [Option("net-useragent", DefaultValue = null, HelpText = "Set the default User-Agent string for HTTP requests")]
         public string NetUserAgent { get; set; }
 
+        [Option("backfill", HelpText = "Process all releases instead of only the most recent")]
+        public bool BackFill { get; set; }
+
         [Option("prerelease", HelpText = "Index GitHub prereleases")]
         public bool PreRelease { get; set; }
 
--- a/Netkan/CmdLineOptions.cs
+++ b/Netkan/CmdLineOptions.cs
@@ -28,8 +28,8 @@ namespace CKAN.NetKAN
         [Option("net-useragent", DefaultValue = null, HelpText = "Set the default User-Agent string for HTTP requests")]
         public string NetUserAgent { get; set; }
 
-        [Option("backfill", HelpText = "Process all releases instead of only the most recent")]
-        public bool BackFill { get; set; }
+        [Option("releases", DefaultValue = "1", HelpText = "Number of releases to inflate, or 'all'")]
+        public string Releases { get; set; }
 
         [Option("prerelease", HelpText = "Index GitHub prereleases")]
         public bool PreRelease { get; set; }
--- a/Netkan/Program.cs
+++ b/Netkan/Program.cs
@@ -73,7 +73,7 @@ namespace CKAN.NetKAN
                         moduleService,
                         Options.GitHubToken,
                         Options.PreRelease,
-                        Options.BackFill
+                        ParseReleases(Options.Releases)
                     );
 
                     IEnumerable<Metadata> ckans = transformer.Transform(netkan);
@@ -113,6 +113,11 @@ namespace CKAN.NetKAN
             }
 
             return ExitOk;
+        }
+
+        private static int? ParseReleases(string val)
+        {
+            return val == "all" ? (int?)null : int.Parse(val);
         }
 
         private static void ProcessArgs(string[] args)
--- a/Netkan/Program.cs
+++ b/Netkan/Program.cs
@@ -1,18 +1,19 @@
 using System;
+using System.Collections.Generic;
 using System.Net;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Text;
-using CKAN.NetKAN.Model;
-using CKAN.NetKAN.Services;
-using CKAN.NetKAN.Transformers;
-using CKAN.NetKAN.Validators;
 using CommandLine;
 using log4net;
 using log4net.Core;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using CKAN.NetKAN.Model;
+using CKAN.NetKAN.Services;
+using CKAN.NetKAN.Transformers;
+using CKAN.NetKAN.Validators;
 
 namespace CKAN.NetKAN
 {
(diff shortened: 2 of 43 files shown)


politas:
Is the intention to trigger this in the bot, or do it manually and create a PR with the new .ckans?

HebaruSan:
At this point it would be manually only. Doing this in the bot would be a drastic increase of bot and cache usage unless we could limit it to only a few mods that we know need it, and I don't have a vision for that right now.

techman83:
We could have a daily backfill run and have a x_netkan_backfill flag that could be set.

politas:
> We could have a daily backfill run and have a x_netkan_backfill flag that could be set.

Would need to be careful that any customisations made to old .ckan files weren't overwritten.

To me, `--backfill` implies a selection process that only processes releases that aren't already registered, but this does all releases provided by the source?

HebaruSan:
> To me, `--backfill` implies a selection process that only processes releases that aren't already registered, but this does all releases provided by the source?

Yep. I'm open to ideas for better names; "already registered" is not something that netkan.exe can determine.

HebaruSan:
Thinking about the possibility of using this automatically, I'm partial toward this idea:

1. The bot would run a normal non-backfill netkan pass by default
2. In those cases where a *new* version is found during the normal pass, the bot would then do a partial backfill just for that one module, checking the last 3-5 releases only (would have to change this pull request to add a number option like `--backfill 3` for this to work)
3. Modifications to existing files would be ignored (to avoid overwriting customizations), but **new** files would be added/committed/pushed

How do folks feel about that idea? It ought to give us the ability to automatically catch "skipped" releases, at the cost of a bit more cache consumed.

techman83:
That sounds like an interesting idea. I don't think it would be super difficult to change the logic a little to allow for the backfill option. A check for untracked files and a function that's separate like `staged_commit`.

politas:
I'm liking the sound of this. Can we have `--releases=[x|all]` as the option, with the default being 1?

HebaruSan:
OK, done. The option is renamed `releases` and accepts an argument of either `all` or a number.

DasSkelett:
Is there a way to advise the bot to do it on a specific mod, or do I have to run it manually and upload the .ckans to the repo myself right now?
Specifically asking for InnerLock, which has several releases for different KSP versions, but has been indexed just now, so only the "latest" release for KSP 1.6.1 is on CKAN.
https://spacedock.info/mod/2113/InnerLock#changelog

politas:
There's going to be a bot upgrade that will kick this off on a regular basis (but not as often as the single release run), but for now, have to do manually. 

There'll probably never be a way to signal the bot to do it.

DasSkelett:
An option to start from a specific release (in the form of nth-to-last maybe) would be nice. Imagine doing a --releases 57 run [_cough_](https://github.com/KSP-CKAN/CKAN-meta/pull/1453) (luckily didn't happen, but the day will come...)  and it throws an error for one release somewhere in the middle. Currently there's no way to start from that release again, you would have to go through all of them again.

PR #2682, opened 2019-02-11
Avoid redundant metadata downloads

### LINKED ISSUE #854: Use If-Modified-Since when fetching metadata
Currently I am refreshing on every start which takes quite a while on my connection.
A fast mechanism to check if the mod-list changed since last refresh would be great - if the check is fast enough it could be done automatically at startup.
Ideas: HEAD of http-response, extra version file, last git-fingerprint(?)

HebaruSan:
## Motivation

Updating the registry can be slow, as it requires downloading a 2.5 MB ZIP or tar.gz file, then extracting it and parsing the JSON of all the modules and populating the registry with them. In GUI there's a further delay as the mod list and filters are updated.

If there haven't been any metadata changes since the last time you refreshed, then all of this is done for naught. It would be nice to skip it in that case.

## Changes

Now we use the `ETag` HTTP response header to determine when a repo's metadata is already up to date. This is an opaque hexadecimal string (with quotes around it) that changes when the `master.tar.gz` file does. It is returned when the file is downloaded, but it can also be retrieved with a much quicker HEAD request. In my testing, this value does indeed change when something in the CKAN-meta repo is modified.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

The `Repository` objects stored in the registry now have a `last_server_etag` property corresponding to the most recent value received for their URLs.

`Repo.UpdateAllRepositories`'s return type is changed from `bool` to an enum with three options: Updated (formerly true), Failed (formerly false), and NoChanges. If all of our repos have the same `ETag` that they had the last time we downloaded them, then it returns NoChanges without updating anything, which is much quicker than doing a full update. Otherwise an update is performed as normal, and the `ETag`s are captured from `WebClient` and saved into the `Repository` objects. (We don't attempt to suppress individual repo updates because we don't have the ability to identify which modules are from which repos. To keep the list complete, updates have to be all or nothing.)

When it encounters a NoChanges return value, GUI skips rebuilding the mod list, which further speeds up redundant refreshes.

Fixes #854.

### DIFF
--- a/Cmdline/Action/Update.cs
+++ b/Cmdline/Action/Update.cs
@@ -150,7 +150,7 @@ namespace CKAN.CmdLine
             RegistryManager registry_manager = RegistryManager.Instance(ksp);
 
             var updated = repository == null
-                ? CKAN.Repo.UpdateAllRepositories(registry_manager, ksp, manager.Cache, user)
+                ? CKAN.Repo.UpdateAllRepositories(registry_manager, ksp, manager.Cache, user) != CKAN.RepoUpdateResult.Failed
                 : CKAN.Repo.Update(registry_manager, ksp, user, repository);
 
             user.RaiseMessage("Updated information on {0} available modules",
--- a/Core/Net/Net.cs
+++ b/Core/Net/Net.cs
@@ -31,6 +31,24 @@ namespace CKAN
         };
 
         /// <summary>
+        /// Make a HEAD request to get the ETag of a URL without downloading it
+        /// </summary>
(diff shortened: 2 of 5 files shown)


techman83:
I have some work to split up the repo to have historical data on separate branches. To make the payload smaller. But some issues with the testing held me up, I should get back to it KSP-CKAN/NetKAN-bot#58

DasSkelett:
Can confirm that it is a lot faster:
(Tested with `time ckan update`)
Before: ~8.3s
After:    ~4.8s
-> 1.7 times faster

PR #2685, opened 2019-02-16
Force redraw of versions ListView on Mono

### LINKED ISSUE #2684: Mod info panel versions tab: Column header invisible
Background
----------

**CKAN Version:** 
Current HEAD (62da77df)
and current release (v1.25.4)

**Operating System:** 
Kubuntu 18.04.1 -> mono 5.18.0.240

Problem
-------

~~1) Open "Metadata" tab in mod info panel (or any one besides the "Versions" tab)
2) Select a mod (don't keep the mod selected which is selected at startup, a different one!)
3) Change to "Versions" tab~~
1) Select any mod **after startup**
2) Open "Versions" tab in the mod info panel -> everything is still fine
3) Change to another tab
4) Select different mod
5) Change back to "Versions" tab

See the column headers.

**What happened instead?**
invisible and unclickable column headers
ckan1

If you keep the versions tab **open** and select a different mod, the column headers reappear.
ckan2

HebaruSan:
## Problem

If you do this under Mono:

1. Open GUI
2. Click any mod
3. Open the Versions tab in mod info
4. Open any other tab in mod info
5. Click another mod
6. Open the Versions tab

... then the column headers won't be drawn:

screenshot

## Cause

Unclear, but CKAN's GUI doesn't do anything particularly tricky with these controls. There's a `TabControl` with a `TabPage` that has a `ListView`, and we mostly let the runtime worry about handling them.

It looks a Mono bug, but I wasn't able to pin down exactly how or where. So we'll be going with a workaround.

## Changes

Now when you switch to the Versions tab, we call `VersionsListView.EndUpdate()` on Mono. This ensures that drawing is enabled and forces a full redraw. In my testing this addresses the problem with drawing the headers.

@DasSkelett, can you please test this and report whether it's fixed for you?

Fixes #2684.

### Side change

The method used by `ModInfoIndexChanged` of comparing `SelectedIndex` to `TabIndex` wasn't robust and required properties to be set up just so; for example, the `TabIndex` of the Versions tab was 1, while its `SelectedIndex` value was 3, so extending that pattern would not have worked without updating the `TabIndex` value. Add to this the complication that the runtime can change `TabIndex` values automatically, and it's just not good.

Now we key off of `TabPage.Name`, which is reliably known and can be compared to constant values.

### DIFF
--- a/GUI/MainAllModVersions.cs
+++ b/GUI/MainAllModVersions.cs
@@ -35,6 +35,15 @@ namespace CKAN
             }
         }
 
+        /// <summary>
+        /// Make the ListView redraw itself.
+        /// Works around a problem where the headers aren't drawn when this tab activates.
+        /// </summary>
+        public void ForceRedraw()
+        {
+            VersionsListView.EndUpdate();
+        }
+
         public GUIMod SelectedModule
         {
             set
--- a/GUI/MainModInfo.cs
+++ b/GUI/MainModInfo.cs
@@ -107,13 +107,6 @@ namespace CKAN
             Main.Instance.ResetFilterAndSelectModOnList(e.Node.Name);
         }
 
-        private void ModuleRelationshipType_SelectedIndexChanged(object sender, EventArgs e)
-        {
-            GUIMod module = SelectedModule;
-            if (module == null) return;
-            UpdateModDependencyGraph(module);
-        }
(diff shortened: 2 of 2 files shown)


DasSkelett:
Works fine :+1:

PR #2690, opened 2019-02-26
Customisable columns in GUI Modlist

### LINKED ISSUE #2687: Feature Request: Customise columns in GUI Modlist
Background
----------
Several changes have added columns to the GUI modlist, providing more information for users. But some users have limited screen real estate and may not be interested in seeing all the new columns (perhaps even some of the older, original columns).

Proposal
---------
CKAN should have a way to customise the columns that appear in the modlist, turning specific columns on or off. We could then make use of this feature to activate certain columns only when there is data (such as the Update/Replace columns)

DasSkelett:
**This PR adds the option to hide columns in the modlist.** as asked by @politas in #2687.

This happens via a right click context menu on the header row of the modlist:
ckan1

The columns are hidden with `column.Visibility = false`, which seems to work perfectly fine in my testing. This doesn't need a reload of the modlist or anything, so it's fast.
On every change of the visibility of a column it is saved to the GUIconfig, which means it's persistend across ckan restarts.

The `Update` and the `Replace` checkbox columns are hidden by default, and only shown if an update / a replacement is available. I could only test updates yet, but replacements should work the same.
Once all updates / replacements have been done, the columns are hidden again.
This all happens in `UpdateModsList()`.

Furthermore, the `Installed version` and the `Install date` columns are hidden with the `Not Installed` filter setting selected, there's no sense in showing them. This happens *without* a GUIconfig change, so once the filter is changed again, those two columns reappear (or stay hidden) based on the previous setting.

many columns hidden updates available

Closes #2687 

---
Small sidefix: if you right-clicked on the modlist header bar before, the context menu for the mod rows opened, despite it shouldn't. It's fixed now by not assigning `ModList.ContextMenuStrip = ModListContextMenuStrip`. Instead, the new column selection menu opens, obviously.

### DIFF
--- a/GUI/MainModList.cs
+++ b/GUI/MainModList.cs
@@ -278,9 +278,7 @@ namespace CKAN
             // Write it to the configuration, else they are hidden agian after a filter change.
             // After the update / replacement, they are hidden again.
             ModList.Columns[1].Visible = has_any_updates;
-            configuration.VisibleColumns[1] = has_any_updates;
             ModList.Columns[2].Visible = has_any_replacements;
-            configuration.VisibleColumns[2] = has_any_replacements;
 
             AddLogMessage("Updating tray...");
             UpdateTrayInfo();
@@ -397,7 +395,8 @@ namespace CKAN
 
             if (col != null)
             {
-                configuration.VisibleColumns[col.Index - 3] = col.Visible = !clickedItem.Checked;
+                col.Visible = !clickedItem.Checked;
+                configuration.SetColumnVisibility(col.Name, !clickedItem.Checked);
                 if (col.Index == 0)
                 {
                     InstallAllCheckbox.Visible = col.Visible;
--- a/GUI/MainModList.cs
+++ b/GUI/MainModList.cs
@@ -245,6 +245,7 @@ namespace CKAN
             AddLogMessage("Updating filters...");
 
             var has_any_updates = gui_mods.Any(mod => mod.HasUpdate);
+            var has_any_replacements = gui_mods.Any(mod => mod.HasReplacement);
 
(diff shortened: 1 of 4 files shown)


HebaruSan:
Let's remove the update/replace columns from the menu since we're managing them automatically.

DasSkelett:
I'm going to set the `Installed` checkbox column (the first one) to always visible then true.
Else the code will get a bit messy, because I access the columns via their index (which is taken from the index of the clicked button).

DasSkelett:
Removed the buttons for the first three columns in the context menu.
Beware: If you've already run the previous commit, make sure to either delete the whole GUIConfig.xml or only the `VisibleColumns` array, because I shortened the array, and "using" the old one will be buggy.

HebaruSan:
This looks pretty good. Created DasSkelett/CKAN#3 with some code style and UI ideas.

DasSkelett:
For the record: @HebaruSan's suggestions are merged and included. Thanks :)

GUI/MainModList.cs
```diff
@@ -273,6 +274,14 @@ private void _UpdateModsList(bool repo_updated, IEnumerable<ModChange> mc)
 
             UpdateFilters(this);
 
+            // Hide update and replacement columns if not needed.
+            // Write it to the configuration, else they are hidden agian after a filter change.
+            // After the update / replacement, they are hidden again.
+            ModList.Columns[1].Visible = has_any_updates;
+            configuration.VisibleColumns[1] = has_any_updates;
```
HebaruSan: Configuration probably should be modified here if we're no longer including update/replace in the list.

HebaruSan:
DasSkelett/CKAN#4 created with a suggestion for the config.

PR #2699, opened 2019-03-16
Right click copy option for mod info links

### LINKED ISSUE #2698: Copy / drag&drop for mod links (Feature Request)
Hello,

would it be possible to add a "right click, copy", a "mark" or a "drag & drop" feature to the links in the mod description?
Background: On one of my systems I don't want to open KSP links in my default browser. I have a second browser open, but it's impossible to open the link somehow in the second browser. In the end I always end up googling the link, even it's direct in front of me.

The really only thing you can do is click the link and this fires up the default browser.
Usually (with other applications) I can just drag a link and drop it in the tab bar of the second browser and it opens fine.
If this is to much work, would it maybe be possible to add a right click "copy" or at least make the link markable with the mouse, so that CTRL-C sould work?

Best Regards
Jebs_SY

PS: Not sure how to mark this as feature request?!

HebaruSan:
## Motivation

It may sometimes be desirable to access a mod's home page URL without opening it in the default browser, for instance you might want to open it in another browser or paste it into a forum post.

Currently right clicking the links works just like left clicking, the URL is opened in the system default browser.

screenshot

## Changes

Now if you right click one of the mod info links, a context menu appears with a "Copy link address" option. Clicking this option copies the URL to the clipboard.

image

Fixes #2698.

### DIFF
--- a/GUI/AboutDialog.cs
+++ b/GUI/AboutDialog.cs
@@ -13,27 +13,27 @@ namespace CKAN
 
         private void linkLabel1_LinkClicked(object sender, LinkLabelLinkClickedEventArgs e)
         {
-            System.Diagnostics.Process.Start("https://github.com/KSP-CKAN/CKAN/blob/master/LICENSE.md");
+            Util.HandleLinkClicked("https://github.com/KSP-CKAN/CKAN/blob/master/LICENSE.md", e);
         }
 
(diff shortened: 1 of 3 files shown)


DasSkelett:
Works on Windows too:
image
What about adding it to the About dialog?

HebaruSan:
Rebased and added About window support.

PR #2709, opened 2019-03-31
Ignore spaces and special chars in mod search

### LINKED ISSUE #2708: Ignore spaces and hyphen in the filters by name
We have CamelCase names: BetterBurnTime, etc
pythen names: Kerbal-Konstructs, etc
space names: Community Category Kit, etc

and they are not replaceable, so, if I start typing: 
"Kerbal Kon" or "KerbalKon" I will not get Kerbal-Konstructs,  
"Better Bu" ≠ "BetterBurnTime", 
"CategoryKit" ≠ Community Category Kit, and so on...

Sometimes it is very annoying.

DasSkelett:
## Problem
If you want to search the mod list (in GUI, commandline or consoleUI), you have to know the name/identifier/author/description exactly, even spaces and special character.
There are **a lot** of mods with spaces and special characters in their name, sometimes even in the identifier indexed in the CKAN.
Examples:
* aati-flags (name) - AATI-Flags (identifier)
(special char)
* [1.3] Capsule Corporation Endeavour (Mars Base Camp)
(special chars and spaces)
* BetterBurnTime
(no spaces)
* and on and on

It's nearly impossible to find a mod on the first try right now, if you don't know the exact name including all spaces, hyphens, brackets, dots...

## Solution
On startup, more precisely during the deserialization of `CkanModule`s on load of registry, 'Searchables' are calculated for every mo and mod version, and saved during runtime in `CkanModule`. The don't get serialized!
Those 'Searchables' are
```
SearchableIdentifier
SearchableName
SearchableAbstract
SearchableDescription
SearchableAuthors (this one is a List<string>)
```
and represent special-characterless versions of their respective counterparts.
This means no spaces, no hyphens, no brackets and so on. Only alphanumerics remain.
They are calculated by removing each non-alphanumeric char from the source strings (regex `[^a-zA-Z0-9]`.
If a mod search is done, the search term gets treated the same way.

This applies to all three interfaces, for all search types (author, description, name...).

---
Furthermore, I've improved the console search a bit. There's now also a `--detail` option (similar to the one in `available`, see #2286) to list more than just the identifier, including the compatibility for KSP versions of the last mod version.
The new `--all` option lists incompatible mods too. Without `--detail`, they're mixed up in the compatibles, but with they are printed as two separate lists.
Third, the `--author <authorName>` option allows to search for mods by a specific author. An empty `search_term` is possible, to list all the mods of an author.

So the search command can now range from
`ckan search kerbal` to
`ckan search --all --detail --author linuxguru RecycledPartsFTmNImprovedAtomicRockets`,
something for everyone.

---
## tl;dr:
* aatiflags or aati flags finds aati-flags
* capsulecorporation endeavour Mars Base Camp finds [1.3] Capsule Corporation Endeavour (Mars Base Camp)
* better burn time finds BetterBurnTime
* `ckan search --all --detail --author linuxguru RecycledPartsFTmNImprovedAtomicRock` finds RecycledPartsFTmNImprovedAtomicRockets by LGG in commandline
* ...

Closes #2708

### DIFF
--- a/Cmdline/Action/Search.cs
+++ b/Cmdline/Action/Search.cs
@@ -16,60 +16,147 @@ namespace CKAN.CmdLine
 
         public int RunCommand(CKAN.KSP ksp, object raw_options)
         {
-            SearchOptions options = (SearchOptions) raw_options;
+            SearchOptions options = (SearchOptions)raw_options;
 
             // Check the input.
-            if (String.IsNullOrWhiteSpace(options.search_term))
+            if (String.IsNullOrWhiteSpace(options.search_term) && String.IsNullOrWhiteSpace(options.author_term))
             {
                 user.RaiseError("No search term?");
 
                 return Exit.BADOPT;
             }
 
-            var matching_mods = PerformSearch(ksp, options.search_term);
+            List<CkanModule> matching_compatible = PerformSearch(ksp, options.search_term, options.author_term, false);
+            List<CkanModule> matching_incompatible = new List<CkanModule>();
+            if (options.all)
+            {
+                matching_incompatible = PerformSearch(ksp, options.search_term, options.author_term, true);
+            }
 
             // Show how many matches we have.
-            user.RaiseMessage("Found " + matching_mods.Count().ToString() + " mods matching \"" + options.search_term + "\".");
+            if (options.all && !String.IsNullOrWhiteSpace(options.author_term))
+            {
+                user.RaiseMessage("Found {0} compatible and {1} incompatible mods matching \"{2}\" by \"{3}\".",
+                    matching_compatible.Count().ToString(),
+                    matching_incompatible.Count().ToString(),
+                    options.search_term,
+                    options.author_term);
+            }
+            else if (options.all && String.IsNullOrWhiteSpace(options.author_term))
+            {
+                user.RaiseMessage("Found {0} compatible and {1} incompatible mods matching \"{2}\".",
+                    matching_compatible.Count().ToString(),
+                    matching_incompatible.Count().ToString(),
+                    options.search_term);
+            }
+            else if (!options.all && !String.IsNullOrWhiteSpace(options.author_term))
+            {
+                user.RaiseMessage("Found {0} compatible mods matching \"{1}\" by \"{2}\".",
+                    matching_compatible.Count().ToString(),
+                    options.search_term,
+                    options.author_term);
+            }
+            else if (!options.all && String.IsNullOrWhiteSpace(options.author_term))
+            {
+                user.RaiseMessage("Found {0} compatible mods matching \"{1}\".",
+                    matching_compatible.Count().ToString(),
+                    options.search_term);
+            }
 
             // Present the results.
-            if (!matching_mods.Any())
+            if (!matching_compatible.Any() && (!options.all || !matching_incompatible.Any()))
             {
                 return Exit.OK;
             }
 
-            // Print each mod on a separate line.
-            foreach (CkanModule mod in matching_mods)
+            if (options.detail)
             {
-                user.RaiseMessage(mod.identifier);
+                user.RaiseMessage("Matching compatible mods:");
+                foreach (CkanModule mod in matching_compatible)
+                {
+                    user.RaiseMessage("* {0} ({1}) - {2} by {3} - {4}",
+                        mod.identifier,
+                        mod.version,
+                        mod.name,
+                        mod.author == null ? "N/A" : String.Join(", ", mod.author),
+                        mod.@abstract);
+                }
+
+                if (matching_incompatible.Any())
+                {
+                    user.RaiseMessage("Matching incompatible mods:");
+                    foreach (CkanModule mod in matching_incompatible)
+                    {
+                        Registry.GetMinMaxVersions(new List<CkanModule> { mod } , out _, out _, out var minKsp, out var maxKsp);
+                        string kspVersion = Versioning.KspVersionRange.VersionSpan(minKsp, maxKsp).ToString();
+
+                        user.RaiseMessage("* {0} ({1} - {2}) - {3} by {4} - {5}",
+                            mod.identifier,
+                            mod.version,
+                            kspVersion,
+                            mod.name,
+                            mod.author == null ? "N/A" : String.Join(", ", mod.author),
+                            mod.@abstract);
+                    }
+                }
+            }
+            else
+            {
+                List<CkanModule> matching = matching_compatible.Concat(matching_incompatible).ToList();
+                matching.Sort((x, y) => string.Compare(x.identifier, y.identifier, StringComparison.Ordinal));
+
+                foreach (CkanModule mod in matching)
+                {
+                    user.RaiseMessage(mod.identifier);
+                }
             }
 
             return Exit.OK;
         }
 
         /// <summary>
-        /// Searches for the term in the list of available modules for the ksp instance. Looks in name, identifier and description fields.
+        /// Searches for the term in the list of compatible or incompatible modules for the ksp instance.
+        /// Looks in name, identifier and description fields, and if given, restricts to authors matching the author term.
         /// </summary>
-        /// <returns>List of mathcing modules.</returns>
+        /// <returns>List of matching modules.</returns>
         /// <param name="ksp">The KSP instance to perform the search for.</param>
         /// <param name="term">The search term. Case insensitive.</param>
-        public List<CkanModule> PerformSearch(CKAN.KSP ksp, string term)
+        public List<CkanModule> PerformSearch(CKAN.KSP ksp, string term, string author = null, bool searchIncompatible = false)
         {
+            // Remove spaces and special characters from the search term.
+            term   = String.IsNullOrWhiteSpace(term)   ? string.Empty : CkanModule.nonAlphaNums.Replace(term, "");
+            author = String.IsNullOrWhiteSpace(author) ? string.Empty : CkanModule.nonAlphaNums.Replace(author, "");
+
             var registry = RegistryManager.Instance(ksp).registry;
-            return registry
-                .Available(ksp.VersionCriteria())
-                .Where((module) =>
+
+            if (!searchIncompatible)
             {
-                // Extract the description. This is an optional field and may be null.
-                string modDesc = string.Empty;
-
-                if (!string.IsNullOrEmpty(module.description))
+                return registry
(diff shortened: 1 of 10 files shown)


politas:
Could we speed this process up by adding fields to GUIMod with the filtered properties, so we aren't constantly applying multiple regexes to every line in GUIModList every time we refresh? This data isn't changing dynamically based on user input.

DasSkelett:
If we add it to GUImod, we can use the feature only for the GUI. But generally I support the idea.
Don't know if this makes sense in CkanModule? That would be used by every UI.

Also we would have to save it for the name, identifier, author and description, maybe as an array?
A wild idea: concatenate all 4 strings together in one field, and only have one search box for everything.

yalov:
one search box and 3 checkboxes, instead of the 3 search boxes?

HebaruSan:
- Don't list out all punctuation characters in the regex, just negate the alphanums (see next bullet point's code citation for an example)
- The regex should be static and compiled for performance:
  https://github.com/KSP-CKAN/CKAN/blob/5331bf9344f6729a6b96f52bea45bdec245a3bb3/ConsoleUI/Toolkit/ConsoleListBox.cs#L427
- It makes sense to precalculate as we do with `GUIMod.Abbreviation`, and I guess I'm OK with putting it in `CkanModule`
- A concatenated string wouldn't work for GUI, since you couldn't use it to search _just_ the name or author. How about a simple set of supplemental properties named to make their purpose clear (and obviously excluded from the JSON serialization)?
  - `SearchableName`
  - `SearchableIdentifier`
  - `SearchableAbstract`
  - `SearchableDescription`
  - `SearchableAuthors`
- (Old issue) Does Cmdline completely fail to consider the `Abstract`? That's where most modules have their descriptions. And GUI ignores the description, looking *only* at the abstract?? Consistency would be nice here.

DasSkelett:
Good suggestions. When to calculate the 'Searchables'?
I'm not happy increasing the startup time even more. Maybe do it in the background, and deactivate the search boxes until it's done?

HebaruSan:
I wouldn't attempt to optimize this further until after you've measured the performance using a static, compiled regex. Some portion of the slowdown you saw before is due to re-creating and compiling the regex over a thousand times, and it could be a large portion.

DasSkelett:
Update: Now I use a static, compiled, negated (non alphanumerics) regex > performance is better.
Not noticable in GUI or consoleUI, a `time ckan search` tells me 4.8s before vs. 5.1s after.

The calculations are done in the constructor of `CkanModule` and on deserialisation of `AvialableModule`. Did I miss any possible source of `CkanModule`s that needs the calculations added?

All UIs now search the abstract as well as the description.

DasSkelett:
I want to put the calculations in a separate function in `CkanModule` which also gets called in `AvailableModule` and maybe remove all but one of the regexes to simplify possible future changes, so hang on please.

DasSkelett:
The calculations are now in a separate method: `CalculateSearchables()`.
Every use of the regex is also done via `CkanModule.nonAlphaNums`, the other ones were removed.
It's better if it's only in one place.
The calculations previously done in the deserialisation of `AvailableModule` are now done via a call to `CalculateSearchables()` in the deserialisation of  `CkanModule`. That's where it belongs to, and is no problem, because a deserialisation of `AvailableModule` implies a deserialisation of  `CkanModule`.

ConsoleUI/ModListScreen.cs
```diff
@@ -48,18 +48,17 @@ public ModListScreen(KSPManager mgr, bool dbg)
...
-                                if (auth.IndexOf(authorFilt, StringComparison.CurrentCultureIgnoreCase) == 0) {
-                                    return true;
-                                }
-                            }
+                        } else {
+                            // Remove special characters from search term
+                            authorFilt = CkanModule.nonAlphaNums.Replace(authorFilt, "");
+                            return m.SearchableAuthors.IndexOf(authorFilt, StringComparison.CurrentCultureIgnoreCase) == 0;
                         }
-                        return false;
```
HebaruSan: I'm guessing that this `return false;` was there for a reason. Why was it removed?

GUI/GUIMod.cs
```diff
@@ -186,8 +199,9 @@ public GUIMod(string identifier, IRegistryQuerier registry, KspVersionCriteria c
             }
 
             // If we have a homepage provided, use that; otherwise use the spacedock page, curse page or the github repo so that users have somewhere to get more info than just the abstract.
-
             Homepage = "N/A";
+
+            SearchableIdentifier = CkanModule.nonAlphaNums.Replace(Identifier, "");
```
HebaruSan: Why ware we calculating this one differently? Can't we just set it to `mod.SearchableIdentifier`?

Core/Types/CkanModule.cs
```diff
@@ -552,6 +566,21 @@ public CkanModule(string json, IGameComparator comparator)
...
+        /// <summary>
+        /// Calculate the mod properties used for searching via Regex.
+        /// </summary>
+        public void CalculateSearchables()
+        {
+            SearchableIdentifier  = identifier  == null ? string.Empty : CkanModule.nonAlphaNums.Replace(identifier, "");
+            SearchableName        = name        == null ? string.Empty : CkanModule.nonAlphaNums.Replace(name, "");
+            SearchableAbstract    = @abstract   == null ? string.Empty : CkanModule.nonAlphaNums.Replace(@abstract, "");
+            SearchableDescription = description == null ? string.Empty : CkanModule.nonAlphaNums.Replace(description, "");
+            SearchableAuthors     = author      == null ? string.Empty : CkanModule.nonAlphaNums.Replace(String.Join("", author), "");
```
HebaruSan: I'm not sure concatenating the authors like this is what we want. If my authors list is `[ "abc", "def", "ghi" ]`, then this would match on "bcde", which isn't part of any author's name and depends on the ordering of the list. Can we apply the regex to each author's name individually but keep them as separate list elements?

DasSkelett: Do you think there's a high chance that someone might encounter it?
Even if, this probably only results in very few false positives.

HebaruSan: Well if someone wants to search for droric with "dro", this will cause Asteroid Day to match incorrectly:

image

But in general why not do it right since we know about it?

DasSkelett: Okay, changing it :+1:

DasSkelett:
@HebaruSan we only print the identifier of the search results in the command line.
https://github.com/KSP-CKAN/CKAN/blob/16eeaf1d5176a9304f7195a49f39e433148ccd01/Cmdline/Action/Search.cs#L40-L44
Wouldn't it make sense to list the mod names too? They are a bit user friendlier, I think.

HebaruSan:
Yes, that would be nicer. The issue that came up last time in #2286 is that people might be running the CmdLine client in scripts that expect the old output, so any new format should probably be controlled by a new command line option.

DasSkelett:
Okay. For consistency I would do it like #2286, via a --detail flag and with identifier, latest version and full name.
Is the abstract needed for the search command? It seems like you know what a mod does if you search for it.

HebaruSan:
Having the abstract could be nice if you're searching for a particular kind of mod rather than identifiers. E.g., you might want to find planet packs and then pick one based on its description.

HebaruSan:
As long as you're going all-out for this, should there also be a way to include incompatible mods in the CmdLine search? It might be nice to see that I could install more planet packs if I marked 1.4 as compatible, for example.

DasSkelett:
Support this. What about naming it `--all`? Also I think we should include the compatible ksp version for these.
By the way, is there a way to search for mods of a single author in the cmdline? Didn't find anything.
Could imagine it like `ckan search "@HebaruSan astro"`, including the @. Due to the regex, the author could only be placed at the beginning of the search term.

HebaruSan:
Or `ckan search --author Hebarusan astro`, might as well take advantage of the command line parameter framework when we can.

DasSkelett:
I've updated the post with infos about all those shiny new options for the commandline search.

Cmdline/Action/Search.cs
```diff
@@ -16,60 +16,150 @@ public Search(IUser user)
...
-            var matching_mods = PerformSearch(ksp, options.search_term);
+            List<CkanModule> matching_compatible = PerformSearch(ksp, options.search_term, options.author_term, false);
+            List<CkanModule> matching_incompatible = new List<CkanModule>();
+            if (options.all)
+            {
+                matching_incompatible = PerformSearch(ksp, options.search_term, options.author_term, true);
+            }
+
+            // The string builder doesn't like null s in the argument list.
+            // options.search_term = options.search_term ?? string.Empty;
```
HebaruSan: Was this not needed? It's still commented.

DasSkelett: Oh yeah, forgot to remove it, thanks.
Is not needed, the string builder indeed actually doesn't care about null strings,
it was that the `user.RaiseMessage("...{0}...{1}...{2}...")` had wrong numbers, like `user.RaiseMessage("...{1}...{3}...{4}...")`. This also throws a `NullReferenceException`.

Cmdline/Action/Search.cs
```diff
@@ -16,60 +16,150 @@ public Search(IUser user)
...
+                        mod.author == null ? "N/A" : String.Join(", ", mod.author),
+                        mod.@abstract);
+                }
+
+                if (matching_incompatible.Any())
+                {
+                    user.RaiseMessage("Matching incompatible mods:");
+                    foreach (CkanModule mod in matching_incompatible)
+                    {
+                        Registry.GetMinMaxVersions(new List<CkanModule> { mod } , out _, out _, out var minKsp, out var maxKsp);
```
HebaruSan: Oh, I didn't know you could discard an `out` param with "_". Cool; there are some places where I'll want to use that (mostly other spots that call `GetMinMaxVersions`).

DasSkelett: Yep, searched for it because I wanted to know if it's possible. There are quite a few answers on stackoverflow mentioning them if you search for `c# discard out parameter` or so.

Cmdline/Action/Search.cs
HebaruSan: Well the code that's *not* commented looks very good to me now. I'd say ready for approve/merge after these lines are removed. Nice work!

DasSkelett: Thanks! Is removed now.

PR #2725, opened 2019-04-11
Save restore position when minimized

### LINKED ISSUE #2717: Window opens offscreen if previously closed while minimized
Background
----------

**CKAN Version:** 
1.26.0

**KSP Version:** 
1.6.1

**Operating System:** 
Windows 10 Pro 1809

No

Problem
-------

Closed CKAN by right-clicking the tasbar icon while it was minimized.

For CKAN to close without issues.

**What happened instead?**
CKAN closes fine but on the next launch the window is nowhere to be seen and the maximize shortcut (Win+Up) does not work.

**Screenshots:**
GUIConfig.xml under Kerbal Space Program\CKAN looks like this when the issue occurs. Setting the highlighted values to 0 fixes the issue.

Annotation 2019-04-04 171001
None

HebaruSan:
## Problem

If you:

1. Open CKAN with the tray icon enabled
2. Minimize it
3. Right click the tray icon
4. Exit

... then the following is saved to `GUIConfig.xml`:

config

And the next time CKAN runs, its window will not appear.

## Cause

Apparently Windows moves minimized windows to (-32000,-32000).

https://stackoverflow.com/questions/1478765/location-coordinates-on-computer-showing-x-32000-y-32000

So when we exit and save the `Location` property while minimized, we're saving that value.

## Changes

Now we ignore `Location` if the window state isn't normal (i.e., we're minimized or maximized), and instead we check `RestoreBounds.Location`, which is where the window's former location to which it will be returned lives. This mirrors how we handle the window size on the next line of code (though in that case the problem was with maximization).

Fixes #2717.

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -307,9 +307,16 @@ namespace CKAN
             base.OnFormClosed(e);
         }
 
-        protected override void OnLoad(EventArgs e)
+        private void SetStartPosition()
         {
-            if (configuration.WindowLoc.X == -1 && configuration.WindowLoc.Y == -1)
+            Screen screen = Util.FindScreen(configuration.WindowLoc, configuration.WindowSize);
+            if (screen == null)
+            {
+                // Start at center of screen if we have an invalid location saved in the config
+                // (such as -32000,-32000, which Windows uses when you're minimized)
+                StartPosition = FormStartPosition.CenterScreen;
+            }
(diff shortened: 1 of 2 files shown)


HebaruSan:
I'm thinking it should also recover from the issue once the file already has -32000 in it. Please hold off on merging till that's added.

HebaruSan:
Updated to default to center of screen when the saved position isn't on any of the known screens.

Issue #2703, opened 2019-03-25
[PRE-RELEASE 2] "Add available updates" crashes CKAN


4x4cheesecake:
Background
----------

**CKAN Version:** 
1.26.0 [PRE-RELEASE 2]

**KSP Version:** 
1.6.1

**Operating System:** 
Win10 x64

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** 
no

Problem
-------

**What steps did you take in CKAN?**
For testing purposes, I've installed an old version of a mod via the "Versions" tab. Then, I tried to update to the latest version by clicking on "Add available updates".
Tried the same with actual updates on an older KSP install (also 1.6.1)

**What did you expect to happen?**
CKAN should add all availabe updates.

**What happened instead?**
CKAN crashed.
This is replicable with any mod I've tried. (Works fine in 1.25.4 though.)
I also want to mention that it works fine when the update is selected manually, it is just the "Add available updates" button which causes the crash.

**Screenshots:**

![image](https://user-images.githubusercontent.com/40007416/54946737-0f4dd500-4f39-11e9-8803-ff7d34709afd.png)

**Full error message:**
 
```
Informationen über das Aufrufen von JIT-Debuggen
anstelle dieses Dialogfelds finden Sie am Ende dieser Meldung.

************** Ausnahmetext **************
System.InvalidOperationException: Die aktuelle Zelle kann nicht auf eine unsichtbare Zelle festgelegt werden.
   bei System.Windows.Forms.DataGridView.set_CurrentCell(DataGridViewCell value)
   bei CKAN.Main.MarkAllUpdatesToolButton_Click(Object sender, EventArgs e)
   bei System.Windows.Forms.ToolStripItem.RaiseEvent(Object key, EventArgs e)
   bei System.Windows.Forms.ToolStripMenuItem.OnClick(EventArgs e)
   bei System.Windows.Forms.ToolStripItem.HandleClick(EventArgs e)
   bei System.Windows.Forms.ToolStripItem.HandleMouseUp(MouseEventArgs e)
   bei System.Windows.Forms.ToolStrip.OnMouseUp(MouseEventArgs mea)
   bei System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   bei System.Windows.Forms.Control.WndProc(Message& m)
   bei System.Windows.Forms.ToolStrip.WndProc(Message& m)
[...]

PR #2727, opened 2019-04-13
Fix home/end, url handler, typing nav, and dependency double click

HebaruSan:
## Problem

If you do any of the following:

- Press Home on the mod list
- Press End on the mod list
- Invoke the URL handler
- Start typing a mod name on the mod list
- Double click a mod in the mod info dependency tree

... this exception occurs:

```
System.InvalidOperationException: Die aktuelle Zelle kann nicht auf eine unsichtbare Zelle festgelegt werden. (Rough translation: the current cell can't be set to an invisible cell)
   bei System.Windows.Forms.DataGridView.set_CurrentCell(DataGridViewCell value)
   bei CKAN.Main.ModList_KeyDown(Object sender, KeyEventArgs e)
   bei System.Windows.Forms.Control.OnKeyDown(KeyEventArgs e)
   bei System.Windows.Forms.DataGridView.OnKeyDown(KeyEventArgs e)
   bei System.Windows.Forms.Control.ProcessKeyEventArgs(Message& m)
   bei System.Windows.Forms.DataGridView.ProcessKeyEventArgs(Message& m)
   bei System.Windows.Forms.Control.WmKeyChar(Message& m)
   bei System.Windows.Forms.Control.WndProc(Message& m)
   bei System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

## Cause

These all work by setting `ModList.CurrentCell` to a row's third cell, for reasons explained here:

https://github.com/KSP-CKAN/CKAN/blob/3bc537637e8f4452c9787699045d61b4d869061f/GUI/Main.cs#L1055-L1056

However, as of #2671, the third cell is Replace instead of Name, and as of #2690, it's hidden most of the time. A hidden cell cannot be the current cell, so an exception is thrown. This is essentially the same problem as #2703, but it's not specific to the mark-all-updates operation.

## Changes

Now we look for usable columns in this order:

1. The current column, if any, to avoid scrolling horizontally
2. The first visible non-checkbox column, if any
3. The Installed checkbox column (always visible fallback)

In addition, the fix for #2704 is modified to use this same logic. I think the original reasoning from the comment quoted above applies here, and so we should avoid setting focus to checkboxes when we can.

Fixes KSP-CKAN/NetKAN#7108.

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -564,7 +564,7 @@ namespace CKAN
             if (Main.Instance.configuration.AutoSortByUpdate)
             {
                 // set new sort column
-                var new_sort_column = ModList.Columns[1];
+                var new_sort_column = ModList.Columns[UpdateCol.Index];
                 var current_sort_column = ModList.Columns[configuration.SortByColumnIndex];
 
                 // Reset the glyph.
@@ -573,7 +573,7 @@ namespace CKAN
                 UpdateFilters(this);
 
                 // Select the top row and scroll the list to it.
-                ModList.CurrentCell = ModList.Rows[0].Cells[1];
+                ModList.CurrentCell = ModList.Rows[0].Cells[SelectableColumnIndex()];
             }
 
             ModList.Refresh();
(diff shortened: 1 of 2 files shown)


DasSkelett:
Why does it show my name for the label remove (and in other occasions)?
Is this because I am the last one who logged in on the Waffel page?

HebaruSan:
Yeah, that Waffle thing likes to make automatic changes for reasons I don't understand. It did it in my name for a while when I first joined as well.

HebaruSan:
Welcome to the team by the way, @DasSkelett! I'm excited about having a more... active... collaborator. :)

DasSkelett:
Thanks. You tell me something, that was quite a hot start with you going on vacation, and @politas releasing a new major version, and all that exactly in those two weeks where I had training every two days for a fire brigade performance test....

PR #2729, opened 2019-04-13
Auto size yes no dialog

### LINKED ISSUE #2718: Text cut off in info window
Background
----------

**CKAN Version:** 
1.26.0

**KSP Version:** 
1.6.1

**Operating System:** 
Win 10 x64

No

Problem
-------

Launched CKAN and selected an "older" KSP install

(not relevant)

**What happened instead?**
CKAN displayed a warning/info window about some changed metadata but the text exceeds the window size and is cut off. The window is not resizeable.

**Screenshots:**

grafik

HebaruSan:
## Problem

If the `YesNoDialog` needs to display a long string, it can be truncated:

screenshot

## Cause

It's just a statically sized form, so only short strings will fit.

## Changes

Now the form is resizable, so if it's unreadable, the user has the option to resize it. The controls have the `Anchor` property set to make them move correctly.

Now when a message is set, the form uses `Graphics.MeasureString` to determine the required height of the label, and uses this to set the size of the form. This ensures that the full message will be visible.

`YesNoDialog` no longer inherits from `FormCompatibility` because it wasn't calling `ApplyFormCompatibilityFixes` anyway. This will have no effect.

Fixes #2718.

### DIFF
--- a/GUI/YesNoDialog.Designer.cs
+++ b/GUI/YesNoDialog.Designer.cs
@@ -37,6 +37,9 @@
             // 
             // panel1
             // 
+            this.panel1.Anchor = ((System.Windows.Forms.AnchorStyles)((((System.Windows.Forms.AnchorStyles.Top | System.Windows.Forms.AnchorStyles.Bottom)
+            | System.Windows.Forms.AnchorStyles.Left)
+            | System.Windows.Forms.AnchorStyles.Right)));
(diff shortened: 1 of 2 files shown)


HebaruSan:
One easy way is to select a bunch of modules to install and then try to quit without applying changes:

image

PR #2732, opened 2019-04-16
Add self to opt-out

cake-pie:
To protect against unwanted proliferation of hot code

### DIFF
--- a/policy/Opt-out-list.md
+++ b/policy/Opt-out-list.md
@@ -10,3 +10,4 @@ RoverDude | BobPalmer | Self-managing metadata
 Sarbian | Sarbian | Self-managing metadata
 Angel-125 | Angel-125 | Installation problems and request by author
 DefiantZombie | DefiantZombie | request by author
+cakepie | cake-pie | Guard against hot code unsuitable for general release 


HebaruSan:
Can we ask why? I'm not sure what "proliferation of hot code" means.

cake-pie:
"Hot code" in the sense of "this just got pushed/merged with only the barest minimum of testing" and liable to be unstable or broken. This is particularly true for new mods currently in my pipeline, that are far from a beta state, let alone suitable for general release.

I may make them available for download via the forums/github, but only with clear warnings to the effect that they are not to be considered production quality.  The intended audience is other modders and tech-savvy power-users.

As such I would like to forestall the possibility of any well-intentioned CKAN user adding my mods to the index before I am ready to provide support for a larger, generally less tech-savvy user base.

HebaruSan:
Fair enough. For what it's worth, current de facto practice is to ask author permission before indexing anyway. It sounds like you're not as strongly antagonistic to CKAN as the typical member of that list, but I can understand wanting to ensure that prototypes don't end up getting wide distribution. Thanks for responding.

PR #2734, opened 2019-04-18
Fix Newly Compatible filter with repo autoupdate

### LINKED ISSUE #2716: Newly Compatible shows all mods, not just those newly compatible
Background
----------

**CKAN Version:**
1.26.0

**KSP Version:**
1.6.1

**Operating System:**
Windows 10 Pro 64 bit

No

Problem
-------
Filter "Newly Compatible" shows same number of mods as "All".

1. Started ckan.exe
2. Let it refresh repository of mods
3. Pressed Filter button

What used to happen in previous version: Newly Compatible only showing those mods that have recently (since last refresh) become compatible with the current version of KSP.

**What happened instead?**
Filter shows all mods.

**Screenshots:**

image

None

HebaruSan:
## Problem

If you have repository auto updates enabled, then the Newly Compatible filter will show all modules:

screenshot

## Cause

Before #2694, GUI performed two mod list updates at startup if repository auto updates were enabled: one with the existing registry data (which the user will never see), and one with the registry data after the repo update. This is not great because mod list updates are one of our slower operations.

After #2694, GUI performed only one mod list update at startup, using the data after the repo update. This improved performance, but it broke the Newly Compatible filter because that filter was based on a comparison of the mod list before and after the repo update. With only do one mod list update, the "before" list in the comparison is empty, and every mod seems to be newly added.

## Changes

Now the "before" part of the before-and-after comparison is generated based on the registry before the repositories are updated, and then passed to the mod list updating code after the repo update. This allows us to populate the filter properly without performing multiple mod list updates.

Fixes #2716.

### DIFF
--- a/GUI/MainRepo.cs
+++ b/GUI/MainRepo.cs
@@ -1,9 +1,11 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Net;
 using System.Timers;
 using System.Linq;
 using Newtonsoft.Json;
+using CKAN.Versioning;
 
 namespace CKAN
 {
@@ -64,7 +66,22 @@ namespace CKAN
             try
             {
                 AddStatusMessage("Updating repositories...");
-                e.Result = Repo.UpdateAllRepositories(RegistryManager.Instance(CurrentInstance), CurrentInstance, Manager.Cache, GUI.user);
+
+                // Note the current mods' compatibility for the NewlyCompatible filter
+                KspVersionCriteria versionCriteria = CurrentInstance.VersionCriteria();
+                IRegistryQuerier registry = RegistryManager.Instance(CurrentInstance).registry;
+                Dictionary<string, bool> oldModules = registry.Available(versionCriteria)
+                    .ToDictionary(m => m.identifier, m => false);
+                registry.Incompatible(versionCriteria)
+                    .Where(m => !oldModules.ContainsKey(m.identifier))
+                    .ToList()
+                    .ForEach(m => oldModules.Add(m.identifier, true));
+
+                RepoUpdateResult result = Repo.UpdateAllRepositories(
+                    RegistryManager.Instance(CurrentInstance),
+                    CurrentInstance, Manager.Cache, GUI.user);
+                e.Result = new KeyValuePair<RepoUpdateResult, Dictionary<string, bool>>(
+                    result, oldModules);
             }
             catch (UriFormatException ex)
             {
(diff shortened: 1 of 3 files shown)


GUI/MainRepo.cs
```diff
@@ -64,7 +66,22 @@ private void UpdateRepo(object sender, DoWorkEventArgs e)
...
                 AddStatusMessage("Updating repositories...");
-                e.Result = Repo.UpdateAllRepositories(RegistryManager.Instance(CurrentInstance), CurrentInstance, Manager.Cache, GUI.user);
+
+                // Note the current mods' compatibility for the NewlyCompatible filter
+                KspVersionCriteria versionCriteria = CurrentInstance.VersionCriteria();
+                IRegistryQuerier registry = RegistryManager.Instance(CurrentInstance).registry;
+                Dictionary<string, bool> oldModules = registry.Available(versionCriteria)
+                    .ToDictionary(m => m.identifier, m => false);
+                registry.Incompatible(versionCriteria)
+                    .Where(m => !oldModules.ContainsKey(m.identifier))
```
DasSkelett: Can mods be in `registry.Incompatible` but not in `registry.Available`?

GUI/MainRepo.cs
```diff
@@ -64,7 +66,22 @@ private void UpdateRepo(object sender, DoWorkEventArgs e)
             try
             {
                 AddStatusMessage("Updating repositories...");
-                e.Result = Repo.UpdateAllRepositories(RegistryManager.Instance(CurrentInstance), CurrentInstance, Manager.Cache, GUI.user);
+
+                // Note the current mods' compatibility for the NewlyCompatible filter
+                KspVersionCriteria versionCriteria = CurrentInstance.VersionCriteria();
+                IRegistryQuerier registry = RegistryManager.Instance(CurrentInstance).registry;
+                Dictionary<string, bool> oldModules = registry.Available(versionCriteria)
+                    .ToDictionary(m => m.identifier, m => false);
```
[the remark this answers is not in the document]
HebaruSan: It's false for every mod in `Available` and true for every mod in `Incomaptible.`
There is no `m.IsIncompatible` yet because `m` is a `CkanModule`, not a `GUIModule`.

GUI/MainRepo.cs
HebaruSan: Yes, that's the idea. The two functions together return non-overlapping sets, the union of which is all modules (see #2216).

https://github.com/KSP-CKAN/CKAN/blob/5331bf9344f6729a6b96f52bea45bdec245a3bb3/Core/Registry/Registry.cs#L476-L526

DasSkelett: Oh I see where my confusion comes from. `Registry.Available` basically means `Registry.Compatible` and only includes the latest compatible version of all mods,
contrary to `Registry.available_modules` which includes all mods and mod versions and is saved in the registry.json?
Seems the naming scheme is a bit mixed up.

HebaruSan: Yup. I started a branch to rename Available to Compatible, but there are complications, such as the `ckan available` command only listing compatible modules. It's not a simple matter to reach a consistent state.

PR #2735, opened 2019-04-18
Compare RelationshipDescriptors with Equals instead of Same

HebaruSan:
## Problem

@Poodmund reported an issue on the forum where this popup appears every time CKAN is launched, even after you click Yes to reinstall:

screenshot

https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1260-baikonur/&do=findComment&comment=3586115

## Cause

At the top level, `Repo.RelationshipsAreEquivalent` uses `RelationshipDescriptor.Same` to compare the relationships of an installed module with those of its freshly updated available counterpart:

https://github.com/KSP-CKAN/CKAN/blob/5331bf9344f6729a6b96f52bea45bdec245a3bb3/Core/Net/Repo.cs#L321-L330

Child class `ModuleRelationshipDescriptor` implements `Same` as a comparison of the name and versions:

https://github.com/KSP-CKAN/CKAN/blob/5331bf9344f6729a6b96f52bea45bdec245a3bb3/Core/Types/CkanModule.cs#L152-L160

Child class `AnyOfRelationshipDescriptor` implements `Same` by comparing the lists contained in the `any_of` property using `SequenceEqual`:

https://github.com/KSP-CKAN/CKAN/blob/5331bf9344f6729a6b96f52bea45bdec245a3bb3/Core/Types/CkanModule.cs#L239-L244

However, when `SequenceEqual` checks the pairs of values in the two lists, [it uses `Equals`, not `Same`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sequenceequal)! This means the nested relationships aren't compared the same way as regular ones, so they might not produce the correct result. This causes CKAN to think an `any_of` relationship has changed when it hasn't, and mistakenly alert @Poodmund to it.

## Changes

Now `RelationshipDescriptor` implements `IEquatable<RelationshipDescriptor>`, and it and its child classes implement `Equals` instead of `Same`, with the same logic as before. Calling code is updated to use `Equals` instead. This will ensure that the intended comparison logic is used when we call `SequenceEqual`.

Fixes @Poodmund's issue.

### DIFF
--- a/Core/Net/Repo.cs
+++ b/Core/Net/Repo.cs
@@ -323,7 +323,7 @@ Do you wish to reinstall now?", sb)))
                 var aRel = aSorted[i];
                 var bRel = bSorted[i];
 
-                if (!aRel.Same(bRel))
+                if (!aRel.Equals(bRel))
                 {
                     return false;
                 }
--- a/Core/Types/CkanModule.cs
+++ b/Core/Types/CkanModule.cs
@@ -13,7 +13,7 @@ using CKAN.Versioning;
 
 namespace CKAN
 {
-    public abstract class RelationshipDescriptor
+    public abstract class RelationshipDescriptor : IEquatable<RelationshipDescriptor>
     {
         public abstract bool MatchesAny(
             IEnumerable<CkanModule> modules,
@@ -25,7 +25,7 @@ namespace CKAN
 
         public abstract List<CkanModule> LatestAvailableWithProvides(IRegistryQuerier registry, KspVersionCriteria crit, IEnumerable<CkanModule> toInstall = null);
 
-        public abstract bool Same(RelationshipDescriptor other);
+        public abstract bool Equals(RelationshipDescriptor other);
 
         public abstract bool ContainsAny(IEnumerable<string> identifiers);
 
@@ -149,7 +149,7 @@ namespace CKAN
             return registry.LatestAvailableWithProvides(name, crit, this, toInstall);
         }
 
-        public override bool Same(RelationshipDescriptor other)
+        public override bool Equals(RelationshipDescriptor other)
         {
             ModuleRelationshipDescriptor modRel = other as ModuleRelationshipDescriptor;
             return modRel != null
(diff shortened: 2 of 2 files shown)


Poodmund:
What do you need from me and how do I acquire such logs?

HebaruSan:
The first thing that would be helpful is a copy of your `KSP/CKAN/registry.json` file. There may be some quirk in your copy of the metadata that isn't present in mine. Beyond that, just filling in the issue template would be useful in answering such questions as game versions, etc.

Poodmund:
Background
----------

**CKAN Version:** 1.26.0


**KSP Version:** 1.6.1.2401


**Operating System:** Windows 10 64bit


**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** Yes, to install mods that are not listed on CKAN. 


Problem
-------

**What steps did you take in CKAN?** Opening CKAN for the first time each boot cycle.


**What did you expect to happen?** CKAN opens and automatically updates the repository list.


**What happened instead?** CKAN loads and hangs with a pop-up appearing stating that a mod has had its metadata changed since the last update and recommends reinstalling it the preserve consistency. If you click Yes or No in the pop-up, CKAN then takes the respective action, updates the repository list and loads into the mod listing screen as per expectations.


**Screenshots:** See above


**CKAN error codes (if applicable):** No error codes

Registry.json: https://www.dropbox.com/s/3icb9v3lms8q9e1/registry.json?dl=0

HebaruSan:
Thanks! Taking a look, the available module exactly matches the installed module, so that rules out a metadata quirk.

Maybe a platform thing; could be that `SequenceEqual` and `Equal` have different default behaviors on Windows and Mono (I'm in Linux right now).

This test build has the fix from this pull request, if you wouldn't mind trying it out and reporting whether the problem is fixed:

- [ckan.zip](https://github.com/KSP-CKAN/CKAN/files/3095880/ckan.zip)

Poodmund:
Testing the above build, CKAN hangs with most of the UI missing for about 5 seconds upon initial application opening then it kicks into gear and starts automatically updating the repo and then loads into the mod list.

The change seems to have resolved my issue and I have hard-cycled the PC 5 times to test and all 5 times CKAN booted into the mod list without showing the pop-up.

Poodmund:
I don't know whether it is related, but trying to launch CKAN today just opens the initial console window, it closes after a second or so, then nothing else launches. Hovering over the window preview in the taskbar just shows a blank white preview of the CKAN window but it never opens.

I tried downgrading the executable to the 1.26.0 release but the same issue is occuring.

Could there be any possible relation?

DasSkelett:
Sounds like you encountered #2717 ?

Poodmund:
Yep, exactly that, thanks for the heads up and sorry for the false alarm. Weird bug though.

PR #2743, opened 2019-04-27
Fix freeze on non CKAN files in cache folder

### LINKED ISSUE #2711: CKAN stuck at loading available modules
Background
----------
CKAN Version: 1.26.0 

KSP Version: 1.4.3.2152

Operating System: Windows 10 x64

Have you made any manual changes to your GameData folder (i.e., not via CKAN)? -> Yes

Problem
-------

What steps did you take in CKAN?
  -> selecting 1.4 as new compatible KSP game version

What did you expect to happen?
  -> Game change the settings to an other compatible mode

What happened instead?
  -> CKAN (freez) stuck at loading available modules in Loading Modules

**Screenshots:**
  ->
Ckan Prob

HebaruSan:
## Problem

Some users reported that CKAN freezes while trying to load the registry at startup.

## Cause

This exception is thrown:

```
System.ArgumentOutOfRangeException: 인덱스 및 길이는 문자열 내의 위치를 참조해야 합니다.
매개 변수 이름: length (Rough translation: The index and length must refer to the position within the string. Parameter name: length)
   위치: System.String.Substring(Int32 startIndex, Int32 length)
   위치: System.Linq.Lookup`2.Create[TSource](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   위치: System.Linq.GroupedEnumerable`3.GetEnumerator()
   위치: System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   위치: System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector)
   위치: CKAN.NetFileCache.GetCachedFilename(Uri url, Nullable`1 remoteTimestamp)
```

... which means that this block from #2563 tried to get a substring of a filename shorter than 8 characters:

https://github.com/KSP-CKAN/CKAN/blob/44d44ea5cd38e2f568d95f2af6d24674e01f7a45/Core/Net/NetFileCache.cs#L175-L180

It appears that some users may have chosen a CKAN cache folder that is not exclusive to CKAN:

screeenshot

## Changes

Now that block will ignore any filenames that don't start with 8 hex digits and a dash. This will both fix the problem with short filenames **and** exclude other files that were not created by CKAN even if they have long names.

Fixes #2711 (I think).

### DIFF
--- a/Core/Net/NetFileCache.cs
+++ b/Core/Net/NetFileCache.cs
@@ -30,6 +30,7 @@ namespace CKAN
         private Dictionary<string, string> cachedFiles;
         private string cachePath;
         private KSPManager manager;
+        private static readonly Regex cacheFileRegex = new Regex("^[0-9A-F]{8}-", RegexOptions.Compiled);
         private static readonly ILog log = LogManager.GetLogger(typeof (NetFileCache));
 
         /// <summary>
@@ -377,7 +378,10 @@ namespace CKAN
                 DirectoryInfo legDir = new DirectoryInfo(legacyDir);
                 files = files.Union(legDir.EnumerateFiles());
             }
-            return files.ToList();
+            return files
+                // Require 8 digit hex prefix followed by dash; any else was not put there by CKAN
+                .Where(fi => cacheFileRegex.IsMatch(fi.Name))
+                .ToList();
         }
 
         /// <summary>


DasSkelett:
Oh wow, didn't think of the possibility that there are other files with shorter names in the cache folder.

There's another issue happening:
Since v1.26, more precisely #2617, (fatal) errors are no longer thrown during startup. This is due to the whole process being a task now, and C# doesn't throw errors happening in tasks without using `await`.
This can be a big problem, because the only way to get a stack trace and error message is with a debugger, which is only an option if you know how to reproduce a bug. In #2711 this was a problem to, we only got the error text with a downgrade to v1.25.
I'll do a PR to fix this.

HebaruSan:
Oh, oops. I assume we can keep the loading screen if we switch to `async`/`await`?

DasSkelett:
Yes. It works exactly the same, but actually throws errors if they occur.

HebaruSan:
OK, good. I've tried probably around five times now to learn `async`/`await`, and I still find threads much easier to understand.

PR #2744, opened 2019-04-28
Combine recommendation and suggestion screens

### LINKED ISSUE #2724: Importing .ckan also selects conflicting recommends
Importing .ckan also selects conflicting recommends

HebaruSan:
*(a.k.a., "Rewrite recommendations... again")*

## Problem

If you import a .ckan file that recommends CustomBarnKit, it'll try to install two mods by default:

- CustomBarnKit
- CustomBarnKit-RO

These conflict with one another, so if the user simply accepts the recommendation, the install will fail with an error.

## Cause

CustomBarnKit-RO `provides` CustomBarnKit, so recommending one pulls in both.

This is important for something like FerramAerospaceResearchContinued, which we want to match a recommendation for FAR.

## Changes

We need the ability to present a recommendation as unchecked. This is impossible within the current code structure, since all recommendations are handled the same way in one big group, and there's no conduit through which we could pass an indication of whether to check the checkbox.

Now the recommendations and suggestions prompts are merged into one big prompt, containing some items that are checked by default (the recommendations) and others that are unchecked by default (the suggestions). This reduces the number of clicks needed to get through an install.

To make this work, the recommendations code is refactored to generate a single sequence of `ListViewItem`s, each of which represents a recommendation or suggestion, and the checked state of which is controlled by more subtle logic. If only one module matches the recommendation, or if there's an exact identifier match, then it'll be checked, otherwise it'll be unchecked. In the original CustomBarnKit scenario, CustomBarnKit will be checked while CustomBarnKit-RO is unchecked. This allows the user to change it if they really want to, while still providing something that will work by default.

Using the new code is somewhat simpler for calling code, since it no longer needs to loop over modules or generate lists of recommendations and suggestions. Instead we can now simply pass the list of source modules and receive back the rows that would go to the dialog as input.

Fixes #2724.

### DIFF
--- a/GUI/Main.Designer.cs
+++ b/GUI/Main.Designer.cs
@@ -117,6 +117,8 @@
             this.RecommendedModsToggleCheckbox = new System.Windows.Forms.CheckBox();
             this.RecommendedDialogLabel = new System.Windows.Forms.Label();
             this.RecommendedModsListView = new System.Windows.Forms.ListView();
+            this.RecommendationsGroup = new System.Windows.Forms.ListViewGroup();
+            this.SuggestionsGroup = new System.Windows.Forms.ListViewGroup();
             this.columnHeader3 = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
             this.columnHeader4 = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
             this.columnHeader5 = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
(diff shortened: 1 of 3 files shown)


GUI/Main.Designer.cs
```diff
@@ -1047,9 +1047,8 @@ private void InitializeComponent()
             this.RecommendedDialogLabel.Name = "RecommendedDialogLabel";
             this.RecommendedDialogLabel.Size = new System.Drawing.Size(627, 20);
             this.RecommendedDialogLabel.TabIndex = 6;
-            this.RecommendedDialogLabel.Text = "The following modules have been recommended by one or more of the chosen modules:" +
-    "";
             // 
+            this.RecommendedDialogLabel.Text = "The following modules have been recommended or suggested by one or more of the chosen modules:";
```
DasSkelett: The tab is missing a legend I think. It's probably hard to guess which mods are recommended and which are suggested if you don't know the checked-unchecked rule.
Changing this sentence to `The following modules have been recommended (mostly checked - upper part) or suggested (unchecked - lower part).` maybe would suffice.

DasSkelett: Apart from that, the changes look good to me. And they work!

HebaruSan: How about some `ListViewGroup`s instead?

image

DasSkelett: Even better! 🎉

HebaruSan:
BTW, submitted mono/mono#14325 to try to fix the group header colors on Linux:

screenshot

PR #2749, opened 2019-04-30
Internationalize the GUI

HebaruSan:
## Motivations

- The CKAN GUI is English-only
- We're split on whether it's en-GB/en-AU or en-US

## Changes

- All user visible properties in the `.Designer.cs` files are moved into the corresponding `.resx` files and set via `resources.ApplyResources`.
- All user visible strings in code are moved to `Properties/Resources.resx` and accessed via `Properties.Resources.NameOfString`.
- All occurrences of American spelling in the default locale are changed to UK/Aussie spelling.
- Supplemental `.en-US.resx` files are added as needed to switch to American spellings for American users.
- The build process is updated to bundle the "satellite assemblies" that the compiler creates into the main EXE.
- New classes `SingleAssemblyResourceManager` and `SingleAssemblyComponentResourceManager` extend the built in resource manager classes to support bundled satellite assemblies.

Now when a UK/Aussie user opens the CKAN GUI, they'll see familiar extra "u"s and words ending in "ise", while Americans will see no extra "u"s and words ending in "ize".

### German translation

@DasSkelett has translated the entire GUI into German!

image

Erfolg!

### Icons and exe size

While poking around at our resource files, @DasSkelett discovered that we have 11 copies of the CKAN icon (roughly one per form), and that each one takes up something like 700 KiB, adding up to fully **half** the size of the compiled exe!!

Now these are moved from the forms' resx files to the project resx file, where one copy is shared. This makes `ckan.exe` less than half the size it was previously, while working exactly the same.

### Known limitations

Only GUI is updated! This PR doesn't touch Core (but a future one might). Lots of visible strings come from Core, but the ones from GUI should all be localizable.


DasSkelett:
Wow, nice! Did you use the VisualStudio functionality to move all the strings?
Should I open a PR to your branch for the German translation, or create a PR here when this one is merged?

HebaruSan:
I used Atom and my mouse to move all the strings :face_with_head_bandage:, but I based the changes on a demo from Microsoft, so it should match up with how VS would do it:
https://code.msdn.microsoft.com/CSWinFormLocalization-59675a87

As for a Deutsch PR, either way is fine with me. I'd be happy to include it here or add it later.

DasSkelett:
Wow, that must have been a lot of work.
Okay, I will do the PR to your branch then, since I found some things to correct (missing strings, formatting...), I can include the in my PR this way.

DasSkelett:
Just an idea: What about an option to set the language manually? Can imagine some peaople wanting it.

And a FYI: You can set the localization yourself in the code if you add those 3 lines (found [here](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.defaultthreadcurrentculture?redirectedfrom=MSDN&view=netframework-4.8#System_Globalization_CultureInfo_DefaultThreadCurrentCulture)) as early as possible (before any localized strings are loaded), e.g. right at the start of `GUI.Main()`:
```
System.Globalization.CultureInfo ci = new System.Globalization.CultureInfo("en-US");
System.Globalization.CultureInfo.DefaultThreadCurrentCulture = ci;
System.Globalization.CultureInfo.DefaultThreadCurrentUICulture = ci;
```
If you want to test the German translation for example later on. If you found a better way to do this, I'm all ears.

HebaruSan:
> If you found a better way to do this

That was exactly how I did it. :)  Except I only set one of those.

politas:
Wow, massive piece of work there, @HebaruSan! Is this ready for our point release, or shall we leave it for the next one?

HebaruSan:
I think it's pretty safe. If there are bugs, they should be of the "this dialog has a strange string here" variety rather than crashes or corrupted instsalls.

DasSkelett:
Can somebody by any chance tell me, why my tranlations work when compiling in VisualStudio (2019), but not when when compiling with the BuildTools (15) installed without VS, nor when compiling on Linux?
To spedifiy: It somehow doesn't include the 
https://github.com/DasSkelett/CKAN/tree/feature/german-translation

And btw, does anybody know how to prevent VS from overwriting all the auto-generated comments?
(See [here](https://github.com/DasSkelett/CKAN/commit/ceae2c58b7f70881fc7eaa0921af0a14f871da9d#diff-ec0064348aa1f8fef1c2c4e358b04e80) for example)

HebaruSan:
The en-US translations seem to be working for me on Mono (after the commit I just pushed):

image

I'll take a look at your branch shortly...

GUI/Main.resx
```diff
@@ -13774,4 +13774,95 @@
         AAD//wAA//8AAP//AAD//wAA//8AAA==
 </value>
   </data>
-</root>
\ No newline at end of file
+  <data name="$this.Localizable" type="System.Boolean, mscorlib">
+    <value>True</value>
+  </data>
+  <data name="$this.Language" type="System.Globalization.CultureInfo, mscorlib">
+    <value>(Default)</value>
```
DasSkelett: What does this do? It's just in this .resx, not in others.
Should I change `(Default)` to something different in the `Main.de-DE.resx`, or discard it, or leave it?

HebaruSan: All of the tutorials/answers that I found online said to set that property. I think it controls which language VS populates when you edit properties?

DasSkelett:
> The en-US translations seem to be working for me on Mono (after the commit I just pushed):

Sorry, missed a important bit: The normal .resx translations work perfectly fine, just the `Properties/Resources.de-DE.resx` (so 'dynamic' strings) doesn't work.

HebaruSan:
The `Properties/Resources.en-US.resx`seems OK on Mono as well (this would say "favourites" otherwise):

image

HebaruSan:
Yeah, I'm seeing some weird stuff in de-DE.

image

I'll let you know if I can figure out why...

DasSkelett:
> Yeah, I'm seeing some weird stuff in de-DE.
Oh that's probably due to https://github.com/mono/mono/pull/12643.

> The Properties/Resources.en-US.resxseems OK on Mono as well (this would say "favourites" otherwise):
Well, it *does* say 'favourites' on my system...

HebaruSan:
BTW, in case it's useful, my command for testing on Linux is:

```bash
LC_ALL=de_DE _build/ckan.exe
```

GUI/Main.resx
DasSkelett: So, just let it as is in `Main.de-DE.resx`?

HebaruSan:
> Oh that's probably due to mono/mono#12643.

Huh, you're right. I have a local build of mono's current HEAD (which includes that PR), and it works there:

image

HebaruSan:
> You tested on my branch, right?

My "favourites" test was on *my* branch, to establish a baseline of whether there's a known-working version here. I'm checking your branch now.

GUI/Main.resx
HebaruSan: I think this property doesn't need to be included in `.de-DE.resx`.

HebaruSan:
This is where `Resources.de-DE.resx` broke, it needs to be a `SingleAssemblyResourceManager` object or it won't check the satellite assemblies bundled in the EXE:

https://github.com/DasSkelett/CKAN/blob/ceae2c58b7f70881fc7eaa0921af0a14f871da9d/GUI/Properties/Resources.Designer.cs#L42

Can you revert all your changes to that file? You should be able to use my version without changes.

DasSkelett:
> Huh, you're right. I have a local build of mono's current HEAD (which includes that PR), and it works there:

I recognize the issues I fixed ;)
It should be fixed since mono-5.16.0.235.

> BTW, in case it's useful, my command for testing on Linux is:
> 
> ```shell
> LC_ALL=de_DE _build/ckan.exe
> ```

Thanks, that might be easier than hardcoding the CultureInfo manually and recompile. Just Windows doesn't offer an option, as always.

HebaruSan:
> I recognize the issues I fixed ;)

Oh nice! I didn't even think to check if that was you. :laughing: Congrats on getting that merged!

DasSkelett:
> 
> 
> This is where `Resources.de-DE.resx` broke, it needs to be a `SingleAssemblyResourceManager` object or it won't check the satellite assemblies bundled in the EXE:
> 
> https://github.com/DasSkelett/CKAN/blob/ceae2c58b7f70881fc7eaa0921af0a14f871da9d/GUI/Properties/Resources.Designer.cs#L42
> 
> Can you revert all your changes to that file? You should be able to use my version without changes.

Will try, thanks 🎉.
Mind if I force-push my branch? Easiest for me is resetting the commit and unstaging the file.

HebaruSan:
> Mind if I force-push my branch?

Please do, I won't be making changes to your branch.

DasSkelett:
**Regarding the `LinkArea` bug** (short summary for those reading this and don't want to check the linked PR: Mono doesn't (yet) take text alignment into account when rendering LinkLabels, which makes all non-upper-left aligned LinkLabels unreadable):
> It should be fixed since mono-5.16.0.235.

That's wrong it seems, since I run `5.20.1.19`which is from April 11th and my PR is not included. So I have no idea how the release mechanism of mono works, and when a PR is included.

**Why is this bug a problem for us/me?**
I changed the layout of the `AboutDialog`in my branch so that every (link)label spans from left to right (excluding the last row with three LinkLabels) and set
```
linkLabel.TextAlign = ContentAlignment.MiddleCenter
```
for easy centering in every language.
But with the mono bug the dialog is unreadable, most links are unclickable.
Basically, the `AboutDialog` is unusable like this.
If we stick to this, we must hope that the next mono release includes that fix and everyone updates to it.

**A workaround/how it was before**: Set the location of every link label for every language (by now only two for this form) manually so they appear centered (which is _very_ fiddly to do!).
But at least you can read something like this.
Another idea: Calculate the string size during runtime and set the position and size accordingly. A bit dirty, but should do it.

HebaruSan:
> If we stick to this, we must hope that the next mono release includes that fix and everyone updates to it.

Or that no one opens the About dialog. :rofl:

HebaruSan:
> all non-upper-left aligned LinkLabels

I suppose we also could switch to upper left alignment. The current layout is acceptable but not really beautiful.

DasSkelett:
So a simple list layout?
```
_________________________________________________
| CKAN -The Comprehensive Kerbal archive network | 
| License                                        |
| Authors                                        |
| Source code                                    |
| Forum thread                                   |
| Homepage                                       |
| Version v1.26.1                                |
|________________________________________________|
```
Like this?
I wouldn't say that's more beautiful, but okay :smile:

HebaruSan:
Yeah, that's roughly what I was imagining. You have to admit it's more beautiful than the bugged out version though. :)

about

Maybe a table with the real URLs shown, for more visual balance...

```
__________________________________________________________________________________
|____________________________________About__________________________________-_▢_X|
| CKAN -The Comprehensive Kerbal archive network                                 | 
| Version v1.26.1                                                                |
|                                                                                |
| License       https://github.com/KSP-CKAN/CKAN/blob/master/LICENSE.md          |
| Authors       https://github.com/KSP-CKAN/CKAN/graphs/contributors             |
| Source code   https://github.com/KSP-CKAN/CKAN/                                |
| Forum thread  http://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan |
| Homepage      http://ksp-ckan.space                                            |
|________________________________________________________________________________|
```

DasSkelett:
I can go with the url table, that looks nice. :+1:

HebaruSan:
BTW, is the `LinkArea` property actually related to this? It sounds like that represents the offset of the characters that are intended to be clickable.

DasSkelett:
Mono divides each `linklabel` into several substrings, either 'clickable' or not. The `LinkArea` property says how to divide the string, at which character the 'clickability' should begin or stop.
Those substrings are then converted to `region`s and later `pieces`, which also contain coordinates for the rendering process. And those pieces should consider any offset caused by a 'non-upper-left' alignment, which they didn't/don't.
And the coordinates of those pieces are later used to determine the areas where you can click, and where not, and where the 'selection rectangle' is drawn.
_But not_ the text. That one is rendered (or how you call that, that's not my area of expertise) by the gdisomethingsomething library of Microsoft, which takes alignment into account. And the text is only drawn if it's inside the are of the pieces \> it's cut somewhere in the middle.

At least that's what I remember from back then ^^
So yeah, the `LinkArea` bug actually is not really a `LinkArea` bug, but it has to do with the separation into pieces.

DasSkelett:
Is there a way to save the little icon we use in the upper left of each form centrally (maybe in `Properties/Resources.resx`?
For reference, it's that part in each .resx:
```
<data name="$this.Icon" type="System.Drawing.Icon, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
 <value> Base64EncodedIconStringGoesHere </value>
</data>
```
It takes ~600 KiB in every standard language (=en-GB) .resx file, that makes around 7 MiB in total.

I think the compiler apparently is smart enough to only include it once in the final .exe, but there's also the problem that if we want to change it, we have to change it for each form (if that's how it works, I'm not sure).
And it would reduce the size of the source code (only 7 MiB, but better than nothing, eh?).

HebaruSan:
Hmm. What if `Main` loaded from its own `resx` file, and then other forms re-used `Main.Instance.Icon`?

DasSkelett:
Works if you add `this.Icon = null` before `resources.ApplyResources(...);` in `Main.Designer.cs`.

DasSkelett:
Okay got some big, good? news: The compiler was **not** smart enough to remove duplicate image strings in resxes. So yeah, just compiled CKAN after removing all those, and you won't believe it, but **CKAN.exe is now 3.5 MiB** small...
From 6.9 MiB down to 3.5 MiB, I just halved the whole exe :tada:.

I'll open a Patreon now so you can buy me some drinks :cocktail:.

HebaruSan:
Wait, a 32x32 bitmap was taking up *how much* space?? 4KB per icon should be more than enough, even for uncompressed 32-bit color. Does the compiled binary contain that gross base64 encoded mess?

DasSkelett:
I just did the calculations, it must have saved them with about 30% compression (depending on the encoding of the characters and stuff, I assumed 7 bit):
6803 lines per icon resx * 80 chars per line * 7 bit/char / 8 bit/byte * 11 times I deleted it / 1024 / 1024 = 5.00 MiB.
1 - (3.5 MiB savings / 5,00 MiB potential space usage) = 0.3. Yep.

Well, if you look at them, they are about 70% 'A's, so maybe they do a simple RLE?

HebaruSan:
Doing a bit more checking, the consensus seems to be that we should move it to `Properties.Resources.AppIcon`. That way we wouldn't be depending on the forms loading in a particular order as with `Main.Instance.Icon`, and it may even be compatible with the VS form editor GUI.

DasSkelett:
> 
> 
> Doing a bit more checking, the consensus seems to be that we should move it to `Properties.Resources.AppIcon`. That way we wouldn't be depending on the forms loading in a particular order as with `Main.Instance.Icon`, and it may even be compatible with the VS form editor GUI.

Welcome that. Do you have some details on how to do that?
WIth `Properties.Resources.AppIcon`, do you mean a file named `Resources.AppIcon` in `GUI/Properties/`, or a `data name="AppIcon" ...` node in `Resources.resx`?

HebaruSan:
I meant a node in `Resources.resx`, copy/pasting the string that we have already.

DasSkelett:
So why don't just reference the .ico file located in `assets/`in `Resources.resx`, as we do with all the other bitmaps?

HebaruSan:
If it works, that would be fine with me. I'm not sure what limitations are imposed by the VS GUI, so I lean towards being conservative when working with GUI control properties.

DasSkelett:
Both ways don't seem to work on the first try :(
I will look at it tomorrow again, it's quite late here. :sleeping:

HebaruSan:
@DasSkelett, FYI, I need to rebase this to internationalize the autoinstall changes anyway, so I'm going to take a crack at including the icon changes as well...

HebaruSan:
OK, these changes are now rebased on top of master, then squashed, then amended with a missing string and the icon clean-up, then  force-pushed. You should be able to update your German branch thusly:

```
git fetch --all
git checkout feature/german-translation
git rebase -i HebaruSan/feature/i18n
```
(And then resolve the conflicts.)

DasSkelett:
Please do so :+1:

DasSkelett:
Created my PR to add my changes to your branch and thus include them in your PR:
https://github.com/HebaruSan/CKAN/pull/5

HebaruSan:
My GNOME Shell theme apparently doesn't put icons on windows anyway, but it does put them in the top bar:

image

Not sure if I can test that issue.

DasSkelett:
Okay, I just found out that the icon problem is **not** introduced by our changes!
Other CKAN versions (i.e. current master@75deab0) have the same problems, at least on my system. It just never noticed it. But as I tested if you can remove the icon of the resource files, I obviously put attention there and spotted it.

DasSkelett:
From my side only the issue with the wrapping menu entry is left now.

grafik

HebaruSan:
> the wrapping menu entry

This issue is unreal! There's nothing different in how we treat that object, but that effect is completely consistent and persistent no matter what changes I make.

... I think we're looking at a one-in-a-million Mono bug related to the exact pixel length of the resource string. I tried changing it to "afsdkamjsldk fdjklpa" (same number of characters with same grouping, but the letters aren't the exact same width) and it started working normally. But as long as the exact text is "Empfehlungen ansehen", it goes weird. Can you think of a slightly different translation for that string?

DasSkelett:
"Empfehlungen installierter Mods", translated to "Recommendations of installed mods" should fit well.

HebaruSan:
:+1: 
Mono is so weird.

image

politas:
So we're all good to merge this, now? Are we sure those weird alignment errors don't appear with different app window sizes, or something like that?

DasSkelett:
> So we're all good to merge this, now?

@HebaruSan needs to rebase once more for #2762.
@HebaruSan, if you want to include the German translation for the new string yourself, I suggest "Durchsuche GameData nach DLCs und manuell installierten Modulen...".

> Are we sure those weird alignment errors don't appear with different app window sizes, or something like that?

I don't remember if I tried resizing the window, but I doubt that would affect the issue. Resizing the dropdown menu didn't fix it.
I can try it out tomorrow.

So after the rebase you have a Go from my side :+1:.

HebaruSan:
Rebased!

> I suggest "Durchsuche GameData nach DLCs und manuell installierten Modulen...".

I decided the English string sounded better without GameData. Does " I suggest "Durchsuche nach DLCs und manuell installierten Modulen..." make sense?

DasSkelett:
"Suche nach DLCs und manuell installierten Modulen..." (without the "Durch") would be a bit more natural.

HebaruSan:
@politas, yes, this is ready now. I haven't merged it because I wasn't sure whether to count it as "reviewed" given that both @DasSkelett and I have made commits to the branch.

DasSkelett:
> I don't remember if I tried resizing the window, but I doubt that would affect the issue. Resizing the dropdown menu at least didn't fix it.
I can try it out tomorrow.

No, resizing the window does indeed nothing. Would have been weird, because I'm sure @HebaruSan  doesn't use the exact same window size, and it happened on his system too.

GUI/Properties/Resources.de-DE.resx
```diff
@@ -0,0 +1,310 @@
...
+  <data name="ManageKspInstancesNotValid" xml:space="preserve"><value>"{0}" ist kein gültiges KSP-Verzeichnis.</value></data>
+  <data name="NewRepoDialogFailed" xml:space="preserve"><value>Die Master-Liste konnte nicht abgerufen werden.</value></data>
+  <data name="SettingsDialogSummmary" xml:space="preserve"><value>{0} Dateien, {1}</value></data>
+  <data name="SettingsDialogSummaryInvalid" xml:space="preserve"><value>Ungültiger Pfad: "{0}"</value></data>
+  <data name="SettingsDialogCacheDescrip" xml:space="preserve"><value>Wähle einen Ordner um CKANs Mod-Downloads zu speichern:</value></data>
+  <data name="SettingsDialogDeleteConfirm" xml:space="preserve"><value>Möchten Sie wirklich {0} Dateien aus dem Cache löschen und damit {1} freigeben?</value></data>
+  <data name="AddAuthTokenTitle" xml:space="preserve"><value>Authentifikationstoken hinzufügen</value></data>
+  <data name="AddAuthTokenToken" xml:space="preserve"><value>Token:</value></data>
+  <data name="AddAuthTokenAccept" xml:space="preserve"><value>&amp;Annehmen</value></data>
+  <data name="AddAuthTokenCancel" xml:space="preserve"><value>&amp;Abbrechen</value></data>
```
Olympic1: These two actions will conflict if you press `A` in the dialog

DasSkelett: @HebaruSan can you move the `&` to the `c` of `Abbrechen`, so `Abbre&amp;chen`?
`C` makes sense too, that's next to `Q` a shortcut often used for quitting/canceling stuff.

HebaruSan: Fixed. Thanks for catching that1

Olympic1:
Is it possible to add the translation files into a separate folder like `en-US` and `de-DE`, because every translation wil add 18 new files. This way it would be easier to find a specific language file and will prevent the GUI folder to get cluttered with `.resx` files.

HebaruSan:
Good idea. Rebased to get the NuGet updates and moved those to a Localization/de-DE folder. I couldn't figure out how to make it work with Resources.de-DE.resx, but those are already in the Properties folder so that shouldn't be as much of a problem.

Olympic1:
For me this is good to merge. If nobody finds any alignment errors anymore for the German translation, this can be merged.

PR #2756, opened 2019-05-08
Cache downloads individually upon completion

### LINKED ISSUE #1954: Don't fail all updates if one download fails
When updating several mods, sometimes one fails to download (probably because of my garbage internet connection). It would be nice if it actually continued and completed the ones that did download OK, rather than seemingly aborting the whole process.

Currently I get an error message, and then only have a Cancel-button which takes me back to the list of mods which looks the same as before. I.e. the ones with new versions are still not listed as up to date.

So yeah.. would be nice to instead of getting an error message and a Cancel-button, a summary of which ones failed and which ones were OK. And every mod that was downloaded OK should've been updated too, maybe except if they depend on one that failed.

### LINKED ISSUE #2746: Add downloads to cache immediately upon individual completion
Background
----------

**CKAN Version:** 
v1.26.0

**KSP Version:** 
1.7

**Operating System:** 
Win10 x64

Nothing other than installing NearFuture Aeronautics, not related with issue.

Description
-------
This is not an urgent request, rather - a minor, maybe even personal one.

temp3

When downloading mods, CKAN wait for **all mods** to be downloaded in temporary folder before copying to 'Download Cache' folder, and does not remove those .tmp files until clean exit of CKAN.

But when downloading multiple large mods, they would easily fill up Ramdisk causing CKAN to hang on download indefinitely. In that case: .tmp files will stay and take up space until reboot.

If CKAN moves **a mod** to 'Download Cache' folder as soon as it's complete and remove corresponding .tmp file, user can cancel download and not loose progress of already completed downloads, or need to care about their Temporary folder capacity.

HebaruSan:
## Problem

If you select many uncached mods to install, they're all downloaded to temp files in your temp folder before being moved to the CKAN download cache. The temp files remain until *all* of the downloads fail or finish. This consumes disk space in the temp folder for longer than strictly necessary and delays the insertion of individual files into the cache.

image

## Cause

`NetAsyncModulesDownloader` and `NetAsyncDownloader` communicate via a callback function that's only called at the end of the download process. So `NetAsyncModulesDownloader` waits for all downloads to complete before ever calling `cache.Store`.

## Changes

- `NetAsyncDownloader.onCompleted ` is replaced by `onOneCompleted`, which is called once per individual download, when that download fails or finishes
- `NetAsyncModulesDownloader.ModuleDownloadsComplete` is replaced by `ModuleDownloadComplete` which handles one download at a time
- The `cache` parameter of `NetAsyncModuleDownloader.DownloadModules` is moved to that class's constructor, and all calling code is updated to accommodate this (the motivation is to be able to use `ModuleDownloadComplete` cleanly as the callback without a lambda)

Note that this change is not guaranteed to fix any real world problem. If you were having a problem with your temp folder filling up before, that exact problem can still happen if the timing of the downloads works out such that they all last a long time. You will always need enough temp folder space available to handle all downloads that you initiate.

Fixes #2746.
Fixes #1954 (probably fixed a long time ago, but now we can say that the possibility of this happening has been eliminated).

### DIFF
--- a/Cmdline/Action/Replace.cs
+++ b/Cmdline/Action/Replace.cs
@@ -155,7 +155,7 @@ namespace CKAN.CmdLine
                 // TODO: These instances all need to go.
                 try
                 {
-                    ModuleInstaller.GetInstance(ksp, manager.Cache, User).Replace(to_replace, replace_ops, new NetAsyncModulesDownloader(User));
+                    ModuleInstaller.GetInstance(ksp, manager.Cache, User).Replace(to_replace, replace_ops, new NetAsyncModulesDownloader(User, manager.Cache));
                     User.RaiseMessage("\r\nDone!\r\n");
                 }
                 catch (DependencyNotSatisfiedKraken ex)
--- a/Cmdline/Action/Upgrade.cs
+++ b/Cmdline/Action/Upgrade.cs
@@ -129,13 +129,13 @@ namespace CKAN.CmdLine
 
                     }
 
-                    ModuleInstaller.GetInstance(ksp, manager.Cache, User).Upgrade(to_upgrade, new NetAsyncModulesDownloader(User));
+                    ModuleInstaller.GetInstance(ksp, manager.Cache, User).Upgrade(to_upgrade, new NetAsyncModulesDownloader(User, manager.Cache));
                 }
                 else
                 {
                     // TODO: These instances all need to go.
                     Search.AdjustModulesCase(ksp, options.modules);
-                    ModuleInstaller.GetInstance(ksp, manager.Cache, User).Upgrade(options.modules, new NetAsyncModulesDownloader(User));
+                    ModuleInstaller.GetInstance(ksp, manager.Cache, User).Upgrade(options.modules, new NetAsyncModulesDownloader(User, manager.Cache));
                 }
             }
             catch (ModuleNotFoundKraken kraken)
--- a/ConsoleUI/InstallScreen.cs
+++ b/ConsoleUI/InstallScreen.cs
@@ -61,7 +61,7 @@ namespace CKAN.ConsoleUI {
                             inst.UninstallList(plan.Remove);
                             plan.Remove.Clear();
                         }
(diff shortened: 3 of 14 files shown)


DasSkelett:
> Note that this change is not guaranteed to fix any real world problem. If you were having a problem with your temp folder filling up before, that exact problem can still happen if the timing of the downloads works out such that they all last a long time. You will always need enough temp folder space available to handle all downloads that you initiate.

We already have the download size of the mods in the metadata, and the .tmps are just the "unnamed" zips, right? Shouldn't it be possible to add a warning before the download starts if all mods together are bigger than the remaining disk space in the temp folder?

HebaruSan:
> Shouldn't it be possible to add a warning before the download starts if all mods together are bigger than the remaining disk space in the temp folder?

Yes... **if** .NET provided a cross platform "free space in folder" function. The only one I've been able to find is `DriveInfo.AvailableFreeSpace`, which I think is only useful on Windows.

DasSkelett:
> which I think is only useful on Windows.

Why? Is it buggy/not working/not implemented on Mono?

HebaruSan:
Other OSes don't have a concept of a "drive" in the same way Windows has "C:" and "D:", and I don't think the runtime provides a cross-platform way to translate from a folder path to a Drive. That was from the last time I looked into it, though, so I'm prepared to be proven wrong.

DasSkelett:
Yeah, I tested it, and it seems that mono can't get a drive info for the current directory, if it's not the exact point where a drive/partition is mounted (that means it works for /media/... or root / itself, but not f.e. for /home/user/documents/ when the home folder is a symlink to another drive.)
And `Path.GetPathRoot()` always returns / .

You're right it seems :/

HebaruSan:
Yeah, I think this last came up when we were consolidating the cache folder. The only reliable way to detect disk overflows is after the fact, with exceptions.

DasSkelett:
Pffft, now I did spot where to dismiss reviews again...
Well, for the next time :smile:

PR #2757, opened 2019-05-09
Generate RPM packages

### LINKED ISSUE #2748: RPM support
Background
----------

**CKAN Version:** 
All(packaging issue)

**KSP Version:** 
N/A

**Operating System:** 
RPM Linux distros

Problem
-------

N/A

I expected to find an RPM download link

**What happened instead?**
There was only a DEB link. This only works on Debian derivatives. Supporting RPM or even Snap or Flatpak would allow non-Debian installations.

**Screenshots:**

HebaruSan:
## Motivation

We generate .deb files as of #2187, but users of Linux distributions that aren't Debian-derived have to use the plain `ckan.exe` file. This means they miss out on:

- Having ckan in the system `$PATH`
- Having a ckan icon in the system app menus
- An out of date man page
- Falling back to ConsoleUI if `$DISPLAY` isn't defined

Some users may even assume that the lack of an .rpm file means they can't use CKAN, even though they can.

## Changes

Now we also generate .rpm files, so Fedora users can enjoy the above benefits. This is done with a simple `Makefile` and a .spec file, sharing a few files with the .deb package. The build system is updated to include the .rpm file as a release asset.

To try it via the command line:

```sh
./build rpm
```

Fixes #2748.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,8 @@ addons:
             - fakeroot
             - dpkg
             - lintian
+            # Stuff for building .rpm files
+            - rpm
 
 # These commands simulate having a graphical display, which is needed
 # for our GUI tests.
@@ -51,6 +53,7 @@ script:
 before_deploy:
     - ./build osx
     - ./build deb
+    - ./build rpm
     - echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
 
 deploy:
@@ -64,6 +67,7 @@ deploy:
         - _build/repack/$BUILD_CONFIGURATION/ckan.exe
         - _build/osx/CKAN.dmg
         - _build/deb/ckan_*.deb
+        - _build/rpm/RPMS/noarch/ckan*.rpm
         - _build/out/AutoUpdater/$BUILD_CONFIGURATION/bin/AutoUpdater.exe
       on:
         repo: KSP-CKAN/CKAN
--- a/build.cake
+++ b/build.cake
@@ -81,6 +81,20 @@ Task("deb-clean")
     .Does(() => StartProcess("make",
         new ProcessSettings { Arguments = "clean", WorkingDirectory = "debian" }));
 
+Task("rpm")
+    .IsDependentOn("Ckan")
+    .Does(() => StartProcess("make",
+        new ProcessSettings { WorkingDirectory = "rpm" }));
+
+Task("rpm-test")
+    .IsDependentOn("Ckan")
+    .Does(() => StartProcess("make",
+        new ProcessSettings { Arguments = "test", WorkingDirectory = "rpm" }));
+
+Task("rpm-clean")
+    .Does(() => StartProcess("make",
+        new ProcessSettings { Arguments = "clean", WorkingDirectory = "rpm" }));
+
 Task("Restore-Nuget")
     .Does(() =>
 {
--- a/rpm/Makefile
+++ b/rpm/Makefile
@@ -1,0 +1,30 @@
+.PHONY: clean test
+
+# Borrow files from .deb
+DEBDIR:=$(shell pwd)/../debian
+
+TOPDIR:=$(shell pwd)/../_build/rpm
+SCRIPTSRC:=$(DEBDIR)/ckan
+MANSRC:=$(DEBDIR)/ckan.1
+DESKTOPSRC:=$(DEBDIR)/ckan.desktop
+ICONSRC:=$(DEBDIR)/ckan.ico
+EXESRC:=$(shell pwd)/../_build/repack/Release/ckan.exe
+CHANGELOGSRC:=../CHANGELOG.md
+VERSION:=$(shell egrep '^\s*\#\#\s+v.*$$' $(CHANGELOGSRC) | head -1 | sed -e 's/^\s*\#\#\s\+v//' )
+RPM:=$(shell pwd)/../_build/rpm/RPMS/noarch/ckan-$(VERSION)-1.noarch.rpm
+
+# rpmbuild tries to use $HOME/rpmbuild by default
+# rpmbuild can't handle relative paths for its topdir
+# rpmbuild assumes we'll copy files into its SOURCES dir for it
+
+$(RPM): $(SCRIPTSRC) $(EXESRC) $(MANSRC) $(DESKTOPSRC) $(ICONSRC)
+	mkdir -p "${TOPDIR}/SOURCES"
+	cp $^ "${TOPDIR}/SOURCES"
+	rpmbuild --define "_topdir ${TOPDIR}" --define "_version $(VERSION)" -bb ckan.spec
+
+clean:
+	rm -r "$(TOPDIR)"
+
+test: $(RPM)
+	rpm -qlp $<
+	rpmlint $<
--- a/rpm/ckan.spec
+++ b/rpm/ckan.spec
@@ -1,0 +1,48 @@
+Name: ckan
+Version: %{_version}
+Release: 1%{?dist}
+Summary: Mod manager for Kerbal Space Program
+URL: https://ksp-ckan.space
+Packager: The CKAN authors <rpm@ksp-ckan.space>
+License: MIT
+AutoReqProv: no
+Requires: mono-core >= 5.0
+BuildArch: noarch
+Source0: ckan
+Source1: ckan.exe
+Source2: ckan.1
+Source3: ckan.desktop
+Source4: ckan.ico
+
+%description
+KSP-CKAN official client.
+Official client for the Comprehensive Kerbal Archive Network (CKAN).
+Acts as a mod manager for Kerbal Space Program mods.
+
+%post
+cert-sync /etc/pki/tls/certs/ca-bundle.crt
+cert-sync --user /etc/pki/tls/cert.pem
+
+%install
+umask 0022
+mkdir -p %{buildroot}%{_bindir}
+cp %{SOURCE0} %{buildroot}%{_bindir}
+mkdir -p %{buildroot}/usr/lib/ckan
+cp %{SOURCE1} %{buildroot}/usr/lib/ckan
+mkdir -p %{buildroot}%{_mandir}/man1
+cp %{SOURCE2} %{buildroot}%{_mandir}/man1
+mkdir -p %{buildroot}%{_datadir}/applications
+cp %{SOURCE3} %{buildroot}%{_datadir}/applications
+mkdir -p %{buildroot}%{_datadir}/icons
+cp %{SOURCE4} %{buildroot}%{_datadir}/icons
+
+%files
+%{_bindir}/*
+/usr/lib/ckan
+%{_datadir}/applications/*
+%{_datadir}/icons/*
+%{_mandir}/man1/*
+
+%changelog
+* Thu May 9 2019 The CKAN authors <rpm@ksp-ckan.space> 1.26.3-1
+- Initial RPM release


HebaruSan:
Build error does not look related to the changes:

```
W: http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/trusty/Release.gpg: Signature by key 15866BAFD9BCC4F3C1E0DFC7D69548E1C17EAB57 uses weak digest algorithm (SHA1)
W: GPG error: https://packagecloud.io/github/git-lfs/ubuntu trusty InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 6B05F25D762E3157
W: The repository 'https://packagecloud.io/github/git-lfs/ubuntu trusty InRelease' is not signed.
W: There is no public key available for the following key IDs:
6B05F25D762E3157 
```

Is the build VM out of date? Maybe a temporary glitch?

HebaruSan:
@enderger, could you please try this test build and report any problems you have with it? I don't have an RPM-based distro available for testing.

- [ckan-1.26.3-1.noarch.rpm.zip](https://github.com/KSP-CKAN/CKAN/files/3163369/ckan-1.26.3-1.noarch.rpm.zip)

enderger:
I will do so at home. It may take awhile, given the upcoming potential flooding.

enderger:
YAST claims that the package is unsigned.

HebaruSan:
Can you bypass that? It's definitely unsigned, since I just compiled it for this test.

enderger:
No other issues occur during install. YAST sees the dependences and I told the package manager to ignore the "error".

enderger:
The thing loads, but seems to think that the run uses Windows 95.

enderger:
Screenshot_20190509_172307

enderger:
Test launch works

HebaruSan:
Yup, that's what WinForms apps look like. Thanks for checking it out!
Mind opening a terminal and confirming that "man ckan" prints the man page?

enderger:
Man page works:
Screenshot_20190509_172935

enderger:
Screenshot_20190509_173210

enderger:
That field's label is incomplete

HebaruSan:
Can you show us? Maybe there's a font difference that's affecting the alignment:

image

enderger:
That's the entire text string

HebaruSan:
OK, but we can't fix it if we can't see the problem :)

enderger:
Screenshot_20190509_174049

HebaruSan:
Huh; *all* of your labels are being truncated by one word. That's weird, because they all have `AutoSize` set to true, so they should all be sizing to fit the text.

image

enderger:
This is happening on KDE, I'll try with a DEB install in a VM

HebaruSan:
Out of curiosity, which version of Mono are you running (`mono --version`)? If it happens to be an old one, maybe this was a bug that got fixed at some point.

DasSkelett:
> This is happening on KDE, I'll try with a DEB install in a VM

I'm using KDE too, and it's working for me. (Not installed with the rpm, but I don't think that's the cause.)

enderger:
The one listed in the DEB. It was autoinstalled with CKAN

HebaruSan:
The .deb and .rpm packages don't specify a precise version for the mono dependency; which one is installed on your system?

enderger:
5.10.1.47

enderger:
The issue spreads to the main screen as well.
Screenshot_20190509_175649

HebaruSan:
Sorry, what are you pointing out in that last screenshot?

enderger:
Updated with a highlight

HebaruSan:
... yup, there were some fixes to label auto-sizing in March 2018 in mono/mono#7401.

[5.10.1.47](https://github.com/mono/mono/releases/tag/mono-5.10.1.47) came out in April 2018, but I think Mono doesn't necessarily release the very latest code they have based on some changes I've had accepted there, so my guess is that bug is the cause of what you're seeing.

enderger:
perhaps it should depend on this: https://en.opensuse.org/MonoDevelop

enderger:
Wait, the name was misleading

HebaruSan:
No, you don't need the development environment to run a compiled application.

enderger:
I thought it was a development version of mono. Hence why I deleted the comment

DasSkelett:
You need https://en.opensuse.org/Mono.
Current version shipped to openSUSE seems to be 5.20.1, that's up to date:
https://software.opensuse.org/package/mono-core

enderger:
Attempting to update my system to get it to work

enderger:
I'm on leap, I need to install manually

enderger:
Got the update manually. The label issue is fixed. I have found no RPM packaging bugs.

enderger:
After a few days, I have run into a few problems:
1. After realizing that Mono didn't update, I have updated Mono and the labels apparently never fixed themselves.
2. If an installation fails, CKAN completely breaks
3. Any attempt to install UnManned Before Manned crashes CKAN

DasSkelett:
> After realizing that Mono didn't update, I have updated Mono and the labels apparently never fixed themselves.

What does that mean? Which version of mono are you running now? Is the label bug fixed as you said before, or not?

> If an installation fails, CKAN completely breaks

In what sense? Does it crash? Does it throw an error message? Is the GUI frozen?

> Any attempt to install UnManned Before Manned crashes CKAN

With an error message?
Right now I obviously get the SSL trust error, because Spacedock has problems with its SSL certificate. But that's all, CKAN doesn't crash with this error.
I can return to the modlist, deselect UbM and continue.

enderger:
I have managed to bump it up to 5.20.1.19(I tried the 1-click install, but Opera didn't want to cooperate and I, being a newbie when it comes to 1-click installs, was not aware of the issue. After finding out about the issue of installation, I went back in to find that the labels were not as fixed as I first thought. After finding a workaround to 1-click installs on Opera, I managed to update. The labels seem to be a bit more broken after the update.
image

As for the CKAN completely breaks, I mean that any attempts to install after the failure results in the software locking up on an "install failed" screen. It also requires using a kill command to close CKAN.

Finally, the crash with Unmanned Before Manned. When CKAN reaches installing the aforementioned mod, CKAN just closes out, no questions asked.

DasSkelett:
> The labels seem to be a bit more broken after the update.

Is the last word still cut off from the labels? Or is it only the dropdown button that's misbehaving now?

> As for the CKAN completely breaks, I mean that any attempts to install after the failure results in the software locking up on an "install failed" screen. It also requires using a kill command to close CKAN.

That's weird. It's not happening on my system.

> EDIT: The UBM mod issue seems to have fixed itself

This indicates that it in fact had something to do with Spacedocks SSL error. CKAN shouldn't crash though, just show an error message.

enderger:
1. All labels still seem to be broken and a few buttons misplace themselves.
2. The crash could be the old version of Mono

HebaruSan:
OK, the questions relevant to this PR are:

1. Are @enderger's issues general issues affecting many people or quirks of his PC?
2. Do they have anything to do with how we built the RPM package?
3. Is Mono's packaging more broken or out of date on RPM distros generally?
4. Is that a reason not to release an RPM package?

My guesses:

1. General
2. No
3. Maybe
4. Maybe

In other words, shipping an RPM package may bring an influx of reports of issues caused by RPM distros themselves rather than by CKAN. Probably best to hold this back till we're sure.

enderger:
I am an OpenSUSE Leap user. This means that I have older packages than some other distros. This also is the case for Debian Stable. I see that Tumbleweed probably could run this better.

enderger:
Due to poor Mono support on OpenSUSE and a few stability concerns, I probably will move to Debian Testing. I will test CKAN on my laptop, which runs Fedora. I can monitor CKAN issues by having a dummy KSP install, as that’s enough to test.

enderger:
My home folder is on it’s own partition, so I should keep my data

DasSkelett:
Quick update for mono on RPM-based distros:

**openSUSE Leap 15.1** [has been released today](https://news.opensuse.org/2019/05/22/opensuse-community-releases-leap-15-1-version/), but they're [still stuck with mono 5.10.1](https://software.opensuse.org/package/mono-core). However, there's a [community package for version 5.20.1](https://build.opensuse.org/package/show/home%3AMihailJP/mono-core) available.

**Fedora** [distributes mono 5.18.1](https://apps.fedoraproject.org/packages/mono-core), that's not too bad.

I think those are the two main distros we should be worried about regarding rpm/mono/ckan compatibility.

This means, if we release a rpm package in the near future, we have to make sure to advise openSUSE Leap users to install a newer version of mono than the main repo offers, for example via the community package.

enderger:
Perhaps integrating an outdated mono warning for earlier mono releases will be good for all versions, mainly for Debian Stable users

HebaruSan:
We have a wiki page for Fedora, but not openSUSE.

https://github.com/KSP-CKAN/CKAN/wiki/Installing-CKAN-on-Fedora

It's pretty out of date though, talks about Mono **2 and 3**! Updating it and adding a page for openSUSE would be useful anyway, and could make an RPM package easier to support. Any volunteers?

HebaruSan:
I've attempted to make the needed wiki updates:

- https://github.com/KSP-CKAN/CKAN/wiki/Installing-CKAN-on-Fedora
- https://github.com/KSP-CKAN/CKAN/wiki/Installing-CKAN-on-openSUSE

It's been almost 20 years since I've used Red Hat or seen openSUSE though, so errors are likely and review would be appreciated. In particular I only know about the `zypper` command from internet searches for how to install software on openSUSE.

DasSkelett:
> I've attempted to make the needed wiki updates:
> 
>     * [KSP-CKAN/CKAN/wiki/Installing-CKAN-on-Fedora](https://github.com/KSP-CKAN/CKAN/wiki/Installing-CKAN-on-Fedora)
> 
>     * [KSP-CKAN/CKAN/wiki/Installing-CKAN-on-openSUSE](https://github.com/KSP-CKAN/CKAN/wiki/Installing-CKAN-on-openSUSE)
> 
> 
> It's been almost 20 years since I've used Red Hat or seen openSUSE though, so errors are likely and review would be appreciated. In particular I only know about the `zypper` command from internet searches for how to install software on openSUSE.

I'm trying out the installation on Fedora right now. Some interesting things:
Fedora has mono in its repos too:
```
[user@localhost]$ sudo yum --showduplicates list mono-devel
Letzte Prüfung auf abgelaufene Metadaten: vor 0:04:52 am Fr 07 Jun 2019 14:41:33 CEST.
Verfügbare Pakete
mono-devel.x86_64        5.18.0.268-0.xamarin.3.epel8        mono-centos8-stable
mono-devel.i686          5.18.1-2.fc30                       fedora             
mono-devel.x86_64        5.18.1-2.fc30                       fedora             
mono-devel.i686          5.18.1-7.fc30                       updates            
mono-devel.x86_64        5.18.1-7.fc30                       updates            
mono-devel.x86_64        5.18.1.0-0.xamarin.1.epel8          mono-centos8-stable
mono-devel.x86_64        5.18.1.3-0.xamarin.1.epel8          mono-centos8-stable
mono-devel.x86_64        5.20.1.19-0.xamarin.3.epel8         mono-centos8-stable
[user@localhost]$ sudo yum --showduplicates list mono-complete
Letzte Prüfung auf abgelaufene Metadaten: vor 0:05:19 am Fr 07 Jun 2019 14:41:33 CEST.
Verfügbare Pakete
mono-complete.x86_64      5.18.0.268-0.xamarin.3.epel8       mono-centos8-stable
mono-complete.x86_64      5.18.1-2.fc30                      fedora             
mono-complete.x86_64      5.18.1-7.fc30                      updates            
mono-complete.x86_64      5.18.1.0-0.xamarin.1.epel8         mono-centos8-stable
mono-complete.x86_64      5.18.1.3-0.xamarin.1.epel8         mono-centos8-stable
mono-complete.x86_64      5.20.1.19-0.xamarin.3.epel8        mono-centos8-stable
```

**I definitely recommend installing mono from the official Fedora repos without adding mono's own repos!**

I'm getting the libgdiplus missing error despite `libgdiplus` being installed (and even after a reboot) when installing from mono's repos:
```
[user@localhost]$ mono ckan.exe
exception inside UnhandledException handler: (null) assembly:/usr/lib/mono/4.5/mscorlib.dll type:NullReferenceException member:(null)

[ERROR] FATAL UNHANDLED EXCEPTION: System.TypeInitializationException: 
The type initializer for 'System.Windows.Forms.XplatUI' threw an exception. 
---> System.TypeInitializationException: The type initializer for 'System.Drawing.GDIPlus' threw an exception. 
---> System.DllNotFoundException: libgdiplus.so.0
```

DasSkelett:
Regarding the RPM: I'm getting the following error:
```
Error: Failed dependencies:
	mono is required by ckan-1.26.3-1.noarch
	mono(system) = 2.0.0.0 is required by ckan-1.26.3-1.noarch
	mono(System.Transactions) = 2.0.0.0 required by ckan-1.26.3-1.noarch
	mono(mscorlib) = 2.0.0.0 is required by ckan-1.26.3-1.noarch
```

Ignoring the dependencies with `rpm -i --nodeps *.rpm` works (obviously mono must be installed beforehand).

HebaruSan:
OK, changed the wiki to install via `yum`, and changed the dependency in the package from `mono` to `mono-complete`.

DasSkelett:
Not exactly a issue with the rpm, but: I wanted to build the rpm and test it, but the build fails in my Fedora VM (not only `./build rpm`, normal `./build` too):
```
Restore-Nuget
MSBuild auto-detection: using msbuild version '14.0' from '/usr/lib/mono/xbuild/14.0/bin'.

>>>> xbuild tool is deprecated and will be removed in future updates, use msbuild instead <<<<

MSBUILD: error MSBUILD0004: Too many project files specified

WARNUNG: Error reading msbuild project information, ensure that your input solution or project file is valid. NETCore and UAP projects will be skipped, only packages.config files will be restored.
```
Then all the packages are downloaded, and then:
```
Versioning/ModuleVersion.cs(193,33): error CS1525: Unexpected symbol `(', expecting `,', `;', or `='
Versioning/ModuleVersion.cs(193,45): error CS1525: Unexpected symbol `string', expecting `,', `;', or `='
Versioning/ModuleVersion.cs(195,92): error CS1525: Unexpected symbol `}', expecting `,' or `;'
Versioning/ModuleVersion.cs(301,14): error CS1519: Unexpected symbol `if' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(301,24): error CS1519: Unexpected symbol `==' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(302,54): error CS1001: Unexpected symbol `(', expecting identifier
Versioning/ModuleVersion.cs(302,27): error CS0106: The modifier `new' is not valid for this item
Versioning/ModuleVersion.cs(302,27): error CS1520: Class, struct, or interface method must have a return type
Versioning/ModuleVersion.cs(302,61): error CS1001: Unexpected symbol `)', expecting identifier
Versioning/ModuleVersion.cs(304,14): error CS1519: Unexpected symbol `if' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(304,31): error CS1519: Unexpected symbol `==' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(304,41): error CS1519: Unexpected symbol `&&' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(304,64): error CS1519: Unexpected symbol `(' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(304,64): error CS8124: Tuple must contain at least two elements
Versioning/ModuleVersion.cs(304,74): error CS1519: Unexpected symbol `)' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(308,25): error CS1519: Unexpected symbol `!=' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(308,39): error CS1519: Unexpected symbol `)' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(309,31): error CS1519: Unexpected symbol `>' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(309,48): error CS1519: Unexpected symbol `1' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(314,14): error CS1519: Unexpected symbol `if' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(314,44): error CS1519: Unexpected symbol `(' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(314,54): error CS1519: Unexpected symbol `out' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(314,63): error CS1519: Unexpected symbol `)' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(315,27): error CS1519: Unexpected symbol `;' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(318,33): error CS1519: Unexpected symbol `=' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(318,43): error CS1519: Unexpected symbol `;' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(319,34): error CS1519: Unexpected symbol `=' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(319,50): error CS1519: Unexpected symbol `;' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(322,47): error CS1519: Unexpected symbol `>' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(322,82): error CS1519: Unexpected symbol `>' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(323,13): error CS9010: Primary constructor body is not allowed
Versioning/ModuleVersion.cs(350,14): error CS1519: Unexpected symbol `if' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(350,45): error CS1519: Unexpected symbol `==' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(351,13): error CS9010: Primary constructor body is not allowed
Versioning/ModuleVersion.cs(351,13): error CS8041: Primary constructor already has a body
Versioning/ModuleVersion.cs(363,35): error CS1519: Unexpected symbol `(' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(363,43): error CS1519: Unexpected symbol `1' in class, struct, or interface member declaration
Versioning/ModuleVersion.cs(393,15): error CS1525: Unexpected symbol `bool', expecting `(' or `type'
Versioning/ModuleVersion.cs(395,12): error CS1525: Unexpected symbol `return', expecting `(' or `type'
Versioning/ModuleVersion.cs(396,9): error CS1514: Unexpected symbol `}', expecting `.' or `{'
Versioning/ModuleVersion.cs(424,15): error CS1525: Unexpected symbol `bool', expecting `(' or `type'
```

Using
```
XBuild Engine Version 14.0
Mono, Version 5.18.1.3
```

On my main/host system I can't build the rpm because I'm missing rpmbuild, but otherwise it obviously works fine.
@HebaruSan do you by any chance know how to fix that error/what's wrong with the setup? I can't find anything useful online.

HebaruSan:
That compile error is happening at a point where we declare a function inside another function (as of #2326), which supposedly is a new feature of C# 7 (and Mono 5):

https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#local-functions
https://www.mono-project.com/docs/about-mono/releases/5.0.0/#c-7-adds-a-number-of-new-features-to-the-c-language

Those release notes say that `csc` supports C#7 and `mcs` doesn't. Can you confirm which is being used?

DasSkelett:
Yep, it's mcs:
```
Tool /usr/lib/mono/4.5/mcs.exe execution started with arguments: [...]
`msc``

**Comment by HebaruSan** [comment]
Is `csc` installed? Maybe it's falling back to `mcs` because it can't find `csc`?

**Comment by DasSkelett** [comment]
Nope, `locate` can't find anything. So the Fedora version of mono seems to lack csc...

**Comment by HebaruSan** [comment]
That's the thing to try to solve, then.

This page makes it sound like Fedora has an aversion to `csc` because it uses binary reference assemblies:

https://fedoraproject.org/wiki/Changes/Mono_5

> Upgrading to Mono 5 has been delayed because Mono 5 compiles by default with csc rather than msc, and makes use of binary reference assemblies.
> 
> Only in the past months an effort was made to build Mono from source, as described in this upstream issue: https://github.com/mono/mono/issues/7445 This issue contains a description how the build was done for Debian so that building from itself is possible, using msc instead of csc, and rebuilding the reference assemblies from source.

... but that only apply to the process for compiling Mono itself. I'm not finding any other references to `csc` being missing on Fedora.

**Comment by DasSkelett** [comment]
That button is definitely too big for me!
Sorry :/

**Comment by DasSkelett** [comment]
Looking at `yum whatprovides "*csc.exe"` only the following package includes csc.exe:
```
[user@localhost CKAN]$ yum whatprovides "*csc.exe"
wine-mono-4.8.3-1.fc30.noarch : Mono library required for Wine
Source      : updates
Matched by:
Others : *csc.exe
```
And that seems to be a release of mono especially for wine:
```
[user@localhost CKAN]$ ls /usr/share/wine/mono/wine-mono-4.8.3/lib/mono/
2.0-api  4.0      4.5        4.5.2-api  4.6.1-api  4.6-api    4.7.2-api  gac   mono-configuration-crypto  msbuild  xbuild-frameworks
3.5-api  4.0-api  4.5.1-api  4.5-api    4.6.2-api  4.7.1-api  4.7-api    lldb  monodoc                    xbuild
```

**Comment by HebaruSan** [comment]
Does Fedora have a bug tracker? They might want to know that their OS can't compile C# 7.

**Comment by DasSkelett** [comment]
Looking around it seems to be https://bugzilla.redhat.com/
or for mono rpms: https://bugzilla.redhat.com/buglist.cgi?classification=Fedora&component=mono&product=Fedora.
I'll try my luck.

**Comment by DasSkelett** [comment]
[Request made](https://bugzilla.redhat.com/show_bug.cgi?id=1718439).
Let's see how that turns out.

At least this "only" affects compiling CKAN, not running it ^^

**Comment by DasSkelett** [comment]
[Bugreport closed:](https://bugzilla.redhat.com/show_bug.cgi?id=1718439#c1)
> The problem is, that we cannot build csc.exe from source yet. This would require Roslyn, msbuild etc. I hope when the effort to bring .NET Core into Fedora succeeds, then msbuild and csc will be available in Fedora as well. [...]
> I cannot find the exact sentence in the packaging guidelines at the moment, but I know we are not allowed to use binaries, that have not been built in Fedora, except for bootstrapping.

I think we both kinda expected this.

But it sounds like there's something going on to bring .NET Core to Fedora, this might allow them to build csc from source.

So I myself now go ahead and install rpm on my main system, this should allow me to build rpms. Then I can test it in the Fedora VM.
For possible future contributors/users who want to build the rpm themselves,
should I add a note to https://github.com/KSP-CKAN/CKAN/wiki/Build-Environment?

**Comment by DasSkelett** [comment]
Okay this is weird.
```
user@localhost CKAN]$ sudo rpm -i *.rpm
Error: Failed dependencies:
	mono(system) = 2.0.0.0 is required by ckan-1.26.3-1.noarch
	mono(System.Transactions) = 2.0.0.0 required by ckan-1.26.3-1.noarch
	mono(mscorlib) = 2.0.0.0 is required by ckan-1.26.3-1.noarch
```
So the only change is that `mono` is no longer on that list.
Querying the dependencies of the rpm file returns this (arrows added by me):
```
[user@localhost CKAN]$ sudo rpm -qp --requires *.rpm
/bin/sh
mono(Microsoft.CSharp) = 4.0.0.0
mono(System) = 2.0.0.0                <---------
mono(System) = 4.0.0.0
mono(System.Configuration) = 4.0.0.0
mono(System.Core) = 4.0.0.0
mono(System.Data) = 4.0.0.0
mono(System.Drawing) = 4.0.0.0
mono(System.Numerics) = 4.0.0.0
mono(System.Runtime.Serialization) = 4.0.0.0
mono(System.Transactions) = 2.0.0.0    <---------
mono(System.Transactions) = 4.0.0.0
mono(System.Web) = 4.0.0.0
mono(System.Windows.Forms) = 4.0.0.0
mono(System.Xml) = 4.0.0.0
mono(System.Xml.Linq) = 4.0.0.0
mono(mscorlib) = 2.0.0.0               <---------
mono(mscorlib) = 4.0.0.0
mono-complete
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
```
So it is asking two times for those three libraries, one time an older version.
I can't read anything like this out of `rpm/ckan.spec`...

**Comment by HebaruSan** [comment]
OK, I set up a Fedora client under VirtualBox. Any idea how to make it mount the shared folder I configured so I can access the generated RPM file? Host:

image

Guest:

```
[osboxes@localhost log]$ lsmod|grep vbox
vboxguest              45056  4
vboxvideo              40960  2
drm_kms_helper        212992  1 vboxvideo
ttm                   114688  1 vboxvideo
drm                   491520  5 drm_kms_helper,vboxvideo,ttm
[osboxes@localhost log]$ ls /ckan
[osboxes@localhost log]$ mount|grep ckan
[osboxes@localhost log]$ sudo mount -t vboxsf CKAN /ckan
/sbin/mount.vboxsf: mounting failed with the error: No such device
[osboxes@localhost log]$ sudo mount -t vboxsf ckan /ckan
/sbin/mount.vboxsf: mounting failed with the error: No such device
```

**Comment by DasSkelett** [comment]
> OK, I set up a Fedora guest under VirtualBox. Any idea how to make it mount the shared folder I configured so I can access the generated RPM file?

I have no idea. I tried everything you mentioned too, and the folder is always created, but it's empty.
Neither does copy-pasting files work, only text.
VB guest additions seem to be installed, and clipboard/drag-and-drop are set to bidirectional, but nothing of that works.

I ended up using the file transfer tool of VirtualBox (under Machine \> File Manager or something like that).

**Comment by HebaruSan** [comment]
The file manager's Create Session button does nothing for me, not even an error message. :(

**Comment by DasSkelett** [comment]
> The file manager's Create Session button does nothing for me, not even an error message. :(

With your guest system credentials entered? Hmm...
Guest additions are installed? (Should be by default with the normal Fesora Workstation edition)

**Comment by HebaruSan** [comment]
OK, I may have broken it by installing from my VirtualBox's ISO. `/var/log/vboxadd-setup.log` says:

```
VBoxClient: error while loading shared libraries: libcrypt.so.1: cannot open shared object file: No such file or directory
```

... and presumably the built-in version would be linked correctly.

**Comment by HebaruSan** [comment]
Ah! I had the username wrong! It's "osboxes", but it's displayed as "osboxes.org" in most places so I was trying that. File manager working now.

**Comment by HebaruSan** [comment]
```
user@localhost CKAN]$ sudo rpm -i *.rpm
Error: Failed dependencies:
	mono(system) = 2.0.0.0 is required by ckan-1.26.3-1.noarch
	mono(System.Transactions) = 2.0.0.0 required by ckan-1.26.3-1.noarch
	mono(mscorlib) = 2.0.0.0 is required by ckan-1.26.3-1.noarch
```

I believe those are "automatic dependencies":

https://rpm.org/user_doc/more_dependencies.html

> Automatic Dependencies
> 
> To reduce the amount of work required by the package builder, RPM scans the file list of a package when it is being built. Any files in the file list which require shared libraries to work (as determined by ldd) cause that package to require the shared library.

So `rpmbuild` is scanning _a Debian system_ for things it might need on Fedora and confusing the heck out of itself in the process. I've just pushed a commit that hopefully turns that off (and depends on `mono-core` instead, since I think we're not supposed to depend on `mono-complete` just like on Debian). You might need to run `./build rpm-clean` to make sure these changes are applied. I can now install the CKAN RPM, but I get certificate errors running it, so I'm going to see whether the old solutions to that help...

HebaruSan:
Added a post-install script to sync the certs, since that fixed the errors I was seeing.

DasSkelett:
Yep, I can give a go for the rpm on Fedora.
Now installs fine even without preinstalled mono via the Software application and/or `yum`.
The rpm command can't install dependencies for you, but a clear `needs mono-core >= 5.0` is returned, so everybody using that should now what to do.

enderger:
If issues persist, maybe opensuse may have better luck(Tumbleweed seems up to date)

HebaruSan:
The issues are fixed on Fedora now, but testing on openSUSE as well is a good idea, downloading VM now...

HebaruSan:
@enderger, your display issues are caused by font hinting.

https://stackoverflow.com/a/445063/2422988

I saw the same problem initially on my openSUSE VM. I installed Fontweak and unchecked this checkbox:

image

And now it works as expected:

image

HebaruSan:
It seems that changing the hint style to full works as well:

image

The default for me was a hint style of 'none' with the checkbox enabled. I suspect that Mono auto sizes its labels based on hinting, but when it renders them the hint style 'none' makes the font larger than expected.

enderger:
Just made a bug report

HebaruSan:
So @enderger have you confirmed that changing the setting to hintfull fixes it for you?

I already have a copy of the mono source, so I might look into submitting a fix for this...

enderger:
I had to change to a different linux distro after my wifi driver went kaput

enderger:
The labels work on the new install, though

HebaruSan:
@enderger So what distro are you using now? Still RPM-based? Are you installing via the RPM?

enderger:
I went with mint, as it works with my wifi card more easily than rpm distros. I can’t move off until the card driver gets better

Olympic1:
@enderger Have you had some time to test this?

HebaruSan:
@Olympic1, I think he switched to a non-RPM distro (Linux Mint is based on Ubuntu is based on Debian), so he can't test this anymore. It looked good on my Fedora and openSUSE VMs though.

PR #2759, opened 2019-05-10
Add scrollbars to metadata tab

DasSkelett:
## Problem
The metadata is the only tab of the mod info panel that doesn't display scrollbars if its content is to big.
On low-res / small monitors or if you don't want to shrink your modlist / upper mod info panel some mod infos may be out of screen.

## Solution
Make the tab scrollable, horizontally and vertically.
This also fixes the problem that links are mostly unreadable. Now they are all oneliners and if they are longer than the panel is wide, a scrollbar appears.

grafik

### DIFF
--- a/GUI/MainModInfo.Designer.cs
+++ b/GUI/MainModInfo.Designer.cs
@@ -194,8 +194,8 @@
             //
             this.MetaDataLowerLayoutPanel.AutoSizeMode = System.Windows.Forms.AutoSizeMode.GrowAndShrink;
             this.MetaDataLowerLayoutPanel.ColumnCount = 2;
-            this.MetaDataLowerLayoutPanel.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle(System.Windows.Forms.SizeType.Percent, 26.16279F));
-            this.MetaDataLowerLayoutPanel.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle(System.Windows.Forms.SizeType.Percent, 73.83721F));
+            this.MetaDataLowerLayoutPanel.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle(System.Windows.Forms.SizeType.Absolute, 90F));
+            this.MetaDataLowerLayoutPanel.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle(System.Windows.Forms.SizeType.AutoSize));
             this.MetaDataLowerLayoutPanel.Controls.Add(this.VersionLabel, 0, 0);
             this.MetaDataLowerLayoutPanel.Controls.Add(this.LicenseLabel, 0, 1);
             this.MetaDataLowerLayoutPanel.Controls.Add(this.AuthorLabel, 0, 2);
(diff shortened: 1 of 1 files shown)


HebaruSan:
Code looks good and works well. I thought there was a consensus that horizontal scrolling is bad, but I was able to find few references to it other than this, and several contradicting it, so I guess that's fine, too:

https://www.nngroup.com/articles/scrolling-and-scrollbars/

DasSkelett:
> I thought there was a consensus that horizontal scrolling is bad

I think it's not optimal, but I don't see good alternatives here.
Line wrapping makes links rather unreadable in my opinion, at the latest when they are cut because they don't fit in the available horizontal space.
Also the way links are divided for the line breaking is catastrophic in most cases, something like
```
www.github.
com/KSP-CKAN/CKAN/pull/
2759
```
is not unusual.

And we already have a big area with horizontal scrolling \> the modlist.

HebaruSan:
Yeah, I gave up when I got to "users have to move their viewport in two dimensions, which makes it hard to cover the entire space" on that page. Covering the entire space is not a consideration. Users will see the labels and the start of the content, and in most cases they won't scroll at all.

HebaruSan:
```
1) Failed : Tests.Core.Net.Net.SpaceDockSSL
  But was:  <System.Net.WebException: The request timed out
  at System.Net.WebClient.DownloadFile (System.Uri address, System.String fileName) [0x000ad] in <1a0168907d2042858c12ff7b2c118a26>:0 
  at System.Net.WebClient.DownloadFile (System.String address, System.String fileName) [0x00016] in <1a0168907d2042858c12ff7b2c118a26>:0 
  at (wrapper remoting-invoke-with-check) System.Net.WebClient.DownloadFile(string,string)
  at CKAN.Net.Download (System.String url, System.String& etag, System.String filename, CKAN.IUser user) [0x00192] in <6983702b63f744fdb5439a57982a2c26>:0 
  at CKAN.Net.Download (System.String url, System.String filename, CKAN.IUser user) [0x00001] in <6983702b63f744fdb5439a57982a2c26>:0 
  at Tests.Core.Net.Net+<>c.<SpaceDockSSL>b__5_0 () [0x00001] in <44c9297a81f343d385cefdb39bfdc841>:0 
  at NUnit.Framework.Constraints.ThrowsConstraint+VoidInvocationDescriptor.Invoke () [0x00000] in <64cd73ddfc8a470aa513c8243b429bdb>:0 
  at NUnit.Framework.Constraints.ThrowsConstraint+ExceptionInterceptor.Intercept (System.Object invocation) [0x00047] in <64cd73ddfc8a470aa513c8243b429bdb>:0 >
  at Tests.Core.Net.Net.SpaceDockSSL () [0x0000a] in <44c9297a81f343d385cefdb39bfdc841>:0 
```

https://forum.kerbalspaceprogram.com/index.php?/topic/184383-spacedock-isnt-working/

DasSkelett:
Our status page looks quite... disturbing right now.
BTW, what are all those
```
Error reading JArray from JsonReader. Current JsonReader item is not an array: StartObject. Path '', line 1, position 1.
```
errors that come and go for some time now @HebaruSan?
I tested some of those netkans and they run fine locally.

HebaruSan:
I have no idea. I tried the same thing with the same results.

Issue #2113, opened 2017-09-21
[Feature] Mods Language Filter


yalov:
Developers introduced many languages, some mod already has user's target language, some hasn't.

I propose:

1. add languages to metadata specification, and show to user at the mod versions panel
2. - introduce language filter (it shows only target language mods) and/or
   - include languages to separate column.  
For example, if user checks target language, then target-language column will appear with [×]/[  ] values (sorting supported). English language mods still is showed. I assume that most users are mostly interested in only one target language.

HebaruSan:
It might be worth localizing CKAN itself first. A user who prefers a non-English locale for mods might also have difficulty navigating an English-only application, and the same setting could drive both the CKAN display and the mod filter.

politas:
So far, the only non-English mods we have are translations, and there are only a handful of them. Localizing CKAN sounds like a pretty big task.

yalov:
There are plenty of not-only-English mods, list at forum has a tendency to become outdated. 
https://forum.kerbalspaceprogram.com/index.php?/topic/158018-addon-localization-home/

Introduce it to SCAN makes it up-to-date at CKAN.

yalov:
@politas  if you want "start using the Mod Versions panel to give us extra functionality" from #2154 reconsider 1. there

HebaruSan:
- CKAN's GUI was localized in #2749, including a German translation.
- We've been collecting the locales of modules in metadata since #2760.
- #2736 would provide a place to put this in the UI.

Issue #2736, opened 2019-04-19
Master Search box


yalov:
in addition to https://github.com/KSP-CKAN/CKAN/pull/2709 I propose to merge the search boxes to one, and to add checkboxes for backward ability to leave only one search type.
Search: [----------------]        [x] by name         [x] by author name          [x] by description

HebaruSan:
Note that that isn't backwards compatible because it doesn't support combined searches using multiple boxes:

![image](https://user-images.githubusercontent.com/1559108/56448012-7a51b800-62fb-11e9-8977-1e342d8d6ac7.png)

... so we'd inevitably have somebody moaning about how we broke their work flow.

yalov:
=)
Search: [----------------]      [x] by name     [x] by author name      [x] by description     [x] by LGG

or another way to settle this is supporting & and | relation:
linux&exten

DasSkelett:
Should your first proposal be a dropdown selection box? We can go with a normal search text box then, like now.

The second one is too complex in my opinion. Users don't want to learn a complete search system just to search a mod in CKAN. Also I don't think an "Or |" relationship makes much sense here.

yalov:
it is usual search text box, the same as now, but searching all fields at once.

Someone indeed may want to search mods by name and author, and I would prefer the name and the desc have one search box, so maybe something like this?

filter by name: [--------------]   [x] include by description                      filter by author: [-----------]

DasSkelett:
Sorry, i was very unclear. I meant the last option "[x] by LGG"

yalov:
it is a checkbox (and a joke) because LGG is the only modder, that have enough mods for inner search

DasSkelett:
Okay, didn't get it...
I can support this compromise:
``` 
filter by name: [--------]   [x] also search description and abstract        filter by author: [--------]
```
I don't think there's no workflow broken, but it would simplify the search.

HebaruSan:
As a case study, GMail has a plain search box with a dropdown icon:

![image](https://user-images.githubusercontent.com/1559108/57077193-edcfce00-6cda-11e9-88b9-dbd45ec7d0bd.png)

If you click the dropdown, advanced options are revealed:

![image](https://user-images.githubusercontent.com/1559108/57077252-16f05e80-6cdb-11e9-80f4-2f5624774f1a.png)

[...]

Issue #538, opened 2014-12-09
Conditional relationships


Ippo343:
I have been thinking about the possibility of having conditional relationships, specifically a conditional recommends or suggests.

Example usage: Entropy is an extension for Dang It that also has a secondary module called EntropyRealChute. The latter obviously depends on RealChute: so the main module Entropy cannot recommend EntropyRealChute, or RealChute becomes selected by default even for users that did not mean to install that.

It would be cool, in this case, to be able to recommend EntropyRealChute only if RealChute is already installed or about to be installed: this way, it's automatically selected for the users to whom it applies and remains hidden for those that don't need it.

Ideally, it would look something like this:

```
"recommends" [
    {
        "name": "EntropyRealChute",
        "if_installed": "RealChute"
    }
]
```

cake-pie:
Being able to specify "if Mod A & Mod B then (recommend/suggest) Mod C" would be tremendously useful, especially if we can define "mod" loosely to include extensions, compatibility patches, etc.

Would like to note that the logic should not be confined to only "when installing Mod A, (recommend/suggest) C if B is also installed". Suppose the example given above is reversed, so Entropy (A) is already installed, then the user adds RealChute (B) -- it is also appropriate to recommend EntropyRealChute (C) at that juncture.

Want to support this feature with several additional use cases:

---

### 1. Indicator Lights & Indicator Lights Community Extensions
Indicator Lights uses patches to add its lights to existing parts, comes with configs for stock parts.  
IndicatorLights Community Extensions contains patches for parts of other mods, contributed by the community,

e.g. End user has (or about to install) Indicator Lights (A) and MKS (B). Strong recommendation for grabbing `IndicatorLights_MKS.cfg` (C) from ILCE.

---

### 2. Connected Living Space configs
- CLS (Mod A) allows specifying behavior of each part via config
- Some part mods (Mods B) bundle their own CLS configs (C) for their own parts
- CLS bundles configs (C) for some part mods (B), but coverage is not exhaustive
  - installing CLS currently includes baggage of extra configs for mods you don't have

#### 2a
With metadata set up properly, CKAN can selectively install from among the configs (C) that are bundled with CLS (A) only when the end user also has the corresponding mods (B) installed. 

#### 2b
Where neither CLS (A) nor the part mod (B) provides CLS configs, a third party could create the necessary config (C) and make it available separately.

---

### 3. Connected Living Space compatibility for parts with dynamic crew capacity
[...]

PR #2760, opened 2019-05-11
Extract locales from downloads

HebaruSan:
## Motivation

#2113 suggested filtering mods by language. To achieve that, we would need to acquire metadata tracking modules' translations. That would be pretty tedious to do manually.

## Changes

Now Netkan will examine the .cfg files in each download to see if they have translation data using some fancy regular expressions. If any translations are found, their locale names will be added to the module's `localizations` property. This is a list of strings that flows through to the `CkanModules.localizations` property in the application and could be used to implement displays and filters in the future. No UI changes are made yet.

I'm mainly seeking feedback on this at this point. Did I miss anything that would prevent this from working? Would we need additional data to implement filters?

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -118,6 +118,12 @@
                 "pattern" : "^[a-z0-9-]+$"
             }
         },
+        "localizations" : {
+            "description" : "A list of KSP locale strings for which this mod includes localizations",
+            "type"        : "array",
+            "items"       : { "type" : "string" },
+            "uniqueItems" : true
+        },
         "depends" : {
             "description" : "Optional list of dependencies",
             "$ref"        : "#/definitions/relationship"
--- a/Core/Types/CkanModule.cs
+++ b/Core/Types/CkanModule.cs
@@ -431,6 +431,9 @@ namespace CKAN
 
         [JsonProperty("install", NullValueHandling = NullValueHandling.Ignore)]
         public ModuleInstallDescriptor[] install;
+        
+        [JsonProperty("localizations", NullValueHandling = NullValueHandling.Ignore)]
+        public string[] localizations;
 
         // Used to see if we're compatible with a given game/KSP version or not.
         private readonly IGameComparator _comparator;
--- a/Netkan/CKAN-netkan.csproj
+++ b/Netkan/CKAN-netkan.csproj
@@ -111,6 +111,7 @@
     <Compile Include="Transformers\InternalCkanTransformer.cs" />
     <Compile Include="Transformers\ITransformer.cs" />
     <Compile Include="Transformers\JenkinsTransformer.cs" />
+    <Compile Include="Transformers\LocalizationsTransformer.cs" />
     <Compile Include="Transformers\MetaNetkanTransformer.cs" />
     <Compile Include="Transformers\NetkanTransformer.cs" />
     <Compile Include="Transformers\OptimusPrimeTransformer.cs" />
--- a/Netkan/Services/IModuleService.cs
+++ b/Netkan/Services/IModuleService.cs
@@ -1,5 +1,7 @@
-﻿using CKAN.NetKAN.Sources.Avc;
+using System.Collections.Generic;
+using ICSharpCode.SharpZipLib.Zip;
 using Newtonsoft.Json.Linq;
+﻿using CKAN.NetKAN.Sources.Avc;
 
 namespace CKAN.NetKAN.Services
 {
@@ -8,5 +10,8 @@ namespace CKAN.NetKAN.Services
         AvcVersion GetInternalAvc(CkanModule module, string filePath, string internalFilePath = null);
         JObject GetInternalCkan(string filePath);
         bool HasInstallableFiles(CkanModule module, string filePath);
+        
+        IEnumerable<InstallableFile> GetConfigFiles(CkanModule module, ZipFile zip);
+        
     }
 }
--- a/Netkan/Services/ModuleService.cs
+++ b/Netkan/Services/ModuleService.cs
@@ -1,4 +1,5 @@
 ﻿using System.IO;
+using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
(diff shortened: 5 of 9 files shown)


DasSkelett:
Feedback-wise I can only attest that this works with every mod I tested so far, and that it is a cool feature :D.

For a later GUI implementation, I think we have to redo our search/filter structure soon.
Either like [you suggested in #2736](https://github.com/KSP-CKAN/CKAN/issues/2736#issuecomment-488666214) and include the locale filter in the dropdown,
or create a completely new dialog for advanced searches, and move everything but a simple name/identifier search box there.
But I think that discussion is better held in #2736.

HebaruSan:
It's probably better to call the property `localizations` or `translations` instead of `locales`.

- `Localizations` is the keyword for the parent node in the .cfg files:
  ```
  Localization
  {
          en-us
          {
  ```
- `locale` doesn't really explain _what_ is provided by the module, whereas `translations` implies that the module is _translated_ into those locales and `localizations` implies that it is _localized_ into them
- `locale` kind of implies that the values should conform to [ISO/IEC_15897](https://en.wikipedia.org/wiki/ISO/IEC_15897) (which KSP does not do—a dash replaces the underscore, and the country codes are lowercase and sometimes omitted):
  https://en.wikipedia.org/wiki/Locale_(computer_software)
  > On POSIX platforms such as Unix, Linux and others, locale identifiers are defined by ISO/IEC 15897

cake-pie:
> Now Netkan will examine the .cfg files in each download to see if they have translation data using some fancy regular expressions. ... I'm mainly seeking feedback on this at this point. Did I miss anything that would prevent this from working?

1. What if a mod has packaged its localization files as optional downloads in separate file(s)?  Will netkan bot look for them automatically (on github, spacedock, etc.) or will additional metadata be required in  `*.netkan`s to tell it where to look (and download)?

2. Will this support cases where a localization patch requires replacing certain files from the main download?  Essentially this would involve separate downloads as per (1) above -- you'd install the main package first, and then run a localization patch that might overwrite some of the original files.

    Case in point, squad's localization patches do replace KSPedia files as well as some part textures that have text on them. No reason why a mod couldn't do the same.

3. While it is pretty neat to automatically scan for localization entries in `*.cfg` files, here are a couple of issues to consider:
    - some mods may have placeholder localization files which may be marked as a certain language, but actually just contain English strings, pending actual translation work. It would be misleading to advertise them as having support for the language.
    - considering that textures, kspedia, etc may also be localized, there could conceivably be mods that have such localized files, but no localization entries in `*.cfg` files. Such mods would not get picked up by the automated process, but it looks like they would still be supported, albeit the metadata would need to be explicitly manually added -- can you confirm?

HebaruSan:
> What if a mod has packaged its localization files as optional downloads in separate file(s)? Will netkan bot look for them automatically (on github, spacedock, etc.) or will additional metadata be required in *.netkans to tell it where to look (and download)?

I imagine we would set up a separate module for each of those optional downloads and mark them as recommended or suggested. Has anyone taken that approach yet? Seems like it would be a big packaging hassle.

> Will this support cases where a localization patch requires replacing certain files from the main download?

I think this would have to be handled as per below...

> albeit the metadata would need to be explicitly manually added -- can you confirm?

Yes, you could put a `localizations` property in a netkan to handle a module with localizations that can't be detected automatically. Which reminds me, such values should not be overwritten by the automated extractor, new commit pushed to account for this. Thanks for your thoughts!

cake-pie:
> I imagine we would set up a separate module for each of those optional downloads and mark them as recommended or suggested. Has anyone taken that approach yet? Seems like it would be a big packaging hassle.

AirlockPlus has had localization cfgs for other languages in a separate zip file ever since localization became a thing in KSP -- it was one of the first to have support for all four languages at the time.

The packaging is not that big of a hassle, really.  
In my case it's partly because I'm kind of a stickler about not having to download/install optional stuff, support patches, etc. unless the end-user actually needs / wants them. Though it does yield a nice benefit in that I can get stats about how many people are downloading support for other languages.

OTOH the use case where localization involves replacing textures, kspedia, etc, would probably necessitate separate downloads. As far as I can imagine it'd be most aptly handled using the main install + patch approach. Instructions for manual installs would be straightforward: "drop main install into GameData, then drop desired language patch into GameData, accept if it asks about overwriting files". CKAN installs can be configured as needed (looks like a good fit for `PROVIDES`).

The alternatives I can think of are:

1. Distribute one download that contains all the language options.
    - Manual installation requires more complex instructions that detail how to pick out and install the files for the language you want.
    - Can CKAN handle different install directives depending on desired language?
    - Potential download file bloat if many textures, etc.

2. Have completely separate installation files for each language.
    - Actually really straightforward for manual installs
    - CKAN would still have to treat each language as a separate module.
    - Excessive if only a small fraction of textures etc need replaced.

If language patches offered as separate installs are treated as separate modules, then it pretty much boils back down to one of the use cases I posted at https://github.com/KSP-CKAN/CKAN/issues/538#issuecomment-487413470  for auto-suggesting optional localization patches based on the end-user's installed KSP language.

HebaruSan:
> AirlockPlus has had localization cfgs for other languages in a separate zip file

Heh, that file is less than 3 KB. We could already create an AirlockPlus-Localization module and make it a suggestion for AirlockPlus (mental note to self...), and then this pull request's code would mark it as containing the appropriate localizations. So that use case is covered.

> Manual installation requires more complex instructions that detail how to pick out and install the files for the language you want
> - Can CKAN handle different install directives depending on desired language?

This case would probably be handled by one module per locale, sharing downloads but with different install stanzas. And since this pull request inspects the files to be installed for the given module rather than all the files in the ZIP, each such module would be tagged with the appropriate locales automatically if it contains .cfg files, or manually in the netkan if it's just other resources. So I think this is covered as well.

> Have completely separate installation files for each language.
> - CKAN would still have to treat each language as a separate module.

Yeah, that's no problem. :+1:

HebaruSan:
This is ready to merge, barring discovery of functional issues or discussion of the schema/spec changes.

PR #2762, opened 2019-05-14
Update GUI modlist if scan detects changes

### LINKED ISSUE #2761: Modlist dosen't update if new AD mods are detected
Background
----------

**CKAN Version:** 
v1.26.3

Yes!

Problem
-------

1) Open CKAN GUI
2) Install mod manually in the GameData folder
3) Click Refresh

CKAN recognizes the newly installed mod and marks it as AD.

**What happened instead?**
CKAN doesn't recognize the mod. It doesn't scan for autoinstalled **if it doesn't fetch from the metadata repo**, that points towards #2682 being the cause of this.
After a restart of CKAN the AD mods show up.

Possible solutions
------------------

1) Don't skip the GameData scan if the the download is skipped.
2) Add a way to force a refresh of the modlist / to scan the KSP directory for AD mods in the GUI, similar to the console command `scan`.

HebaruSan:
## Problem

If you install a DLL manually while GUI is running, clicking Refresh will not mark that mod as AD.

## Cause

As of #2682, we avoid refreshing the mod list if the repos have not changed.

The DLC/DLL scanning functions are fire-and-forget; they don't tell us whether their data has changed. So either we refresh everything every time, or we miss scan changes. Currently it's the latter.

## Changes

- Now `KSP.ScanGameData` returns a boolean, true if changed and false if same
- Now we call `ScanGameData` from the repo update's background worker, with a status message describing its function
- Now the background worker reports `RepoUpdateResult.Updated` if the scan found changes but the repos were the same, which will cause a mod list update

Fixes #2761.

### DIFF
--- a/Core/KSP.cs
+++ b/Core/KSP.cs
@@ -404,13 +404,17 @@ namespace CKAN
         /// Clears the registry of DLL data, and refreshes it by scanning GameData.
         /// This operates as a transaction.
         /// This *saves* the registry upon completion.
+        /// TODO: This would likely be better in the Registry class itself.
         /// </summary>
-        // TODO: This would likely be better in the Registry class itself.
-        public void ScanGameData()
+        /// <returns>
+        /// True if found anything different, false if same as before
+        /// </returns>
+        public bool ScanGameData()
         {
             var manager = RegistryManager.Instance(this);
             using (TransactionScope tx = CkanTransaction.CreateTransactionScope())
             {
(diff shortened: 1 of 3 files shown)


DasSkelett:
Ha, was just about to ask thy you made `RegisterDll()` a bool ;)

HebaruSan:
Early draft, forgot to un-add from the commit. :)

PR #2768, opened 2019-05-25
Detect Breaking Ground DLC

### LINKED ISSUE #2765: New DLC "Breaking Ground" will be out soon
https://forum.kerbalspaceprogram.com/index.php?/topic/184279-kerbal-space-program-breaks-ground-with-a-new-dlc/

- Surface features
- Placeable over-time experiments
- Robotic parts

Estimated release May 30. Cost $14.99.

We'll need a follow-up to #2326 to detect this. I think it's safe to say that there's more ambient excitement about this than the previous DLC, so my guess is there'll be more mods depending on it.

dbent:
(Probably) Fixes #2765

Refactors the common logic of `MakingHistoryDlcDetector` into a `StandardDlcDetectorBase` and creates a new `BreakingGroundDlcDetector`.

Quoting from the documentation comment:

> Base class for DLC Detectors that follow standard conventions.
>
> "Standard conventions" is defined as detecting installation by the presence of directory with the name `IdentifierBaseName` in the `[GameData]/SquadExpansion` directory, detecting version by parsing a version line in a `readme.txt` file in the same directory, and having an identifier of `IdentifierBaseName-DLC`.

Assuming Breaking Ground in fact does follow these conventions then this should be all that's required to support it. This can probably be merged after Breaking Ground is released and we confirm that it does follow the conventions.

### DIFF
--- a/Core/DLC/BreakingGroundDlcDetector.cs
+++ b/Core/DLC/BreakingGroundDlcDetector.cs
@@ -1,0 +1,13 @@
+﻿using System.Collections.Generic;
+
+namespace CKAN.DLC
+{
+    /// <summary>
+    /// Represents an object that can detect the presence of the official Making History DLC in a KSP installation.
+    /// </summary>
+    public sealed class BreakingGroundDlcDetector : StandardDlcDetectorBase
+    {
+        public BreakingGroundDlcDetector()
+            : base("BreakingGround", "Serenity") { }
+    }
+}
--- a/Core/CKAN-core.csproj
+++ b/Core/CKAN-core.csproj
@@ -73,7 +73,9 @@
     <Compile Include="Converters\JsonSingleOrArrayConverter.cs" />
     <Compile Include="Converters\JsonRelationshipConverter.cs" />
     <Compile Include="DLC\IDlcDetector.cs" />
+    <Compile Include="DLC\BreakingGroundDlcDetector.cs" />
     <Compile Include="DLC\MakingHistoryDlcDetector.cs" />
+    <Compile Include="DLC\StandardDlcDetectorBase.cs" />
     <Compile Include="Exporters\BbCodeExporter.cs" />
     <Compile Include="Exporters\DelimeterSeperatedValueExporter.cs" />
     <Compile Include="Exporters\Exporter.cs" />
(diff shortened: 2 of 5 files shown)


Olympic1:
`FakeInstance` should be reworked for the new DLC so it could create a fake install with both of them installed.

https://github.com/KSP-CKAN/CKAN/blob/b996c59304dd575286e235cf978667e3150998c1/Core/KSPManager.cs#L207

The tests in `CKAN/Tests/Core/Registry/Registry.cs` shouldn't need an update if the naming convention stays the same.

HebaruSan:
Looks solid, will take it for a test drive once the DLC is released. :+1:

dbent:
@Olympic1 I'll probably add support for that this weekend.

Core/DLC/BreakingGroundDlcDetector.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace CKAN.DLC
+{
+    /// <summary>
+    /// Represents an object that can detect the presence of the official Making History DLC in a KSP installation.
+    /// </summary>
+    public sealed class BreakingGroundDlcDetector : StandardDlcDetectorBase
+    {
+        public BreakingGroundDlcDetector()
+            : base("BreakingGround", new Dictionary<string, string>()
```
HebaruSan: Apparently the folder name for this DLC is `Serenity`, so looking for a `BreakingGround` folder doesn't work. And we probably don't want to make modders depend on `Serenity-DLC`, so can we split out the identifier from the folder?

dbent: Well that's... interesting. I'll have an update tonight once I'm home.

dbent:
@HebaruSan `StandardDlcDetectorBase` now has an overloaded constructor which allows you to pass in a `directoryBaseName`. If the other ctor is used, `directoryBaseName` defaults to the same as `identifierBaseName`.

I've also removed the canonical version dictionary for... Serenity, as the first release is in fact `1.0.0` and not `1.0`.

HebaruSan:
Confirmed that it detects the DLC correctly with the latest changes. :+1: 

@dbent, you mentioned updating the fake instance stuff. Are you planning to add that to this PR? If so, I can wait for further review, but otherwise I think this is ready to merge.

dbent:
I'll do that in a separate PR.

DasSkelett:
@dbent, are you already working on the instance faking update? Else I could look into it tomorrow, if you'd prefer that.

dbent:
@DasSkelett I have not, have at it.

PR #2772, opened 2019-06-05
Unignore version controlled folders

HebaruSan:
CKAN's .gitignore file currently ignores two directories that have files committed in them:

```
bin
├── changes-since-last-release.pl
├── ckan-build-info.py
├── ckan-build.py
├── ckan_github_utils.py
├── ckan-release-maker.py
├── ckan-release-promoter.py
├── ckan-validate.py
├── close-old-support-tickets.pl
├── outdated_metadata.pl
└── requirements.txt

Screenshots
├── ckan-main-1.22.1.png
└── KSP1.18.0.png
```

`bin` is caught by the attempt to ignore Bin folders created by the compiler. Now we re-include the root-level bin folder after that, since it contains scripts rather than compiled C# code.

`Screenshots` just doesn't need to be in there. Now it's removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 /_build/
 /.vs/
-/Screenshots/
 /tools
 test-results
 *.userprefs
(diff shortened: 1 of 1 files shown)


DasSkelett:
I never really looked into that bin folder.
Interesting scripts in there :D

Issue #2770, opened 2019-05-27
Show DLC recommendations in the GUI / Improve handling in commandline


DasSkelett:
Background
----------

**CKAN Version:** 
v1.26.2

**KSP Version:** 
1.4+

Enhancement
----------
Came up in https://github.com/KSP-CKAN/NetKAN/pull/7205.
"Water Launch Sites" seems to be the first mod that recommends a DLC.
Looking at the mod, it makes sense.
It works, it is shown in the relationships tab, and nothing crashes :)
But it is not shown in the recommendations/suggestions tab.

Also, the console outputs the following on installation:
```
4246 [1] INFO CKAN.RelationshipResolver (null) - MakingHistory-DLC is recommended/suggested but it is not listed in the index,
or not available for your version of KSP.
```

Question is: do we want to show DLC recommendations in the rec.-tab?
I imagine it as just a text entry, obviously without the ability to mark it for installation.

At least the console output should be changed IMO, to a  short heads-up message, not that generic "not available" info, which is wrong in this case.

HebaruSan:
If we're going to interrupt the install flow, it would be nice if there was more of a point to it than just displaying the string "MakingHistory-DLC" somewhere and making the user click Continue.

E.g., the KSP store has some pages where you can buy MH. Getting these on-screen would give the user the option to act on the knowledge that it's recommended:

- https://www.kerbalspaceprogram.com/product/kerbal-space-program-making-history/
- https://www.kerbalspaceprogram.com/product/kerbal-space-program-making-history-steam/

HebaruSan:
Maybe we could extend the `kind` property to allow `DLC` as an option, and then build an entry in CKAN-meta to hold info for each DLC?

```json
{
    "spec_version": "v1.28",
    "identifier":   "MakingHistory-DLC",
    "name":         "Making History",
    "kind":         "DLC",
    "resources": {
        "homepage": "https://www.kerbalspaceprogram.com/product/kerbal-space-program-making-history/"
    }
}
```

That would take care of getting the identifier into the list of available modules, and we could check the `kind` property wherever special treatment was needed.

DasSkelett:
> Maybe we could extend the `kind` property to allow `DLC` as an option, and then build an entry in CKAN-meta to hold info for each DLC?
> 
> ```json
> {
>     "spec_version": "v1.28",
>     "identifier":   "MakingHistory-DLC",
>     "name":         "Making History",
>     "kind":         "DLC",
[...]

PR #2773, opened 2019-06-06
Support BreakingGround-DLC in instance faking

### LINKED ISSUE #2791: [Bug] Out-of-order checks during Fake creation
Background
----------
* Operating System: Win10
* CKAN Version: 1.26.2
* KSP Version: 1.7.2.2556

No

Problem
-------
During creation of a fake instance, the name for the fake instance is checked only after CKAN puts files (including a lock) into the destination path. Changing the name after it complains still doesn't allow for the creation of the fake at that location, because now the destination isn't blank -- and the user can't delete the registry.locked file CKAN put there without exiting CKAN.

**Steps to reproduce**
1. Make new folders named 'one' and 'two' on desktop
2. File, manage ksp instances, new ksp instance, clone or fake new instance.
3. Fake new instance, name 'test1', path into 'one'.
4. Create and switch to instance.
5. File, manage instances, new instance, clone/fake.
6. Fake new instance, name 'test1', path into 'two', press create.
7. CKAN puts files into 'two' folder, then returns error about 'name already used' or 'name taken'.
8. Change name from 'test1' to 'test2'. press create.
9. Now, CKAN complains the folder is non-empty.
10. Attempt to delete files from folder.
11. Fail, because open handle on registry.locked file.

**Expected behavior**
CKAN checks requisite conditions like name uniqueness before altering path on disk, and/or cleans up after itself (rolling back its file creations) if the creation process does not successfully complete.

DasSkelett:
Got a bit bigger than I thought.
(And it's getting bigger and bigger)

Changes
-----
Important stuff first:
Console command for faking changed!
It is now `ckan ksp fake <name> <path> <version> [--MakingHistory <version>] [--BreakingGround <version>]`
instead of previous `ckan ksp fake <name> <path> <version> [MakingHistoryVersion]`.
I think this is more future-proof.
I am going to try my luck adjusting KSP-CKAN/xKAN-meta_testing.

**`IDlcDetector`**:
I added `string InstallPath()`, it returns the install path relative to the GameDir, that means for example `"GameData/SquadExpansion/Serenity"`.
Depending on how [this idea of mine](https://github.com/KSP-CKAN/CKAN/issues/2770#issuecomment-498413897) will be implemented (if at all), we might restructure the whole thing again. But I think this is the best solution for now, I'm open for feedback.

I added `bool AllowedOnBaseVersion(KspVersion)`, it returns whether a DLC is allowed to be installed or faked on a given basegame version.

I added
```csharp
string IdentifierBaseName { get; }
KspVersion ReleaseGameVersion { get; }
```
to allow accessing those two from outside the interface implementations.
I changed the two fields with these names inside `StandardDlcDetectorBase` to getter without setter (this effectively is like the `readonly` modificator before).

**`CloneFakeInstanceDialog`**:
The GUI `CloneFakeInstanceDialog` has been adjusted:
CloneFakeInstanceDialog
If Squad releases another 10 DLCs we might get out of space and need a new system, but for now it works.

**`FakeInstance()`**:
`FakeInstance()` is now a transaction.
Before the DLC just wasn't faked if f.e. the KSP version was < 1.4.0.
This left the user in an unfavorable situation, he had to delete the new folder and run `ckan ksp forget` once and try again to include the DLC. now it's all or nothing.

`FakeInstance()` takes the DLCs that should be faked as `Dictionary<IDlcDetector, KspVersion>`. Future DLC releases _shouldn't_ need a change in this fuction.

Also `CopyDirectory()` is now a real transaction that includes IO operations, if I understand stock .NET transactions right they don't include them, but using `ChinhDo.Transactions` with `TxFileManager` should fix this.
Correct me if I got something wrong there (or anywhere else too of course).

Fixes #2791 

~Oh yeah, #2749 needs another rebase ^^~

### DIFF
--- a/Cmdline/Action/KSP.cs
+++ b/Cmdline/Action/KSP.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using CKAN.Versioning;
@@ -54,8 +55,7 @@ namespace CKAN.CmdLine
                     {
                         // First the commands with three string arguments
                         case "fake":
-                            ht.AddPreOptionsLine($"Usage: ckan ksp {verb} [options] name path version dlcVersion");
-                            ht.AddPreOptionsLine($"Choose dlcVersion \"none\" if you want no simulated dlc.");
+                            ht.AddPreOptionsLine($"Usage: ckan ksp {verb} [options] name path version [--MakingHistory <version>] [--BreakingGround <version>]");
                             break;
 
                         case "clone":
@@ -124,8 +124,14 @@ namespace CKAN.CmdLine
             [ValueOption(0)] public string name { get; set; }
             [ValueOption(1)] public string path { get; set; }
             [ValueOption(2)] public string version { get; set; }
-            [ValueOption(3)] public string dlcVersion { get; set; }
-            [Option("set-default", DefaultValue = false, HelpText = "Set the new instance to the default one.")] public bool setToDefault { get; set; }
+
+            [Option("MakingHistory", DefaultValue = "none", HelpText = "The version of the Making History DLC to be faked.")]
+            public string makingHistoryVersion { get; set; }
+            [Option("BreakingGround", DefaultValue = "none", HelpText = "The version of the Breaking Ground DLC to be faked.")]
+            public string breakingGroundVersion { get; set; }
+
+            [Option("set-default", DefaultValue = false, HelpText = "Set the new instance as the default one.")]
+            public bool setDefault { get; set; }
         }
 
         // This is required by ISubCommand
@@ -303,8 +309,8 @@ namespace CKAN.CmdLine
 
             // Parse all options
             string instanceNameOrPath = options.nameOrPath;
-            string new_name = options.new_name;
-            string new_path = options.new_path;
+            string newName = options.new_name;
+            string newPath = options.new_path;
 
 
             log.Info("Cloning the KSP instance: " + options.nameOrPath);
@@ -320,16 +326,16 @@ namespace CKAN.CmdLine
                         if (instance.Name == instanceNameOrPath)
                         {
                             // Found it, now clone it.
-                            Manager.CloneInstance(instance, new_name, new_path);
+                            Manager.CloneInstance(instance, newName, newPath);
                             break;
                         }
                     }
                 }
                 // Try to use instanceNameOrPath as a path and create a new KSP object.
                 // If it's valid, go on.
-                else if (new CKAN.KSP(instanceNameOrPath, new_name, User) is CKAN.KSP instance && instance.Valid)
+                else if (new CKAN.KSP(instanceNameOrPath, newName, User) is CKAN.KSP instance && instance.Valid)
                 {
-                    Manager.CloneInstance(instance, new_name, new_path);
+                    Manager.CloneInstance(instance, newName, newPath);
 
                 }
                 // There is no instance with this name or at this path.
@@ -370,12 +376,16 @@ namespace CKAN.CmdLine
                 ListInstalls();
                 return Exit.ERROR;
             }
-
+            catch (InstanceNameTakenKraken kraken)
+            {
+                User.RaiseError("This instance name is already taken: {0}", kraken.instName);
+                return Exit.BADOPT;
+            }
 
             // Test if the instance was added to the registry.
             // No need to test if valid, because this is done in AddInstance(),
             // so if something went wrong, HasInstance is false.
-            if (Manager.HasInstance(new_name))
+            if (Manager.HasInstance(newName))
             {
                 return Exit.OK;
             }
@@ -506,10 +516,26 @@ namespace CKAN.CmdLine
         /// </summary>
         private int FakeNewKSPInstall (FakeOptions options)
         {
+            int error()
+            {
+                log.Debug("Instance faking failed, see console output for details.");
+                User.RaiseMessage("--Error--");
+                return Exit.ERROR;
+            }
+
+            int badArgument()
+            {
+                log.Debug("Instance faking failed: bad argument(s). See console output for details.");
+                User.RaiseMessage("--Error: bad argument(s)--");
+                return Exit.BADOPT;
+            }
+
+
             if (options.name == null || options.path == null || options.version == null)
             {
-                User.RaiseMessage("ksp fake <name> <path> <version> [dlcVersion] - argument(s) missing");
-                return Exit.BADOPT;
+                User.RaiseMessage("ksp fake <name> <path> <version> " +
+                	"[--MakingHistory <version>] [--BreakingGround <version>] - argument(s) missing");
+                return badArgument();
             }
 
             log.Debug("Parsing arguments...");
@@ -517,16 +543,33 @@ namespace CKAN.CmdLine
             string installName = options.name;
             string path = options.path;
             KspVersion version;
-            bool setToDefault = options.setToDefault;
-            // dlcVersion is null if a user wants no simulated DLC
-            string dlcVersion;
-            if (options.dlcVersion == null || options.dlcVersion.ToLower() == "none")
+            bool setDefault = options.setDefault;
+
+            // options.<DLC>Version is "none" if the DLC should not be simulated.
+            Dictionary<DLC.IDlcDetector, KspVersion> dlcs = new Dictionary<DLC.IDlcDetector, KspVersion>();
+            if (options.makingHistoryVersion != null && options.makingHistoryVersion.ToLower() != "none")
             {
-                dlcVersion = null;
+                if (KspVersion.TryParse(options.makingHistoryVersion, out KspVersion ver))
+                {
+                    dlcs.Add(new DLC.MakingHistoryDlcDetector(), ver);
+                }
+                else
+                {
+                    User.RaiseError("Please check the Making History DLC version argument - Format it like Maj.Min.Patch - e.g. 1.1.0");
+                    return badArgument();
+                }
             }
-            else
+            if (options.breakingGroundVersion != null && options.breakingGroundVersion.ToLower() != "none")
             {
-                dlcVersion = options.dlcVersion;
+                if (KspVersion.TryParse(options.breakingGroundVersion, out KspVersion ver))
+                {
+                    dlcs.Add(new DLC.BreakingGroundDlcDetector(), ver);
+                }
+                else
+                {
+                    User.RaiseError("Please check the Breaking Ground DLC version argument - Format it like Maj.Min.Patch - e.g. 1.1.0");
+                    return badArgument();
+                }
             }
 
             // Parse the choosen KSP version
@@ -538,7 +581,7 @@ namespace CKAN.CmdLine
             {
                 // Thrown if there is anything besides numbers and points in the version string or a different syntactic error.
                 User.RaiseError("Please check the version argument - Format it like Maj.Min.Patch[.Build] - e.g. 1.6.0 or 1.2.2.1622");
-                return Exit.BADOPT;
+                return badArgument();
             }
 
             // Get the full version including build number.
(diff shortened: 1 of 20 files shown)


Core/Utilities.cs
```diff
@@ -1,6 +1,7 @@
-﻿using System;
+using System;
```
DasSkelett: Here was a BOM (0xFEFF) at the beginning, I removed it.

DasSkelett:
I see that the fact that DLCs weren't faked without error message if the version was too low was used in the validation scripts.
But I still believe that we should do all or nothing.

We can use the `ckan compare` command to determine whether to fake a DLC or not in the scripts.
Right now it outputs a very human readable answer:
`"1.1.0" is lower than "1.4.1".`
but if we add another option (`--machine-readable` or so) to output -1, 0 or 1 it should be more usable.

HebaruSan:
The scripts have a function `versions_less_or_equal` to do version comparisons.

DasSkelett:
> The scripts have a function `versions_less_or_equal` to do version comparisons.

That's cool, thanks!

HebaruSan:
> Oh yeah, #2749 needs another rebase ^^

The shoe is on the other foot!

DasSkelett:
> The shoe is on the other foot!

Darn it, feared this would happen.
Will take some days until I'm home again.

HebaruSan:
Created DasSkelett/CKAN#5. GitHub reports conflicts, which shouldn't surprise it since rebasing was the point.

DasSkelett:
'Merged' the PR, thanks.

Cmdline/Action/KSP.cs
```diff
@@ -550,36 +564,48 @@ private int FakeNewKSPInstall (FakeOptions options)
...
             User.RaiseMessage(String.Format("Creating new fake KSP install {0} at {1} with version {2}", installName, path, version.ToString()));
+            log.Debug("Faking instance...");
 
             try
             {
                 // Pass all arguments to CKAN.KSPManager.FakeInstance() and create a new one.
-                Manager.FakeInstance(installName, path, version, dlcVersion);
-                if (setToDefault)
+                Manager.FakeInstance(installName, path, version, dlcVersions);
+                if (setDefault)
```
HebaruSan: We are missing curly braces here. Only the `RaiseMessage` call is controlled by the `if`, the `SetAutoStart` will happen unconditionally.

Cmdline/Action/KSP.cs
```diff
@@ -124,8 +123,13 @@ internal class FakeOptions : CommonOptions
             [ValueOption(0)] public string name { get; set; }
             [ValueOption(1)] public string path { get; set; }
             [ValueOption(2)] public string version { get; set; }
-            [ValueOption(3)] public string dlcVersion { get; set; }
-            [Option("set-default", DefaultValue = false, HelpText = "Set the new instance to the default one.")] public bool setToDefault { get; set; }
+
+            [Option("MakingHistory", DefaultValue = "none", HelpText = "The version of the DLC to be faked.")]
+            public string makingHistoryVersion { get; set; }
+            [Option("BreakingGround", DefaultValue = "none", HelpText = "The version of the DLC to be faked.")]
```
HebaruSan: Let's make the help texts different. "... of the Making History DLC" ?

Core/KSPManager.cs
```diff
@@ -200,67 +202,96 @@ public void CloneInstance (KSP existing_instance, string new_name, string new_pa
...
-
         /// <summary>
         /// Create a new fake KSP instance
         /// </summary>
         /// <param name="new_name">The name for the new instance.</param>
         /// <param name="new_path">The loaction of the new instance.</param>
         /// <param name="version">The version of the new instance. Should have a build number.</param>
-        /// <param name="dlcVersion">The version of the DLC. Null if DLC should be faked.</param>
-        public void FakeInstance(string new_name, string new_path, KspVersion version, string dlcVersion = null)
+        /// <param name="dlcVersions">The versions of the DLCs. Null / empty strings if DLCs should not be faked.</param>
```
HebaruSan: I'm not a big fan of an array of strings as a parameter type, with the array indices meaningful and `null` padding values required in some cases. It'll require us to update this function every time a new DLC is released, and it's prone to array out of bounds errors and so forth. And the current parameter documentation doesn't explain how to use it.

What do you think about changing this parameter to a `Dictionary<IDlcDetector, KspVersion>` ? Then we could `foreach` through them to get the key value pairs, and use `InstallPath()` on the key to create the folders in a generic way. If you want to retain the game version validation, we could add a `bool AllowedOnBaseVersion(KspVersion)` method to `IDlcDetector` to do that generically as well. Then when a new DLC is released we would only have to update the calling code to add a new pair to the dictionary. Generating that parameter should be relatively simple:

```csharp
    new Dictionary<IDlcDetector, KspVersion>() {
        { new MakingHistoryDlcDetector(),  new KspVersion("1.1.0") },
        { new BreakingGroundDlcDetector(), new KspVersion("1.0.0") },
    }
```

Cmdline/Action/KSP.cs
DasSkelett: A bit of Python here and there can't hurt.

... actually, it does in this case.

Core/KSPManager.cs
DasSkelett: Sounds like a much better solution than mine. 👍

DasSkelett: I would add a new `private readonly KspVersion ReleaseGameVersion;` to the `StandardDlcDetectorBase` to save the KSP version there, and set it in the constructors like `IdentifierBaseName` and `DirectoryBaseName`.
I think it makes sense to have it in an obvious, code-accessible place, and not only buried somewhere inside `bool AllowedOnBaseVersion()`.
Does that seem reasonable to you?

HebaruSan: Definitely make whatever changes you think are appropriate for the `AllowedOnBaseVersion` concept, that was just meant to be very vague pseudocode. :+1:

DasSkelett: I'll try my best 😆

DasSkelett: One more thing:
Right now we output the DLC version as an `UnmanagedModuleVersion` in the detectors.
Should we 'upgrade' that to `KspVersion`? It allows better handling, and if I'm not mistaken it always matched the pattern, and probably will do so in the future too.

HebaruSan: I guess that sounds fine to me, but let's ask @dbent in case there are subtleties that I'm not noticing.

dbent: The version object type is used as a sentinel value in a few different places to distinguish between modules CKAN installed itself and those that were autodetected. I don't particularly like that strategy myself but it's a legacy from how the original autodetect code was written.

If you return a `ModuleVersion` instead of an `UnmanagedModuleVersion` then everywhere that checks if the version object is an `UnmanagedModuleVersion` will have to be updated. IIRC, the changes were numerous and sensitive enough that it wasn't worth the effort at the time.

DasSkelett: Okay, thanks @dbent. So not something I want to do in this PR, if at all ^^
If this PR stays as is, we would be double-tracked, with all the instance faking stuff using `KspVersion` and the mod managing and other stuff using `UnmanagedModuleVersion`.
Well, it works, but I don't make use of it being `KspVersion`, so I probably should change it (back) to `UnmanagedModuleVersion` for the sake of consistency.

Core/KSP.cs
```diff
@@ -440,6 +445,7 @@ public bool ScanGameData()
 
                 bool dlcChanged = manager.ScanDlc();
                 manager.Save(false);
+                manager.Dispose();
```
DasSkelett: @HebaruSan, is this safe to call here?
Problem was, `SetupCkanDirectories()` calls `ScanGameData()`,
which creates a new `RegistryManager.Instance()`, which creates a lock file.
This interferes with transactions.

Since this `RegistryManager` object is instantiated inside `ScanGameData()`, it can't be accessed after leaving the function scope, yes? This should make it safe to dispose it.

I tried to add a new KSP dir without a dir, which calls `ScanGameData()` too, and that still works fine :D

HebaruSan: > Since this RegistryManager object is instantiated inside ScanGameData(), it can't be accessed after leaving the function scope, yes?

No, any other code can access the same manager via `RegistryManager.Instance(ksp)` if it has a reference to the same `KSP ksp` object. In fact, it's possible that some other code *already* has such a reference and is depending on it not being disposed (GUI calls this function at the start of `UpdateRepo` and probably would prefer to keep its lock).

DasSkelett: Oh well, I see it's a bad idea.
Then I'll search another way to remove the lock file in failed transactions or prevent it from being created at all.

DasSkelett: I solved it with passing a bool to `KSP()` and `SetupCkanDirectories()` signaling whether `ScanGameData()` should be called. It's the easiest way I think.

Core/KSPManager.cs
DasSkelett: On another thought, I could just leave it as is. The format checks should be done either way, and it should simplify a future change to `KspVersion` in the underlying module managing code.

Core/DLC/IDlcDetector.cs
```diff
@@ -7,6 +7,9 @@ namespace CKAN.DLC
     /// </summary>
     public interface IDlcDetector
     {
+        string IdentifierBaseName { get; }
+        KspVersion ReleaseGameVersion { get; }
```
HebaruSan: Please explicitly mark these as `public` or `private`.

HebaruSan:
(Please ignore my deleted comment about marking things public or private. I forgot I was looking at an Interface.)

Core/KSPManager.cs
```diff
@@ -200,67 +206,91 @@ public void CloneInstance (KSP existing_instance, string new_name, string new_pa
...
+                        }
+                    }
+
+                    // Add the new instance to the registry
+                    KSP new_instance = new KSP(newPath, newName, User, false);
+                    AddInstance(new_instance);
+                }
+                // Thrown by AddInstance() if created instance is not valid.
+                // Thrown f.e. if a write operation didn't complete for unknown reasons.
+                catch (NotKSPDirKraken kraken)
```
HebaruSan: Unless I'm missing something, we don't need to catch and re-throw here. That's for when you want to throw a different exception or take some action in addition. If we just remove the try/catch blocks, the effect should be the same.

Core/KSP.cs
```diff
@@ -71,24 +72,29 @@ public KSP(string gameDir, string name, IUser user)
...
+            TxFileManager txFileMgr = new TxFileManager();
+
             if (!Directory.Exists(CkanDir()))
             {
                 User.RaiseMessage("Setting up CKAN for the first time...");
                 User.RaiseMessage("Creating {0}", CkanDir());
-                Directory.CreateDirectory(CkanDir());
+                txFileMgr.CreateDirectory(CkanDir());
 
                 User.RaiseMessage("Scanning for installed mods...");
```
HebaruSan: Very minor, but we probably don't need this `RaiseMessage` if `scan` is false.

Core/KSPManager.cs
HebaruSan: The new DLC creation implementation looks good now. :+1:

HebaruSan:
Another minor point, the name error ends with "--Done--" while the folder error ends with "--Error--". Looks like this is because `Manager.HasInstance(installName)` will return true thanks to the instance that's already using the name?

```
$ _build/ckan.exe ksp fake Dakar /tmp/invalid2 1.5.1 --MakingHistory 1.0.0 --BreakingGround none
Creating new fake KSP install Dakar at /tmp/invalid2 with version 1.5.1.2335
This instance name is already taken: Dakar
--Done--
$ _build/ckan.exe ksp fake Dakar2 /tmp/invalid 1.5.1 --MakingHistory 1.0.0 --BreakingGround none
Creating new fake KSP install Dakar2 at /tmp/invalid with version 1.5.1.2335
The specified folder already exists and is not empty.
--Error--
```

Core/KSPManager.cs
DasSkelett: Yeah, I added that for clarity, to point out what call can throw which exception and why. Don't know if that's a 'don't' in C#.

HebaruSan: Oh, OK. Nevermind then!

DasSkelett:
> Another minor point, the name error ends with "--Done--" while the folder error ends with "--Error--". Looks like this is because `Manager.HasInstance(installName)` will return true thanks to the instance that's already using the name?

It's because I just forgot to put a `return` in that specific kraken catch, so the execution continues and yes, `Manager.HasInstance()` finally returns `true`.

Tests/Core/KSPManager.cs
```diff
@@ -195,20 +196,29 @@ public void FakeInstance_InNotEmptyFolder_ThrowsBadInstallLocationKraken()
         }
 
         [Test]
-        public void FakeInstance_ValidArgumentsWithDLC_ManagerHasValidInstance()
```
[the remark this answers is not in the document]
DasSkelett: Is that new test as you imagined?

DasSkelett:
Okay, some more changes:
* Added new kraken: `WrongKspVersionKraken`, to be used if a given KSP version is valid, but not allowed for an action (like instance faking and older base game version).
* Renamed `IncorrectKspVersionKraken` to `BadKspVersionKraken` to better distinct it from the new one. To be used if a version is not valid, i.e. not in the build map or malformatted.
* Added another test to check if `FakeInstance()` throws `WrongKspVersionkraken` for base game versions that are too old for the DLC(s) requested to fake.
* Redid the exit code handling in `Cmdline.KSP.FakeNewInstance()` a bit.

Tests/Core/KSPManager.cs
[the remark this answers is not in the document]
HebaruSan: Yes, looks good, thanks!

DasSkelett:
Does the bot skip network-dependent tests?

HebaruSan:
I believe so:

https://github.com/KSP-CKAN/CKAN/blob/1ac6261f0f81a07213c91c967c444ed451324990/.travis.yml#L43

DasSkelett:
Yep, just discovered that flag. Was wondering, because on my machine I obviously fail the two Curseforge tests.

PR #2787, opened 2019-06-17
Fix 'ManageKspInstances' dialog logic

DasSkelett:
Problem
-----
The current behavior of `ManageKspInstances` is really weird.
* As soon as you open the dialog, the current `AutostartInstance` is reset to null.
* It saves that instance as default/autostart, that is selected when you click the `Select` button, and not the one where you activate the checkbox, against all intuition.
* What this means is that you can never ever select one instance as your default one, but use a different one for now. You'll always end up setting either the wrong one or none at all as autostart.
* As a result the two checkboxes in `CloneFakeKspDialog` don't work at all.

Solution
------
* Don't call `Instance.Manager.ClearAutoStart()` before opening the instance management dialog.
* Update autostart instance on checkbox click, not on button click.
* Update the checkbox (so whether it shows as checked or not) in `KSPInstancesListView_SelectedIndexChanged()`
* Add a right click context menu to the instance list view to open the root directory of the selected instance (I really wanted that for a long time)
* Close the dialog when `CloneFakeKspDialog` exits with `Switch to new instance` checked and switch to the new instance.

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -557,6 +557,7 @@ namespace CKAN
                 || !RegistryManager.Instance(CurrentInstance).registry.HasAnyAvailable();
             if (allowRepoUpdate && repoUpdateNeeded)
             {
+                ModList.Rows.Clear();
                 UpdateRepo();
 
                 // Update the filters after UpdateRepo() completed.
@@ -575,6 +576,7 @@ namespace CKAN
                 UpdateModsList();
                 Filter((GUIModFilter)configuration.ActiveFilter);
             }
+
             ChangeSet = null;
             Conflicts = null;
         }
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -530,8 +530,8 @@ namespace CKAN
         /// <summary>
         /// React to switching to a new game instance
         /// </summary>
-        /// <param name="onStartup">true if this is the initial load and should trigger auto repo updates, false otherwise</param>
-        private void CurrentInstanceUpdated(bool onStartup)
+        /// <param name="allowRepoUpdate">true if a repo update is allowed if needed (e.g. on initial load), false otherwise</param>
+        private void CurrentInstanceUpdated(bool allowRepoUpdate)
         {
             Util.Invoke(this, () =>
             {
@@ -555,18 +555,28 @@ namespace CKAN
 
             bool repoUpdateNeeded = configuration.RefreshOnStartup
                 || !RegistryManager.Instance(CurrentInstance).registry.HasAnyAvailable();
-            if (onStartup && repoUpdateNeeded)
+            if (allowRepoUpdate && repoUpdateNeeded)
             {
                 UpdateRepo();
+
+                // Update the filters after UpdateRepo() completed.
+                // Since this happens with a backgroundworker, Filter() is added as callback for RunWorkerCompleted.
+                // Remove it again after it ran, else it stays there and is added again and again.
+                void filterUpdate (object sender, RunWorkerCompletedEventArgs e)
+                {
+                    Filter((GUIModFilter)configuration.ActiveFilter);
+                    m_UpdateRepoWorker.RunWorkerCompleted -= filterUpdate;
+                }
+
+                m_UpdateRepoWorker.RunWorkerCompleted += filterUpdate;
             }
             else
             {
                 UpdateModsList();
+                Filter((GUIModFilter)configuration.ActiveFilter);
             }
             ChangeSet = null;
             Conflicts = null;
-
-            Filter((GUIModFilter)configuration.ActiveFilter);
         }
 
         public void UpdateCKAN()
@@ -818,6 +828,10 @@ namespace CKAN
             // Triggers mainModList.ModFiltersUpdated()
             mainModList.ModFilter = filter;
 
+            // Save new filter to the configuration.
+            configuration.ActiveFilter = (int)mainModList.ModFilter;
+            configuration.Save();
+
             // Ask the configuration which columns to show.
             foreach (DataGridViewColumn col in ModList.Columns)
             {
@@ -1031,11 +1045,13 @@ namespace CKAN
 
         private void manageKspInstancesMenuItem_Click(object sender, EventArgs e)
         {
-            Instance.Manager.ClearAutoStart();
             var old_instance = Instance.CurrentInstance;
             var result = new ManageKspInstances(!actuallyVisible).ShowDialog();
             if (result == DialogResult.OK && !Equals(old_instance, Instance.CurrentInstance))
-                Instance.CurrentInstanceUpdated(false);
+            {
+                ModList.ClearSelection();
+                CurrentInstanceUpdated(true);
+            }
         }
 
         private void openKspDirectoryToolStripMenuItem_Click(object sender, EventArgs e)
--- a/GUI/MainModList.cs
+++ b/GUI/MainModList.cs
@@ -204,8 +204,7 @@ namespace CKAN
             {
                 foreach (GUIMod gm in gui_mods)
                 {
-                    bool oldIncompat;
-                    if (old_modules.TryGetValue(gm.Identifier, out oldIncompat))
+                    if (old_modules.TryGetValue(gm.Identifier, out bool oldIncompat))
                     {
                         // Found it; check if newly compatible
                         if (!gm.IsIncompatible && oldIncompat)
--- a/GUI/ManageKspInstances.cs
+++ b/GUI/ManageKspInstances.cs
@@ -39,17 +39,18 @@ namespace CKAN
             {
                 _manager.FindAndRegisterDefaultInstance();
             }
-            
+
             // Set the renderer for the AddNewMenu
             if (Platform.IsMono)
             {
                 this.AddNewMenu.Renderer = new FlatToolStripRenderer();
+                this.InstanceListContextMenuStrip.Renderer = new FlatToolStripRenderer();
             }
 
             UpdateInstancesList();
             UpdateButtonState();
         }
-        
+
         public void UpdateInstancesList()
         {
             KSPInstancesListView.Items.Clear();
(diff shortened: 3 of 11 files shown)


GUI/ManageKspInstances.cs
```diff
@@ -127,25 +135,52 @@ private void UseSelectedInstance()
...
             }
             catch (NotKSPDirKraken k)
             {
-                GUI.user.RaiseError(Properties.Resources.ManageKspInstancesNotValid,
-                    new object[] { k.path });
+                GUI.user.RaiseError(Properties.Resources.ManageKspInstancesNotValid, k.path);
+            }
+        }
+
+        private void SetAsDefaultCheckbox_Click(object sender, EventArgs e)
```
HebaruSan: Could we react to `_CheckboxChanged` instead of `_Click`? The checkbox already has its own built in logic to change the value on clicking, and it's weird that we want to do the same in our code as well.

DasSkelett: I wanted to do that too, but it seems that this line
```csharp
SetAsDefaultCheckbox.Checked = _manager.AutoStartInstance?.Equals(instName) ?? false;
```
of `KSPInstancesListView_SelectedIndexChanged` triggers it too.

DasSkelett: Looking at it again, I'm not sure whether it's really that line that's causing trouble, but nevertheless the checkbox doesn't work anymore if I change it to react to `CheckedChanged` :/

HebaruSan: OK, I'm fine with this unless/until there's a problem...

GUI/Main.cs
```diff
@@ -1011,11 +1011,10 @@ private void exportModListToolStripMenuItem_Click(object sender, EventArgs e)
 
         private void manageKspInstancesMenuItem_Click(object sender, EventArgs e)
         {
-            Instance.Manager.ClearAutoStart();
             var old_instance = Instance.CurrentInstance;
             var result = new ManageKspInstances(!actuallyVisible).ShowDialog();
             if (result == DialogResult.OK && !Equals(old_instance, Instance.CurrentInstance))
-                Instance.CurrentInstanceUpdated(false);
+                Instance.CurrentInstanceUpdated(true);
```
HebaruSan: I guess `onStartup` was a bad name for that parameter since this is clearly not startup. Would you mind renaming it, something like `allowRepoUpdate`?

DasSkelett:
I also added `ModList.ClearSelection()` to hide the mod info panel during the instance switch.

HebaruSan:
Shouldn't the `ClearSelection` call happen *before* `CurrentInstanceUpdated` then?

DasSkelett:
Yeah, I realized that too after I pushed it. But interestingly it still works...
I'll change it though.

DasSkelett:
Okay I'm pretty sure there's some thread racing going on.
Putting it before `Instance.CurrentInstanceUpdated()` makes it not working anymore,
unless I'm putting some break points in (one at `UpdateRepo()` and one at `ModList_SelectedIndexChanged()`).

I'm taking a deeper dive now.

DasSkelett:
I think I got it: Under certain conditions, `UpdateFilters()` can be called before `m_UpdateRepoWorker.RunWorkerAsync()` completed.
And because `UpdateFilters()` also calls `Main.set_ActiveModInfo()`, the panel is shown on the wait page.

DasSkelett:
I added `Filter()` as a callback on `M_UpdateRepoWorker.RunWorkerCompleted`.
Seems to work and is the easiest for now. I also made the `Forget` button disabled if the selected instance is the one 'running' currently.

For me this is ready to merge.

GUI/Main.cs
```diff
@@ -555,18 +555,22 @@ private void CurrentInstanceUpdated(bool onStartup)
...
                 || !RegistryManager.Instance(CurrentInstance).registry.HasAnyAvailable();
-            if (onStartup && repoUpdateNeeded)
+            if (allowRepoUpdate && repoUpdateNeeded)
             {
                 UpdateRepo();
+
+                // Update the filters AFTER UpdateRepo() completed.
+                // Since this happens with a backgroundworker, Filter() is added as callback for RunWorkerCompleted.
+                void filterUpdate (object sender, RunWorkerCompletedEventArgs e) { Filter((GUIModFilter)configuration.ActiveFilter); }
+                m_UpdateRepoWorker.RunWorkerCompleted += filterUpdate;
```
HebaruSan: Won't this add the handler every time this block runs? If we do this multiple times, I think you'd have redundant actions when the worker completes.

GUI/Main.cs
```diff
@@ -1031,11 +1035,13 @@ private void exportModListToolStripMenuItem_Click(object sender, EventArgs e)
...
         private void manageKspInstancesMenuItem_Click(object sender, EventArgs e)
         {
-            Instance.Manager.ClearAutoStart();
             var old_instance = Instance.CurrentInstance;
             var result = new ManageKspInstances(!actuallyVisible).ShowDialog();
             if (result == DialogResult.OK && !Equals(old_instance, Instance.CurrentInstance))
-                Instance.CurrentInstanceUpdated(false);
+            {
+                ModList.ClearSelection();
+                Instance.CurrentInstanceUpdated(true);
```
HebaruSan: (Old gripe) We don't need `Instance` here. This code is already running in the instance.

GUI/Properties/Resources.resx
```diff
@@ -7116,6 +7116,7 @@ Open settings now?</value></data>
   <data name="MainWaitPleaseWait" xml:space="preserve"><value>Please wait</value></data>
   <data name="MainWaitDone" xml:space="preserve"><value>All done!</value></data>
   <data name="ManageKspInstancesNotValid" xml:space="preserve"><value>Directory {0} is not valid KSP directory.</value></data>
```
HebaruSan: (Another complaint about old code) Can you please change to, "is not **a** valid KSP directory" ?

GUI/Main.cs
DasSkelett: That's... uhhhm... unfortunate.
Let me try to move the filter update before `UpdateRepo()`, see if that works too.

DasSkelett: Okay I just discovered that we only save the current filter `OnFormClosing()`.
It makes sense saving it before each instance change, doesn't it?

HebaruSan: Another option is to remove the handler after it runs with `-=`, or just add it once globally somewhere.

DasSkelett: I don't want to add it globally, since it is designed to explicitly run when the instance changed (by loading the old filter from the config),
which, if I wouldn't have added that the new filter is saved with each filter change, would "reset" the filter again to what it was at CKAN startup with each `UpdateRepo()` call, so even after refreshes and mod installs and so on.

So thanks to the config save now I _could_ add it globally, but it doesn't fit there IMO.
I'm going with removing it again.

HebaruSan:
My 1.3.1 instance has 4 mods:

```
$ ckan list --ksp=KSP-1.3.1

KSP found at /media/DataTrove/Games/KSP-1.3.1

KSP Version: 1.3.1.1891

Installed Modules:

- EnvironmentalVisualEnhancements 2:EVE-1.2.2-1
- ModuleManager 3.0.4
^ StockVisualEnhancements 3:1.3.0.4
^ SVE-LowResolution 3:1.0.7

Legend: -: Up to date. X: Incompatible. ^: Upgradable. >: Replaceable
        A: Autodetected. ?: Unknown. *: Broken. 
```

And my 1.4.1 instance has none:

```
$ ckan list --ksp=1.4.1

KSP found at /media/DataTrove/Games/KSP-1.4.1

KSP Version: 1.4.1.2089

Installed Modules:

A MakingHistory-DLC 1.0.0 (unmanaged)

Legend: -: Up to date. X: Incompatible. ^: Upgradable. >: Replaceable
        A: Autodetected. ?: Unknown. *: Broken. 
```

I set up these two instances to auto update at startup to test this, and then I used the dialog to switch between them repeatedly. Neither one loads its installed mods properly; they show the same installed mods as whatever instance was loaded before (in this case my steam install):

image

image

Instances without auto update appear to be working OK.

DasSkelett:
Okay, I found the cause faster than I thought:

It's a mix of calling `CurrentInstanceUpdated(bool allowRepoUpdate)` with `true` now:
https://github.com/KSP-CKAN/CKAN/blob/209689164169052a1f56a9f013f6e4279a84b524/GUI/Main.cs#L1046-L1055

And #2682, which made the GUI modlist only refresh when there were changes in the repo after an repo update, which aren't if you switch repeatedly in a  short time.

Thus, `repoUpdateNeeded` (due to auto refresh) and `allowRepoUpdate` are true,
which doesn't call `UpdateModsList()`, and instead hopes for `UpdateRepo()` to call it (which does not either, see above).
https://github.com/KSP-CKAN/CKAN/blob/209689164169052a1f56a9f013f6e4279a84b524/GUI/Main.cs#L556-L577

So we just need to call `UpdateModsList()` in both cases / outside the if-else.

However, it would be nice if `UpdateModsList()` is not called twice in case _there are_ repo changes. Don't know how to do that yet.

DasSkelett:
I solved it with an argument `bool forceModListUpdate = false` in `UpdateRepo()`, see my new commit (_not_ force-push diff).

I'm not especially happy with that solution, but it's the only way that came to my mind. Open to suggestions.

HebaruSan:
> I'm not especially happy with that solution, but it's the only way that came to my mind. Open to suggestions.

What about clearing the mod list first, so this bit handles the refresh?

https://github.com/KSP-CKAN/CKAN/blob/7b00d472297f05e30cb48135f32ce2da473e7239/GUI/MainRepo.cs#L116-L120

DasSkelett:
Seems to work :+1:

GUI/MainModList.cs
```diff
@@ -204,8 +204,7 @@ private void _UpdateModsList(IEnumerable<ModChange> mc, Dictionary<string, bool>
             {
                 foreach (GUIMod gm in gui_mods)
                 {
-                    bool oldIncompat;
-                    if (old_modules.TryGetValue(gm.Identifier, out oldIncompat))
+                    if (old_modules.TryGetValue(gm.Identifier, out bool oldIncompat))
```
HebaruSan: Ooh, I like that `out bool` syntax! I can think of many uses for that.
(I think this is the second time I've learned something useful about `out` params from reviewing your code? :1st_place_medal:)

PR #2798, opened 2019-06-23
Refactor Netkan for SQS mode

### LINKED ISSUE #2789: NetKAN Bot Re-Architechture
### History
NetKAN started its life as a Perl script in late 2014, followed by a single python/bash script, another Perl script which then inspired NetKAN-bot. Which is a modular Perl application with a bunch of libraries and a bunch of scripts to do things like regular indexing, webhooks and mirroring to archive.org.

### Problems
NetKAN-bot has served its purpose since writing it around 4 years ago, however with the growth of the number of mods along with now generating the majority of the metadata via the bot it is really starting to creak under the pressure. The application from a build perspective is reasonably straight forward, the infrastructure is not and makes deploying changes a bit of a pain.

### Rationale
One of the things that really drove the development for NetKAN-bot was ensuring the metadata produced complied with our schema. @HebaruSan has done a bunch of excellent work in #2788, meaning we could drop a of logic making for a far more simple indexer. Splitting these up and using a FIFO queue also means we don't have to be so defensive about our git repo getting messed up as we don't need to check outputs or even touch the file system if the payload and file on the disk are identical.

Initial Brain Dump
==================

NetKAN bot will be replaced by a bunch of distinct containers with Specific roles. 

### Inflator

@DasSkelett  / @HebaruSan  - How much work do you think would be in this? I'm happy to do the Docker/AWS archicture as I pretty much breathe this stuff at work, C# is not something I do a lot of.

- Processes SQS queue
  * Incomming Message example
     ```
    Payload: ModIdentifier
    Attr Refresh: True/False (this is to signify a repo pull is required)
    ```
  * Outgoing Message example
    ```
    Payload: CKAN JSON Output (if success)
    Attr Identifier: ModIdentifier
    Attr Success: True/False
    Attr Error Message: Message/Exception if there is one
    Attr Check Time: timestamp
    ```

Initially there will be 2 incomming and 2 outgoing

Incomming:
- Webhooks
- Scheduled

Outgoing:
- Master
- Staged

**Implementation**

Will require some new features added to NetKAN:

- Ability to run as a foreground process, that watches an SQS queue
- Read sqs queues, prioritise webhooks
- Netkan Repo pull changes when refresh is set true on an sqs message, a call out to shell would be fine at a minmum
- Upon inflation publish output to sqs master, unless it's marked for staging
- Manage cache aging. Currently it's a sceduled task, but netkan could age out files and when refresh is set it could mark for refresh of the cached download.

Assumptions that can be made

- As long as it'll run in fg, packaging into a docker container is trivial
- Cache storage path will be provided
- SQS Access will not require keys, the instance will be granted access via IAM

### Indexer

Takes care of processing the outgoing messages, check if the payload is different, update, commit, push + udpate status.

The idea behind having a separate queue is to avoid the current situation of needing to be real careful of our repo state, if we process them seperately we can manage this much much easier.

It will also update the status, likely a DynamoDB table. Which we can in the interim produce a json file periodically and upload to S3.

### Scheduler

Periodically check instance cpu credit levels, submit all NetKANs to the queue for inflation.

### WebHook Processing

This may actually be a lambda script behind load balancer or it may reside on the instance (load balancers are kinda expensive for this small use case). But it will take care of receiving inflation requests and creating an SQS Message on the webhook queue and CKAN-meta changes can spawn messages for the Mirroring to take place.

### Mirror processing

This will take care of processing the mirror queue and submitting applicable mods to archive.org for preservation.

Summary
=======

It's likely we can achieve this in a few steps and the idea will be to build automated processes for deployment of changes.

1. Minimal Operation
  - NetKAN Daemon mode with Git/SQS Processing
  - Indexer
  - Scheduler
2. Webhooks
  - They can quite happily run as they are until we've built something to process them
3. Mirroring
  - Along with the webhooks, they'll happily run as they are for now

Extra
- Consider how we want to impelment repo splitting
- Consider how to backfill x releases

HebaruSan:
## Motivation

See #2789; we want to refactor the NetKAN-bot to use AWS SQS so it can be managed and iterated more easily.

- https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/welcome.html
- https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

## Changes

This pull request builds upon #2788 and should not be merged until after that one is merged. Click on the second commit to view the diff of the changes that are unique to this PR.

Now a new command format is supported. Note that you must set the credentials environment variables for this to work, values available upon request from @techman83:

```bash
export AWS_ACCESS_KEY_ID='(access key here)'
export AWS_SECRET_ACCESS_KEY='(secret key here)'
export AWS_REGION='us-west-2'
mono netkan.exe --queues InboundDev.fifo,OutboundDev.fifo
```

This new command will authenticate with AWS SQS and retrieve inflation requests from  the input queue in this format:

```
Payload: JSON contents of a .netkan file
Attr Releases: Number of releases to retrieve (optional, 1 if omitted)
```

Such requests may be generated for testing with this Python script, which also requires the above environment variables and accepts a path to a .netkan file as its parameter:
- https://gist.github.com/techman83/0604f4dc9f849aac605e46f494de9403

This JSON is then inflated via existing Netkan logic. A message is then sent to the output`queue in this format:

```
Payload: JSON contents of the generated .ckan file
Attr ModIdentifier: Identifier of the module
Attr Staged: True if the netkan uses staging, false for immediate commit to master
Attr Success: True if successful, false if there were errors
Attr ErrorMessage: Error message if any, otherwise this attribute is omitted
Attr CheckTime: Timestamp when message was sent
Attr FileName: output-dir/identifier-version.ckan string explaining where to save the file
```

The output queue may be inspected with this Python script:
- https://gist.github.com/HebaruSan/42117ed50a415fa76bfaee4846a0c9f2

After this it will check the input queue again for more inflation requests, forever.

Fixes #2789.

## Open questions / left to-do

This isn't 100% complete yet.

- [x] Merge #2788
- [x] `Releases` as an input attribute (to be confirmed with design committee first)
  - [x] `Releases` as a parameter internally without re-creating the `NetkanTranslator`
- [x] Get region from environment or config file instead of hard coding in C#
- [x] Do we need a way to stop the loop? Currently it's a `while(true)`
- [x] Unknown unknowns


DasSkelett:
If I spot this right, when the queue is empty `getNetkan()` returns null, and the if-clause makes the loop repeat immediately.
What about a short break, to not spam out requests (and maybe further save some cpu credits)?

HebaruSan:
The break/delay happens during the `client.ReceiveMessage` call already; it will sit there for 20 seconds before returning unless a message comes in. It's not obvious how that works without seeing it; I'll PM you the credentials on the forum for testing.

DasSkelett:
Thanks. And I should read the documentation now ^^

DasSkelett:
The queue names are also hardcoded atm, IMO it makes sense to outsource them too (they have to be changed before merge anyway).
They aren't listed as supported env-vars, so either create own env-vars and read them, or with a config file, or whatsoever.

HebaruSan:
Well we're going to have to hard-code *something*. Either we know the queue names and use them to look up the URLs, or we know some other keys and use them to look up the names and use them to look up the URLs, etc. I'll defer to @techman83 on how this is typically handled.

DasSkelett:
Okay.
Just tested a failing netkan (`WoodSpaceprogram`).
First, the log message at line 50 in `Queuehandler.cs`, `Sending failure: Could not find Wood Space Program entry in zipfile to install` is a bit misleading, it sounds like the send process failed, which isn't the case at that point.
I would word it `Sending inflation failure: ...` or alike.

Second, sending fails indeed: `Send failed: The request must contain the parameter MessageBody.`
So `SendMessageRequest` doesn't like a real `null` for `MessageBody`.

HebaruSan:
Good points, I'm going to try deleting all the messages so I can see new ones...

HebaruSan:
I have no idea why Amazon decided an empty body was so unacceptable that it should throw an exception; it invites silly things like sending "{}" as the body. Which is what we do now.

DasSkelett:
Oh, I thought it was a null object that was passed in `SendMessageRequest()`, but it was an empty string. That's really picky, yep.

Otherwise, the PR looks fine for me. I'm impressed that you were so fast.

techman83:
> Well we're going to have to hard-code _something_. Either we know the queue names and use them to look up the URLs, or we know some other keys and use them to look up the names and use them to look up the URLs, etc. I'll defer to @techman83 on how this is typically handled.

I think we should pass the queue names in as CLI argments or environment variables, I think it's acceptable to look up the queue URLs when constructing the class.

> Releases

I have a few ideas, but I think we should open a separate feature. There is a lot to do in the first pass, but we'll be in a much better place to rapidly get changes into the wild.

> Code review

I've had a none C# dev look over the code and I think this is excellent! Likely some extra requirements will shake out as we use it in anger, but this definitely meets and surpasses the MVP.

Netkan/Processors/QueueHandler.cs
```diff
@@ -0,0 +1,206 @@
...
+            inflator = new Inflator(cacheDir, overwriteCache, githubToken, prerelease);
+
+            inputQueueURL  = getQueueUrl(inputQueueName);
+            outputQueueURL = getQueueUrl(outputQueueName);
+            log.DebugFormat("Queue URLs: {0}, {1}", inputQueueURL, outputQueueURL);
+        }
+
+        public void Process()
+        {
+            while (true)
```
techman83: From my perspective, this is a GoodEnough&trade; starting point. If you're interested you could do something with signal handlers, so that we can bail out gracefully when a term signal is sent. But that's also rather linux specific, I don't know how cross platform the logic is (though we only care about linux here).

Netkan/Processors/QueueHandler.cs
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        private Message getFromQueue(string url)
+        {
+            log.DebugFormat("Looking for message from {0}", url);
+            var resp = client.ReceiveMessage(new ReceiveMessageRequest()
+            {
+                QueueUrl            = url,
+                MaxNumberOfMessages = 1,
+                VisibilityTimeout   = (int)TimeSpan.FromMinutes(15).TotalSeconds,
```
techman83: This is really good. If we run multiple inflators we can use this to avoid duplication of inflation.

Netkan/Processors/QueueHandler.cs
```diff
@@ -0,0 +1,206 @@
...
+                        StringValue = err
+                    }
+                );
+            }            
+            
+            SendMessageRequest msg = new SendMessageRequest()
+            {
+                QueueUrl               = outputQueueURL,
+                MessageGroupId         = "1",
+                MessageDeduplicationId = Path.GetRandomFileName(),
```
techman83: This ID is important to avoid duplicate messages hitting the queue. We could enable deduplication based on content (avoiding the need to generate this) which might be better? What I've done with the hack script is to MD5 the content before I send it.

Netkan/Processors/QueueHandler.cs
```diff
@@ -0,0 +1,206 @@
...
+            }
+        }
+
+        private string serializeCkan(Metadata ckan)
+        {
+            if (ckan == null)
+            {
+                // SendMessage doesn't like empty bodies, so send an empty JSON object
+                return "{}";
+            }
```
techman83: We could send the error message as a json object? The body is likely required as part of their queue deduplication efforts. I don't really have a strong opinion either way here.

Netkan/Processors/QueueHandler.cs
HebaruSan: OK, as long as it won't be a problem for managing the container. I know it's easy enough to kill a process, but if there's a cleaner way to shut down, just let me know.

Netkan/Processors/QueueHandler.cs
HebaruSan: Yup (and apparently [it's required if you want to be able to delete messages as well](https://github.com/aws/aws-sdk-js/issues/1279#issuecomment-268830676)).

What about `MaxNumberOfMessages`? Should we try to handle more?

Netkan/Processors/QueueHandler.cs
HebaruSan: Yup, that's why I went with `Path.GetRandomFileName()`, "[a cryptographically strong, random string](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getrandomfilename)". I don't want the queue logic getting smart on us and deleting "duplicates" that aren't actually duplicates, because from what I can tell, we won't be generating any duplicates.

Netkan/Processors/QueueHandler.cs
HebaruSan: We could try something like that, but if our payload is defined as the .ckan JSON, then we'd be sending JSON that doesn't match that definition some of the time.

Or some kind of dual-purpose wrapper object that could contain either a .ckan or an error in the style of a C union, but then we're making up a JSON format that we don't have a schema for. And it starts to look like the outer layer of that object would be redundant with message attributes anyway.

```json
{
    "identifier": "AwesomeMod",
    "ckan": {
        "spec_version": 1,
        etc...
    },
    "error": "Didn't work"
}
```

Do you know how this sort of thing is handled in other queue setups? We can't be the only project with a process that may fail to generate the expected output.

Netkan/Processors/QueueHandler.cs
techman83: > Yup (and apparently [it's required if you want to be able to delete messages as well](https://github.com/aws/aws-sdk-js/issues/1279#issuecomment-268830676)).
Ohhh, interesting. Good find!

> What about `MaxNumberOfMessages`? Should we try to handle more?
It would be more efficient message wise to handle them in batches of 10. We would need to consider visibility timeouts and what not. If anything maybe allow for multiple, but set max messages to 1. Then when we've shaken out the logistics our basic layout is for multiple messages.

Netkan/Processors/QueueHandler.cs
techman83: It might be unlikely that we'll generate duplicates, but I feel we should trust the queue to do the RightThing&trade;. In dev it might be desirable to allow us to produce metadata over and over that is the same, but in production I think we should ensure that if we do send a duplicate message that our `MessageDeduplicationId` is generated from the body in a repeatable fashion.

Netkan/Processors/QueueHandler.cs
techman83: Yeah, I think you're right here. The body should be the body and essentially an error is a blank object, that way we're not overloading what it means.

HebaruSan:
`InboundDev.fifo` isn't working for me anymore. Nothing shows up after I run `send_netkan.py`. Did we break it?

HebaruSan:
Changes in latest commit:

- `--sqs` flag replaced with `--queues InboundDev.fifo,OutboundDev.fifo` to specify queue names
- Retrieve and process up to 10 messages at once
- If input message contains Release attribute, will process that many releases
- Delete messages after processing rather than after receiving to make killing the process safer

Netkan/Processors/QueueHandler.cs
HebaruSan: Refactored to get 10 at once. It seems to work.

Netkan/Processors/QueueHandler.cs
HebaruSan: Well, we can't just hash the body (or use `ContentBasedDeduplication`), because all of the error messages will have a body of "{}". They'd be falsely marked as duplicates if multiple were sent within 5 minutes, and we'd potentially miss out an an error message for the status page.

The place to worry about duplicates for this container is upstream. The Inflator does the most CPU-, network-, and RAM-intensive work. If we've received and processed a duplicate message from `InboundDev.fifo,` we've already "lost'. Trying to generate dedup IDs at this point is shutting the barn door after the horse has escaped.

HebaruSan:
Pushed just now:

- Get the region from the environment. Turns out this works, but you need to use AWS_REGION instead of AWS_DEFAULT_REGION; instructions in PR text updated.

Netkan/Processors/QueueHandler.cs
techman83: That's a really fair point and in reality this will suit our use case. The client has retry mechanisms built in and there are scenarios where the message made it but the client didn't get the response the first.

And you're very right, the indexer will just go, nothing to see here and drop it. Thanks for persisting with my design stubbornness :-)
Nothing-to-See-15a34a2fc727c8

techman83:
> ~`InboundDev.fifo` isn't working for me anymore. Nothing shows up after I run `send_netkan.py`. Did we break it?~
> 
> EDIT: Working again now. Odd.

That's likely based around our discussion on the `MessageDeduplicationId`, my little script hashes the body. I think this is likely the correct for generating content for the inbound queue in production, but for testing a random string would be better.

> Changes in latest commit:
> 
> * `--sqs` flag replaced with `--queues InboundDev.fifo,OutboundDev.fifo` to specify queue names
> * Retrieve and process up to 10 messages at once
> * If input message contains Releases attribute, will process that many releases
> * Delete messages after processing rather than after receiving to make killing the process safer

Those are some awesome additions! I think for now we should let this shake out and I'm happy for it to be merged when ready. We can see how it all shakes out and make any improvements as needed.

> Pushed just now:
> 
> * Get the region from the environment. Turns out this works, but you need to use AWS_REGION instead of AWS_DEFAULT_REGION; instructions in PR text updated.

I see. Looks like a bunch of the SDKs use `AWS_REGION`, just not the CLI or the boto3 library. Good catch :-)

Netkan/Processors/QueueHandler.cs
```diff
@@ -0,0 +1,197 @@
...
+                        StringValue = success.ToString()
+                    }
+                },
+                {
+                    "CheckTime",
+                    new MessageAttributeValue()
+                    {
+                        DataType    = "String",
+                        StringValue = DateTime.UtcNow.ToString("s", CultureInfo.InvariantCulture)
+                    }
```
techman83: As an extra attribute, I think what would really be good to have is the resulting ckan filename. Currently I do some funky git gyrations to figure it out, which really isn't necessary if NetKAN can tell me.

That way the indexer is always going to do exactly what NetKAN would do.

HebaruSan: We can do that. With or without the output dir parameter?

HebaruSan: Added FileName attribute, which will include the output dir parameter if any. Let me know if you want it changed to exclude the output dir.

HebaruSan:
The last blocker for this PR is code review for dependency #2788, which we need to ensure that the generated metadata is valid. If anybody is interested in having these changes merged, consider taking a look at that PR.

DasSkelett:
Yeah, I didn't really jump on #2788 because it's line-wise a lot, but I will look at it now.

HebaruSan:
> Yeah, I didn't really jump on #2788 because it's line-wise a lot, but I will look at it now.

Thanks!! Yes, it looks like a lot, but it has the advantage of being made up of many much smaller pieces, each of which is pretty simple.

PR #2808, opened 2019-06-26
Refresh builds.json when validating netkans

HebaruSan:
## Problem

After #2788, lots of modules report this error:

[image](http://status.ksp-ckan.space/)

## Cause

We use `KspBuildMap` to generate the list of known game versions. This class loads the list of known game version from the Win32Registry. If the list isn't in the registry, then it will try downloading the remote build map from:

- https://raw.githubusercontent.com/KSP-CKAN/CKAN-meta/master/builds.json

In the regular client, a forced refresh call occurs here when updating the registry, via that weird code-obfuscating `Autofac` stuff that some people like for some reason:

https://github.com/KSP-CKAN/CKAN/blob/9d8c3ed13d464ad009904ab607be39c160db4714/Core/Net/Repo.cs#L100-L103

There's currently no equivalent of that call in Netkan, so the bot has a copy of this file saved in its registry from October 2018, which includes KSP 1.4.1 but not 1.5.1. So all modules compatible with 1.4.1 or earlier are working, and the rest are erroring out.

## Changes

Now we refresh the build map before using it, which forces it to download the remote build map. This will ensure that as soon as we add a build to the above file, the bot will know about it.

### DIFF
--- a/Netkan/Validators/MatchesKnownGameVersionsValidator.cs
+++ b/Netkan/Validators/MatchesKnownGameVersionsValidator.cs
@@ -6,14 +6,22 @@ namespace CKAN.NetKAN.Validators
 {
     internal sealed class MatchesKnownGameVersionsValidator : IValidator
     {
+        public MatchesKnownGameVersionsValidator()
+        {
+            buildMap = new KspBuildMap(new Win32Registry());
+        }
+
         public void Validate(Metadata metadata)
         {
             var mod = CkanModule.FromJson(metadata.Json().ToString());
-            var knownVersions = new KspBuildMap(new Win32Registry()).KnownVersions;
-            if (!mod.IsCompatibleKSP(new KspVersionCriteria(null, knownVersions)))
+            // Get latest builds from server
+            buildMap.Refresh();
+            if (!mod.IsCompatibleKSP(new KspVersionCriteria(null, buildMap.KnownVersions)))
             {
                 throw new Kraken($"{metadata.Identifier} doesn't match any valid game version");
             }
         }
+
+        private KspBuildMap buildMap;
     }
 }


techman83:
My only comment is that we'll be pinging this endpoint ~2000 times per index run. When we switch over to queues, we'll want some way to handle this also.

HebaruSan:
Yeah, ideally we'd use the embedded build map, since it requires no additional network traffic and will always be as up to date as netkan.exe is, which is "best of both worlds." But `KspBuildMap`'s design makes that awkward to do; the embedded build map is only used as a last resort fallback if we can't get the remote build map and we've *never* retrieved it before.

Actually, that's interesting; currently the bot downloads a new netkan.exe at the start of each pass, but if we're running a long-duration `netkan.exe --queues Input,Output` command over multiple passes, we'd need to figure out a new way of refreshing netkan.exe when we update it. I assume the container structure will facilitate this somehow.

techman83:
> Actually, that's interesting; currently the bot downloads a new netkan.exe at the start of each pass, but if we're running a long-duration `netkan.exe --queues Input,Output` command over multiple passes, we'd need to figure out a new way of refreshing netkan.exe when we update it. I assume the container structure will facilitate this somehow.

It will check on every run, only downloading when it's different. As for the container infrastructure, it'll be a 3 line docker file for NetKAN (I might request some assistance for making NetKAN take env variables, but we'll cross that bridge later), which travis can build/push to docker hub and then I can generate a set of travis AWS keys IAM locked to ping the Elastic Container Service API to force a reload of the container.  We do something similar at work to keep our internal services up to date. Code gets developed, tested and when a PR is merged it builds/pushes/deploys as part of the Continuous Deployment pipeline.

We can encrypt the keys using the Travis tools and worst case if they leaked, someone could repeatedly force the container to be redeployed. Which would be annoying, but not allow them to spin up 1000 crypto miners and give us a massive bill.

HebaruSan:
> travis can build/push to docker hub and then I can generate a set of travis AWS keys IAM locked to ping the Elastic Container Service API to force a reload of the container. We do something similar at work to keep our internal services up to date. Code gets developed, tested and when a PR is merged it builds/pushes/deploys as part of the Continuous Deployment pipeline.

That sounds awesome! So the "new bot" would switch over to new code as soon as it was compiled, and pick up where the previous incarnation left off thanks to the queues! Love it!

techman83:
> That sounds awesome! So the "new bot" would switch over to new code as soon as it was compiled, and pick up where the previous incarnation left off thanks to the queues! Love it!

Yup :smile:

PR #2812, opened 2019-06-26
Use embedded buildmap in Netkan

### LINKED ISSUE #2809: [Feature] Avoid Spurious build.json download in NetKAN
Problem
-------
#2808 ensures NetKAN is up to date, but this file doesn't change super regularly. So the process will refresh it > 1800 times per index run, which is less than ideal.

Suggestions
-----------
We could use a special message type + webhook to handle this, but we're a little ways off from having the webhooks moved over to the new infra. 

For now we could put a time stamp in the refresh logic and short cut it if it has been checked in the last 4 hours. Which will mean it gets checked once per index run.

HebaruSan:
## Problem

As of #2808, we now load the remote build map every time netkan.exe runs in order to validate that generated metadata matches at least one known game version. This means that per bot pass, we download https://raw.githubusercontent.com/KSP-CKAN/CKAN-meta/master/builds.json 1800 times. This is much more frequent than necessary considering that weeks or months typically pass between KSP versions.

## Changes

Now `KspBuildMap` is rearranged to allow client code to choose the Remote, Cached, or Embedded build map. The default is the same as before: load Cached with Remote and then Embedded as fallbacks if nothing special is done, otherwise load Remote with Cached and then Embedded as fallbacks if `Refresh` is called. However, now there's also the option to **only** load the Embedded build map, which is used in the Netkan game version validator. This will reduce network usage by the bot.

Fixes #2809.

### DIFF
--- a/Core/GameVersionProviders/IKspBuildMap.cs
+++ b/Core/GameVersionProviders/IKspBuildMap.cs
@@ -3,12 +3,19 @@ using System.Collections.Generic;
 
 namespace CKAN.GameVersionProviders
 {
+    public enum BuildMapSource
+    {
+        Embedded,
+        Remote,
+        Cache,
+    };
+
     public interface IKspBuildMap
     {
         KspVersion this[string buildId] { get; }
 
         List<KspVersion> KnownVersions { get; }
 
-        void Refresh();
+        void Refresh(BuildMapSource source = BuildMapSource.Remote);
     }
 }
--- a/Core/GameVersionProviders/KspBuildMap.cs
+++ b/Core/GameVersionProviders/KspBuildMap.cs
@@ -2,9 +2,9 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
-using CKAN.Versioning;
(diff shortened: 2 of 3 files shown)


techman83:
I am probably missing something, but where does the build map get embedded for NetKAN?

HebaruSan:
It's included in the core csproj:

https://github.com/KSP-CKAN/CKAN/blob/c3ed4b3eac7d057fc5c39cb7dde01187b01c421e/Core/CKAN-core.csproj#L151-L153

https://github.com/KSP-CKAN/CKAN/blob/master/Core/builds.json

The result is that both `ckan.exe` and `netkan.exe` end up with an internal copy of `builds.json` that is current as of when they were compiled. The downside is that if you use an old executable it may get out of date (hence the preference for the remote build map), but for netkan.exe as used by the bot this is a non-issue because mechanisms to keep `netkan.exe` up to date that are and will be in place.

techman83:
Ok, that's awesome and the rest LGTM!

PR #2815, opened 2019-06-28
Fix Netkan schema resource name on Windows

HebaruSan:
The embedded schema is given a weird name when compiling on Windows:

image

This makes `netkan.exe` fail to run.

Now we set the `LogicalName` to be what it already is on Mono, `CKAN.NetKAN.CKAN.schema`.

### DIFF
--- a/Netkan/CKAN-netkan.csproj
+++ b/Netkan/CKAN-netkan.csproj
@@ -159,7 +159,9 @@
     <Compile Include="Validators\VersionStrictValidator.cs" />
   </ItemGroup>
   <ItemGroup>
-    <EmbeddedResource Include="../CKAN.schema" />
+    <EmbeddedResource Include="../CKAN.schema">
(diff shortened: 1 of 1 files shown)


Netkan/CKAN-netkan.csproj
```diff
@@ -159,7 +159,9 @@
     <Compile Include="Validators\VersionStrictValidator.cs" />
   </ItemGroup>
   <ItemGroup>
-    <EmbeddedResource Include="../CKAN.schema" />
+    <EmbeddedResource Include="../CKAN.schema">
```
techman83: Just a very minor comment, the path is a little inconsistent with the majority of the file. Though I note that the before build also uses forward slashes.

I come from a forward slash world, so prefer `../foo` but being C# I'm guessing the common convention is `..\bar`

techman83:
Aside from my minor nit pick, LGTM and merging with thanks!

PR #2819, opened 2019-06-29
Netkan option to validate .ckan file

HebaruSan:
## Motivation

https://github.com/KSP-CKAN/xKAN-meta_testing/pull/51#issuecomment-506614403

> Also, in theory NetKAN can consume a CKAN and output a validated CKAN. Maybe md5 before/after and we can catch when a CKAN is handcrafted (which we never want to do).

This would help to unify our pull request validation logic, as the CKAN-meta script could use `netkan.exe` for validation just as the NetKAN script does.

## Changes

Now a new command format is supported:

```bash
mono netkan.exe --validate-ckan Astrogator/Astrogator-v0.9.2.ckan
```

This will load the indicated .ckan file and run Netkan's standard validators on it (see #2788). If there are any problems, the usual error will be printed and the process will exit with a failure code.

If the .ckan file is valid, it will be printed with standard pretty formatting and property sorting, as if it had just been generated by `netkan.exe`, and the process will exit with a success code.

### DIFF
--- a/Netkan/CmdLineOptions.cs
+++ b/Netkan/CmdLineOptions.cs
@@ -40,6 +40,9 @@ namespace CKAN.NetKAN
         [Option("queues", HelpText = "Input,Output queue names for Queue Inflator mode")]
         public string Queues { get; set; }
 
+        [Option("validate-ckan", HelpText = "Name of .ckan file to check for errors")]
+        public string ValidateCkan { get; set; }
+
         [Option("version", HelpText = "Display the netkan version number and exit")]
         public bool Version { get; set; }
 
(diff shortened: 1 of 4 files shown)


techman83:
This is awesome! We have the hashing functions already, maybe an --output-sha256 that we can compare (or maybe it can do it internally?). Not needed to merge this.

DasSkelett:
> If the .ckan file is valid, it will be printed with standard pretty formatting and property sorting, as if it had just been generated by `netkan.exe`, and the process will exit with a success code.

Could we have an option to overwrite the .ckan if it is valid, to overwrite it with the pretty formatting? E.g. a `[-o]` flag.

techman83:
Technically that's NetKANs job 😉 I think maybe I didn't think through the option and got caught up in the excitement 😅

Although there is hope, as part of the validate command checksum the input and output to see if they match, potentially giving a different exit code for each exception type.

Issue #2771, opened 2019-06-01
Port to .NET Core


Abuseware:
**This isn't issue!**

Please, can you (sup)port this awesome tool from classic .NET to .NET Core?

I know, there isn't any GUI toolkit by default (so far), but the GUI part is nearly unusable on Mono anyway, especially on HiDPI displays. Porting should be pretty straightforward, almost all interfaces you're using is implemented in at least one library compatible with .NET Core.

HebaruSan:
Why?

Abuseware:
Because .NET Core is waaay faster in comparison to Mono? Because it has more predictable and reliable codebase shared between platforms. Mono isn't near MS implementation in any way, and has long story of bad written, buggy and slow code which only make problems with programs. It even has problems with 64 bit builds - we have 64 bit CPUs in PC for something like 16 years now...

HebaruSan:
> there isn't any GUI toolkit by default

So would this change mean that the GUI would no longer run?

Abuseware:
It mean that with proper management it will be fully compatible with both versions of framework, GUI for classic .NET and console only for Core, as separate executables. Most of the work is replacing libraries with compatible ones and adding a target in project build settings. I don't have Windows, don't have Visual Studio, so I can't deliver reliable and tested code for pull request.

politas:
I do all my dev work on KSP-CKAN using monodevelop on Linux. Visual Studio is not required to contribute.

Abuseware:
Good for you for using MonoDevelop. As far as I know, there is only subtle support for .net core, even in comparison to vs code/sublime text/atom.

Second, [here](https://youtu.be/mDr1S2jr0gg) is link to screen capture of CKAN GUI under MacOS, as you can see, is fulfilled with display bugs, font rendering issues, and this is with "Retina" mode disabled at all. While it's on there are more glitches, and even few of windows (like settings) goes center on screen, while title bar and bottom part of window disappears under and above desktop, and there isn't any obvious method of scaling it down from this state...

Third. It works only on mono compiled via Xamarin, with the default build (bundled in homebrew for example) doesn't start it at all - there was few bug reports for that, by ex. [this](https://github.com/KSP-CKAN/CKAN/issues/2630). And this isn't CKAN or MacOS issue, it's buggy Mono.

[...]

PR #2820, opened 2019-06-29
Use New JSON Config File; Add .NET Standard 2.0 Target to CKAN-core

### LINKED ISSUE #2799: [Feature] Refactor CKAN-core to use .NET standard 2.0
I am considering creating a new GUI for CKAN which is native to macOS. I would like to make the GUI using Xamarin.Mac and Xamarin.Forms. In order to compile CKAN for Xamarin.Mac, CKAN-core must be moved to .NET Standard 2.0 (instead of .NET Framework 4.5).

If I make a pull request refactoring CKAN-core, would the project be open to accepting it? Furthermore, if I were to go further and actually create the Xamarin GUI, would the project be open to accepting that as well?

I have done some preliminary work towards such a refactor already, and so my understanding is that the refactor involves the following changes:

1. Core/CKAN-core.csproj must be entirely rewritten.
2. Within CKAN-core, SHA1Cng must be replaced with another SHA1 provider because SHA1Cng is not supported in .NET Standard 2.0. The other .NET providers are supported, so this change is effectively trivial.
3. Several dependencies must be swapped out for drop-in replacements on Nuget which target .NET Standard
3. ChinhDo.Transactions.FileManager must be replaced/rewritten to support .NET Standard 2.0 (A blueprint for how to do this can be found here: https://github.com/VirtoCommerce/vc-platform-core/issues/51)
4. The remaining CKAN components must be moved from .NET 4.5 to .NET 4.6.1.

jbrot:
This pull request creates two major changes: first, it replaces the use of the Windows registry with a JSON configuration file. This is more idiomatic on platforms other than Windows, and is more portable as the registry is not included in .NET Standard.

The second major change made is to allow CKAN-core to be compiled for .NET Standard 2.0. .NET Standard is an API specification that allows for broad compatibility across .NET implementations. Note that .NET Standard is _not_ a runtime, and so to actually run (and test) the .NET Standard version of CKAN-core, we must pick a .NET Standard 2.0 compliant implementation.

The default target of CKAN, .NET Framework 4.5 is not .NET Standard 2.0 compliant. The first .NET Standard 2.0 compliant version of .NET Framework is .NET Framework 4.6.1. To that end, I added a second target to the Tests project in the same way I added a second target to CKAN-core. The new Tests target runs the .NET Standard 2.0 version of CKAN-core on Microsoft's new cross-platform .NET Core 2.1. 

You can switch between the already existing .NET Framework 4.5 target and the new .NET Core 2.1 target through the build configuration. The `Debug` and `Release` configurations build all of the components against .NET Framework 4.5 (nothing has changed here). The new `Debug_NetCore` and `Release_NetCore` configurations will just build the .NET Standard 2.0 version of CKAN-core and the .NET Core 2.1 version of Tests.

The new configurations work perfectly with Cake. For instance, you can build and test CKAN-core against .NET Core 2.1 by running `./build test --configuration=Debug_NetCore.` Visual Studio Mac does not yet support projects with multiple target frameworks, so it can only build the old configurations. If you select the `Debug_NetCore` or `Release_NetCore` configuration in Visual Studio Mac, it will get confused and not build. I have not tested the new configurations in Visual Studio on Windows, but I suspect it may not like the new configurations either.

Note that this pull request lays the groundwork for running the entirety of CKAN on .NET Core, but only actually configures CKAN-core to run on .NET Core. To actually run CKAN on .NET Core, one would need to modify CKAN-ConsoleUI to allow it to be built for .NET Standard 2.0 and modify CKAN-cmdline to run on .NET Core. If more people show interest in #2771 I can make these changes, but at the moment I am more interested in creating a better macOS GUI and will leave them for someone else.

Fixes #2799


DasSkelett:
```
Utilities.cs(31,104): error CS0246: The type or namespace name 'TxFileManager' could not be found (are you missing a using directive or an assembly reference?) [/home/travis/build/KSP-CKAN/CKAN/Core/CKAN-core.csproj]
```
That's the reason it's failing.

jbrot:
Yep. It appears that I may have messed up some of the conditional directives. The .NET 45 target is building successfully, though, which is a good sign. I will hopefully publish a corrected version relatively soon.

DasSkelett:
Well, from a quick first look, all three UIs still work fine.
However, netkan seems to have problems loading the build map:
`378 [1] WARN CKAN.GameVersionProviders.KspBuildMap (null) - Could not refresh the build map from any source`

Core/CKAN-core.csproj
```diff
@@ -1,159 +1,64 @@
...
+    <PackageReference Include="log4net" Version="2.0.8" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
+    <PackageReference Include="TxFileManager.NETStandard" Version="2.0.0" />
+    <PackageReference Include="Shaman.CurlSharp" Version="0.6.3.1" />
+    <PackageReference Include="System.Security.Permissions" Version="4.5.0" />
+    <PackageReference Include="Microsoft.Win32.Registry" Version="4.5.0" />
+    <PackageReference Include="Microsoft.CSharp" Version="4.5.0" />
   </ItemGroup>
-  <ItemGroup>
-    <EmbeddedResource Include="builds.json" />
```
HebaruSan: We need this embedded resource.

jbrot: Ah! That explains why netkan stopped working.

DasSkelett:
And mono < 16.0 has some problems...
```
Converters/JsonIgnoreBadUrlConverter.cs(11,46): error CS0012: The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. [/home/travis/build/KSP-CKAN/CKAN/Core/CKAN-core.csproj]
Types/CkanModule.cs(296,10): error CS0012: The type 'Attribute' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. [/home/travis/build/KSP-CKAN/CKAN/Core/CKAN-core.csproj]
CkanTransaction.cs(11,23): error CS0246: The type or namespace name 'TransactionScope' could not be found (are you missing a using directive or an assembly reference?) [/home/travis/build/KSP-CKAN/CKAN/Core/CKAN-core.csproj]
CkanTransaction.cs(16,24): error CS0246: The type or namespace name 'TransactionOptions' could not be found (are you missing a using directive or an assembly reference?) [/home/travis/build/KSP-CKAN/CKAN/Core/CKAN-core.csproj]
```
And so on.

HebaruSan:
I don't understand the motivation or consequences of this, so requesting review from @Olympic1 and @dbent in the hope that they do.

Core/CKAN-core.csproj
DasSkelett: :+1: the new commit fixed it.

jbrot:
@HebaruSan I discussed my personal motivation for this in #2799, and the most immediate consequence is that CKAN.dll becomes significantly more portable. My intention was that the current .NET 4.5 target is unaffected. However, there is a lot of subtlety with build files so there may be unintentional consequences, which I hope should be caught during the review process now.

HebaruSan:
Allow me to rephrase: I don't know what .NET Standard 2.0 is or why it matters. It's not something I've had to worry about before.

jbrot:
.NET Standard is just a list of APIs.

C# code gets compiled into IL (Intermediate Language) which is interpreted by an implementation of Microsoft's CLI (Common Language Infrastructure). Currently, there exist several such implementations. The most commonly known are the traditional .NET Framework and Mono. Some less common implementations include Xamarin.Droid, Xamarin.iOS, and .NET Core (essentially a faster Mono, but made by Microsoft).

This creates a problem: each of these implementations supports a slightly different set of the .NET APIs, which means writing code that runs on all of them is challenging—even though this is something developers would really like, especially for core libraries.

.NET Standard is Microsoft's attempt to solve this problem. It is a list of the .NET APIs Microsoft considers essential. The way it works is that developers trying to write portable code only use the APIs that are in .NET Standard, and then each of the implementations of the CLI makes sure that they support all of the .NET Standard APIs. Thus, if your code compiles against .NET Standard, it can be run on all of the .NET Standard compliant CLI implementations.

Currently, all of the major CLI implementations (every one I listed earlier) is .NET Standard 2.0 compliant, so by making CKAN.dll compile against .NET Standard, you can now use CKAN.dll in each of these implementations. This is relevant to me personally because I am trying to create a Xamarin GUI, which runs on a different CLI implementation than the currently targeted .NET Framework. Another user a few weeks ago wanted to run CKAN on .NET Core, ostensibly for the performance gains, and this also allows them to do that—although, the front end would need to be adapted to work, which is an entirely separate beast.

Hopefully this clears things up a little bit, please let me know if any of this doesn't make sense and I can try to clarify. :)

HebaruSan:
Thanks for the primer! That brings me (hopefully) up to speed.

Can we update the .travis.yml file to make it build and test on these other implementations? Currently it would still just try four versions of Mono, so if we added a new dependency that didn't work in .NET Core, we wouldn't know.

Core/KSP.cs
```diff
@@ -7,7 +7,11 @@
 using System.Text.RegularExpressions;
 using System.Transactions;
 using Autofac;
+#if NETSTANDARD2_0
```
HebaruSan: I am really not excited about maintaining `#if` statements. Is there no other way to do this?

jbrot:
It should build for the versions of mono in the travis.yml now (at least it did on my machine—famous last words, I know). Configuring the CKAN.dll unit tests to run on .NET core seems reasonable, and I will look into that in the coming days.

In theory, everything except for the Winforms GUI should be able to be ported to .NET core, but that is a bit of an undertaking and would require rewriting every components .csproj in an analogous fashion to how I rewrote CKAN-core.csproj. I'm not strictly opposed to doing this, but I'm not super gung-ho about it either.

I agree the \#if directives are not ideal. I will look into if there is another way to deal with them. Currently, they are necessary because the .NET Standard and .NET Framework targets depend on slightly different versions of TxFilemanager, which results in the difference in namespace. I will try to either unify the version referenced, or find a way to reduce the needed \#if directives, preferably to only one location.

Cmdline/CKAN-cmdline.csproj
```diff
@@ -92,7 +92,7 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\Core\CKAN-core.csproj">
-      <Project>{3B9AEA22-FA3B-4E43-9283-EABDD81CF271}</Project>
+      <Project>{EBE1FF06-FAB5-4EFD-A65D-7085C5281919}</Project>
```
HebaruSan: This is another case of me not knowing enough about this, but does the project UUID *need* to change? Maybe it's necessary, but if it's not, my preference would be to keep it the same.

Netkan/CKAN-netkan.csproj
```diff
@@ -160,7 +160,6 @@
   </ItemGroup>
   <ItemGroup>
     <EmbeddedResource Include="..\CKAN.schema">
-      <LogicalName>CKAN.NetKAN.CKAN.schema</LogicalName>
```
HebaruSan: Was there a problem with this `LogicalName` element? It was added to fix something.

Cmdline/CKAN-cmdline.csproj
jbrot: I’m not sure. Visual Studio did that automatically, so I will look into why that happened.

Netkan/CKAN-netkan.csproj
jbrot: Likewise, Visual Studio did this automatically. Since it was added to fix an issue I’m the past, I will definitely add it back since there is no need to remove it.

Cmdline/CKAN-cmdline.csproj
jbrot: So the reason why it changed is because I removed the line specifying the project's GUID, causing Visual Studio to automatically assign it a new GUID. I removed the line because it is optional in the new csproj format. However, removing one line from the csproj is certainly not sufficient reason to change the GUID. :smile

Netkan/CKAN-netkan.csproj
HebaruSan: This is fixed too. Thanks!

DasSkelett:
Soooo... how can I specify which target I want to compile Core against?
```
./build --target="netstandard2.0"
```
gives me
```
Error: One or more errors occurred. (The target 'netstandard2.0' was not found.)
The target 'netstandard2.0' was not found.
```
Equally with `net45`.

jbrot:
What’s published right now is not quite right, mainly because you can’t do what you just tried to do. My next commit will finish implementing the ability to target .Net Core. It’s taking me longer than I expected to get it working, though, because of some difficulties with NUnit on .Net Core. At this point, I expect I won’t be ready to publish this commit until much later today or early tomorrow.

jbrot:
From a build configuration perspective, I think the addition of .Net support is now complete. However, despite things compiling, they don't actually work just yet. The problem is that the Win32 Registry is not supported by .Net Core.

Fortunately, barring one instance in KSPPathUtils, all direct interactions with the registry are located in the Win32Registry class. Therefore, it would not be a particularly large undertaking to replace the registry internally. My gut reaction is to ensure all uses go through the IWin32Registry interface, and then do some kind of dependency injection to keep things working the same way on windows, while using a config file on the unix-like platforms.

This seems like slightly beyond the stated goals of this pull request, though, so it probably makes sense to create a separate issue and pull request for discussing what to do about the registry. I see three paths forward: 1) Rework the registry as part of this pull request; 2) Mark the .Net Core builds in the travis.yml with allow-failures, complete this pull request, and then rework the registry in a separate pull request; and 3) Rework the registry in a separate pull request, and keep this pull request in limbo until after the registry pull request gets merged. A fourth, unfortunate option is to say that the registry interface should not be touched and abandon this whole undertaking, although I would really prefer not to do that. :wink:

HebaruSan:
Supposedly `Microsoft.Win32.Registry` is available as a NuGet package, does that help us at all here?

https://github.com/dotnet/corefx/issues/31516#issuecomment-409764023

I've tried to read through that issue and I can't quite follow it enough to say.

HebaruSan:
> I see three paths forward:

Regarding next steps, I think we generally try to avoid leaving any of our projects' master branches in a "broken" state. You never know when the project lead might pop in to publish a new release. So that would imply going for option 1 or 3 in your list.

HebaruSan:
I'm getting confused again as to the point of this "core" stuff. Wasn't .NET already supposed to be cross platform from the beginning? And indeed, we already have Mono for non-Windows platforms. It seems like we're just hunting for a library that does what Mono already does to simulate the Windows registry.

HebaruSan:
@jbrot, these last few commits look great. One thought occurred to me:  We could migrate **all** users to the new JSON file, so the registry would no longer be needed after a few releases. The Windows version could load from the registry (as a fallback) and then save to the new JSON file. Any thoughts? Good/bad idea?

jbrot:
@HebaruSan That seems reasonable. It just occurred to me that I need to write migration code, anyways, so that on non-Windows operating systems the new config file gets deployed seamlessly.

Core/Net/Net.cs
```diff
@@ -99,6 +100,7 @@ public static string Download(string url, out string etag, string filename = nul
             catch (Exception ex)
             {
                 log.InfoFormat("Download failed, trying with curlsharp...");
+                Console.WriteLine("Download failed, trying with curlsharp...");
```
HebaruSan: Please don't push these. We don't want debugging output in `Core.Net`.

jbrot: They are definitely not intended to be merged. However, I want them to be run on the build server, as I am struggling to reproduce the failed unit test locally.

HebaruSan: Ahh, OK. Carry on then.

jbrot:
I've identified the issue with the .Net Core unit test. It turns out that the test URL "https://spacedock.info/mod/132/Contract%20Reward%20Modifier/download/2.1" redirects to "http://spacedock.info/content/DMagic_247/Contract_Reward_Modifier/Contract_Reward_Modifier-2.1.zip". Importantly, the https url gets redirected to http. The .Net Core implementation of HttpWebRequest correctly identifies this as a security issue, and refuses to redirect automatically. 

When the HttpWebRequest fails, CKAN then tries to use curl. Both on my computer and on an Ubuntu VM I set up, the curl fallback works flawlessly. The only reason why I noticed something was wrong at all was because on the build server, the curl fallback fails with a cryptic SslConnectError. I believe this may be a bug in the version of curl used on the build server, and can probably be fixed by updating curl.

So, I see three ways around this. The most direct way would be to create a subclass of WebClient which will follow the redirect and ignore the change from https to http. This is not terribly involved, but if I'm doing a refactor of Net.cs, it should probably be to make it use HttpClient instead of WebClient. Note that the comment around line 390 of Net.cs says "Maybe one day we'll be able to use HttpClient" indicating that, at the time this code was written, HttpClient was not an option. As far as I can tell, HttpClient is available in .NET 4.5 / Mono 4.0.0, so it should be fair game—but perhaps there is some other reason I'm not seeing.

An easier way to fix the issue is to simply change the version of curl on the build server, fixing the SslConnect Error. This just masks the issue and doesn't actually resolve it. Furthermore, on Windows, there is no curl fallback, so this won't work everywhere. However, the issue only shows up on the .NET Core version, which will not be seeing widespread use in the immediate future, and certainly not on Windows—so this is not as problematic as it first appears. N.B., it might be a good idea to have the build server run the unit tests on windows and macOS as well, but that is definitely out of scope here.

Finally, if we are willing to just mask the issue with a curl update, then there's an even easier way to mask the issue: just disable this specific unit test on .NET Core. Please let me know your thoughts.

HebaruSan:
Permalink to the test in question:  https://github.com/KSP-CKAN/CKAN/blob/56eded6390f3f63787c06ca9e9e01c2fabb43729/Tests/Core/Net/Net.cs#L48-L60

Starting from the root cause, we should tell SpaceDock their redirects are messed up:
https://github.com/KSP-SpaceDock/SpaceDock/
Though a pull request with a fix (plus a PM to VITAS on the forum or Discord) would probably get more traction than an issue given typical response times.

Then, maybe the fallback to curl should not happen for a security related exception? I don't know off hand how easy it would be to make that distinction, but if it's possible, that seems like a more responsible way to handle that case. WebClient isn't failing due to some weird reason, it's explicitly deciding it's not safe to handle the request. This would naturally break that test on all platforms though, so maybe we should change it to `Assert.Throws` until SpaceDock's redirects are fixed?

If you want to convert us over to `HttpClient`, more power to you, but I think that's extreme overkill.

I think we should update the libcurl, but only after the other stuff is resolved in a sensible way.

These are just my initial reactions, and I'd like to hear what others think.

HebaruSan:
> This would naturally break that test on all platforms though

Did I misunderstand that? It works fine on Windows?

jbrot:
I completely agree that talking to SpaceDock directly is probably the best course of action. Although, it may be the case that the redirect to HTTP is intentional to allow for caching. Two points of clarification:

1. On further research, the curl version on the build server is modern, and downloads the URL without difficulty when invoked on the command line, so the source of the "SslConnectError" remains elusive.
2. Although the redirect from HTTPS to HTTP always occurs, it only gets flagged as an error on .NET Core (the new build path this PR creates). On the old build path, and in curl, this redirect happens without question.

HebaruSan:
> the redirect from HTTPS to HTTP always occurs, it only gets flagged as an error on .NET Core 

Hmm, that makes the correct course of action far more murky. Either we suppress an exception that one platform throws, or we accept that our code behaves differently on different platforms.

Thank goodness for cross platform "standards"!

HebaruSan:
Have I dug the right part of the log out for the exception that's unique to .NET Core?

```
  Error Message:
  But was:  <System.Net.WebException: The remote server returned an error: (302) FOUND.
   at System.Net.HttpWebRequest.GetResponse()
   at System.Net.WebClient.GetWebResponse(WebRequest request)
   at System.Net.WebClient.DownloadBits(WebRequest request, Stream writeStream)
   at System.Net.WebClient.DownloadFile(Uri address, String fileName)
   at System.Net.WebClient.DownloadFile(String address, String fileName)
   at CKAN.Net.Download(String url, String& etag, String filename, IUser user) in /home/travis/build/KSP-CKAN/CKAN/Core/Net/Net.cs:line 96
   at CKAN.Net.Download(String url, String filename, IUser user) in /home/travis/build/KSP-CKAN/CKAN/Core/Net/Net.cs:line 75
   at Tests.Core.Net.Net.<>c.<SpaceDockSSL>b__5_0() in /home/travis/build/KSP-CKAN/CKAN/Tests/Core/Net/Net.cs:line 54
   at NUnit.Framework.Internal.Reflect.DynamicInvokeWithTransparentExceptions(Delegate delegate)
```

My initial idea to catch this one exception separately wouldn't work if that's it, since just about everything WebClient raises is a WebException...

jbrot:
> Have I dug the right part of the log out for the exception that's unique to .NET Core?

Yep, that's the correct part. I've also just published a commit which explicitly handles the redirect, which seems sufficient to me.

techman83:
> 2\. Although the redirect from HTTPS to HTTP always occurs, it only gets flagged as an error on .NET Core (the new build path this PR creates). On the old build path, and in curl, this redirect happens without question. This is by design, and different from .NET Framework behavior as documented here: [dotnet/corefx#19183](https://github.com/dotnet/corefx/issues/19183)

There is a good reason that a redirect http -> https fails. In the browser this will throw up an error. We could use something hosted on GitHub as a work around, but we should probably work out why spacedock is not giving a https download link.

https://github.com/KSP-SpaceDock/SpaceDock

I wouldn't be in favour of working around what's realistically a security hole in a 3rd partie's host. But there are some pragmatic considerations we probably need to think about.

jbrot:
With those last unit tests up, I think this PR is now feature-complete. :smile:

DasSkelett:
> Should I squash the commits after we finish up code review/resolving the HTTPS->HTTP issue?

I already prepared myself to do the Linus and [complain about the commit history](https://lkml.org/lkml/2019/6/10/838) :smile:
So yes, I think squashing the commits makes sense before merge.

It would be great if you could extend the PR summary a bit, and also include instructions on how to build for .Net Standard/Core, so you don't have to fight through the cake config trying to figure that out.

But otherwise, great work! Thanks a lot!

jbrot:
The commits are now nice and pretty, and I've updated the PR description to be much more through and accurate. :smile:

build.cake
```diff
@@ -8,6 +8,7 @@ using Semver;
 
 var target = Argument<string>("target", "Default");
 var configuration = Argument<string>("configuration", "Debug");
+var buildFramework = configuration.EndsWith("_NetCore") ? "netcoreapp2.1" : "net45";
```
DasSkelett: You are testing here for `EndsWith("_NetCore")`, in the CKAN-Core.csproj you are testing for `EndsWith("NetCore")`.
It's not critical, but it would be nice if it's consistent.

DasSkelett:
Trying to build with mono gives me the following:
```
./build --configuration=Debug_NetCore
Alle in "packages.config" aufgeführten Pakete sind bereits installiert.
Setup
Generate-GlobalAssemblyVersionInfo
Restore-DotNetCore
An error occurred when executing task 'Restore-DotNetCore'.
Teardown
[quote]

Error: One or more errors occurred. (.NET Core CLI: Could not locate executable.)
.NET Core CLI: Could not locate executable.
```

jbrot:
> Trying to build with mono gives me the following:

That is the expected behavior. .NET Core is a separate program that you need to install in order to compile against it. You can download it [here](https://dotnet.microsoft.com/download/archives) or from your package manager (it’s usually called something like ‘dotnet’). Note that even with .NET Core installed, you will still need mono if you want to build with cake. If you just have .NET Core you can run ‘dotnet test’ from the repo root to build and test directly.

HebaruSan:
The dotnet runtime package in Ubuntu 19.04 has two unsatisfiable dependencies, but this worked:

https://dotnet.microsoft.com/download/linux-package-manager/ubuntu19-04/runtime-2.2.0

HebaruSan:
The `config.json` file has a JSON encoded string as the value of its `KSPBuilds` property, so it's JSON-within-JSON. Can we do anything about that?

```json
{
  "KspInstances": [],
  "AuthTokens": [],
  "AutoStartInstance": null,
  "DownloadCacheDir": "/home/user/.local/share/CKAN/downloads",
  "CacheSizeLimit": null,
  "RefreshRate": null,
  "KSPBuilds": "{\n  \"builds\": {\n    \"464\": \"0.23.5.464\",\n    \"559\": \"0.24.2.559\",\n    \"642\": \"0.25.0.642\",\n    \"705\": \"0.90.0.705\",\n    \"830\": \"1.0.0.830\",\n    \"842\": \"1.0.2.842\",\n    \"861\": \"1.0.4.861\",\n    \"1024\": \"1.0.5.1024\",\n    \"1028\": \"1.0.5.1028\",\n    \"1172\": \"1.1.0.1172\",\n    \"1174\": \"1.1.0.1174\",\n    \"1180\": \"1.1.0.1180\",\n    \"1183\": \"1.1.0.1183\",\n    \"1196\": \"1.1.0.1196\",\n    \"1203\": \"1.1.0.1203\",\n    \"1209\": \"1.1.0.1209\",\n    \"1215\": \"1.1.0.1215\",\n    \"1224\": \"1.1.0.1224\",\n    \"1228\": \"1.1.0.1228\",\n    \"1230\": \"1.1.0.1230\",\n    \"1250\": \"1.1.1.1250\",\n    \"1260\": \"1.1.2.1260\",\n    \"1289\": \"1.1.3.1289\",\n    \"1473\": \"1.2.0.1473\",\n    \"1479\": \"1.2.0.1479\",\n    \"1485\": \"1.2.0.1485\",\n    \"1486\": \"1.2.0.1486\",\n    \"1489\": \"1.2.0.1489\",\n    \"1494\": \"1.2.0.1494\",\n    \"1499\": \"1.2.0.1499\",\n    \"1500\": \"1.2.0.1500\",\n    \"1509\": \"1.2.0.1509\",\n    \"1517\": \"1.2.0.1517\",\n    \"1520\": \"1.2.0.1520\",\n    \"1523\": \"1.2.0.1523\",\n    \"1532\": \"1.2.0.1532\",\n    \"1539\": \"1.2.0.1539\",\n    \"1540\": \"1.2.0.1540\",\n    \"1546\": \"1.2.0.1546\",\n    \"1548\": \"1.2.0.1548\",\n    \"1553\": \"1.2.0.1553\",\n    \"1563\": \"1.2.0.1563\",\n    \"1564\": \"1.2.0.1564\",\n    \"1569\": \"1.2.0.1569\",\n    \"1574\": \"1.2.0.1574\",\n    \"1576\": \"1.2.0.1576\",\n    \"1583\": \"1.2.0.1583\",\n    \"1584\": \"1.2.0.1584\",\n    \"1586\": \"1.2.0.1586\",\n    \"1604\": \"1.2.1.1604\",\n    \"1622\": \"1.2.2.1622\",\n    \"1727\": \"1.2.9.1727\",\n    \"1730\": \"1.2.9.1730\",\n    \"1737\": \"1.2.9.1737\",\n    \"1738\": \"1.2.9.1738\",\n    \"1743\": \"1.2.9.1743\",\n    \"1750\": \"1.2.9.1750\",\n    \"1758\": \"1.2.9.1758\",\n    \"1764\": \"1.2.9.1764\",\n    \"1773\": \"1.2.9.1773\",\n    \"1781\": \"1.2.9.1781\",\n    \"1790\": \"1.2.9.1790\",\n    \"1796\": \"1.2.9.1796\",\n    \"1800\": \"1.2.9.1800\",\n    \"1804\": \"1.3.0.1804\",\n    \"1836\": \"1.3.1.1836\",\n    \"1847\": \"1.3.1.1847\",\n    \"1855\": \"1.3.1.1855\",\n    \"1863\": \"1.3.1.1863\",\n    \"1891\": \"1.3.1.1891\",\n    \"2077\": \"1.4.0.2077\",\n    \"2089\": \"1.4.1.2089\",\n    \"2110\": \"1.4.2.2110\",\n    \"2152\": \"1.4.3.2152\",\n    \"2215\": \"1.4.4.2215\",\n    \"2243\": \"1.4.5.2243\",\n    \"2256\": \"1.4.5.2256\",\n    \"2332\": \"1.5.0.2332\",\n    \"2335\": \"1.5.1.2335\",\n    \"2395\": \"1.6.0.2395\",\n    \"2401\": \"1.6.1.2401\",\n    \"2483\": \"1.7.0.2483\",\n    \"2539\": \"1.7.1.2539\",\n    \"2555\": \"1.7.2.2555\",\n    \"2556\": \"1.7.2.2556\",\n    \"2594\": \"1.7.3.2594\"\n  }\n}\n"
}
```

jbrot:
> The `config.json` file has a JSON encoded string as the value of its `KSPBuilds` property, so it's JSON-within-JSON. Can we do anything about that?

I implemented it that way because I didn't want to change the IWin32Registry interface, and the build map is stored as a string in the registry. Looking at it now, I realize that you are correct—it doesn't make much sense to have a JSON object stored as a string within another JSON object. I'll try to make this change sometime over the weekend,

DasSkelett:
> .NET Core is a separate program that you need to install in order to compile against it.

Oh I somehow thought mono bundles it.

> The dotnet runtime package in Ubuntu 19.04 has two unsatisfiable dependencies, but this worked:

Thanks!

Also regarding the registry/config: Should we delete the old registry entries after migration?
I don't like leaving old, unused entries in the registry.

jbrot:
@DasSkelett I just drafted a commit to delete the registry entries after migration. I'm hesitant to push it, though, because the migration code runs as part of the test suite. So, if this gets added, then whenever you test this code your registry entries will get purge, resetting any other version of CKAN you may have installed. This may not actually be that big of a deal, though, since the registry values do all get migrated before being deleted and the new configuration file shouldn't be changed by running the test suite.

Olympic1:
For me this works on Windows (building and running)

DasSkelett:
Works for me on Linux - even with mono 6.0 ;) - too.
Another squash before or during merge would be nice.

jbrot:
Just did another squash. Let me know if there are any other changes you'd like made, or if it needs to be squashed further.

Core/Configuration/JsonConfiguration.cs
```diff
@@ -0,0 +1,419 @@
...
+            var instances = registry.GetInstances();
+            lock (_lock)
+            {
+                config.KspInstances = instances.Select(instance => new KspInstance
+                {
+                    Name = instance.Item1,
+                    Path = instance.Item2
+                }).ToList();
+
+                SaveConfig();
```
HebaruSan: Shouldn't SaveConfig be at the end of this function, after all the properties are set?

Core/Configuration/JsonConfiguration.cs
```diff
@@ -0,0 +1,419 @@
...
+                // try to migrate from the registry.
+                config = new Config();
+                config.KspInstances = new List<KspInstance>();
+                config.AuthTokens = new List<AuthToken>();
+
+                // Ensure the directory exists
+                new FileInfo(configFile).Directory.Create();
+
+                // Write the configuration to the disk
+                SaveConfig();
```
HebaruSan: `Migrate` will also call `SaveConfig`. Should this be in an `#else` block?

Core/Configuration/JsonConfiguration.cs
```diff
@@ -0,0 +1,419 @@
...
+
+#if !NETSTANDARD
+                // If we are not running on .NET Standard, try to migrate from the real registry
+                Migrate();
```
HebaruSan: Won't this overwrite the "new" config in the JSON file with the old values from the registry (which we can no longer edit) every time? I don't see any checks for whether the old or new settings are present.

Core/Configuration/JsonConfiguration.cs
```diff
@@ -0,0 +1,419 @@
...
+                    config.AuthTokens = new List<AuthToken>();
+                }
+            }
+            catch (Exception ex) when (ex is FileNotFoundException || ex is DirectoryNotFoundException)
+            {
+                // This runs if the configuration does not exist. We will create a new configuration and
+                // try to migrate from the registry.
+                config = new Config();
+                config.KspInstances = new List<KspInstance>();
+                config.AuthTokens = new List<AuthToken>();
```
HebaruSan: The `Config` class should initialize these `List`s itself rather than relying on external code to set them.

Core/Configuration/JsonConfiguration.cs
```diff
@@ -0,0 +1,419 @@
...
+
+                if (config.KspInstances == null)
+                {
+                    config.KspInstances = new List<KspInstance>();
+
+                }
+
+                if (config.AuthTokens == null)
+                {
+                    config.AuthTokens = new List<AuthToken>();
```
HebaruSan: The `Config` class should initialize these `List`s itself rather than relying on external code to set them.

Core/Configuration/JsonConfiguration.cs
```diff
@@ -0,0 +1,419 @@
...
+                return config.AuthTokens.Select(token => token.Host);
+            }
+        }
+
+
+        public bool TryGetAuthToken (string host, out string token)
+        {
+            lock (_lock)
+            {
+                foreach (AuthToken t in config.AuthTokens)
```
HebaruSan: This block would be simpler if `AuthTokens` was a `Dictionary<string, string>`. We would just call `config.AuthTokens.TryGetValue(host, out token)` and handle the missing case in an `else`.

Core/Configuration/RegistryConfiguration.cs
```diff
@@ -3,30 +3,25 @@
...
         private const           string CKAN_KEY           = @"HKEY_CURRENT_USER\Software\CKAN";
         private static readonly string CKAN_KEY_NO_PREFIX = StripPrefixKey(CKAN_KEY);
 
         private const           string authTokenKey         = CKAN_KEY + @"\AuthTokens";
         private static readonly string authTokenKeyNoPrefix = StripPrefixKey(authTokenKey);
 
-        static Win32Registry()
+        static RegistryConfiguration()
         {
             ConstructKey(CKAN_KEY_NO_PREFIX);
```
HebaruSan: If a user installs CKAN for the very first time with no JSON file **and** nothing in the registry, I think this will end up initializing the registry key with no data or default data, never to be used, since the `Migrate` logic relies on constructing a `RegistryConfiguration` object. Can we add a way to check whether the key exists without doing that?

Core/Configuration/JsonConfiguration.cs
jbrot: That's a good point. The current design was chosen to make the config file clearer with host and token explicitly labelled, but I don't think much clarity would be lost by switching to a dictionary.

Core/Configuration/JsonConfiguration.cs
jbrot: From what I've seen, you don't usually add any code beyond public properties to JSON serialized classes. I'm not exactly sure how constructor code plays out with serialization, so I will look into it. Provided you can actually move this code into the Config class, I will make that change as it will certainly make the code more clear.

Core/Configuration/RegistryConfiguration.cs
jbrot: Good call. I completely forgot to account for that. As it's currently written, the code works, but that would certainly be more correct.

jbrot: How do you feel about deleting the registry keys after migration? I had a brief discussion with @DasSkelett about it earlier, and so right now the code for that is written but not pushed to this branch.

Core/Configuration/JsonConfiguration.cs
jbrot: Currently SaveConfig() gets called after every write operation, which is analogous to how the registry works, where all changes are immediately persistent. I have the SaveConfig() here to go along with that pattern. I could probably change Migrate() to only save once at the end, although I'm not sure if the added complexity is justified.

Core/Configuration/RegistryConfiguration.cs
HebaruSan: > How do you feel about deleting the registry keys after migration?

It's appealing from an architectural cleanliness perspective, but there are also advantages to leaving the old data in place in case something goes wrong, and until the new code establishes a track record of stability, I'd vote for not deleting.

DasSkelett: On another thought, @HebaruSan is probably right, we should keep it for now, to ease testing and switching around. And since deleting the old registry entries will definitely break backwards-compatibility, it makes sense to release it only in a new major release.

HebaruSan:
We will need to update the pull request validation scripts after this, they currently assume they can reset this config by clearing Mono's fake registry:

- https://github.com/KSP-CKAN/xKAN-meta_testing/blob/81ad931206c6ff67d915d548e1f83db13874ce6b/NetKAN/build.sh#L50-L58
- https://github.com/KSP-CKAN/xKAN-meta_testing/blob/81ad931206c6ff67d915d548e1f83db13874ce6b/CKAN-meta/build.sh#L45-L53

(Not to be fixed in this PR, just a note for the future.)

Core/Configuration/JsonConfiguration.cs
jbrot: AuthTokens and KspInstances can be explicitly set to null in the config file, so these checks are still necessary, even with the values being initialized by Config.

jbrot:
@HebaruSan The last commit resolves nearly all of your comments. The one area that I did not resolve was with saving the config file excessively when migrating. I think I may have an elegant solution to that, which I will try to implement tomorrow.

.travis.yml
```diff
@@ -40,7 +49,7 @@ before_install:
 
 script:
     - ./build --configuration=$BUILD_CONFIGURATION
-    - ./build test+only --configuration=$BUILD_CONFIGURATION --where="cat!=FlakyNetwork"
+    - ./build test+only --configuration=$BUILD_CONFIGURATION --where="Category!=FlakyNetwork"
```
politas: I'm probably missing it, but I can't see anywhere else that this parameter name is changed?

jbrot: The NUnit test driver understands both cat and Category, but the .NET Core test driver only understands Category. I figured it was easier to use the common name instead of providing different strings. If the condition needs to get more complicated, though, you will probably need to have different strings.

Core/Configuration/JsonConfiguration.cs
jbrot: I have modified Migrate so that it only calls SaveConfig once at the end of the function.

Core/Configuration/JsonConfiguration.cs
jbrot: I'd like to keep the code in preprocessor blocks to a minimum, so I'd rather just call SaveConfig twice if we are migrating. If you think it's important, though, I'm more than happy to put it in an \#else block.

jbrot:
Anything else I can do to facilitate this getting merged?

HebaruSan:
> Anything else I can do to facilitate this getting merged?

Nope, two reviewer approvals is plenty. I just merged it, but GitHub hasn't figured that out yet. Thanks for taking this on!

Issue #2276, opened 2018-02-06
GUI uses nearly the same name for four different change set variables


HebaruSan:
## Problem

These are all members of `CKAN.Main` in GUI:

https://github.com/KSP-CKAN/CKAN/blob/bb028923969f731df7b2a66b592ad4c5e9f9baa0/GUI/Main.cs#L82

https://github.com/KSP-CKAN/CKAN/blob/bb028923969f731df7b2a66b592ad4c5e9f9baa0/GUI/Main.cs#L85

https://github.com/KSP-CKAN/CKAN/blob/bb028923969f731df7b2a66b592ad4c5e9f9baa0/GUI/MainChangeset.cs#L12

https://github.com/KSP-CKAN/CKAN/blob/bb028923969f731df7b2a66b592ad4c5e9f9baa0/GUI/MainChangeset.cs#L14

All four names are available for use inside of `CKAN.Main.UpdateChangesDialog`. To call this confusing would be an understatement. In terms of meaning, these names are indistinguishable, and at the syntactic level 3 of them vary only in their capitalization. If a programmer needs to read or write from the current change set, how likely is he to be able to figure out which variable does what? If a programmer is trying to investigate a problem with change sets, how easy will it be to discern the purpose of code using these variables?

## Suggestions

The three member variables should be consolidated into one. It's reasonable to have one private member variable tracking change sets, but two or three is asking for trouble.

When used as a parameter, the name should have a prefix describing how it's different from the member variable; for example it could be called `newChangeSet`. There would be no much less of accidentally typing newChangeSet when you meant to use the main changeSet variable, or vice versa.

politas:
I have always found this "feature" of the codebase a bit odd. My programming background strongly favours clearly described variable names. I think there are many other cases in our code where similar things happen, including some where the only difference is that one variable uses the "wrong" naming scheme.

HebaruSan:
This line is setting the wrong one:

https://github.com/KSP-CKAN/CKAN/blob/ddf9ac36eed0edfddd6c29f109be28bf23b6c41e/GUI/MainInstall.cs#L73-L76

HebaruSan:
We eliminated `Main.changeSet` in #2966 with the migration of the change set tab's logic into a `UserControl`, but we replaced it with a `Main.Changeset` control. So we still have several objects with nearly the same name, but now one of them is a `UserControl` class rather than `List<ModChange>`.

politas:
[...]

PR #2821, opened 2019-07-01
Queue module version changes in change set

### LINKED ISSUE #2714: Feature Request: Ability to install any specific version of a mod
Background
----------

**CKAN Version:** 1.26.0

**KSP Version:** 1.4.5

**Operating System:** W7

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** Yes

Problem
-------

**What steps did you take in CKAN?** Clicked on a mod version which is listed as incompatible with my current install, but in fact does work.

**What did you expect to happen?** A dialogue window to offer me the option to install the mod version.

**What happened instead?** Absolutely nothing.

### LINKED ISSUE #2715: Feature Request: Ability to queue mod version changes in the changelist
Background
----------

**CKAN Version:** 1.26.0

**KSP Version:** 1.4.5

**Operating System:** W7

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** Yes

Problem
-------

**What steps did you take in CKAN?** Double clicked a mod version to install as a downgrade from my currently installed version

**What did you expect to happen?** That the mod version would be queued in the changelist, allowing the selection of multiple concurrent operations on the install folder, to be executed after all selections had been made.

**What happened instead?** The mod version was installed immediately, rather than allowing the selection of multiple mod version changes.

**Why is this a problem?** Selecting version changes one at a time is relatively time consuming, compared to queueing all downloads/installs to be executed concurrently.

### LINKED ISSUE #2822: [Feature] Force install
Problem
-------
To install old but working modifications I have to open the list of versions.

Suggestions
-----------
I propose to add the "Force install" item to the context menu of the list of modifications, which will mark the modification for installation in spite of any restrictions.

HebaruSan:
## Background

Originally, each row in the GUI mod list only allowed the user to install one version of the corresponding module: the most recent version compatible with the user's game version. Installing incompatible modules was completely impossible. Change sets could only handle installing the latest compatible version, because `ModChange` holds a reference to `GUIMod`, which represents **all** versions of a module rather than a particular version.

Later, #2364 added a double click option to compatible versions in the mod info version list to install older versions. To get around the version limitations of change sets, it prompted the user for confirmation and then jumped to the progress screen of the installation flow, skipping the change set confirmation and other early steps. In keeping with the previous total impossibility of installing incompatible modules, double clicking incompatible modules had no effect.

## Motivation

Some users have expressed a preference for queuing module version changes along with normal installation/removal, so they can all be applied together. This makes sense (but is challenging to do).

In addition, the limitation to compatible versions is not particularly popular. Users would prefer to be allowed to install incompatible versions.

## Changes

- The double click installation is removed. See below for what replaces it.
- `ModChange` now holds a reference to a `CkanModule` instead of a `GUIMod`, so you can queue up a specific version of a module in the change set rather than just the identifier abstractly or the latest compatible. Updating the calling code for this was trivial, and in most cases an improvement as a `GUIMod` object was being faked up unnecessarily.
- `GUIMod` is no longer implicitly convertible to `CkanModule`, because there are multiple possible choices and you should think about which one you're using.
- `GUIMod` now has a public `SelectedMod` property that tracks which version the user wants to install and helps to drive the change set logic. This must be stored somewhere outside the UI so we can display the right selection if you switch to another row and back.
- `GUIMod.GetRequestedChange` is replaced by `GUIMod.GetRequestedChanges`, which can return multiple values to handle situations when we want to remove and install versions of the same mod. Conveniently this also simplifies how `ComputeUserChangeSet` collates the changes from multiple rows; instead of checking `.HasValue` and then selecting `.Value`, we simply use `.SelectMany` to join all the lists, which are allowed to be empty.
- The Versions list now has checkboxes that represent the value of `GUIMod.SelectedMod`  (and the column is 25px wider to compensate, not shown in old screenshot below): 
  image
  These can be checked and unchecked to choose versions to install, and those choices will appear in the change set. Switching from an older version to the *latest* version triggers the existing functionality for Upgrading (and vice versa).
- If you try to choose an incompatible module, a very scary confirmation warning appears:
  image
  German translation:
  > {0} wird von Ihrer aktuellen Spielversion nicht unterstützt und funktioniert möglicherweise überhaupt nicht. Wenn Sie Probleme damit haben, sollten Sie die Betreuer NICHT um Hilfe bitten.
  > 
  > Möchten Sie es wirklich installieren?
  
  If you click Install, it will allow you to install whatever:
  image
  Hopefully nobody will complain to mod authors if they break their install with this.
  And hopefully users will still report inaccurate metadata rather than bypassing it and forgetting about it.
- The change set can now show removal of one version of a module and installation of another version, for example:
  image

Fixes #2714. Fixes #2715. Fixes #2822.

### Companion fixes

- Upgrading modules now prints more messages, in the style of installation (also affects CmdLine and ConsoleUI):
  image
- The installation progress bar now counts up to 60% while mods are being installed, because 70%-100% are for post-install bookkeeping. The jump back from 100% to 70% looked weird.
- The install-all checkbox (in the header of the Install column) now re-checks itself whenever your change set becomes empty, for convenience after you use it to uninstall everything.
- `InstallModuleDriver` now clears `ChangeSet` instead of `changeSet` at the end, see https://github.com/KSP-CKAN/CKAN/issues/2276#issuecomment-506990808.

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -210,7 +210,8 @@ namespace CKAN
             {
                 for (int i = 0; i < modsToInstall.Count; i++)
                 {
-                    int percent_complete = (i * 100) / modsToInstall.Count;
+                    // The post-install steps start at 70%, so count up to 60% for installation
+                    int percent_complete = (i * 60) / modsToInstall.Count;
 
                     User.RaiseProgress(String.Format("Installing mod \"{0}\"", modsToInstall[i]),
                                          percent_complete);
@@ -984,20 +985,29 @@ namespace CKAN
 
             using (var tx = CkanTransaction.CreateTransactionScope())
             {
-
+                int totSteps = (remove?.Count() ?? 0)
+                             + (add?.Count()    ?? 0);
+                int step = 0;
                 foreach (InstalledModule instMod in remove)
                 {
+                    // The post-install steps start at 80%, so count up to 70% for installation
+                    int percent_complete = (step++ * 70) / totSteps;
+                    User.RaiseProgress($"Removing \"{instMod}\"", percent_complete);
                     Uninstall(instMod.Module.identifier);
                 }
 
                 foreach (CkanModule module in add)
                 {
                     var previous = remove?.FirstOrDefault(im => im.Module.identifier == module.identifier);
+                    int percent_complete = (step++ * 70) / totSteps;
+                    User.RaiseProgress($"Installing \"{module}\"", percent_complete);
                     Install(module, previous?.AutoInstalled ?? false);
                 }
 
+                User.RaiseProgress("Updating registry", 80);
                 registry_manager.Save(enforceConsistency);
 
+                User.RaiseProgress("Committing filesystem changes", 90);
                 tx.Complete();
 
                 EnforceCacheSizeLimit();
@@ -1022,6 +1032,8 @@ namespace CKAN
         /// </summary>
         public void Upgrade(IEnumerable<CkanModule> modules, IDownloader netAsyncDownloader, bool enforceConsistency = true)
         {
+            User.RaiseMessage("About to upgrade...\r\n");
+
             // Start by making sure we've downloaded everything.
             DownloadModules(modules, netAsyncDownloader);
 
@@ -1073,6 +1085,7 @@ namespace CKAN
                 to_remove,
                 enforceConsistency
             );
+            User.RaiseProgress("Done!", 100);
         }
 
         /// <summary>
--- a/GUI/GUIMod.cs
+++ b/GUI/GUIMod.cs
@@ -2,14 +2,64 @@ using System;
 using System.Collections.Generic;
 using System.Windows.Forms;
 using System.Linq;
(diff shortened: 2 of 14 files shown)


Olympic1:
Needs a rebase

DasSkelett:
Yes, German is fine!

HebaruSan:
Rebase completed.

Blu3wolf:
Would this be the place to request a config option somewhere to disable the extra dialog window? Potentially somewhere hard to find, so as to not invalidate the whole point of adding a scary dialog window in the first place. 

I'm all for decreasing nuisance reports to modders, they are at least 90% responsible for the loss of pjf. I'm also all for decreasing user RSI though, and adding extra clicks for operations (especially repetitive ones) is not the best way to achieve that.

HebaruSan:
Can you share some of your planned change sets that include enough incompatible modules to induce repetitive stress injuries?

Blu3wolf:
Certainly. I'll do so around Tuesday I expect, as I'm away from the computer until then.

PR #2823, opened 2019-07-01
Override V to v with x_netkan_force_v

HebaruSan:
## Background

`x_netkan_force_v` prepends a `v` onto a module's version during inflation. This is helpful for when a mod author releases "v1.1" followed by "1.2", which would otherwise mess up CKAN's version sorting.

## Problem

If a mod author freely switches between lowercase 'v' with capital 'V', we end up with almost as bad of a mess:

https://github.com/KSP-CKAN/CKAN-meta/tree/master/xScience

image

## Changes

Now `x_netkan_force_v` removes a capital 'V' prefix if found. The above becomes `xScience-v5.17.ckan` as intended.


techman83:
Love your work! Merging with thanks!

PR #2824, opened 2019-07-02
Auto-epoch based on queue message attribute

HebaruSan:
This PR grew out of a sidebar discussion on #2789. Please discuss it here instead of there, to allow that issue to stay focused on its main goals.

## Problem

As things stand, CKAN-meta will inevitably accumulate inaccurate metadata every time a mod author releases an out-of-order version, which includes things like dropping a 'v' prefix or changing it to/from capital 'V', numbering like `["0.6", "0.65", "0.7"]`, accidentally dropping the `0.` prefix of a version, fat-fingering an extra '.' in a version number, etc, Users will get "stuck" on older versions because the newer release isn't considered newer by CKAN.

There are no safeguards against such things (and no alerts that it has happened), and all require manual clean-up in the form of updating the `x_netkan_epoch` property in .netkan files and sometimes manually updating versions and filenames of .ckan files. It also relies on someone to *notice* the problem in the first place, which often takes a while. Recent examples of such manual fixes:

- KSP-CKAN/NetKAN#7260
- KSP-CKAN/NetKAN#7275
- KSP-CKAN/NetKAN#7278
- KSP-CKAN/CKAN-meta#1477
- KSP-CKAN/NetKAN#7280
- KSP-CKAN/NetKAN#7284
- KSP-CKAN/NetKAN#7285
- KSP-CKAN/CKAN-meta#1481
- KSP-CKAN/NetKAN#7286
- KSP-CKAN/CKAN-meta#1482
- KSP-CKAN/NetKAN#7288
- KSP-CKAN/NetKAN#7289
- KSP-CKAN/CKAN-meta#1484
- KSP-CKAN/NetKAN#7290
- KSP-CKAN/CKAN-meta#1485
- KSP-CKAN/CKAN-meta#1486
- KSP-CKAN/NetKAN#7291
- KSP-CKAN/NetKAN#7292
- KSP-CKAN/CKAN-meta#1488
- KSP-CKAN/NetKAN#7293
- KSP-CKAN/CKAN-meta#1489
- KSP-CKAN/NetKAN#7294
- KSP-CKAN/CKAN-meta#1490
- KSP-CKAN/NetKAN#7295
- KSP-CKAN/CKAN-meta#1491
- KSP-CKAN/NetKAN#7296
- KSP-CKAN/CKAN-meta#1492
- KSP-CKAN/NetKAN#7297
- KSP-CKAN/CKAN-meta#1493
- KSP-CKAN/NetKAN#7298
- KSP-CKAN/CKAN-meta#1494
- KSP-CKAN/NetKAN#7299
- KSP-CKAN/CKAN-meta#1495
- KSP-CKAN/NetKAN#7300
- KSP-CKAN/NetKAN#7301
- KSP-CKAN/NetKAN#7302
- KSP-CKAN/CKAN-meta#1496
- KSP-CKAN/NetKAN#7306
- KSP-CKAN/CKAN-meta#1498
- https://github.com/KSP-CKAN/NetKAN/issues/7318#issuecomment-513243948
- KSP-CKAN/NetKAN#7520
- KSP-CKAN/CKAN-meta#1554

This property is also prone to errors, as a few times an author has re-homed a netkan via a metanetkan and dropped `x_netkan_epoch` entirely because they don't know what it does, and we had to tell them to add it back because it's important.

## Changes

Now the Netkan queue Inflator from #2798 supports an additional optional input attribute:

```
Attr HighestVersion: The highest version string of this module that is currently indexed
```

The `version` property of a newly inflated module will be compared to this value:
- If it is higher, then we are indexing a new release with a good version string
- If it is the same, then we are updating an existing release
- If it is lower, then **we have detected an out of order release**!

When we detect an out of order release, the version epoch is increased automatically until the new release's version is greater (new version) or equal (same version update). The "main" part of the version remains the same.

If/when the scheduler component chooses to set this attribute (by checking the existing files in CKAN-meta), then this method will index out-of-order releases with correct epochs automatically. This will fix the above-mentioned problems with bad metadata.

Tests are included.

## Caveats

We will need a way to suppress this behavior. It makes sense for the typical mod, but a few mods (KSPInterstellarExtended and Kopernicus-Backports) intentionally post "older" releases that are meant to be treated as such. I imagine this would take the form of a new .netkan property:

```json
    "x_netkan_allow_out_of_order": true,
```

If set, the scheduler would skip checking the CKAN-meta files to find the highest version and not set the HighestVersion attribute. The Inflator would then fall back to using the `x_netkan_epoch` property as it does today.

### DIFF
--- a/Core/Versioning/ModuleVersion.cs
+++ b/Core/Versioning/ModuleVersion.cs
@@ -59,6 +59,14 @@ namespace CKAN.Versioning
             _string = version;
         }
 
+        /// <returns>
+        /// New module version with same version as 'this' but with one greater epoch
+        /// </returns>
+        public ModuleVersion IncrementEpoch()
+        {
+            return new ModuleVersion($"{_epoch + 1}:{_version}");
+        }
+
         /// <summary>
         /// Converts the value of the current <see cref="ModuleVersion"/> object to its equivalent
         /// <see cref="String"/> representation.
--- a/Netkan/Processors/Inflator.cs
+++ b/Netkan/Processors/Inflator.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using Autofac;
 using log4net;
 using CKAN.Configuration;
+using CKAN.Versioning;
 using CKAN.NetKAN.Model;
 using CKAN.NetKAN.Services;
 using CKAN.NetKAN.Transformers;
@@ -30,7 +31,7 @@ namespace CKAN.NetKAN.Processors
             transformer   = new NetkanTransformer(http, fileService, moduleService, githubToken, prerelease);
         }
 
-        internal IEnumerable<Metadata> Inflate(string filename, Metadata netkan, int? releases)
+        internal IEnumerable<Metadata> Inflate(string filename, Metadata netkan, int? releases, ModuleVersion previousHighVer)
         {
             log.DebugFormat("Inflating {0}", filename);
             try
@@ -42,7 +43,7 @@ namespace CKAN.NetKAN.Processors
                 log.Info("Input successfully passed pre-validation");
 
                 IEnumerable<Metadata> ckans = transformer
-                    .Transform(netkan, new TransformOptions(releases))
+                    .Transform(netkan, new TransformOptions(releases, previousHighVer))
                     .ToList();
                 log.Info("Finished transformation");
 
--- a/Netkan/Processors/Inflator.cs
+++ b/Netkan/Processors/Inflator.cs
@@ -31,7 +31,7 @@ namespace CKAN.NetKAN.Processors
             transformer   = new NetkanTransformer(http, fileService, moduleService, githubToken, prerelease);
         }
 
-        internal IEnumerable<Metadata> Inflate(string filename, Metadata netkan, int? releases, ModuleVersion previousHighVer)
+        internal IEnumerable<Metadata> Inflate(string filename, Metadata netkan, TransformOptions opts)
         {
             log.DebugFormat("Inflating {0}", filename);
             try
@@ -43,7 +43,7 @@ namespace CKAN.NetKAN.Processors
                 log.Info("Input successfully passed pre-validation");
 
                 IEnumerable<Metadata> ckans = transformer
-                    .Transform(netkan, new TransformOptions(releases, previousHighVer))
+                    .Transform(netkan, opts)
                     .ToList();
                 log.Info("Finished transformation");
 
--- a/Netkan/Processors/QueueHandler.cs
+++ b/Netkan/Processors/QueueHandler.cs
@@ -10,6 +10,7 @@ using log4net;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using CKAN.Versioning;
+using CKAN.NetKAN.Transformers;
 using CKAN.NetKAN.Model;
 
 namespace CKAN.NetKAN.Processors
@@ -115,9 +116,10 @@ namespace CKAN.NetKAN.Processors
             IEnumerable<Metadata> ckans = null;
             bool   caught        = false;
             string caughtMessage = null;
+            var    opts          = new TransformOptions(releases, highVer);
             try
             {
-                ckans = inflator.Inflate($"{netkan.Identifier}.netkan", netkan, releases, highVer);
+                ckans = inflator.Inflate($"{netkan.Identifier}.netkan", netkan, opts);
             }
             catch (Exception e)
             {
@@ -130,20 +132,25 @@ namespace CKAN.NetKAN.Processors
             }
             if (caught)
             {
-                yield return inflationMessage(null, netkan, false, caughtMessage);
+                yield return inflationMessage(null, netkan, opts, false, caughtMessage);
             }
             if (ckans != null)
             {
                 foreach (Metadata ckan in ckans)
                 {
                     log.InfoFormat("Sending {0}-{1}", ckan.Identifier, ckan.Version);
-                    yield return inflationMessage(ckan, netkan, true);
+                    yield return inflationMessage(ckan, netkan, opts, true);
                 }
             }
         }
 
-        private SendMessageBatchRequestEntry inflationMessage(Metadata ckan, Metadata netkan, bool success, string err = null)
+        private SendMessageBatchRequestEntry inflationMessage(Metadata ckan, Metadata netkan, TransformOptions opts, bool success, string err = null)
         {
+            bool staged = netkan.Staged || opts.Staged;
+            string stagingReason = 
+                  !string.IsNullOrEmpty(netkan.StagingReason) ? netkan.StagingReason
+                : !string.IsNullOrEmpty(opts.StagingReason)   ? opts.StagingReason
+                : null;
             var attribs = new Dictionary<string, MessageAttributeValue>()
             {
                 {
@@ -159,7 +166,7 @@ namespace CKAN.NetKAN.Processors
                     new MessageAttributeValue()
                     {
                         DataType    = "String",
-                        StringValue = netkan.Staged.ToString()
+                        StringValue = staged.ToString()
                     }
                 },
                 {
@@ -201,14 +208,14 @@ namespace CKAN.NetKAN.Processors
                     }
                 );
             }
-            if (netkan.Staged && !string.IsNullOrEmpty(netkan.StagingReason))
+            if (staged && stagingReason != null)
             {
                 attribs.Add(
                     "StagingReason",
                     new MessageAttributeValue()
                     {
                         DataType    = "String",
-                        StringValue = netkan.StagingReason
+                        StringValue = stagingReason,
                     }
                 );
             }
--- a/Netkan/Processors/QueueHandler.cs
+++ b/Netkan/Processors/QueueHandler.cs
@@ -9,6 +9,7 @@ using Amazon.SQS.Model;
 using log4net;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using CKAN.Versioning;
 using CKAN.NetKAN.Model;
 
 namespace CKAN.NetKAN.Processors
@@ -103,13 +104,20 @@ namespace CKAN.NetKAN.Processors
                 releases = int.Parse(releasesAttr.StringValue);
             }
 
+            ModuleVersion highVer = null;
+            MessageAttributeValue highVerAttr;
+            if (msg.MessageAttributes.TryGetValue("HighestVersion", out highVerAttr))
+            {
+                highVer = new ModuleVersion(highVerAttr.StringValue);
+            }
+
             log.InfoFormat("Inflating {0}", netkan.Identifier);
             IEnumerable<Metadata> ckans = null;
             bool   caught        = false;
             string caughtMessage = null;
             try
             {
-                ckans = inflator.Inflate($"{netkan.Identifier}.netkan", netkan, releases);
+                ckans = inflator.Inflate($"{netkan.Identifier}.netkan", netkan, releases, highVer);
             }
             catch (Exception e)
             {
--- a/Netkan/Program.cs
+++ b/Netkan/Program.cs
@@ -87,7 +87,7 @@ namespace CKAN.NetKAN
                         Options.GitHubToken,
                         Options.PreRelease
                     );
-                    var ckans = inf.Inflate(Options.File, netkan, ParseReleases(Options.Releases), null);
+                    var ckans = inf.Inflate(Options.File, netkan, new TransformOptions(ParseReleases(Options.Releases), null));
                     foreach (Metadata ckan in ckans)
                     {
                         WriteCkan(ckan);
--- a/Netkan/Program.cs
+++ b/Netkan/Program.cs
@@ -87,7 +87,7 @@ namespace CKAN.NetKAN
                         Options.GitHubToken,
                         Options.PreRelease
                     );
-                    var ckans = inf.Inflate(Options.File, netkan, ParseReleases(Options.Releases));
+                    var ckans = inf.Inflate(Options.File, netkan, ParseReleases(Options.Releases), null);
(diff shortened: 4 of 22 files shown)


DasSkelett:
So basically we will "ignore" the epoch for most mods, only using it as the starting value for the epoch incrementation counter.
In this case I would prefer to drop the `x_netkan_epoch` fully from the netkan files (except those with `x_netkan_allow_out_of_order`), else this will lead to a lot of confusion for everyone who doesn't know the internals of the bot / indexing mechanism.

So spec-wise, this means making `x_netkan_epoch` dependent on `x_netkan_allow_out_of_order`, or combine the two to `"x_netkan_use_manual_epoch" : <epoch_integer>`.

HebaruSan:
Added a spec update to say that `x_netkan_epoch` specifies the minimum epoch and that it may be set or incremented automatically.

I don't want to add `x_netkan_allow_out_of_order` to the spec in this PR since its functionality cannot be implemented in this PR.

HebaruSan:
What am I talking about, of course `x_netkan_allow_out_of_order` can be implemented here. Added in the latest commit. What can't be implemented here is the optimization of the scheduler skipping the version scan step, but now that will be purely an optional optimization on top of `netkan.exe`'s behavior rather than necessary to make `x_netkan_allow_out_of_order` work at all.

techman83:
My only comment is, that we would be putting a lot version logic into the scheduler. Potentially doing a fair amount of heavy lifting, to figure out the latest version. 

I know versions have historically been problematic, but I'm conscious of getting into a race condition of versioning here.

I'd like to see us dig into the Metadata a bit and see big of a problem we're having and is it something we can solve by setting problematic mods to staging or having a different mechanism for flagging potential issues. One of the things we can potentially do with the new infrastructure is better signal when something is a problem and proactively raise an Issue.

The code looks tidy with good test coverage and doco. If I've missed something or I'm being too conservative about the solution, I'm very happy to be corrected :slightly_smiling_face:

HebaruSan:
Maybe a little personal background:  I've been considering taking a break for a while, so when I fix issues such as those listed in the OP, I'm struck by how much manual intervention the current data model requires. CKAN *needs* a human to be actively inspecting metadata and reacting to problem reports from users, mostly due to out of order versions, where the manual clean-up is to increment the epoch and clean up the versions that were mis-filed under the old epoch, something that is eminently automatizable. Maybe I'm too much of a perfectionist, but if we ever want to get to a point where CKAN quietly runs itself in the background, this is 80% case on which to focus.

> My only comment is, that we would be putting a lot version logic into the scheduler. Potentially doing a fair amount of heavy lifting, to figure out the latest version.

How would you feel about a `netkan.exe --find-latest-version AwesomeMod` command for this? We could avoid duplicating the version logic in Python by having the Scheduler run that command instead. It would use existing logic to load each .ckan file in a subfolder of CKAN-meta and compare the versions to find the maximum. Then all the Scheduler would need is copies of `netkan.exe` and CKAN-meta.

> is it something we can solve by setting problematic mods to staging

No, the problematic mods are not known ahead of time (or to put it another way, **all** mods are potentially problematic). Our data integrity depends on mod authors to input version numbers sensibly, and we use `x_netkan_epoch` to work around the resultant "human error" problems.

> having a different mechanism for flagging potential issues

I'll give that a think. The committer component that listens to the Inflator's output would also be able to detect out of order versions, but I wanted to put the logic earlier than that so we could reduce the amount of manual intervention needed rather than just making it a smoother work flow.

I'm happy to let this one simmer for a while. There's no point in merging it until the new infrastructure is almost ready to go live anyway, since these spec updates would be documenting things that don't in fact work yet.

HebaruSan:
A fresh example from just two days ago, courtesy of CraftImport and @linuxgurugamer:

image

(There's an extra period between the 0 and the 8.)

HebaruSan:
I've embarked upon a more in-depth survey, and we are not **anywhere close** to keeping up with the manual maintenance. Bad data is rife, but "out of sight out of mind." I'll continue tracking the fixes with links in the OP here.

DasSkelett:
Okay, I think I can compromise on the notice in the spec. Most mod authors probably don't care about diverting epochs, else they would also care whether their mods are well-ordered.
But I would find a way to easily see especially big auto incrementations nice to have. I can imagine another attribute in the outgoing SQS:
```
Attr AutoIncrements: The number of automated epoch bumps for the generated CKAN
```
Which could be added to the status json and be displayed on the status page.

This way we would have an easy way to spot potential problems and can update the epoch in the netkans from time to time if the difference grows too high.

techman83:
> Maybe a little personal background: I've been considering taking a break for a while, so when I fix issues such as those listed in the OP, I'm struck by how much manual intervention the current data model requires. CKAN _needs_ a human to be actively inspecting metadata and reacting to problem reports from users, mostly due to out of order versions, where the manual clean-up is to increment the epoch and clean up the versions that were mis-filed under the old epoch, something that is eminently automatizable. Maybe I'm too much of a perfectionist, but if we ever want to get to a point where CKAN quietly runs itself in the background, this is the 80% case on which to focus.

Contributor burn out is something that is always at the forefront of my mind. If you need a break take one :slightly_smiling_face: You've been on the coal face for a while and it's easy for me to miss the number of manual interventions required when it isn't my primary focus. You've also made some incredible contributions and improved many long standing issues.

One of the really big successes of the CKAN has been the ability to not force mod authors to do anything special, but it also means that we have to go to great lengths to ensure the metadata is correct. On reflection, there have been numerous metadata wranglers (including yourself) over the years that have done an amazing job to keep up. You're absolutely right about the mission to reduce this load as much as possible.

> How would you feel about a `netkan.exe --find-latest-version AwesomeMod` command for this? We could avoid duplicating the version logic in Python by having the Scheduler run that command instead. It would use existing logic to load each .ckan file in a subfolder of CKAN-meta and compare the versions to find the maximum. Then all the Scheduler would need is copies of `netkan.exe` and CKAN-meta.

I think at this stage, lets keep it in python. It'll help us move fast, iterate and keep the scheduler light. It also will keep the burden lower on those who work on CKAN more focused on that part of the code base and I'm hoping moving to Python with a much tighter scope of each of the Infrastructure functions will make it easier for new contributors to get involved with the infrastructure side.

> Okay, I think I can compromise on the notice in the spec. Most mod authors probably don't care about diverting epochs, else they would also care whether their mods are well-ordered.
> But I would find a way to easily see especially big auto incrementations nice to have. I can imagine another attribute in the outgoing SQS

Versions are hard. Everyone has an opinion on how they work and it can be really hard to be consistent when doing things by hand.

@HebaruSan has more than demonstrated that we have a problem to address. I'm pondering maybe we set a flag of some description (maybe staged? maybe it's own flag? epoched?), so we can raise PRs initially. Let me think on it, but for now I'm gunna give this the thumbs up because it fits our IfItsBetterMergedIt&trade; philosophy.

HebaruSan:
> One of the really big successes of the CKAN has been the ability to not force mod authors to do anything special

I find that truly impressive as well. Debian assigns each package a maintainer who has to track patches and compile it and so on. CKAN not only uses the upstream packages unaltered, but it indexes them automatically! It's amazing that this works as well as it does.

> lets keep it in python. It'll help us move fast, iterate and keep the scheduler light

Sounds good. My only concern is that the version comparison logic is fairly complex (hence the idea to re-use it directly), but whatever works is fine with me.

https://github.com/KSP-CKAN/CKAN/blob/14ab1522e410c2bb0860d60b8986f625ec42a78e/Core/Versioning/ModuleVersion.cs#L191-L365

> it fits our IfItsBetterMergedIt™ philosophy

For what it's worth, I come from an "adversarial code review" environment where the goal is to find as many problems as possible and address them to the satisfaction of the reviewer before anything gets merged, so I'm quite happy to field objections and questions. Please don't feel like you have to approve anything of mine before you're satisfied with it. :ok_hand: 

> Versions are hard. Everyone has an opinion on how they work and it can be really hard to be consistent when doing things by hand.

I think most authors would be surprised that adding or removing a 'v' prefix would be significant, for example. I think most of them assume that we would just naturally sort versions in the order they were released chronologically, rather than according to version numbers.

> I'm pondering maybe we set a flag of some description (maybe staged? maybe it's own flag? epoched?), so we can raise PRs initially.

Interesting idea, almost certainly do-able. Set a flag when the epoch incrementing code gets called, flow that through to the Staged attribute. As long as the downstream bot component handles that attribute, we'd get PRs for out of order releases. Nice!

techman83:
> I find that truly impressive as well. Debian assigns each package a maintainer who has to track patches and compile it and so on. CKAN not only uses the upstream packages unaltered, but it indexes them automatically! It's amazing that this works as well as it does.

I did a short talk on indexing and archiving at LCA2018 -> https://www.youtube.com/watch?v=Y2e8wjcQo8g

> Sounds good. My only concern is that the version comparison logic is fairly complex (hence the idea to re-use it directly), but whatever works is fine with me.

No probs, if it gets too hard we can come back to it. But I think with some tests it should be ok, any edge cases that crop up we can write more tests. It's likely the first pass won't include the functionality, but once we're running and stable, adding it won't take much work.

> Interesting idea, almost certainly do-able. Set a flag when the epoch incrementing code gets called, flow that through to the Staged attribute. As long as the downstream bot component handles that attribute, we'd get PRs for out of order releases. Nice!

Yeah, I think at least initially we might want to be involved with epochs and we'll already have the functionality to do it.

HebaruSan:
> I did a short talk on indexing and archiving at LCA2018 -> https://www.youtube.com/watch?v=Y2e8wjcQo8g

I've seen it - who did you think that 'like' was from! :grin:

HebaruSan:
> No probs, if it gets too hard we can come back to it. But I think with some tests it should be ok, any edge cases that crop up we can write more tests. It's likely the first pass won't include the functionality, but once we're running and stable, adding it won't take much work.

Turns out that `mono ckan.exe compare --machine-readable versionA versionB` already exposes the version comparison logic (thanks to #2773), so even if we decide it's too hard to re-do the logic in Python, there's no need for a new command.

techman83:
Yeah, only problem is that calling subprocesses is expensive, which we'll have to do many times per mod. I have looked into binding directly to C# with python, but it looked very dependent on the code and what it's dependencies were. The version comparison is probably pretty tight in scope without dependencies, so building a ckan_version.dll is probably doable. But I think we can defer that for now. I've got the bulk of the harder logic in the new indexer done, mostly tying it all together left now.

HebaruSan:
Rebased to resolve conflicts now that the Scheduler might be ready soon.

HebaruSan:
> do we want to set the staging flag + reason when an out of order release is detected?

At least in the early stages, yes, I agree that would be a good idea. I'll make a note to add that.

HebaruSan:
OK, an auto-epoch event should now turn on staging downstream. It was a little awkward because there wasn't an easy way to send data outside of the inflated metadata itself. I ended up adding the staging data to `TransformOptions`, which was the closes thing we had to an out-of-band channel.

HebaruSan:
Before we merge this, we should make sure that `x_netkan_epoch` is added to any mods that regularly do backports, otherwise they might get unnecessarily epoched (yes, the staging will help, but it would be better to get valid metadata and not have to clean it up manually). Short list off the top of my head:

- KSPInterstellarExtended
- Kopernicus-Backports

@DasSkelett, can you think of any others?

DasSkelett:
> Before we merge this, we should make sure that `x_netkan_allow_out_of_order` is added to any mods that regularly do backports, otherwise they might get unnecessarily epoched (yes, the staging will help, but it would be better to get valid metadata and not have to clean it up manually). Short list off the top of my head:
> 
>     * KSPInterstellarExtended
> 
>     * Kopernicus-Backports
> 
> 
> @DasSkelett, can you think of any others? Maybe Kerbalism?

Kerbalism seems to only have multi-version releases lately, without need for "backports". It _should_ be fine with auto-epoching.

[Historian-Expanded](https://github.com/Aelfhe1m/Historian-Expanded/releases) had two releases for different KSP versions closely after another a few days ago, but they were in the right order, and this hasn't happened before, so not sure.
Otherwise the only two coming to my mind are the two you already mentioned.

HebaruSan:
Auto PRs for future reference:

- KSP-CKAN/CKAN-meta#1557
- KSP-CKAN/CKAN-meta#1558
- KSP-CKAN/CKAN-meta#1559
- KSP-CKAN/CKAN-meta#1560
- KSP-CKAN/CKAN-meta#1561
- KSP-CKAN/CKAN-meta#1562
- KSP-CKAN/CKAN-meta#1563
- KSP-CKAN/CKAN-meta#1564
- KSP-CKAN/CKAN-meta#1565
- KSP-CKAN/CKAN-meta#1566
- KSP-CKAN/CKAN-meta#1567
- KSP-CKAN/CKAN-meta#1568
- KSP-CKAN/CKAN-meta#1569
- KSP-CKAN/CKAN-meta#1570
- KSP-CKAN/CKAN-meta#1571
- KSP-CKAN/CKAN-meta#1572
- KSP-CKAN/CKAN-meta#1575
- KSP-CKAN/CKAN-meta#1576
- KSP-CKAN/CKAN-meta#1577
- KSP-CKAN/CKAN-meta#1578
- KSP-CKAN/CKAN-meta#1579
- KSP-CKAN/CKAN-meta#1580
- KSP-CKAN/CKAN-meta#1581
- KSP-CKAN/CKAN-meta#1582
- KSP-CKAN/CKAN-meta#1583
- KSP-CKAN/CKAN-meta#1584
- KSP-CKAN/CKAN-meta#1585
- KSP-CKAN/CKAN-meta#1586
- KSP-CKAN/CKAN-meta#1587
- KSP-CKAN/CKAN-meta#1588
- KSP-CKAN/CKAN-meta#1589

Manual clean-up/prep:

- KSP-CKAN/CKAN-meta#1574
- KSP-CKAN/NetKAN#7528
- KSP-CKAN/NetKAN#7530
- KSP-CKAN/NetKAN#7531

PR #2189, opened 2017-11-20
Adapt Curse API to new widget

HebaruSan:
Closes #2178.

This pull request updates the Curse API code to work with the latest changes to the widget.

- The base URL is changed
- No longer need json suffix in URL
- `files` is now an array rather than a dictionary
- `authors` is now moved into `members` array with `username` as property name
- The version is now embedded in `name`

After these changes, the tests all pass and `netkan.exe` generates files that match previous versions except for specifics of URLs.

PR #2825, opened 2019-07-03
Fix Curse API download URLs

HebaruSan:
## Problem

Curse-hosted modules aren't inflating:

image

## Cause

This URL is a 404 as of some recent change to Curse or its API or something:

https://github.com/KSP-CKAN/CKAN/blob/5323e54cee3279e85ad199cc172f6488b8ae8cfd/Netkan/Sources/Curse/CurseFile.cs#L39

In #2189, this was briefly changed to:

https://github.com/KSP-CKAN/CKAN/blob/b643d129b242081fbfe3742bea0d802ffc61274f/Netkan/Sources/Curse/CurseFile.cs#L32

... which still works! But this was reverted during code review.

## Changes

Now we use that `url + "/file"` formula again.

Fixes KSP-CKAN/NetKAN#7268. (I didn't think it was possible!)

### DIFF
--- a/Netkan/Sources/Curse/CurseFile.cs
+++ b/Netkan/Sources/Curse/CurseFile.cs
@@ -36,7 +36,7 @@ namespace CKAN.NetKAN.Sources.Curse
         {
             if (string.IsNullOrWhiteSpace(_downloadUrl))
             {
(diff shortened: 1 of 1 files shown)


techman83:
Awesome! LGTM :slightly_smiling_face:

PR #2835, opened 2019-07-22
Send staging reason to SQS queue

HebaruSan:
## Motivation

In https://github.com/KSP-CKAN/CKAN/issues/2789#issuecomment-513427895, @techman83 pointed out that the SQS queue code in #2798 does not handle the `x_netkan_staging_reason` property from KSP-CKAN/NetKAN-bot#88.

## Changes

Now the `x_netkan_staging_reason` property flows through into a `Metadata.StagingReason` property, which will be passed to a new message attribute if set:

```
Attr StagingReason: The reason that staging is enabled for this module (if it is enabled), to help pull request reviewers determine whether to merge
```

Notes:

- If the module is staged but `x_netkan_staging_reason` is not set, the attribute will not be set
- If the module is **not** staged but `x_netkan_staging_reason` **is** set, the attribute will **not** be set

So the downstream code should be prepared to handle a staged module without this attribute.

`x_netkan_staging` and `x_netkan_staging_reason` are added to the spec (according to their current behavior).

### DIFF
--- a/Netkan/Model/Metadata.cs
+++ b/Netkan/Model/Metadata.cs
@@ -6,13 +6,14 @@ namespace CKAN.NetKAN.Model
 {
     internal sealed class Metadata
     {
-        private const string KrefPropertyName        = "$kref";
-        private const string VrefPropertyName        = "$vref";
-        private const string SpecVersionPropertyName = "spec_version";
-        private const string VersionPropertyName     = "version";
-        private const string DownloadPropertyName    = "download";
-        public  const string UpdatedPropertyName     = "x_netkan_asset_updated";
-        private const string StagedPropertyName      = "x_netkan_staging";
+        private const string KrefPropertyName          = "$kref";
+        private const string VrefPropertyName          = "$vref";
+        private const string SpecVersionPropertyName   = "spec_version";
+        private const string VersionPropertyName       = "version";
+        private const string DownloadPropertyName      = "download";
+        public  const string UpdatedPropertyName       = "x_netkan_asset_updated";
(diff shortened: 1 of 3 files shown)


HebaruSan:
Give me a moment to add the staging properties to the spec...

techman83:
My only issue here is, does it belong in the resulting metadata? But that's from a purist standpoint and I'm ok if this is the path of least resistance to get it done.

Also from what I undestand x_blah things are ignored by the client, so does it need to be in the spec?

HebaruSan:
> My only issue here is, does it belong in the resulting metadata?

I thought the goal was to get it into the body of the pull request. Strings like "Check that the version matches the forum thread" would not be useful to have in CKAN-meta.

HebaruSan:
> Also from what I undestand x_blah things are ignored by the client, so does it need to be in the spec?

Spec.md has a section for netkan syntax, including several `x_netkan_*` properties.

https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#netkan-fields

I think it makes sense to document the staging properties so authors and metadata maintainers can make use of them as needed.

techman83:
> > My only issue here is, does it belong in the resulting metadata?
> 
> I thought the goal was to get it into the body of the pull request. Strings like "Check that the version matches the forum thread" would not be useful to have in CKAN-meta.

Ohhhhh, I ~might~ definitely have misunderstood! Been a very long week :sleeping: 

I'm not as familiar with the inner workings of NetKAN as I would like :slightly_smiling_face:

PR #2838, opened 2019-08-06
Add Dockerfile + Deployment for NetKAN Inflator

techman83:
This is in support of #2789, building and pushing the NetKAN Inflator container to Docker Hub when changes are pushed to master, keeping our container up to date.

A future PR will contain the service redeployment as well, but that requires a few more things to be finished off first and can be easily added.

### DIFF
--- a/bin/build_netkan_container.sh
+++ b/bin/build_netkan_container.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+cp -v Dockerfile.netkan _build/repack/$BUILD_CONFIGURATION/.
+(cd _build/repack/$BUILD_CONFIGURATION/ && docker build . -f Dockerfile.netkan -t kspckan/inflator)
+docker tag "kspckan/inflator" "kspckan/inflator:latest"
+docker push "kspckan/inflator:latest"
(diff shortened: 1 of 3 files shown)


bin/build_netkan_container.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+cp -v Dockerfile.netkan _build/repack/$BUILD_CONFIGURATION/.
```
HebaruSan: What do you think about doing these operations in a `_build/docker/inflator` folder instead of overloading `repack`? That way the docker stuff would be more isolated and not make implicit assumptions about how the rest of the build process works (i.e., that the `repack` folder will contain `netkan.exe`). And it seems more in the spirit of how the `_build` folder is currently organized, with each step in its own path.

HebaruSan:
Submitted #2839 with suggestions for this.

bin/build_netkan_container.sh
techman83: I think that's a wonderful idea! I'm not super across cake, but I know docker + bash!

PR #2853, opened 2019-08-25
Use Debian machine-readable copyright format

### LINKED ISSUE #2851: Make `copyright` file in .deb package machine-readable
Problem
-------
On Kubuntu 19.04, there's a GUI for opening .deb packages that claims that the license is "proprietary".

screenshot of .deb app

It would be better if apps that read .deb metadata would display the correct license for the product.

Suggestions
-----------
What CKAN can do is update the `copyright` file to use the [recommended format](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/). I'll try to get a pull request in this weekend.

bfrobin446:
As discussed under #2851, we aren't aware of any standard tool that parses the `copyright` file from inside a .deb package. But the CKAN metadata specification references the list of well-known license names from the Debian machine-readable license spec (https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/), so it seems like good citizenship to follow the upstream spec ourselves.

Fixes #2851.

### DIFF
--- a/debian/copyright
+++ b/debian/copyright
@@ -4,7 +4,7 @@ Upstream-Contact: http://webchat.esper.net/?channels=ckan
 Source: https://github.com/KSP-CKAN/CKAN/
 
 Files: *
-Copyright: 2014-2017, the Comprehensive Kerbal Archive Network (CKAN) Authors:
+Copyright: 2014-2019, the Comprehensive Kerbal Archive Network (CKAN) Authors:
     https://github.com/KSP-CKAN/CKAN/graphs/contributors
 Comment: You can use the CKAN and its associated files under the MIT license,
     reproduced below. This includes the right to sublicense under compatible
--- a/debian/copyright
+++ b/debian/copyright
@@ -6,10 +6,10 @@ Source: https://github.com/KSP-CKAN/CKAN/
 Files: *
 Copyright: 2014-2017, the Comprehensive Kerbal Archive Network (CKAN) Authors:
     https://github.com/KSP-CKAN/CKAN/graphs/contributors
-License: Expat
-    You can use the CKAN and its associated files under the MIT license,
+Comment: You can use the CKAN and its associated files under the MIT license,
     reproduced below. This includes the right to sublicense under compatible
     licenses.
+License: Expat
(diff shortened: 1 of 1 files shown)


debian/copyright
```diff
@@ -1,14 +1,15 @@
...
-The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: CKAN
+Upstream-Contact: http://webchat.esper.net/?channels=ckan
+Source: https://github.com/KSP-CKAN/CKAN/
+
+Files: *
+Copyright: 2014-2017, the Comprehensive Kerbal Archive Network (CKAN) Authors:
```
HebaruSan: We should probably update the 2017 endpoint to 2019 now.

debian/copyright
```diff
@@ -1,14 +1,15 @@
...
+    https://github.com/KSP-CKAN/CKAN/graphs/contributors
+License: Expat
+    You can use the CKAN and its associated files under the MIT license,
```
HebaruSan: What do you think about moving the 3 lines "You can use... under compatible licenses" to a `Comment:` clause? I tend towards thinking that they're more properly excluded from the License itself.

bfrobin446: Done.

debian/copyright
bfrobin446: Ideally, instead of updating the file every time there's a change in a new year, the Makefile could read the date of the current git commit and plug that year into the package with `sed`. But that's enough of an ask that I might as well leave it set to the current year until somebody gets the urge to automate it.

PR #2854, opened 2019-08-25
Don't invalidate invisible rows

### LINKED ISSUE #2852: [Bug] Unhandled exception when selecting / adding astronomer's visual pack
Background
----------
* Operating System: Windows 10 Pro 1803
* CKAN Version: 1.26.4
* KSP Version: 1.7.3.2594

No

Problem
-------
CKAN gives an unhandled exception when clicking the checkbox to add astronomer's visual pack, or any of the texture files. This does not happen with other mods in the "compatible" list

**Steps to reproduce**
* Install `EnvironmentalVisualEnhancements-HR` (or any other module providing `EnvironmentalVisualEnhancements-Config` other than AVP)
* Go to... the compatible list in CKAN
* Filter by mod names containing "astrono"
* Click on... "astronomers visual pack"
* An exception is thrown while it's trying to alert you to the conflict

```
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
Parameter name: rowIndex
   at System.Windows.Forms.DataGridView.InvalidateRow(Int32 rowIndex)
   at CKAN.Main.ConflictsUpdated(Dictionary`2 prevConflicts)
   at CKAN.Main.set_Conflicts(Dictionary`2 value)
   at CKAN.Main.<UpdateChangeSetAndConflicts>d__63.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at CKAN.Main.<ModList_CellValueChanged>d__284.MoveNext()
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.<>c.<ThrowAsync>b__6_0(Object state)
```

HebaruSan:
## Problem

1. Install `EnvironmentalVisualEnhancements-HR` (or any other module providing `EnvironmentalVisualEnhancements-Config` other than AVP)
2. Filter by mod names containing "astrono"
3. Check the box for Astronomers Visual Pack
4. An exception is thrown

screenshot

## Cause

When you try to install a mod that conflicts with one that's already installed, we highlight both of their rows red. This is done by setting the row's background and then calling `ModList.InvalidateRow` to force the GUI to obey that change immediately.

screenshot2

In this case, one of the rows is not visible due to the filter; the thrown exception indicates that `InvalidateRow` is being called for a row that isn't in the grid, and .NET doesn't like that.

In #2766 we specifically updated this code to cover hidden conflicts. How was this missed? My guess currently is that Mono's `InvalidateRow` has some sanity checks that .NET lacks. The screenshots pertaining to discussion of that change are from Linux.

## Changes

Now we only attempt `InvalidateRow` if the row is visible. This will avoid the exception.
(We will still set the color unconditionally, which should ensure we don't lose #2766's fix.)

Fixes #2852.

### DIFF
--- a/GUI/Main.cs
+++ b/GUI/Main.cs
@@ -113,7 +113,10 @@ namespace CKAN
                         cell.ToolTipText = null;
                     }
                     row.DefaultCellStyle.BackColor = Color.Empty;
-                    ModList.InvalidateRow(row.Index);
+                    if (row.Visible)
+                    {
+                        ModList.InvalidateRow(row.Index);
+                    }
                 }
             }
             if (Conflicts != null)
@@ -130,7 +133,10 @@ namespace CKAN
                         cell.ToolTipText = conflict_text;
                     }
                     row.DefaultCellStyle.BackColor = Color.LightCoral;
-                    ModList.InvalidateRow(row.Index);
+                    if (row.Visible)
+                    {
+                        ModList.InvalidateRow(row.Index);
+                    }
                 }
             }
         }


GUI/Main.cs
```diff
@@ -113,7 +113,10 @@ private void ConflictsUpdated(Dictionary<GUIMod, string> prevConflicts)
                         cell.ToolTipText = null;
                     }
                     row.DefaultCellStyle.BackColor = Color.Empty;
-                    ModList.InvalidateRow(row.Index);
+                    if (row.Visible)
+                    {
+                        ModList.InvalidateRow(row.Index);
```
politas: Why are we invalidating rows that _were_ conflicts if the next step marks all currently conflicting mods? Won't this leave us with no longer conflicting rows being invalidated?  The comment you previously added above says:

```
// Mark old conflicts as non-conflicted
// (rows that are _still_ conflicted will be marked as such in the next loop)
```

HebaruSan: "Invalidate" in this case just means "this row needs to be redrawn".
Rows that previously were conflicted and no longer are *should* be invalidated, because previously they had a red background and they now have a white background and should be redrawn.

PR #2855, opened 2019-08-27
Speed up autoremove search

### LINKED ISSUE #2846: [Bug] Slow response on clicking install checkbox with many mods installed
Background
----------
* Operating System: Win10 Pro build 1903
* CKAN Version: 1.26.4
* KSP Version: 1.7.3.2594

nope

Problem
-------
I have extreme lag when selecting or deselecting a mod to update, install or uninstall. ckan will go unresponsive then recover, afterwards the mod will select in about 10-15 seconds later per mod I clicked on. on a side note, ckan is eating up to 25-35% of my cpu while during the whole process, and seems to be waiting for my hdd.

All other functions work as expected.

**Steps to reproduce**
* Install 300+ mods
* Click the install checkbox for another mod
* It takes a while to respond

screenshot

HebaruSan:
## Problem

If you install many mods (200+) and click the checkbox to queue up one more to install, GUI takes a while to respond. I observed about 4.5 seconds with 282 mods, and a user reports 10-15 seconds with 363 mods.

## Causes

A few `Console.WriteLine` statements revealed the following upon clicking:

```
2019-08-27T00:36:28.1441350+00:00 UpdateChangeSetAndConflicts
2019-08-27T00:36:28.1441350+00:00 ComputeUserChangeSet
2019-08-27T00:36:29.2111960+00:00 ComputeUserChangeSet after FindRemovableAutoInstalled
2019-08-27T00:36:29.2131961+00:00 UpdateChangeSetAndConflicts after ComputeUserChangeSet
2019-08-27T00:36:29.2131961+00:00 ComputeChangeSetFromModList
2019-08-27T00:36:29.2141962+00:00 ComputeChangeSetFromModList before FindReverseDependencies
2019-08-27T00:36:29.2141962+00:00 ComputeChangeSetFromModList before FindRemovableAutoInstalled
2019-08-27T00:36:30.2972581+00:00 ComputeChangeSetFromModList before RelationshipResolver
2019-08-27T00:36:30.3072587+00:00 ComputeChangeSetFromModList after RelationshipResolver
2019-08-27T00:36:30.3082587+00:00 ComputeChangeSetFromModList end
2019-08-27T00:36:30.3112589+00:00 UpdateChangeSetAndConflicts end

2019-08-27T00:36:30.3122590+00:00 UpdateChangeSetAndConflicts
2019-08-27T00:36:30.3122590+00:00 ComputeUserChangeSet
2019-08-27T00:36:31.4493240+00:00 ComputeUserChangeSet after FindRemovableAutoInstalled
2019-08-27T00:36:31.4503241+00:00 UpdateChangeSetAndConflicts after ComputeUserChangeSet
2019-08-27T00:36:31.4503241+00:00 ComputeChangeSetFromModList
2019-08-27T00:36:31.4503241+00:00 ComputeChangeSetFromModList before FindReverseDependencies
2019-08-27T00:36:31.4503241+00:00 ComputeChangeSetFromModList before FindRemovableAutoInstalled
2019-08-27T00:36:32.6093904+00:00 ComputeChangeSetFromModList before RelationshipResolver
2019-08-27T00:36:32.6143907+00:00 ComputeChangeSetFromModList after RelationshipResolver
2019-08-27T00:36:32.6153907+00:00 ComputeChangeSetFromModList end
2019-08-27T00:36:32.6153907+00:00 UpdateChangeSetAndConflicts end
```

Apparent problems:

- `UpdateChangeSetAndConflicts` is called twice, only once is needed
- `FindRemovableAutoInstalled` takes 1-1.5 seconds each time it is called (4x); this is most of the slowness

Looking at `FindRemovableAutoInstalled` reveals several opportunities for improvements:

https://github.com/KSP-CKAN/CKAN/blob/a162c8b1dfa98fcc9a5d459a2d075737f26a0286/Core/Registry/Registry.cs#L1131-L1142

- `FindReverseDependencies` is called on all installed modules, but it only needs to be called on *auto-installed* modules
- There's a linear array search for matching identifiers in the check that all rev deps are auto-installed (the `.All` call)
- `FindReverseDependencies` traverses the **whole** reverse dependency tree, but we only care about enough entries to find one that isn't auto-installed

## Changes

- Now `UpdateChangeSetAndConflicts` is only called once per click
- Now `FindReverseDependencies` uses `IEnumerable<>` and `yield return` to generate its results lazily, so it will not traverse the entire reverse dependency tree if the calling code doesn't need it, avoiding costly `FindUnsatisfiedDepends` calls
- Now `FindRemovableAutoInstalled` only checks `FindReverseDependencies` for auto-installed modules, which should be a small fraction of the total
- Now `FindRemovableAutoInstalled`'s linear array search is replaced with an `IsSupersetOf` call to take advantage of `FindReverseDependencies`'s new lazy evaluation (it will short circuit as soon as we find **one** rev dep that isn't auto-installed)

In my test case, these changes brought the time to process a click from 4.5 seconds to 0.25 seconds, or about a 95% reduction. Mileage will vary with the details of installed mod lists.

@shdwlrd, if you'd like to try a test build with these changes:
-  [ckan.zip](https://github.com/KSP-CKAN/CKAN/files/3544617/ckan.zip)

Fixes #2846.

### DIFF
--- a/Core/Registry/IRegistryQuerier.cs
+++ b/Core/Registry/IRegistryQuerier.cs
@@ -75,7 +75,7 @@ namespace CKAN
         /// <summary>
         /// Finds and returns all modules that could not exist without the listed modules installed, including themselves.
         /// </summary>
-        HashSet<string> FindReverseDependencies(IEnumerable<string> modules);
+        IEnumerable<string> FindReverseDependencies(IEnumerable<string> modules);
 
         /// <summary>
         /// Find auto-installed modules that have no depending modules
--- a/Core/Registry/Registry.cs
+++ b/Core/Registry/Registry.cs
@@ -1067,52 +1067,62 @@ namespace CKAN
 
         /// <summary>
         /// Finds and returns all modules that could not exist without the listed modules installed, including themselves.
-        /// Acts recursively.
+        /// Acts recursively and lazily.
         /// </summary>
-        internal static HashSet<string> FindReverseDependencies(
+        internal static IEnumerable<string> FindReverseDependencies(
             IEnumerable<string> modules_to_remove,
             IEnumerable<CkanModule> orig_installed,
             IEnumerable<string> dlls,
             IDictionary<string, UnmanagedModuleVersion> dlc
         )
         {
-            if (!modules_to_remove.Any())
+            // The empty list has no reverse dependencies
+            // (Don't remove broken modules if we're only installing)
+            if (modules_to_remove.Any())
             {
-                // The empty list has no reverse dependencies
-                // (Don't remove broken modules if we're only installing)
-                return new HashSet<string>();
-            }
-            while (true)
-            {
-                // Make our hypothetical install, and remove the listed modules from it.
-                HashSet<CkanModule> hypothetical = new HashSet<CkanModule>(orig_installed); // Clone because we alter hypothetical.
-                hypothetical.RemoveWhere(mod => modules_to_remove.Contains(mod.identifier));
-
-                log.DebugFormat("Started with {0}, removing {1}, and keeping {2}; our dlls are {3}", string.Join(", ", orig_installed), string.Join(", ", modules_to_remove), string.Join(", ", hypothetical), string.Join(", ", dlls));
-
-                // Find what would break with this configuration.
-                var broken = SanityChecker.FindUnsatisfiedDepends(hypothetical, dlls.ToHashSet(), dlc)
-                    .Select(x => x.Key.identifier).ToHashSet();
-
-                // If nothing else would break, it's just the list of modules we're removing.
-                HashSet<string> to_remove = new HashSet<string>(modules_to_remove);
-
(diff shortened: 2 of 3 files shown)


DasSkelett:
Why use IEnumerables instead of explicit types?

HebaruSan:
That's the only way I know to use `yield return` to implement a lazily evaluated sequence. If I change it back to `HashSet`:

image

DasSkelett:
I see, some weird Microsoft philosophy.
Can't understand why they don't allow a signature with return types _implementing_ the interface, I mean that's essentially what _will_ be returned in the end, right?

HebaruSan:
> I mean that's essentially what will be returned in the end, right?

No, this function now never generates an object with a collection inside of it. It runs pieces of the code inside the function to return each element as it's requested. If the calling code doesn't try to access the full sequence, then it won't be generated.

HebaruSan:
@shdwlrd, thanks for calling this problem to our attention!

PR #2856, opened 2019-08-27
Find portable installs when enforcing cache limits

HebaruSan:
## Problem

A user mentioned this exception with a "portable" install:

```
System.InvalidOperationException: Sequence contains no elements
   at System.Linq.Enumerable.Aggregate[TSource](IEnumerable`1 source, Func`3 func)
   at CKAN.NetFileCache.EnforceSizeLimit(Int64 bytes, Registry registry)
   at CKAN.ModuleInstaller.EnforceCacheSizeLimit()
   at CKAN.ModuleInstaller.InstallList(ICollection`1 modules, RelationshipResolverOptions options, IDownloader downloader, Boolean ConfirmPrompt)
   at CKAN.Main.InstallMods(Object sender, DoWorkEventArgs e)
   at System.ComponentModel.BackgroundWorker.OnDoWork(DoWorkEventArgs e)
```

https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1264-orion/&do=findComment&comment=3660841

## Cause

`NetFileCache.EnforceSizeLimit` assumes that the current instance will be in `manager.Instances`.

https://github.com/KSP-CKAN/CKAN/blob/a162c8b1dfa98fcc9a5d459a2d075737f26a0286/Core/Net/NetFileCache.cs#L294-L298

"Portable" instances are an exception to this:

https://github.com/KSP-CKAN/CKAN/blob/a162c8b1dfa98fcc9a5d459a2d075737f26a0286/Core/KSPManager.cs#L66

If all you have is a portable install, `manager.Instances` will be empty, and `Aggregate` will throw that exception.

## Changes

Now we use `Aggregate`'s optional first parameter to set a default element based on the current instance's version criteria, which should cover portables, falling back to an empty range if not found.

### DIFF
--- a/Core/Net/NetFileCache.cs
+++ b/Core/Net/NetFileCache.cs
@@ -295,7 +295,11 @@ namespace CKAN
                 KspVersionCriteria aggregateCriteria = manager?.Instances.Values
                     .Where(ksp => ksp.Valid)
                     .Select(ksp => ksp.VersionCriteria())
-                    .Aggregate((a, b) => a.Union(b));
+                    .Aggregate(
+                        manager?.CurrentInstance?.VersionCriteria()
+                            ?? new KspVersionCriteria(null),
+                        (combinedCrit, nextCrit) => combinedCrit.Union(nextCrit)
+                    );
 
                 // This object lets us find the modules associated with a cached file
                 Dictionary<string, List<CkanModule>> hashMap = registry.GetDownloadHashIndex();


DasSkelett:
Sooo...
What are portable installs? :thinking:

HebaruSan:
A portable install is when you put `ckan.exe` in your KSP folder and run it from there. We check to see whether the current folder is a KSP folder, and if so, then we make an ad hoc instance that doesn't get saved to the registry (or the new JSON config file). I think the idea was that you could set this up on a USB drive and then be able to use CKAN on multiple computers conveniently.

https://github.com/KSP-CKAN/CKAN/blob/9da096b39e1604345c93c2cc551b6d7e3c649219/Core/KSP.cs#L175-L190

https://github.com/KSP-CKAN/CKAN/blob/9da096b39e1604345c93c2cc551b6d7e3c649219/Core/KSPManager.cs#L83-L94

Core/Net/NetFileCache.cs
```diff
@@ -295,7 +295,11 @@ public void EnforceSizeLimit(long bytes, Registry registry)
                 KspVersionCriteria aggregateCriteria = manager?.Instances.Values
                     .Where(ksp => ksp.Valid)
                     .Select(ksp => ksp.VersionCriteria())
-                    .Aggregate((a, b) => a.Union(b));
+                    .Aggregate(
+                        manager?.CurrentInstance?.VersionCriteria()
+                            ?? new KspVersionCriteria(null),
+                        (a, b) => a.Union(b)
```
DasSkelett: Can you change `a` and `b` to more descriptive names, like `combinedCriteria` and `nextCriteria` or so?
Helps someone like me who didn't know what `Aggregate` does until five minutes ago understanding the code.

HebaruSan: Done. (It's just a poorly named `fold`.)

PR #2869, opened 2019-09-14
Allow totally incompatible modules in changeset

### LINKED ISSUE #2868: Cannot force install a mod which is not supported [Bug]
Background
----------
* CKAN Version: 1.26.4
* KSP Version 1.7.3 (basically fresh install, just played a bit without mods yet)

Problem
-------
Selecting any version of a mod to install, where no version supports the current version, doesn't make it appear in the change sets. For example when I select the any version of `PlanetShine`, no change entry appears. And if it is the only change, the tab also won't appear.

Selecting a mod which has support for the current version (e.g. `[x] Science! Continued`), does properly work and shows me the queued installation with the selected version. As a note I get the warning that the mod doesn't support the current version in both cases (`PlanetShine` and `[x] Science! Continued`).

For reference, if I read it correctly this feature was requested with #2822 and implemented in #2821.

HebaruSan:
## Problem

In #2821 we added the ability to select a specific version of a module to install in GUI rather than only installing the latest compatible. As part of this, incompatible versions were allowed as well.

However, you could only install an incompatible version if at least one other version of the module was compatible. Modules with **no** compatible versions (let's call them "totally incompatible") could be selected but would not show up in the changeset and could not be installed.

## Cause

Totally incompatible modules were excluded in two ways, in line with the old assumptions about installability:

- `MainModList.ComputeUserChangeSet` only checked modules for which `IsInstallable` is true (it's false for totally incompatible modules)
- `GUIMod.GetRequestedChanges` only allowed installing a new module if the install checkbox was visible and checked (it's `-` for totally incompatible modules), and changing from one version to another only worked if the module was already installed

## Changes

- Now `MainModList.ComputeUserChangeSet` checks all modules
- Now `GUIMod.GetRequestedChanges` can return an installation change even if the installation checkbox isn't available, as long as `InstalledMod` and `SelectedMod` aren't the same. As part of this, the old install/uninstall logic and the newer version-switching logic are unified to a single block.

@xZise, if you'd like to try this change, here's a test build:
- [ckan.zip](https://github.com/KSP-CKAN/CKAN/files/3613031/ckan.zip)

Fixes #2868.

### DIFF
--- a/GUI/GUIMod.cs
+++ b/GUI/GUIMod.cs
@@ -288,16 +288,11 @@ namespace CKAN
 
         public IEnumerable<KeyValuePair<CkanModule, GUIModChangeType>> GetRequestedChanges()
         {
-            if (IsInstalled && !IsInstallChecked)
-            {
-                // Uninstall the version we have installed
-                yield return new KeyValuePair<CkanModule, GUIModChangeType>(InstalledMod.Module, GUIModChangeType.Remove);
-            }
-            else if (!IsInstalled && IsInstallChecked)
-            {
-                yield return new KeyValuePair<CkanModule, GUIModChangeType>(SelectedMod ?? Mod, GUIModChangeType.Install);
-            }
-            else if (IsInstalled && (IsInstallChecked && HasUpdate && IsUpgradeChecked))
+            bool selectedIsInstalled = SelectedMod?.Equals(InstalledMod?.Module)
+                ?? InstalledMod?.Module.Equals(SelectedMod)
+                // Both null
+                ?? true;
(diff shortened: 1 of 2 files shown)


DasSkelett:
Hmmm, not working for me. Always throws the following error:
```
AlmostRealSolarSystem 1.2.3 depends on SigmaDimensions, which isn't compatible with your current version of KSP
```

**even if** I select another mod to install, f.e. `[x] Science!`.

So the selected mod isn't updated in the background (in the UI it is).

DasSkelett:
Scratch that, problem was that ARSS was still selected in the versions tab...

PR #2875, opened 2019-09-26
Batch inflator responses

### LINKED ISSUE #2874: [Feature] Send inflated modules to SQS queue in batches
## Background

The new bot architecture from #2789 is live and working well!

One of its goals is to increase efficiency and decrease costs. The cost of running the bot is at least partly based on how many SQS requests are made, so we should try to minimize the number of SQS requests we perform.

## Problem

Currently the Inflator from #2798 uses more SQS requests than strictly necessary. Per pass, it:

- Requests 10 netkans in one batch from its upstream queue
- Inflates them
- Sends 1 SQS request to its downstream queue per module
- Deletes each input message with 1 SQS request per message

## Suggestion

The queue handler should be updated to group together responses from the same input batch and send them on to the output queue together as a batch.

https://github.com/KSP-CKAN/CKAN/blob/2f0c7a37bbda13e51bebb3a4602a8f9094807090/Netkan/Processors/QueueHandler.cs#L172-L182

Deletions should also be batched.

https://github.com/KSP-CKAN/CKAN/blob/2f0c7a37bbda13e51bebb3a4602a8f9094807090/Netkan/Processors/QueueHandler.cs#L91-L97

Example of batch sending code:

https://github.com/KSP-CKAN/NetKAN-Infra/blob/805d9fc15a8016297c823980faf42126e12fe0ee/netkan/netkan/scheduler.py#L61-L81

HebaruSan:
## Motivation

See #2874; currently the Inflator sends output messages and deletes input messages one at a time as it processes each input batch of 10 messages. It would be cheaper in terms of SQS queue costs to send and delete in batches the same way we receive the input in batches.

## Changes

Now all the response objects for a given batch of input message are collected together and sent in as few batches as possible, and the input messages are deleted in a batch. This should cut down the number of SQS requests used by the Inflator (hopefully by a lot).

This will delay each batch's updates until everything in the batch is finished processing, but we think this is a trade-off worth making.

Fixes #2874.

### DIFF
--- a/Netkan/Processors/QueueHandler.cs
+++ b/Netkan/Processors/QueueHandler.cs
@@ -55,50 +55,72 @@ namespace CKAN.NetKAN.Processors
             {
                 log.Debug("No metadata in queue");
             }
-            else foreach (var msg in resp.Messages)
+            else
             {
-                log.DebugFormat("Metadata returned: {0}", msg.Body);
-                var netkan = new Metadata(JObject.Parse(msg.Body));
-
-                int releases = 1;
-                MessageAttributeValue releasesAttr;
-                if (msg.MessageAttributes.TryGetValue("Releases", out releasesAttr))
+                // Might be >10 if Releases>1
+                var responses = resp.Messages.SelectMany(Inflate).ToList();
+                for (int i = 0; i < responses.Count; i += howMany)
                 {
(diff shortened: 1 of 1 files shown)


techman83:
LGTM, at least to my vaguely C# parsing eyse :slightly_smiling_face:

PR #2886, opened 2019-10-09
Use current metadata for installed module compatibility

### LINKED ISSUE #2883: [Bug] Recently-compatible installed mods missing from compatible filter
Background
----------
* Operating System: windows 10
* CKAN Version: v1.26.4
* KSP Version: v1.7.3.2594

No

Problem
-------
**Describe the bug**

Hello, I'm the owner of 2 mods:

- **Mk1-Cabin-Hatch**
https://forum.kerbalspaceprogram.com/index.php?/topic/151443-1xx-mk1-cabin-hatch-v020-3-november-2016/page/3/

- **Ablative-Airbrake**
https://forum.kerbalspaceprogram.com/index.php?/topic/157353-1xx-ablative-airbrake-v030-7-october-2017/

I installed them a long time ago, and now I can't see them with compatibility filter unless i set it to allow KSP version 1.3, but my MOD.version file on my github page describe it as compatible with KSP 1.7.3 and CKAN see it well (in the Max KSP version collumn)
However the MOD.version file in the GameData/MOD is set to old KSP version, but i update the MOD.version file on my github pages for KSP-AVC.

Uninstalling and reinstalling fixes the problem.

**Steps to reproduce**

1. Open CKAN with KSP 1.3.1 installed when AblativeAirbrake and Mk1CabinHatch were only compatible with KSP 1.3.1 (approx. March 2018 and earlier)
2. Install AblativeAirbrake and Mk1CabinHatch
3. Upgrade KSP to 1.7.3
4. Wait for AblativeAirbrake and Mk1CabinHatch to add support up to KSP 1.7 (approx. May 2019 and later)
5. Settings > Compatible  KSP version > tick only 1.7
6. set filter to "compatible", 
7. search for my mods "Mk1-Cabin-Hatch" or "Ablative-Airbrake", they are not visible
8. Settings > Compatible  KSP version > tick 1.3
9. my mods are now visible

CKAN folder generated by the above steps for testing:
- [CKAN.zip](https://github.com/KSP-CKAN/CKAN/files/3698848/CKAN.zip)

HebaruSan:
## Problem

If a module's compatibility metadata changes after you install it, the GUI filters will always reflect the compatibility at time of install. For example, AblativeAirbrakes 0.3.0 only supported KSP 1.3.1 in March 2018, but now supports 1.2.2–1.7.3. If you installed it back then and kept it installed to the present, it would only be counted as compatible in GUI if you have 1.3.1 marked as compatible, but if you uninstall and/or reinstall, it will be considered compatible with all of 1.2.2–1.7.3.

## Cause

`GUIMod` has multiple layered constructors; at the bottom level you can create a `GUIMod` if you only know the identifier, above that you can create a `GUIMod` from a `CkanModule`, and at the top you can create a `GUIMod` using an `InstalledModule`. Each layer calls the level below it and then fills in extra info; the `InstalledModule` layer passes its source module (metadata from time of install), and the `CkanModule` layer passes its identifier.

The handling of compatibility in this hierarchy had some flaws. By default, everything was assumed to be compatible, which could be overridden by a constructor parameter or based on the `CkanModule` parameter's compatibility info. In the case of an `InstalledModule`, this meant that an incompatible source module would mark the overall module as incompatible, even if its current available module had subsequently been updated in CKAN-meta to be compatible.

## Changes

Now the `incompatible` constructor parameters are nullable and `null` by default. If set to true or false, we use that value. If not passed in (i.e., still `null` by the time we get to the identifier level), then we use the identifier to look up the compatibility from the registry based on the current metadata for all versions of the module.

To avoid paying the cost of this computation for compatible modules, `MainModList` now passes `incompatible`=`false` for known-compatible modules. This saves a few operations per compatible module. This function was already passing `true` for incompatible modules, and it will leave the parameter `null` for installed modules to force a check of the current metadata.

The `CkanModule`-level `GUIMod` constructor no longer checks compatibility because it was not strictly correct and is now redundant. The identifier-level constructor that it calls will set the correct value before the `CkanModule`-level constructor runs.

Several functions are changed to return `IEnumerable` instead of `List` for possible performance improvements when a search stops early (e.g., if the new `.All` call finds a false value early on, the rest of the list does not need to be generated).

Fixes #2883.

### DIFF
--- a/Cmdline/Action/Available.cs
+++ b/Cmdline/Action/Available.cs
@@ -15,7 +15,7 @@ namespace CKAN.CmdLine
         {
             AvailableOptions opts      = (AvailableOptions)raw_options;
             IRegistryQuerier registry  = RegistryManager.Instance(ksp).registry;
-            List<CkanModule> available = registry.Available(ksp.VersionCriteria());
+            var              available = registry.Available(ksp.VersionCriteria());
 
             user.RaiseMessage("Mods available for KSP {0}", ksp.Version());
             user.RaiseMessage("");
--- a/Cmdline/Action/Search.cs
+++ b/Cmdline/Action/Search.cs
@@ -186,7 +186,7 @@ namespace CKAN.CmdLine
         {
             IRegistryQuerier registry = RegistryManager.Instance(ksp).registry;
             // Get the list of all compatible and incompatible mods
-            List<CkanModule> mods = registry.Available(ksp.VersionCriteria());
+            List<CkanModule> mods = registry.Available(ksp.VersionCriteria()).ToList();
             mods.AddRange(registry.Incompatible(ksp.VersionCriteria()));
             for (int i = 0; i < modules.Count; ++i)
             {
--- a/Cmdline/Action/Update.cs
+++ b/Cmdline/Action/Update.cs
@@ -39,7 +39,7 @@ namespace CKAN.CmdLine
             {
                 // Get a list of available modules prior to the update.
                 var registry = RegistryManager.Instance(ksp).registry;
-                available_prior = registry.Available(ksp.VersionCriteria());
+                available_prior = registry.Available(ksp.VersionCriteria()).ToList();
             }
 
             // If no repository is selected, select all.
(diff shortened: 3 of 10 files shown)


DasSkelett:
Let me try if I can get the merging right...

PR #2887, opened 2019-10-09
Normalize install paths

### LINKED ISSUE #2880: [Bug] Superfluous metadata changed message with non-normalized install_to
already third time I got

possibly an every bot iteration

HebaruSan:
## Problem

If you install a module with an `install_to` that ends with a slash (`/`), you will receive the changed-metadata and reinstallation prompt every time you perform a non-redundant refresh (i.e., something has actually changed in CKAN-meta).

screenshot

All such extant paths were normalized in KSP-CKAN/NetKAN#7419 and KSP-CKAN/CKAN-meta#1510, but now we will address the root causes.

## Cause

1. The client downloads the non-normalized metadata from CKAN-meta:

| Available module | Installed module |
| --- | --- |
| `GameData/002_CommunityPartsTitles/Extras/` | (none) |

2. The user installs the module, and `ModuleInstaller` normalizes the `install_to` at installation, which affects the instance of the `CkanModule` stored in the registry:

| Available module | Installed module |
| --- | --- |
| `GameData/002_CommunityPartsTitles/Extras` | (none) |

3. The available module is cloned to produce the installed module:

| Available module | Installed module |
| --- | --- |
| `GameData/002_CommunityPartsTitles/Extras` | `GameData/002_CommunityPartsTitles/Extras` |

4. Later, the user updates the registry and again receives the non-normalized metadata:

| Available module | Installed module |
| --- | --- |
| `GameData/002_CommunityPartsTitles/Extras/` | `GameData/002_CommunityPartsTitles/Extras` |

Since these values do not match, CKAN thinks a significant change has been made to this module and prompts for reinstallation.

## Changes

- Now `ModuleInstaller` doesn't change the original install stanza, it just puts the normalized path in a local variable and uses it from there. This will prevent unintentional changes to the registry.
- Now the logic to compare install stanzas normalizes them before comparing. This means that if one stanza has a normalized path and the other has a non-normalized path, they can still be equal if both have the same normalized form.
- Now we normalize `file`, `find`, and `install_to` after deserialization. (But since these values can be set in ways other than deserialization, the existing normalization call is not removed.) This will cause values in the registry to be normalized more consistently.
- Now Netkan will print an error if `file`, `find`, or `install_to` are not normalized in its input. This will ensure that we can catch such metadata in the future at the pull request stage.

Side fixes:

- Now the logic to compare install stanzas is moved from `Repo` to `ModuleInstallDescriptor.Equals`, in the install stanza class. This is where such code usually would go.
- Now the logic to compare install stanzas checks whether `find` or `find_regexp` have changed. These were just missing before and may have allowed some metadata changes to occur without prompting for reinstallation (although I do not recall seeing such a change go through).

Fixes #2880.

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -409,34 +409,32 @@ namespace CKAN
             var files = new List<InstallableFile>();
 
             // Normalize the path before doing everything else
-            // TODO: This really should happen in the ModuleInstallDescriptor itself.
-            stanza.install_to = KSPPathUtils.NormalizePath(stanza.install_to);
+            string install_to = KSPPathUtils.NormalizePath(stanza.install_to);
 
             // Convert our stanza to a standard `file` type. This is a no-op if it's
             // already the basic type.
-
             stanza = stanza.ConvertFindToFile(zipfile);
 
-            if (stanza.install_to == "GameData" || stanza.install_to.StartsWith("GameData/"))
+            if (install_to == "GameData" || install_to.StartsWith("GameData/"))
             {
                 // The installation path can be either "GameData" or a sub-directory of "GameData"
                 // but it cannot contain updirs
-                if (stanza.install_to.Contains("/../") || stanza.install_to.EndsWith("/.."))
-                    throw new BadInstallLocationKraken("Invalid installation path: " + stanza.install_to);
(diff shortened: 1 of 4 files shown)


DasSkelett:
He, already prepared myself to suggest adding a validator to NetKAN, but of course you thought of it yourself ;)

Otherwise looks good, just going take it on a short test run.

PR #2889, opened 2019-10-10
Redeploy Inflator after builds

HebaruSan:
## Motivation

Currently the Inflator's image is rebuilt after commits to master, but it still has to be redeployed manually. It would be quicker and more convenient to redeploy automatically.

## Changes

Now @techman83's command to redeploy the Inflator is added after the image rebuild. It pulls the Docker image for the NetKAN-Infra project and uses it to run the `redeploy-service` command. Ideally we might do this without downloading the image, but there is no Cake package for ECS.

These environment variables should be set on Travis:

- `AWS_ACCESS_KEY_ID`
- `AWS_SECRET_ACCESS_KEY`
- `AWS_DEFAULT_REGION`


DasSkelett:
Syntax looks good, but that's about everything I can judge and test in this PR :smile:

techman83:
Yep, I'll build the rest of the bits on top of that to make it all work! Though the `AWS_SECRET_ACCESS_KEY` should be a secret parameter, the doco for cake.docker isn't incredibly clear how that works (at least to someone who hasn't touched C# much in 4 years).

PR #2270, opened 2018-01-30
Use arch=32 for OSX

HebaruSan:
CKAN's GUI apparently does not work in 64-bit on OSX (see #2269 and [the wiki](https://github.com/KSP-CKAN/CKAN/wiki/Installing-CKAN-on-OSX)).

This pull request adds `--arch=32` to the command line of mono in the `CKAN.app` bundle, and removes `--debug` because it's not recommended on the wiki.

(This doesn't fix #2269 because that user isn't using the app bundle.)

PR #2893, opened 2019-10-19
[macOS] Only run Mono in 32bit mode when the GUI is launched

DasSkelett:
## Motivation
The release of macOS Catalina marked the death of 32-bit applications on Apple's platform.
This means, our GUI is no longer runnable on macOS due to Mono's WinForms implementation being 32-bit only: 

#2270 hardcoded `--arch=32` as argument for Mono, which now prevents users from using the CUI or TUI, if they want to.

## Changes
Now there's a check for arguments in the launch script for macOS. If it's none or `gui`, the flag `--arch=32` is used for Mono, else it's `--arch=64`.

Mono's man page says the following, just FYI:
```
--arch=32, --arch=64
(Mac  OS  X only): Selects the bitness of the Mono binary used, if available.
If the binary used is already for the selected bitness, nothing changes.
If not, the execution switches to a binary with the selected bitness suffix installed side by side
(for example, '/bin/mono --arch=64' will switch to '/bin/mono64' if '/bin/mono' is a 32-bit build).
```

If I understood it right that .apps behave like executables, it should enable to run `ckan.app consoleui` and more again.
Correct me if I'm wrong please.

## Problems
This breaks commands like `ckan --verbose` (on macOS only, of course), which normally should run the GUI with verbose logging in the console. But I don't think a lot of people used this, and they can still add `gui` in the middle to make it work again, so `ckan gui --verbose`.

Also, I did NOT test this, since I'm lacking Apple hardware and as far as I know does Apple a good job preventing macOS to work in VMs.

### DIFF
--- a/macosx/CKAN
+++ b/macosx/CKAN
@@ -22,6 +22,18 @@ else
     # Mono found, so we can run CKAN
     # The exe is called ckan.exe
     ASSEMBLY=ckan.exe
+
+    MONO_ARGS="--arch=32"
+    for ARG in "$@"
+    do
+        # If we want to run the GUI, set mono to 32bit mode. If not, go 64bit.
+        if [[ ! $ARG =~ ^- && $ARG != gui ]]
+        then
+            MONO_ARGS="--arch=64"
+            break
+        fi
+    done
+
     # The script and ckan.exe are in the same folder, go there now
     MACOS_PATH="$(cd "$(dirname "$0")" && pwd)"
     cd "$MACOS_PATH"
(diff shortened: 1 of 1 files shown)


macosx/CKAN
```diff
@@ -22,12 +22,25 @@ else
...
     # The exe is called ckan.exe
     ASSEMBLY=ckan.exe
+
+    MONO_ARGS="--arch=32"
+    for ARG in "$@"
+    do
+        # If we want to run the GUI, set mono to 32bit mode. If not, go 64bit.
+        if [[ ! $ARG =~ ^- && $ARG != gui ]]
+        then
+            MONO_ARGS="--arch=64"
```
HebaruSan: What do you think about clearing the variable instead, in case there's a user out there with a 32-bit only Mac from 2006?

DasSkelett: I didn't want to do that because of this:
> the execution switches to a binary with the selected bitness suffix installed side by side (for example, '/bin/mono --arch=64' will switch to '/bin/mono64' if '/bin/mono' is a 32-bit build).

So to _force_ the 64 bitness, no matter how it's installed.

The help text also says:
> Selects the bitness of the Mono binary used, **if available**

That sounds like if 64-bit mono isn't available, the command will be ignored.

PR #2903, opened 2019-10-25
Replace temp filename with URL in netkan ZIP error

HebaruSan:
## Problem

These ZIP file format errors include temp file names, which are not useful for investigation because we can't access the temp files on the bot server:

image

They are also different every time. Ideally an error message should be consistent. As of KSP-CKAN/NetKAN-Infra#68, these messages will be sent to Discord when they change, so it will be more annoying to have them re-sent every time when the underlying error condition is the same.

## Changes

Now the message will print the source URL instead of the temp file name:

`2755 [1] FATAL CKAN.NetKAN.Program (null) - https://github.com/formicant/TrimIndicator/releases/download/1.8.0.0/TrimIndicator-1.8.0.0.zip is not a valid ZIP file: Error in step EntryHeader for TrimIndicator/TrimIndicator.dll: Exception during test - 'Central header and local header file name mismatch'`

This will be constant from one pass to the next, as well as being more useful for investigation.

### DIFF
--- a/Netkan/Services/CachingHttpService.cs
+++ b/Netkan/Services/CachingHttpService.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using log4net;
 
 namespace CKAN.NetKAN.Services
 {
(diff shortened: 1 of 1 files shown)


DasSkelett:
Can you put a `Log.Debug($"{downloadedFile} is not a valid zip file");` before the throw? If someone wants to find where the file is saved locally, he can get the info in a `--debug` run.

PR #2904, opened 2019-10-25
Restructure Net.Download and Net.DownloadText

DasSkelett:
## Motivation
#2879 tried to get better insights why calls to GitHub by the bot were failing.
It was a partial success, we got the returned status code, but it also broke the custom error handling for SpaceDock for example.

## Changes
First, I enabled the possibility to pass a header writeback function to CurlSharp. This allows catching the header of the response for logging purposes.
Passing it is optional.

Secondly, I created a new Kraken, `NativeAndCurlDownloadFailedKraken`. This one is used in both above mentioned download methods. It has some nice additions to a plain Kraken/Exception (like accessing the response (header and/or content) or the status code) and is more usable than `WebException`s.

Thirdly, I added some logging to the two download methods. If the debug-flag is set, the entire header and the content (in case of `DownloadText`) is written to the output. Plus, it is save in the `NativeAndCurlDownloadFailedKraken`, so accessing and deciding based on them higher up in the stack is easier.
Also , parsing the returned json containing a error message from SpaceDock/CurseForge now works again. This fixes a failing unit test.

Furthermore in the two last commits, I introduced two custom error messages for GitHub if the rate limit is exceeded (based on return code 403 + `X-RateLimit-Remaining` header),
and if CurseForge is blocking CKAN (based on return code 403).

### DIFF
--- a/Netkan/Program.cs
+++ b/Netkan/Program.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Net;
 using System.Diagnostics;
 using System.IO;
@@ -11,9 +10,7 @@ using log4net.Core;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using CKAN.NetKAN.Model;
-using CKAN.NetKAN.Services;
 using CKAN.NetKAN.Transformers;
-using CKAN.NetKAN.Validators;
 using CKAN.NetKAN.Processors;
 
 namespace CKAN.NetKAN
@@ -106,7 +103,7 @@ namespace CKAN.NetKAN
             }
             catch (Exception e)
             {
-                e = e.GetBaseException() ?? e;
+                e = e.GetBaseException();
 
                 Log.Fatal(e.Message);
 
--- a/Core/Net/Curl.cs
+++ b/Core/Net/Curl.cs
@@ -33,7 +33,7 @@ namespace CKAN
 
         /// <summary>
         /// Release any resources used by libcurl. NOT THREADSAFE AT ALL.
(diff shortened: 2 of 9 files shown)


Netkan/Program.cs
```diff
@@ -106,13 +106,13 @@ public static int Main(string[] args)
...
             {
-                e = e.GetBaseException() ?? e;
+                e = e.GetBaseException();
 
                 Log.Fatal(e.Message);
 
                 if (Options == null || Options.Debug)
                 {
-                    Log.Fatal(e.StackTrace);
+                    Log.Fatal(e.ToString());
```
HebaruSan: Can we keep the stack trace? It might not be included in every implementation of `ToString`.

DasSkelett: According to [this section](https://docs.microsoft.com/en-us/dotnet/api/system.exception.tostring?view=netframework-4.8#remarks) it is by default. Except `ToString()` is overridden in one of the krakens...

HebaruSan: Can we just leave it as it was? We already print `e.Message` above, so I'm not sure why we needed a change here.

DasSkelett: Right, I think I did this because I missed to set `kraken.message` in the constructors of the new `NativeAndCurlDownloadFailedKraken`, that's why I made the custom `ToString()` and changes this line here.

I added later that the message is actually set in the constructor, so it should be fine indeed if I revert this. Thanks!

DasSkelett:
Okay, so now it always logs the answer if the download fails, and only with debug flag if the download succeeds.
This means that @techman83  does not have to do an extra debug run to get something useful from the logs.

Issue #2914, opened 2019-11-07
[Bug] Null reference in LatestAvailable via _UpdateModsList


aggie88:
Background
----------
* Operating System: Win 10 pro 
* CKAN Version: 1.24.0+195ed164d95f
* KSP Version: 

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
no

Problem
-------
When loading get the message unhandled exception has occurred in your application

**Steps to reproduce**
* Install Ksp-avc, Restock, Restock+

**CKAN error code (if applicable):**

```
System.ArgumentNullException: Value cannot be null.
Parameter name: key
   at System.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at CKAN.Registry.LatestAvailable(String module, KspVersionCriteria ksp_version, RelationshipDescriptor relationship_descriptor)
   at CKAN.Registry.LatestAvailableWithProvides(String module, KspVersionCriteria ksp_version, IEnumerable`1 available_for_current_version, RelationshipDescriptor relationship_descriptor)
   at CKAN.Registry.allDependenciesCompatible(CkanModule mod, KspVersionCriteria ksp_version, CkanModule[] modules_for_current_version)
   at CKAN.Registry.Available(KspVersionCriteria ksp_version)
   at CKAN.Main._UpdateModsList(Boolean repo_updated, List`1 mc)
   at CKAN.Main.PostInstallMods(Object sender, RunWorkerCompletedEventArgs e)
   at System.ComponentModel.BackgroundWorker.OnRunWorkerCompleted(RunWorkerCompletedEventArgs e) 
```

HebaruSan:
Which mods did you install before this?

aggie88:
Ksp-avc
Restock
Restock+

aggie88:
All updated via ckan last night

jamespglaze:
I am seeing the same exception, I have converted to .txt and attached the .ckan files from today when I saw the error, and Nov 3rd when I I last launched and updated:



Stack trace:

System.ArgumentNullException: Value cannot be null.
Parameter name: key
   at System.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
[...]

PR #2915, opened 2019-11-07
Apply netkan validators to ckans

HebaruSan:
## Problem

KerbalAtomics-1-1.1.0.ckan used `any_of` with a `spec_version` of `v1.4`: https://github.com/KSP-CKAN/CKAN-meta/commit/967423fbfe09bdd1d8af4a42ea46fa716b2d92bb

This means this module loads in CKAN versions that don't support `any_of` and crashes them, see #2914.

This should have failed with this error:

> spec_version v1.26+ required for 'any_of'

But it got through.

## Cause

When we inflate a module, we:

1. Load the netkan
2. Apply the netkan validators
3. Inflate
4. Apply the ckan validators

The `any_of` check is in the netkan validators, so any netkan with a `spec_version` below `v1.26` that tries to use `any_of` will be flagged.

But KerbalAtomics' relationships aren't in its netkan, they're in its *meta*netkan! Metanetkans are loaded and applied in the inflation stage, after netkan validators have already been checked. So the `any_of` check is missed, and metanetkans are allowed to engage in shenanigans.

## Changes

Now we apply the netkan validators to the final ckans. This will prevent metanetkans from bypassing validation.

### DIFF
--- a/Netkan/Processors/Inflator.cs
+++ b/Netkan/Processors/Inflator.cs
@@ -48,6 +48,7 @@ namespace CKAN.NetKAN.Processors
 
                 foreach (Metadata ckan in ckans)
                 {
+                    netkanValidator.Validate(ckan);
                     ckanValidator.ValidateCkan(ckan, netkan);
                 }
                 log.Info("Output successfully passed post-validation");


techman83:
Yep that all makes sense and that seems like a sensible way to address the problem.

PR #2922, opened 2019-11-15
Add owners of parent GitHub repos to authors

HebaruSan:
## Motivation

The `author` property of a module can be a string or an array of strings. It's courteous to credit everyone who has maintained a mod over the years by listing them in an array.

It is somewhat common for a mod on GitHub to go out of maintenance and be adopted by a new maintainer on a fork of the original repo. When this happens, we switch the `$kref` to the new repo, and in order to preserve the full list of authors, we have to manually edit the `author` property to include them. It's better to rely on automatic data generation where possible, or at least to have the option.

## Changes

Now if a repo is a fork-of-a-fork-of-a-fork-of..., we will traverse the parent repository chain all the way to the root and add all of the repository owners as co-authors. This will give a more accurate picture of who has contributed when a repo is forked.

Note that manual overrides of the `author` property still work the same, so this logic will not blow away any manually maintained author lists. To minimize the performance hit, we will not make any extra GitHub API calls when the `author` property is manually set in the netkan.

### DIFF
--- a/Netkan/Extensions/JObjectExtensions.cs
+++ b/Netkan/Extensions/JObjectExtensions.cs
@@ -30,6 +30,26 @@ namespace CKAN.NetKAN.Extensions
         }
         
         /// <summary>
+        /// Write a property to a <see cref="JObject"/> only if it does not already exist and the value is not null.
+        /// The value is generated on-demand by calling tokenCallback, only when we need it.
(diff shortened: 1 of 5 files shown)


DasSkelett:
This is a cool idea!

PR #2925, opened 2019-11-21
Add language selection option to settings

DasSkelett:
## Motivation
Quite a few forum users asked for an option to change the language in CKAN after the release that introduced localization.

## Changes
I added a combobox to the settings that allows changing the language.
The selected language is saved in the "Core's" configuration (so the JSON config and also the Win32Registry, although the second one is no longer in use, but it was easy and fast to do).
It is saved in the [RFC 5646](https://tools.ietf.org/html/rfc5646) / [RFC 4647](https://tools.ietf.org/html/rfc4647) format, the same that is used as appendix for the resource files (=`de-DE`,`en-US`,`en-GB`).

## How it works:
By default, the `Language` field in the config is `null`.
### First start / no language saved in config
When the CKAN GUI starts, it checks the currently saved language. If it's null, it'll try to save whatever language CKAN was launched in the config.
If the launch language is not one that is explicitly supported by CKAN, the config value stays `null`, and the GUI continues in the launch language.
grafik
grafik

Now when the user opens the settings, he will see a "blank" combobox. When he clicks it and selects on of the currently three supported languages, it is saved in the config. A restart is required to apply the languages (I didn't want to bother with re-initializing **all** GUI controls, and nothing wrong with a restart of the client).
grafik

### Subsequent starts  / language set in the config
When CKAN starts and finds a language set in the config, it'll apply it as `System.Threading.Thread.CurrentUICulture`, **before** any controls are initialized.
grafik

When the user opens the settings, the current language will be selected by default in the combobox.
grafik

The CKAN-supported languages are accessible via `string[] CKAN.Utilities.AvailableLanguages`.

---

I also added an additional check in `GUIConfig.LoadOrCreateConfiguration()`  to create a new config if the file exists but is empty.

### DIFF
--- a/GUI/SettingsDialog.resx
+++ b/GUI/SettingsDialog.resx
@@ -160,9 +160,10 @@
   <data name="RefreshPostLabel.Text" xml:space="preserve"><value>minute(s)</value></data>
   <data name="PauseRefreshCheckBox.Text" xml:space="preserve"><value>Pause refreshing</value></data>
   <data name="MoreSettingsGroupBox.Text" xml:space="preserve"><value>More Settings</value></data>
-  <data name="AutoSortUpdateCheckBox.Text" xml:space="preserve"><value>Automatically sort by "Update"-column when clicking "Add available updates"</value></data>
+  <data name="LanguageSelectionLabel.Text" xml:space="preserve"><value>Language (requires restart):</value></data>
   <data name="RefreshOnStartupCheckbox.Text" xml:space="preserve"><value>Update repositories on launch</value></data>
-  <data name="HideEpochsCheckbox.Text" xml:space="preserve"><value>Hide epoch numbers in mod list (Requires Restart)</value></data>
-  <data name="HideVCheckbox.Text" xml:space="preserve"><value>Hide "v" in mod list (Requires Restart)</value></data>
+  <data name="HideEpochsCheckbox.Text" xml:space="preserve"><value>Hide epoch numbers in mod list (requires restart)</value></data>
+  <data name="HideVCheckbox.Text" xml:space="preserve"><value>Hide "v" in mod list (requires restart)</value></data>
+  <data name="AutoSortUpdateCheckBox.Text" xml:space="preserve"><value>Automatically sort by "Update"-column when clicking "Add available updates"</value></data>
   <data name="$this.Text" xml:space="preserve"><value>Settings</value></data>
 </root>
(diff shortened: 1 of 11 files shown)


HebaruSan:
The resizing/scrolling is not working at all well in my testing. It defaults to a bad size, it doesn't fill the space if I make it larger, it chops off the left part of the window without being able to scroll back sometimes, etc. I think we would receive a lot of complaints if we shipped it like this.

If you want to do something to fit smaller screens, what about a tab list at the left, with one tab per panel? It's a pretty common UI pattern for settings windows nowadays.

image

image

GUI/SettingsDialog.resx
```diff
@@ -160,9 +160,10 @@
   <data name="RefreshPostLabel.Text" xml:space="preserve"><value>minute(s)</value></data>
   <data name="PauseRefreshCheckBox.Text" xml:space="preserve"><value>Pause refreshing</value></data>
   <data name="MoreSettingsGroupBox.Text" xml:space="preserve"><value>More Settings</value></data>
-  <data name="AutoSortUpdateCheckBox.Text" xml:space="preserve"><value>Automatically sort by "Update"-column when clicking "Add available updates"</value></data>
+  <data name="LanguageSelectionLabel.Text" xml:space="preserve"><value>Language (Requires Restart):</value></data>
```
HebaruSan:
```suggestion
  <data name="LanguageSelectionLabel.Text" xml:space="preserve"><value>Language (requires restart):</value></data>
```

The parenthesized part doesn't need capitalization in English

DasSkelett: I did it because all the other occurrences were capitalized too.
I can change all of them :+1:

DasSkelett:
> The resizing/scrolling is not working at all well in my testing. It defaults to a bad size, it doesn't fill the space if I make it larger, it chops off the left part of the window without being able to scroll back sometimes, etc. I think we would receive a lot of complaints if we shipped it like this.
> 
> If you want to do something to fit smaller screens, what about a tab list at the left, with one tab per panel? It's a pretty common UI pattern for settings windows nowadays.

Then I'll remove the resizing stuff from this PR for now, and maybe try that in another one.

GUI/SettingsDialog.resx
HebaruSan: Hmm, I don't think anything else needs to change... :worried:

DasSkelett: https://github.com/DasSkelett/CKAN/blob/2506ce06a09a73bee419bf563c53d7779561170b/GUI/SettingsDialog.resx#L165-L166
What about those?

HebaruSan: Oh! Yes, change those as well. :relieved:

DasSkelett:
Rebased to remove the resizing stuff, and changed the `Requires Restart` to lower case.

PR #2934, opened 2019-11-26
Notify on Deployment Success

techman83:
Travis was definitely a little spammy, but the lack having a notification post deployment success is useful.

This limits the notification to the conditions that match our deployment conditions. We could drop the check for master if we want to be notified once per set of builds. Though we should consider moving to [stages](https://docs.travis-ci.com/user/build-stages). Which I might do on NetKAN-infra as an example first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -109,6 +109,11 @@ deploy:
         branch: master
         condition: $BUILD_CONFIGURATION = Release && $(mono --version | perl -ne'/version (\d+\.\d+\.\d+)/ and print $1') = $BUILD_RELEASE_MONO_VERSION
 
+after_success:
+  - wget https://raw.githubusercontent.com/DiscordHooks/travis-ci-discord-webhook/master/send.sh
+  - chmod +x send.sh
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -109,6 +109,11 @@ deploy:
         branch: master
         condition: $BUILD_CONFIGURATION = Release && $(mono --version | perl -ne'/version (\d+\.\d+\.\d+)/ and print $1') = $BUILD_RELEASE_MONO_VERSION
 
+after_success:
+  - wget https://raw.githubusercontent.com/DiscordHooks/travis-ci-discord-webhook/master/send.sh
```
HebaruSan: Suggest moving the `wget` command inside the `if` so we don't download the script if we don't need it.

techman83: It'll probably not need it when I refactor the lot to use stages and I _know_ this will work. Breaking it up across the lines didn't fit the few minutes I had spare to restore something I do keep an eye on.

HebaruSan: OK, I look forward to seeing the stages stuff.

PR #2937, opened 2019-11-30
Obey system colors for dark theme support

### LINKED ISSUE #2881: [Feature] Dark mode
Suggestions
-----------
It should be great if you could add a dark mode for the main windows late at night the window is killing my eyes ;)

Edit :
Oopsy may by the GUI redisign could be enough ;)
https://github.com/KSP-CKAN/CKAN/issues/2848

HebaruSan:
## Problem

You can configure the legacy system colors on Windows to change how WinForms apps display. In Windows 7:

1. Right click desktop
2. Personalize
3. Window Color (in the bottom middle)
4. Advanced appearance settings...
5. Choose the colors that you want to use

CKAN doesn't look very good in this mode:

bad

Note that it looks OK if you choose one of the more extreme "high contrast" themes, but these also change your window borders and everything else, so it's not a great solution:

high contrast

## Causes

- A number of the WinForms controls default to a hard coded style instead of obeying the system theme
- The status bar needed its BackColor/ForeColor properties set

## Changes

- The mod list has visual styles turned off and its header color set to obey the system colors
- `ThemedListView` and `ThemedTabControl` classes are created as drop-in replacements for `ListView` and `TabControl` but tweaked to obey system colors
- The status bar's BackColor and ForeColor are set

This makes dark themes on Windows look better:

better

It also results in a minor degradation of appearance in a normal theme because the gradients for inactive tabs and list headers are suppressed:

light

Needs to be tested on Linux. I think it should look the same, but there could be surprises.

Fixes #2881.

### DIFF
--- a/GUI/CKAN-GUI.csproj
+++ b/GUI/CKAN-GUI.csproj
@@ -249,6 +249,8 @@
     <Compile Include="SingleAssemblyComponentResourceManager.cs" />
     <Compile Include="SingleAssemblyResourceManager.cs" />
     <Compile Include="TabController.cs" />
+    <Compile Include="ThemedTabControl.cs" />
+    <Compile Include="ThemedListView.cs" />
     <Compile Include="URLHandlers.cs" />
     <Compile Include="Util.cs" />
     <Compile Include="X11.cs" />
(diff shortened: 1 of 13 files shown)


DasSkelett:
On Linux, it is now really hard to recognize the active tab in tab panels:

grafik
grafik
grafik

HebaruSan:
I guess .NET considers the tab's "bounds" to be the space *inside* the border, while Mono considers it to include the border? Fix pushed, need to test in Windows. Hopefully we don't have to put `Platform.IsMono` checks in here to make it work on both.

HebaruSan:
Worse than I hoped, better than I feared; there's a terrible looking 1px line under the tabs now in Windows:

image

... but otherwise it's serviceable. Hopefully translating the box down 1px will solve that.

HebaruSan:
Yup, offsetting down by 1 worked.

PR #2941, opened 2019-12-11
Return failure on failed commands for headless prompt

HebaruSan:
## Motivation

I want to use `ckan prompt` to speed up the PR validation scripts by combining multiple commands, see KSP-CKAN/xKAN-meta_testing#55.

## Problem

Currently if a command executed via `ckan prompt` fails, errors are printed but the process keeps running and eventually returns success at exit. This makes sense for an interactive mode where a user is reading the errors and can try again, but for what I have in mind it means that failed installations would be ignored and marked as successful.

## Changes

Now if a command fails within `ckan prompt`, and we're running in headless mode, the process will exit with the failed command's return value. This will allow it to be used in scripts that check for success or failure of operations.

### DIFF
--- a/Cmdline/Action/Prompt.cs
+++ b/Cmdline/Action/Prompt.cs
@@ -35,7 +35,13 @@ namespace CKAN.CmdLine
                 {
                     // Parse input as if it was a normal command line,
                     // but with a persistent KSPManager object.
-                    MainClass.Execute(manager, opts, command.Split(' '));
+                    int cmdExitCode = MainClass.Execute(manager, opts, command.Split(' '));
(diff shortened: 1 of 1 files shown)


DasSkelett:
Theoretically, this _could_ break somebody's workflow, _if_ anyone is using headless prompts in some CI setup or whatever and wants it to continue even if one step fails.
We could add another option like `--abort-on-error` to implement this.

...but I highly suspect anyone is using it like this, so only do it if you want to. It's just something that came up thinking about it.

HebaruSan:
- No one is doing that :stuck_out_tongue_closed_eyes: 
- If they are, they probably don't know that their errors are being ignored :bomb: 
- And they should be smart enough to read our release notes before updating to a new version :thinking:

PR #2944, opened 2019-12-16
Pause Inflator if its output queue fills up

HebaruSan:
## Problem

The bot froze for 18+ hours today and had to be hard reset. The Webhooks, Scheduler, and Inflator were apparently still running, but the Indexer was not responsive; the Inflator's output queue had 11,000 entries in it.

## Cause

After #2942, the bot began a pass that would re-index nearly every non-frozen module because we changed the sort order of some metadata properties. This apparently overloaded the disk IO operations capacity of the Indexer, which has to create commits, create and check out branches, push to GitHub, etc. Meanwhile, while the Indexer is struggling, the Scheduler and Inflator are churning away, consuming their usual levels of resources as if everything was fine, which contributes to starving the Indexer of resources.

The Scheduler prudently has code to check that its input queue is not full before it starts a pass:

https://github.com/KSP-CKAN/NetKAN-Infra/blob/c2871402588322c756ec26f3e95abc55f29c2fe7/netkan/netkan/scheduler.py#L83-L92

The Inflator currently lacks such safeguards, so if something holds up the Indexer, the Inflator will happily keep queuing more and more messages, and the Scheduler will keep running because the Inflator's input queue is kept empty, even though those module inflations aren't making it to completion.

## Changes

Now the Inflator will check its output queue before it starts inflating a batch of messages. If there are fewer than 100 messages, it will proceed with inflation as usual. Otherwise, it will pause and check again in 20 seconds.

If the Indexer starts having trouble for a long period again, this behavior will prevent the Inflator from over-filling its output queue and consuming CPU and disk resources. As a consequence, the Inflator's *input* queue will also retain a large number of entries, which will cause the Scheduler to skip its runs. Overall this will decrease consumption of system resources and allow the Indexer to catch up.

### DIFF
--- a/Netkan/Processors/QueueHandler.cs
+++ b/Netkan/Processors/QueueHandler.cs
@@ -4,6 +4,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Threading;
 using Amazon.SQS;
 using Amazon.SQS.Model;
 using log4net;
@@ -31,9 +32,27 @@ namespace CKAN.NetKAN.Processors
         {
             while (true)
             {
-                // 10 messages, 30 minutes to allow time to handle them all
-                handleMessages(inputQueueURL, 10, 30);
+                if (numQueued(outputQueueURL) < 100)
+                {
+                    // 10 messages, 30 minutes to allow time to handle them all
+                    handleMessages(inputQueueURL, 10, 30);
+                }
+                else
+                {
+                    // Wait 20 seconds to let the Indexer catch up
+                    log.Info("Output queue too full, waiting...");
+                    Thread.Sleep(20 * 1000);
+                }
             }
+        }
+
+        private int numQueued(string queueURL)
+        {
+            return client.GetQueueAttributes(new GetQueueAttributesRequest()
+            {
+                QueueUrl       = queueURL,
+                AttributeNames = new List<string>() { "ApproximateNumberOfMessages" }
+            })?.ApproximateNumberOfMessages ?? 0;
         }
 
         private string getQueueUrl(string name)


HebaruSan:
For what it's worth, I see this as a necessary guardrail, not a sufficient solution to the problem we had today. If the problem is really the Scheduler's passes over CKAN-meta to find max versions, then it's important to suppress that behavior when the system as a whole is struggling, and this would accomplish that.

techman83:
I want to think on this a little. Mostly because I think if we want to check our output Queue, this is the wrong place to put it. If we were to guard against this, it should be in the scheduler, rather than the inflator adding a lot of unnecessary queue calls (checking the queue length is considered receiving a message, which counts towards our usage).

It's also very very unusual for the Indexing process to be the place that the bottle neck in our processes. If anything, we probably should be checking our disk burst credits, because if our indexing process has backed up, something has gone quite wrong.

HebaruSan:
> checking the queue length is considered receiving a message, which counts towards our usage

We could check less often, say, run 10-50 batches in between checks? That would be a pretty minimal additional load while still detecting an aberrant condition before it gets too out of hand.

But if you're committed to removing the Scheduler's queue checks in KSP-CKAN/NetKAN-Infra#112, then this wouldn't make as much sense anymore, since backing up the input queue would no longer pause the Scheduler. Maybe we could keep that just in case?

> It's also very very unusual for the Indexing process to be the place that the bottle neck in our processes.

True, but now we know how to make it happen. I'd like to be able to do mass commits if we have a reason to, rather than being constrained by it.

techman83:
> We could check less often, say, run 10-50 batches in between checks? That would be a pretty minimal additional load while still detecting an aberrant condition before it gets too out of hand.

I think adding a check here is adding unnecessary complication and another place we need to check when something isn't working. We encountered a very unusual situation, as we've never exhausted disk io before, which is a testament to your recent work on making it more efficient.

> True, but now we know how to make it happen. I'd like to be able to do mass commits if we have a reason to, rather than being constrained by it.

Absolutely agree. And I think the scheduler is the right place to ensure this is ok. It should be responsible for keeping tabs on things as we can easily back it off. We should be able to spin up the scheduler say every half an hour and will either schedule or it will wait until the next run.

HebaruSan:
OK, abandoning this for now. If it looks appealing at some point in the future, well, git makes it easy to come back and revive it.

techman83:
> OK, abandoning this for now. If it looks appealing at some point in the future, well, git makes it easy to come back and revive it.

Agreed. I think doing it now is jumping the gun a little right now, but happy to come back to it if we find ourselves in the same situation again.

PR #2946, opened 2019-12-17
Coerce GitHub URLs into the authenticated API in Netkan

### LINKED ISSUE #2945: [Bug] NetKAN being rate limited by unauthenticated requests
Background
----------

One thing preventing us from indexing more frequently is rate limiting by GitHub. The process should go nowhere near hitting the 5000/hr limit. In theory we could run more often than we're able to actually process them and possibly not hit the limit.

```
New inflation error for HalfRSS-Textures2048: GitHub API rate limit exceeded.
New inflation error for HalfRSS-Textures4096: GitHub API rate limit exceeded.
New inflation error for HalfRSS-Textures8192: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-Afterburner: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-AutoBalancingLandingLeg: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-DisableTempGauges: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-DailyFunds: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-DestroyAll: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-EditorCamUtilities: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-FillSpotsWithTourists: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-FPSLimiter: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-FPSViewer: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-KerbalScienceExchange: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-ModifiedExplosionPotential: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-PhysicalTimeRatioViewer: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-RecoverAll: GitHub API rate limit exceeded.
New inflation error for KerboKatzUtilities: GitHub API rate limit exceeded.
```

Problem
-------
Checking our unauthenticated limit, we are definitely exceeding it.

> For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the user making requests.

```json
{
  "resources": {
    "core": {
      "limit": 60,
      "remaining": 0,
      "reset": 1576563098
    },
    "search": {
      "limit": 10,
      "remaining": 10,
      "reset": 1576561641
    },
    "graphql": {
      "limit": 0,
      "remaining": 0,
      "reset": 1576565181
    },
    "integration_manifest": {
      "limit": 5000,
      "remaining": 5000,
      "reset": 1576565181
    }
  },
  "rate": {
    "limit": 60,
    "remaining": 0,
    "reset": 1576563098
  }
}
```

HebaruSan:
## Problem

The bot intermittently reports "GitHub API rate limit exceeded" for various modules, most commonly Telemachus, but also recently many of the KerboKatz utilities and the Half/QuarterRSS textures. This is a custom error message created in #2904 that indicates the server headers include `X-RateLimit-Remaining: 0`.

## Cause

As @techman83 noted in #2945, the rate limit API reported 0 remaining unauthenticated API usages (but we are nowhere near the limit for *authenticated* usages):

```
{
  "resources": {
    "core": {
      "limit": 60,
      "remaining": 0,
      "reset": 1576563098
    },
    "search": {
      "limit": 10,
      "remaining": 10,
      "reset": 1576561641
    },
    "graphql": {
      "limit": 0,
      "remaining": 0,
      "reset": 1576565181
    },
    "integration_manifest": {
      "limit": 5000,
      "remaining": 5000,
      "reset": 1576565181
    }
  },
  "rate": {
    "limit": 60,
    "remaining": 0,
    "reset": 1576563098
  }
}
```

This means that when this error occurs, the bot has accessed too many GitHub URLs without using a token. Normally these would consist of:

- Metanetkans (121 of 122 hosted on GitHub)
- KSP-AVC remote version files (many hosted on GitHub, others on ksp-avc.cybutek.net)
- KSP-AVC version file used as a kref (not hosted on GitHub because we would just use a `github` kref)

Unlike krefs where we construct GitHub API calls from other known properties, these URLs are read out of string properties that can contain **any** full URL, but often one from GitHub. Up to now we only use generic code to retrieve the URL normally, without any special processing.

## Changes

Now when Netkan retrieves such URLs, it checks for the known GItHub hosts and URL patterns. If we find a match, we retrieve the file via the API instead. If we have a token, it will be used. This will reduce the number of unauthenticated GitHub calls we make.

Note that the core and clients are functionally unchanged and will still access all URLs the same way as before. We're just updating Netkan.

Fixes #2945.

### DIFF
--- a/Netkan/Sources/Github/GithubApi.cs
+++ b/Netkan/Sources/Github/GithubApi.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
+﻿using System.Text.RegularExpressions;
 using log4net;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
@@ -20,6 +21,26 @@ namespace CKAN.NetKAN.Sources.Github
 
         private readonly IHttpService _http;
         private readonly string       _oauthToken;
+
+        private const string rawMediaType = "application/vnd.github.v3.raw";
+
+        // https://github.com/<OWNER>/<REPO>/blob/<BRANCH>/<PATH>
+        // https://github.com/<OWNER>/<REPO>/tree/<BRANCH>/<PATH>
+        // https://github.com/<OWNER>/<REPO>/raw/<BRANCH>/<PATH>
+        private static readonly Regex githubUrlRegex = new Regex(
+            @"^/(?<owner>[^/]+)/(?<repo>[^/]+)/(?<type>[^/]+)/(?<branch>[^/]+)/(?<path>.+)$",
+            RegexOptions.Compiled);
+
+        private static readonly HashSet<string> urlTypes = new HashSet<string>()
+        {
+            "blob", "tree", "raw"
+        };
+
+        // https://raw.githubusercontent.com/<OWNER>/<REPO>/<BRANCH>/<PATH>
+        private static readonly Regex githubUserContentUrlRegex = new Regex(
+            @"^/(?<owner>[^/]+)/(?<repo>[^/]+)/(?<branch>[^/]+)/(?<path>.+)$",
+            RegexOptions.Compiled
+        );
 
         public GithubApi(IHttpService http, string oauthToken = null)
         {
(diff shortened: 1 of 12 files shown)


Netkan/Sources/Github/GithubApi.cs
```diff
@@ -100,14 +121,79 @@ public List<GithubUser> getOrgMembers(GithubUser organization)
...
+        /// <returns>
+        /// null if the URL isn't on GitHub, otherwise the contents of the download
+        /// </returns>
+        public string DownloadText(Uri url)
+        {
+            string ghOwner  = null;
+            string ghRepo   = null;
+            string ghBranch = null;
+            string ghPath   = null;
+            if (TryGetGitHubPath(url, out ghOwner, out ghRepo, out ghBranch, out ghPath))
```
DasSkelett: If you want to save some lines, you can also do inline declaration:
```suggestion
            if (TryGetGitHubPath(url, out string ghOwner, out string ghRepo, out string ghBranch, out string ghPath))
```

DasSkelett:
I don't understand the GitHub `/rate_limit` endpoint. It tells me two different values in the header and in the body:
```
$ curl -v -H "Authorization: token <token>" "https://api.github.com/rate_limit"

< X-RateLimit-Limit: 5000
< X-RateLimit-Remaining: 4964
< X-RateLimit-Reset: 1576623587

"rate": {
"limit": 5000,
"remaining": 5000,
"reset": 1576625285
}
```

Netkan/Sources/Github/GithubApi.cs
HebaruSan: That's a nice language feature!

DasSkelett: It is!

techman83:
Brilliant! We can put a check into our scheduler and skip runs that are below X remaining API calls.

HebaruSan:
Hmm, there was no way this was ever going to help with "GitHub API rate limit exceeded", because that error message is **only** raised for (authenticated) API calls already!

techman83:
Is that correct? Whenever I've checked it was only our un-authenticated calls that were exhausted, not authenticated. Unless we're hitting some other throttling?

HebaruSan:
Yup, the error message is in `GitHubApi.Call`, which is specifically for the API and always passes the token:

https://github.com/KSP-CKAN/CKAN/blob/78438b30f5bcb9bc819e8ab1064148cd88ea3e71/Netkan/Sources/Github/GithubApi.cs#L185-L202

Other throttling is quite possible.

PR #2947, opened 2019-12-18
Only stage auto-epoch when creating new file

HebaruSan:
## Motivation

A lot of auto-epoching pull requests have been updates for versions that were created in previous PRs, see KSP-CKAN/CKAN-meta#1621 and KSP-CKAN/CKAN-meta#1694 for an example. In this case we have already confirmed in the first PR that the version should exist, so manual review isn't needed for later ones.

## Changes

Now we only enable staging if the resulting version after auto-epoching is greater than the previous highest version. If we're updating a version that already exists, staging isn't needed.

### DIFF
--- a/Netkan/CmdLineOptions.cs
+++ b/Netkan/CmdLineOptions.cs
@@ -40,6 +40,9 @@ namespace CKAN.NetKAN
         [Option("queues", HelpText = "Input,Output queue names for Queue Inflator mode")]
         public string Queues { get; set; }
 
+        [Option("highest-version", HelpText = "Highest known version for auto-epoching")]
+        public string HighestVersion { get; set; }
+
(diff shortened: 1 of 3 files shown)


DasSkelett:
Code-wise it looks fine. Though I'm thinking whether we should add a `--highest-version` commandline option so we can test changes like this in action without needing to set up the dev queues and stuff?

HebaruSan:
Good idea, added.

techman83:
I'm not finished my coffee. How does netkan know there is a new or existing file?

HebaruSan:
If the file is new, its version will be higher than the highest version.
If the file exists, it will be equal to the highest version.

(If it's lower, it will be auto-epoched until one of the above is true.)

techman83:
Ok Cool! Makes sense :slightly_smiling_face:

PR #2950, opened 2019-12-24
Request fewer GitHub releases and cache string URLs in Netkan

### LINKED ISSUE #2949: [Bug] NetKAN being rate limited by authenticated requests
## Problem

#2945 reported these inflation errors as being caused by unauthenticated API requests:

```
New inflation error for HalfRSS-Textures2048: GitHub API rate limit exceeded.
New inflation error for HalfRSS-Textures4096: GitHub API rate limit exceeded.
New inflation error for HalfRSS-Textures8192: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-Afterburner: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-AutoBalancingLandingLeg: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-DisableTempGauges: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-DailyFunds: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-DestroyAll: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-EditorCamUtilities: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-FillSpotsWithTourists: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-FPSLimiter: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-FPSViewer: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-KerbalScienceExchange: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-ModifiedExplosionPotential: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-PhysicalTimeRatioViewer: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-RecoverAll: GitHub API rate limit exceeded.
New inflation error for KerboKatzUtilities: GitHub API rate limit exceeded.
```

In fact, they are all **authenticated**. The `GitHub API rate limit exceeded` string occurs in `GitHubApi.Call`, which makes authenticated API requests:

https://github.com/KSP-CKAN/CKAN/blob/78438b30f5bcb9bc819e8ab1064148cd88ea3e71/Netkan/Sources/Github/GithubApi.cs#L185-L202

So when we captured all the unauthenticated GitHub calls and made them authenticated in #2946, this problem got *worse*.  Where before it was intermittent, now there are 11 mods on the status page that persistently suffer from it (plus 6 Stock Visual Enahncements modules that are mis-reported 404s).

http://status.ksp-ckan.space/

## Cause

Further research suggests that GitHub may trigger rate limiting not because the documented limits were exceeded, but because an *individual request* is too expensive to complete:

https://www.reddit.com/r/github/comments/bif6io/abuse_detection_mechanism_after_5_searches/en8r2x6/

> GitHub seems to classify most code searches as computationally intensive therefore they trigger the abuse detection.

If that's what's happening, this would be our culprit:

https://github.com/KSP-CKAN/CKAN/blob/78438b30f5bcb9bc819e8ab1064148cd88ea3e71/Netkan/Sources/Github/GithubApi.cs#L65

Retrieving up to 100 releases (which would include **all** releases for most mods) is a lot of database and network traffic, and in most cases not necessary. This was added in #2681 so the `--releases all` option would not miss releases.

## Suggestion

We could instead retrieve fewer releases per request and then retrieve additional groups if we need them. This would be tricky to get right, but would make the 80% case less intensive and hopefully possibly help with those errors.

HebaruSan:
## Problem

```
New inflation error for HalfRSS-Textures2048: GitHub API rate limit exceeded.
New inflation error for HalfRSS-Textures4096: GitHub API rate limit exceeded.
New inflation error for HalfRSS-Textures8192: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-Afterburner: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-AutoBalancingLandingLeg: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-DisableTempGauges: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-DailyFunds: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-DestroyAll: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-EditorCamUtilities: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-FillSpotsWithTourists: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-FPSLimiter: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-FPSViewer: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-KerbalScienceExchange: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-ModifiedExplosionPotential: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-PhysicalTimeRatioViewer: GitHub API rate limit exceeded.
New inflation error for KerboKatzSmallUtilities-RecoverAll: GitHub API rate limit exceeded.
New inflation error for KerboKatzUtilities: GitHub API rate limit exceeded.
```

## Cause

The exact reason this happens is uncertain, but we suspect we're triggering GitHub's abuse prevention limits:

- We're probably requesting too many GitHub releases at once (100)
- We're probably requesting the same API URLs too many times in a short period

## Changes

- Now we request 10 GitHub releases instead of 100, and if we need more (rare!), we request 10 more, and so on. This will cut down on the server load to handle our GitHub API requests.
- Now `CachingHttpService.DownloadText` caches its results for 3 minutes, so any repeated requests for the same URL (e.g., if we inflate all of the KerboKatz modules in rapid succession) will only hit the server once. This will avoid the possibly abuse-y behavior of hitting the same URL over and over, and also probably speed up the Inflator a little bit (since memory is a lot faster than network).

Fixes #2949 (hopefully!).

### DIFF
--- a/Netkan/Services/CachingHttpService.cs
+++ b/Netkan/Services/CachingHttpService.cs
@@ -9,6 +9,10 @@ namespace CKAN.NetKAN.Services
         private readonly NetFileCache _cache;
         private          HashSet<Uri> _requestedURLs  = new HashSet<Uri>();
         private          bool         _overwriteCache = false;
+        private Dictionary<Uri, StringCacheEntry> _stringCache = new Dictionary<Uri, StringCacheEntry>();
+
+        // Re-use string value URLs within 2 minutes
+        private static readonly TimeSpan stringCacheLifetime = new TimeSpan(0, 2, 0);
 
         public CachingHttpService(NetFileCache cache, bool overwrite = false)
         {
@@ -78,11 +82,35 @@ namespace CKAN.NetKAN.Services
 
         public string DownloadText(Uri url)
         {
-            return Net.DownloadText(url);
+            return TryGetCached(url, () => Net.DownloadText(url));
         }
         public string DownloadText(Uri url, string authToken, string mimeType = null)
         {
-            return Net.DownloadText(url, authToken, mimeType);
+            return TryGetCached(url, () => Net.DownloadText(url, authToken, mimeType));
+        }
+        
+        private string TryGetCached(Uri url, Func<string> uncached)
+        {
+            if (_stringCache.TryGetValue(url, out StringCacheEntry entry))
+            {
+                if (DateTime.Now - entry.Timestamp < stringCacheLifetime)
+                {
+                    // Re-use recent cached request of this URL
+                    return entry.Value;
+                }
+                else
+                {
+                    // Too old, purge it
+                    _stringCache.Remove(url);
+                }
+            }
+            string val = uncached();
+            _stringCache.Add(url, new StringCacheEntry()
+            {
+                Value     = val,
+                Timestamp = DateTime.Now
+            });
+            return val;
         }
 
         public IEnumerable<Uri> RequestedURLs { get { return _requestedURLs; } }
@@ -93,4 +121,11 @@ namespace CKAN.NetKAN.Services
 
         private static readonly ILog log = LogManager.GetLogger(typeof(CachingHttpService));
     }
+
+    public class StringCacheEntry
+    {
+        public string   Value;
+        public DateTime Timestamp;
+    }
+    
 }
--- a/Netkan/Sources/Github/GithubApi.cs
+++ b/Netkan/Sources/Github/GithubApi.cs
@@ -62,54 +62,64 @@ namespace CKAN.NetKAN.Sources.Github
 
         public IEnumerable<GithubRelease> GetAllReleases(GithubRef reference)
         {
-            var json = Call($"repos/{reference.Repository}/releases?per_page=100");
-            Log.Debug("Parsing JSON...");
-            var releases = JArray.Parse(json);
-
-            // Finding the most recent *stable* release means filtering
-            // out on pre-releases.
-
-            foreach (var release in releases)
+            const int perPage = 10;
+            for (int page = 1; true; ++page)
             {
-                // First, check for prerelease status...
-                if (reference.UsePrerelease == (bool)release["prerelease"])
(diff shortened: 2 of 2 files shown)


HebaruSan:
I want to see whether this helps, but my reviewers are busy for holiday break. :christmas_tree: 
Merging and will keep an eye and revert if the bot stops working.

DasSkelett:
Hehe. I did take a short look over the code just now, but without testing or anything.
It looks good, maybe it could've used some debug log statements to see whether the cache has been hit or not, but function-wise I think it's good.

HebaruSan:
Nailed it!

image

So just like in #2928, if we get rate limited when the limit is not exceeded, it's because of repeated calls to the same URL. Which should be impossible now.

HebaruSan:
Hmm, I think this froze the Inflator after about 2 days, maybe we used too much memory. We can purge more pro-actively, but that has to be done carefully to avoid taking up a ton of CPU.

PR #2959, opened 2020-01-08
Don't check ZIP health in ModuleInstaller.GetModuleContentsList()

DasSkelett:
## Problem
When scrolling through the mod list with the arrow keys with the file contents tab opened, the GUI will freeze when you encounter biggish cached mods with a lot of files.
It is noticeable with zips around 3 MiB (depending on count of contained files and PC of course) and really disruptive with mods above 50 MiB.

I traced the issue down to the health checks of the zip files done in `NetFileCache.GetCachedZip()`.

## Changes
Use `GetCachedFilename()` instead of `GetCachedZip()` in `ModuleInstaller.GetModuleContentsList()`, which does the same (searching the file and returning the file name if it is cached) without the checks.
To account for the possibility of bit flips and other data degradation problems, catch possible exceptions when opening the zip and return `null`, which would also be returned if the health checks failed.

`GetModuleContentsList()` is only used in `_UpdateModContentsTree()`, so nothing else will be affected.

I also removed another BOM.

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -255,15 +255,22 @@ namespace CKAN
         // TODO: Return files relative to GameRoot
         public IEnumerable<string> GetModuleContentsList(CkanModule module)
         {
-            string filename = Cache.GetCachedZip(module);
+            string filename = Cache.GetCachedFilename(module);
 
             if (filename == null)
             {
                 return null;
             }
 
-            return FindInstallableFiles(module, filename, ksp)
-                .Select(x => x.destination);
+            try
+            {
+                return FindInstallableFiles(module, filename, ksp)
+                    .Select(x => x.destination);
+            }
+            catch (ZipException)
+            {
+                return null;
+            }
         }
 
         /// <summary>
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -252,7 +252,6 @@ namespace CKAN
         ///
         /// Intended for previews.
         /// </summary>
-        // TODO: Return files relative to GameRoot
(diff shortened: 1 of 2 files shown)


Core/ModuleInstaller.cs
```diff
@@ -255,15 +255,22 @@ public void InstallList(ICollection<CkanModule> modules, RelationshipResolverOpt
         // TODO: Return files relative to GameRoot
```
DasSkelett: That TODO could be solved by a simple `ksp.ToRelativeGameDir()`, does anything speak against adding it here?

Core/ModuleInstaller.cs
```diff
@@ -252,18 +252,24 @@ public void InstallList(ICollection<CkanModule> modules, RelationshipResolverOpt
...
             {
                 return null;
             }
 
-            return FindInstallableFiles(module, filename, ksp)
-                .Select(x => x.destination);
+            try
+            {
+                return FindInstallableFiles(module, filename, ksp)
+                    .Select(x => ksp.ToRelativeGameDir(x.destination));
```
HebaruSan: I think `ToRelativeGameDir` will break `MainModInfo.OpenFileBrowser` unless a corresponding change is made there; double clicking the file contents attempts to open the node text in the file browser.

Core/ModuleInstaller.cs
HebaruSan: It makes for a better UI, but see my other comment, we need to update modinfo to fix the double click.

Core/ModuleInstaller.cs
DasSkelett: Amended to the commit, this should address it.

PR #2960, opened 2020-01-08
Add "supports" relationships to recommendations screen

### LINKED ISSUE #1104: Let's clarify what the `supports` relationship means.
The spec has a [pretty short explanation](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#supports) for the `supports` relationship, which came out of #11.
It looks like we need this to be clarified, as KSP-CKAN/NetKAN#1565 shows there's at least two interpretations.

So, I'd like to clarify just what we mean when we use the supports relationship. To me, it's always meant "X works with Y" and in particular "X gives consideration to Y". Using the example which has just come up a mod might "support" the Community Tech Tree, meaning the author has specifically coded their mod to use the extra nodes if it's found to be installed.

At the same time, I think it's reasonable to say that the "Community Tech Supports X". Even though all the effort may have been done in X to ensure compatibility, if it was done in consultation with with the Community Tech Tree, then we'd expect the CTT not to change in a way that causes X to break.

Finally, I'd like to express a very strong preference for `supports` to remain an information-only relationship. Indeed, I can't imagine how it would work as anything else, since all our other bases are already covered in our relationship model (`depends` - you _must_ install this mod; `recommends` - you _should_ install this mod; `suggests` - you _might consider_ installing this mod, `supports` - hey look, a mod!).

Discussion very much welcome. Tagging @dbent in particular, who's very much been in the front-line here.

HebaruSan:
## Background

The spec includes a [`supports`] relationship for indicating compatibility with another mod, but not a recommendation or suggestion that the mod be installed. For example, a parts mod that includes cfg files that enable its parts to be scaled with TweakScale might do this:

[`supports`]: https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#supports

```json
    "supports": [
        { "name": "TweakScale" }
    ],
```

Other examples include mods that add parts to specific tech trees, mods that create props or multi-function display pages for IVAs, etc.

Currently this relationship does not drive any functionality; it's only displayed in the Relationships tab of mod info, as a smiley face:

image

## Motivation

Since mod authors do sometimes fill in this relationship, it would be nice if it did something (sensible). I think it makes sense to treat it as "inverse suggests," such that:

- If I install a tech tree, I would be able to see which mods integrate with that tech tree
- If I install RPM, I would be able to see mods that provide MFD pages
- If I install TweakScale, I would be able to see part mods that allow scaling

Etc.

## Changes

Now during module installation, the recommendations/suggestions screen can contain a third section, "Supported By". This lists modules that support modules that you're installing, defaulting to unchecked, and the user can check the boxes to install, just like suggestions.

For example, if you install KIS, RemoteTech, and TweakScale, you see the several mods that support them:

image

This will make it easier for users to discover mods related to the ones they're installing.

Fixes #1104.

### DIFF
--- a/ConsoleUI/Toolkit/ConsolePopupMenu.cs
+++ b/ConsoleUI/Toolkit/ConsolePopupMenu.cs
@@ -189,6 +189,7 @@ namespace CKAN.ConsoleUI.Toolkit {
         /// <param name="exec">Function to call if the user chooses this option</param>
         /// <param name="radio">If set, this option is a radio button, and this function returns its value</param>
         /// <param name="submenu">Submenu to open for this option</param>
+        /// <param name="enabled">true if this option should be drawn normally and allowed for selection, false to draw it grayed out and not allow selection</param>
         public ConsoleMenuOption(string cap, string key, string tt, bool close,
                 Func<bool> exec = null, Func<bool> radio = null, ConsolePopupMenu submenu = null,
                 bool enabled = true)
--- a/Core/Registry/Tags/ModuleTagList.cs
+++ b/Core/Registry/Tags/ModuleTagList.cs
@@ -58,7 +58,7 @@ namespace CKAN
             {
                 return JsonConvert.DeserializeObject<ModuleTagList>(File.ReadAllText(path));
             }
-            catch (FileNotFoundException ex)
+            catch (FileNotFoundException)
             {
                 return null;
             }
--- a/Core/Types/ModuleInstallDescriptor.cs
+++ b/Core/Types/ModuleInstallDescriptor.cs
@@ -12,7 +12,7 @@ using System.IO;
 namespace CKAN
 {
     [JsonObject(MemberSerialization.OptIn)]
-    public class ModuleInstallDescriptor : ICloneable
+    public class ModuleInstallDescriptor : ICloneable, IEquatable<ModuleInstallDescriptor>
     {
 
         #region Properties
@@ -122,11 +122,23 @@ namespace CKAN
         /// </summary>
         /// <param name="other">The other stanza for comparison</param>
         /// <returns>
-        /// True if they're equivalent, false if they're different
+        /// True if they're equivalent, false if they're different.
         /// </returns>
         public override bool Equals(object other)
         {
-            ModuleInstallDescriptor otherStanza = other as ModuleInstallDescriptor;
+            return Equals(other as ModuleInstallDescriptor);
+        }
+        
+        /// <summary>
+        /// Compare two install stanzas
+        /// </summary>
+        /// <param name="other">The other stanza for comparison</param>
+        /// <returns>
+        /// True if they're equivalent, false if they're different.
+        /// IEquatable<> uses this for more efficient comparisons.
+        /// </returns>
+        public bool Equals(ModuleInstallDescriptor otherStanza)
+        {
             if (otherStanza == null)
                 // Not even the right type!
                 return false;
@@ -163,6 +175,27 @@ namespace CKAN
                 && !include_only_regexp.SequenceEqual(otherStanza.include_only_regexp))
                 return false;
             return true;
+        }
+        
+        public override int GetHashCode()
+        {
+            // Tuple.Create only handles up to 8 params, we have 10+
+            return Tuple.Create(
+                Tuple.Create(
+                    file,
+                    find,
+                    find_regexp,
+                    find_matches_files,
+                    install_to,
+                    @as
+                ),
+                Tuple.Create(                
+                    filter,
+                    filter_regexp,
+                    include_only,
+                    include_only_regexp
+                )
+            ).GetHashCode();
         }
 
         /// <summary>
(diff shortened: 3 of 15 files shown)


DasSkelett:
Neat, I think that's a good use for the `supoports` relationship. It fixes the problem of missing reverse recommendations.

We might want to revisit the conflict handling in the RecSug view. Currently it allows you to select two conflicting modules and continue, but errors out rather ugly before the installation:
RecSug view
Error during installation
It reports an "unknown error" and you should try it again, with the "Cancel" button inactive. You have to find out that you can simply switch back to the mod list via the tab headers, but the error reason is still unknown to the user.

HebaruSan:
On Windows the Retry button is there, and clicking it jumps back to the changeset so you can make different choices on the recommendations screen:

image

I'll make a more concerted effort to fix the Retry button's weird positioning problems on Mono in a little while...

HebaruSan:
The Retry button sends GUI into a weird state on Mono in which the changeset's tab is selected in the TabControl but the changeset tab page isn't displayed (the progress screen remains visible). That should be impossible.

HebaruSan:
After a bit of wrestling, I think the Retry button now works on Mono.

- Set Retry's `Enabled` property instead of `Visible` since I think making it invisible broke Mono's implementation of layout anchoring (also `TabController` does some nasty stuff with `Visible` on Mac, so better not to rely on it)
- Always enable Retry on failure
- Don't hide changeset tab on failure (this fixes the above issue of the tab and tab page getting out of sync)
- Don't need `Invoke` to call `ShowTab` because it does that internally
- Move `TabController`'s private `_WhateverTab` functions into the `WhateverTab` functions for readability
- Fix some compiler warnings

DasSkelett:
Neat! IMHO in the end it would be cool to have a MainModlist-like behaviour with row colouring and warning message directly on selection. But I think that goes beyond this feature PR.

Poodmund:
Reverse recommendations, which is basically what 'supports' behaves as is not manageable but the mod authors end of the mods that are being supported. 

From the Discord:

> Got a query about how the metadata is being handled when recommended/suggested/supported mods are being listed. When you install ReStock or ReStock+ it suggests/recommends two mods but the remotely hosted metadata contains no mention of these recommended mods.
> 
> Is it because these mods list the ReStock and ReStockPlus identifiers in there 'supports' node?
> 
> If so, its a bit misleading that the 2nd column in the list says 'Recommended or Suggested by:' as you can have a mod that has no affiliation 'supporting' another mod and then CKAN says its "Recommended" or "Suggested" by it... when it isn't.
> 
> I would have expected different behaviour; for example I wouldn't expect it to list Magpie Mods for download when choosing ReStock, however, you'd expect to see ReStock listed as a supported mod if you choose Magpie Mods.

```
MOD A

MOD B - supports MOD A

-----

Install MOD A

No recommend, suggest or support lists.

-----

Install MOD B

List that it supports MOD A
```

> The way it acts currently, you could create a mod called something like 'MOD A Killer' that disables all of the MOD A mod and list in its metadata that it supports MOD A. Then if somebody chose to install MOD A, it would list 'MOD A Killer' as a recommended/suggested/supported mod and the MOD A dev team would have no say in whether this mod is listed for download after chosing to install MOD A... does this all make sense?

I would say this this functionality is anti-developer/author as you have no control over who 'supports' a mod you list on CKAN and those mods that are supposedly 'supporting' your mod can potentially cause issues for you as the mod author, whether it be compatibility, functionality or you just do not condone the mod as the author.

HebaruSan:
I don't think it's very sporting to whip out the "anti-developer" cudgel to get your way.

Are there actual problems caused by this? Or is it hypothetical?

HebaruSan:
>  its a bit misleading that the 2nd column in the list says 'Recommended or Suggested by:' 

If we found a better name for that column, would that address your concerns?

Poodmund:
After discussing this a little bit with some other mod authors the general response was that no one realised that this was being handled/displayed to the end user in this way and the responses were along the lines of, "not exactly a fan", "as long as its obvious they're not recommended by our mods" and "it needs a better word than supports" which kind of echoes along the same concerns that I had previously.

To respond to your post above, I can see how this relationship feature would be informational and handy to the end user to identify mods that are set up to use a framework mod, as you pointed out previously as the main intention behind the feature, and a mod author commented that if this feature was fully transparent in that its not an endorsement of the upstream mod to the downstream mod, then its a good way to get the downstream mod visibility... which is the whole point right? So I think it could be a case of making it super clear what 'supports' signifies.

Personally I think a clear way of doing this would be to remove the 'Supports' list from the page listing with 'Recommends' and 'Suggests' and moving it to its own independent list page which would allow for different column headers and page descriptors to fully clarify what is being listed. This is currently what appears when installing the B9 Parts Pack:

Untitled-1

My suggestion would instead just list 'Recommends' and 'Suggests' on a single page (as these are mods identified by the mod author):

Untitled-2

And then list 'Supports' independently with full clarity:

Untitled-3

This ultimately gives the end user all the information as to what is being listed and why it is being listed by CKAN in a way that doesn't disrupt the current install sequence dramatically whilst retaining the 'Supports' functionality.

Thoughts?

HebaruSan:
I definitely agree we should clarify the language on that screen to account for the `supports` relationship. I'll take care of that, and in the meantime discussion of other changes can continue.

PR #2964, opened 2020-01-16
Use Mono 6.6 and more recent versions of everything else

HebaruSan:
## Problem

Recent net-core builds have been failing.

```
Error: The assembly name is invalid.
```

Today my Ubuntu upgraded from Mono 6.6 to Mono 6.8, and suddenly I was affected locally by the same problem for *all* builds.

## Cause

See cake-build/cake#2695, it looks like a check was added to Mono in mono/mono#16446 to reject full pathnames for assemblies, but unfortunately parts of `Microsoft.CodeAnalysis` (closed source?) appear to pass full pathnames, which breaks Cake. This will probably need to be fixed on the Mono side, but I don't yet have enough information for a useful report.

The net-core builds were set to `mono: latest` instead of a specific version like everything else, so when this (apparently) broken version shipped we started using it immediately.

## Changes

To fix the error:

- Now our net-core builds use Mono 6.6 instead of `latest`

General upkeep:

- Now we build on bionic instead of xenial
- Versions of Mono older than 5.16 are dropped, 6.0 through 6.6 are added, and we'll ship the one from 6.6
- NuGet and Cake and Cake.SemVer versions are updated
- net-core version is updated from 2.1 to 3.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: xenial
+dist: bionic
 language: csharp
 
 sudo: required
@@ -10,7 +10,7 @@ services:
 
 env:
     global:
-        - BUILD_RELEASE_MONO_VERSION=5.16.0
+        - BUILD_RELEASE_MONO_VERSION=6.6.0
         - DOCKERHUB_USERNAME=kspckanbuilder
         - secure: "UGXJG9jB9tGwjJXxG0Beu4Poz0leuiCBItnfTTKRm1a/NxXf1eoOH9p9icY5Ur2xHwqh0uWSznuL2aNr58CXtzTcMpXrcUhRsO63FB3Cz6tSdZEb+pKQJ23zmC9929DklKRGUT2D/eBcJcnV+/eAtkarrfLBU1avUQAVJgqXvMI="
         - AWS_DEFAULT_REGION=us-west-2
@@ -21,19 +21,21 @@ env:
         - BUILD_CONFIGURATION=Release
 
 mono:
+    - 6.6.0
+    - 6.4.0
+    - 6.0.0
     - 5.20.1
     - 5.16.0
-    - 5.14.0
-    - 5.12.0
 
 matrix:
     include:
         - env: BUILD_CONFIGURATION=Debug_NetCore
-          mono: latest
-          dotnet: 2.1.802
+          # Cake is broken on Mono 6.8.0
+          mono: 6.6.0
+          dotnet: 3.1.1
         - env: BUILD_CONFIGURATION=Release_NetCore
-          mono: latest
-          dotnet: 2.1.802
+          mono: 6.6.0
+          dotnet: 3.1.1
 
 addons:
     apt:
--- a/Tests/Tests.csproj
+++ b/Tests/Tests.csproj
@@ -9,7 +9,7 @@
   <Choose>
     <When Condition="$(Configuration.EndsWith('NetCore'))">
       <PropertyGroup>
-        <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
+        <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
       </PropertyGroup>
     </When>
     <Otherwise>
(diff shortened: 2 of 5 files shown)


DasSkelett:
I guess you tested on Linux? I currently don't have access to my Linux machine, but on Windows it works fine for me after updating the VS build tools (had no NetCore 3.1 installed yet).

HebaruSan:
I can't test yet, I still have Mono 6.8 and can't compile anything. Working on a downgrade attempt now.

DasSkelett:
I think last time I tried to downgrade Mono I just went with `apt remove mono-complete` and `apt install mono-complete=6.6` in the end.

HebaruSan:
```
The following packages have unmet dependencies:
 mono-complete : Depends: mono-runtime (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: mono-runtime-sgen (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: mono-llvm-support (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: libmono-2.0-1 (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: libmono-profiler (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: mono-utils (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: mono-devel (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: mono-mcs (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: mono-roslyn (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: mono-csharp-shell (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: mono-4.0-gac (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: mono-4.0-service (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: monodoc-base (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: monodoc-manual (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: libmono-cil-dev (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
                 Depends: ca-certificates-mono (= 6.6.0.161-0xamarin7+ubuntu1804b1) but 6.8.0.96-0xamarin4+ubuntu1804b1 is to be installed
E: Unable to correct problems, you have held broken packages.
```

HebaruSan:
(I have **zero** held packages, broken or otherwise.)

DasSkelett:
I remember something similar.
It's kinda funny that our dependency resolution system works better than apt's.

Typing out every dependency with the right version could work (but is not optimal, an `apt-mark manual` afterwards might be useful).

HebaruSan:
OK, I managed to restore a Mono 6.6 install, and I'm able to compile and run CKAN.

HebaruSan:
I'll give @Olympic1 another few days to weigh in, in case I've missed something important about these versions.

politas:
I'm still away from home. Will test on Manjaro when I get home.

PR #2966, opened 2020-01-18
Move GUI files into subfolders, split controls out from Main

HebaruSan:
## Motivation and problems

Lots of annoyances in GUI code:

- The `CKAN/GUI` source folder is almost entirely unorganized, with all GUI files just dumped there together, which makes it hard to find a given class
- The `Main` class includes lots of controls and code that belong in independent controls, which can make it hard to find a given control (or even remember which ones are or aren't in `Main`)
- `GUI.user` is used as a global variable, but it will not work properly if `Main.Instance` isn't set, which is bad
- Similarly, `GUIUser` relied on `Main.Instance` for most of its functions, which is not guaranteed to be ready to fulfill them
- GUI and ConsoleUI each have their own separate logic for recommendations and suggestions
- `MainModInfo` and `MainAllModVersions` are *not* part of the `Main` class, but they have `Main` in their names anyway
- `ManageKspInstances` is the only dialog whose name doesn't end in `Dialog`
- `HintTextBox` has a `.Designer.cs` file, but it also created and added a control in its constructor outside that file, which isn't how that's supposed to work
- The recommendations tab uses `Monitor.Wait` and `Monitor.Pulse` to signal completion, whereas the provides tab uses `TaskCompletionSource`, even though both are used for the same purpose

## Changes

- New subdirs of GUI:
  - Model - files that just hold data for other classes to use
  - Main - files for the `Main` class
  - Controls - files providing `UserControl`s
  - Dialogs - files providing `Form`s (except `Main`, which is special)
  - Plugins - logic for CKAN plugins
- New controls broken out of `Main` (into the `Controls` directory):
  - `Changeset` - the change set tab
  - `ChooseProvidedMods` - the tab for choosing provided mods
  - `ChooseRecommendedMods` - the tab for choosing recommendations and suggestions
  - `Wait` - the tab for the progress bar
    - This required refactoring `ModInfo`'s downloading code; it now fires an `event` when you click the download button, which triggers a callback in `Main` that drives `Wait`
- `GUI.user` is deleted and replaced by `Main.currentUser`, passed around as needed
- `GUIUser`'s constructor now accepts as parameters the two objects it needs to provide its functions, `Main` and `Wait`
- GUI's recommendations business logic is moved to Core's `MainInstaller.FindRecommendations` and used by both GUI and ConsoleUI
- Behold, `ModInfo`, `AllModVersions`, and `ManageKspInstancesDialog`
- `HintTextBox.Designer.cs` sets up all the contained controls
- The recommendations tab is switched over to `TaskCompletionSource`

The files are now easier to browse, `Main` is smaller and easier to browse, and several important chunks of UI are encapsulated and provide their functionality via explicit public interfaces rather than having their controls changed from all over the place.

There should be no functional changes here other than ConsoleUI now including "Supported By" modules thanks to the shared `FindRecommendations` function. There are some subtle visual changes in the new controls to make them work like the `DeleteDirectories` control with respect to layout/scaling (since the `Anchor` property doesn't work well in `UserControl`s).

### Considered and not done, possibly for future

- Break the mod list out into a control, would be cleaner but probably really complicated
- Share `FindRecommendations` with CmdLine


HebaruSan:
The build error is the one we're fixing in #2964.

PR #2969, opened 2020-01-23
Fix CKAN-installed modules shown as AD in some cases

### LINKED ISSUE #2262: Mod Installations being forced to AS, Autodetect

### LINKED ISSUE #2864: [Bug] All mods installed via CKAN listed as AD
Background
----------
* Operating System:  Windows 10 Home 1903 18362.295
* CKAN Version: v1.26.4
* KSP Version: 1.7.3

No manual changes to Gamedata folder

Problem
-------
I have installed 20 mods with CKAN, but after installation CKAN lists all of them as AD.

**Steps to reproduce**
Select a mod
Install said mod
No error message, but the mod will be installed and listed as AD

**Expected behavior**
Select a mod
Install said mod
Able to unistall said mod with CKAN

**Screenshots**
image
All of these were installed manually.

### LINKED ISSUE #2967: [Bug] All mods installed via CKAN immediately listed as AD
Background
----------
* Operating System: Windows 10 Home build 18363.592
* CKAN Version: 1.26.6
* KSP Version: 1.8.1.2694 (with both expansions installed)

No manual changes to GameData.

Problem
-------
Upon installing any mod via CKAN, it immediately gets marked as "AD" in the interface, as if it were an autodetected, manually-installed mod.

**Steps to reproduce**

1. Install CKAN.exe in the KSP game install folder.
2. Start CKAN, open File > Manage KSP Instances; this will list no instances.
3. Add the KSP instance where the CKAN.exe file resides as a new instance and set it as default.
4. Use CKAN to install any mod.
5. Mod will install without error, but will immediately be marked "AD".

**Expected behavior**

CKAN-installed mods should be marked appropriately.

HebaruSan:
### Switching installs A-B-A

1. In GUI, start with or switch to any KSP install A
2. Switch to install B
3. Switch back to install A
4. Install modules
5. The modules will appear as either not installed (if they have no DLL) or as "AD" as if they had been manually installed (if they have a DLL)

### Adding portable install to registry

1. Put ckan.exe in the root folder for a KSP install
2. Run that CKAN to work with that KSP as a "portable" instance
3. Open the Manage Instances window and add the portable install's path as a permanent registered instance
4. Switch to it
5. There will be a weird delay when you click a mod list row as the same registry data is re-loaded
6. Install modules
7. The modules will appear as either not installed (if they have no DLL) or as "AD" as if they had been manually installed (if they have a DLL)

### Path case sensitivity

1. On a Linux or Mac system, create two KSP installs that have the same path except for the use of capital and lowercase letters. For example, `/my/games/ksp` and `/my/games/KSP`.
2. Add both to CKAN
3. Install something in each instance
4. Everything you install will be installed to only **one** of the instances, regardless of which one you thought you were in at the time

## Causes

#1828 updated `KSPManager.SetCurrentInstance` to dispose `RegistryManager` instances when switching to a new instance. This releases the registry lock, which makes sense when we're done working with an instance. It also removes the object from the instance cache, so if we try to retrieve a `RegistryManager` later for the same path, a new object is created.

`ModuleInstaller` held a persistent reference to a `RegistryManager` that it acquired itself, and it kept this reference even after the `RegistryManager` object was disposed. So if you asked it to do something with that instance later, it would notify the wrong `Registry` object of its changes, which would not be reflected in the UI or saved to disk. (Technically, the changed registry is saved to disk and then overwritten by the unchanged one, but :man_shrugging: .)

`KSPManager.SetCurrentInstance` disposes the previous current instance even if it has the same path as the new one. This results in an unnecessary unlocking, re-locking, and re-loading of a registry object that will not have changed.

`ModuleInstaller.GetInstance`, uniquely among all of CKAN's instance retrieval functions, casts its path keys to lowercase before using them. This means that `ModuleInstaller.GetInstance("/my/games/ksp")` and `ModuleInstaller.GetInstance("/my/games/KSP")` return the same object. This is not appropriate in a cross-platform application.

## Changes

Now `ModuleInstaller` doesn't retain a persistent reference to `RegistryManager`; instead, each of its functions that requires a `RegistryManager` or a `Registry` now accepts those objects as a parameter. This more clearly reflects the functional dependencies of these operations, and passes responsibility to the calling code to make sure these objects are up to date. Lots of calling code is updated to pass these new parameters.

In addition, `KSPManager.SetCurrentInstance` will no longer dispose an old `RegistryManager` if it shares its gamedir with the new instance. This will eliminate the delay when switching from a portable install to the same folder as a permanent install.

Finally, `ModuleInstaller.instances` is no longer treats its keys as case insensitive. This will fix the case sensitivity problem on Unix-like systems, and no ill effects are expected because `RegistryManager.Instance` already works this way.

Fixes #2262.
Fixes #2864.
Fixes #2967.

### DIFF
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -127,7 +127,8 @@ namespace CKAN.CmdLine
                 install_ops.without_enforce_consistency = true;
             }
 
-            IRegistryQuerier registry = RegistryManager.Instance(ksp).registry;
+            RegistryManager regMgr = RegistryManager.Instance(ksp);
+            IRegistryQuerier registry = regMgr.registry;
             List<string> modules = options.modules;
 
             for (bool done = false; !done; )
@@ -136,7 +137,7 @@ namespace CKAN.CmdLine
                 try
                 {
                     var installer = ModuleInstaller.GetInstance(ksp, manager.Cache, user);
-                    installer.InstallList(modules, install_ops);
+                    installer.InstallList(modules, install_ops, regMgr);
                     user.RaiseMessage("\r\n");
                     done = true;
                 }
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -128,7 +128,6 @@ namespace CKAN.CmdLine
             }
 
             RegistryManager regMgr = RegistryManager.Instance(ksp);
-            IRegistryQuerier registry = regMgr.registry;
             List<string> modules = options.modules;
 
             for (bool done = false; !done; )
--- a/Core/KSP.cs
+++ b/Core/KSP.cs
@@ -492,7 +492,7 @@ namespace CKAN
 
         public bool Equals(KSP other)
         {
-            return other != null ? gameDir.Equals(other.GameDir()) : base.Equals(other);
+            return other != null && gameDir.Equals(other.GameDir());
         }
 
         public override bool Equals(object obj)
--- a/Core/KSP.cs
+++ b/Core/KSP.cs
@@ -21,7 +21,7 @@ namespace CKAN
     /// <summary>
     /// Everything for dealing with KSP itself.
     /// </summary>
-    public class KSP
+    public class KSP : IEquatable<KSP>
     {
         /// <summary>
         /// List of DLLs that should never be added to the autodetect list.
@@ -490,10 +490,14 @@ namespace CKAN
             return "KSP Install: " + gameDir;
         }
 
+        public bool Equals(KSP other)
+        {
+            return other != null ? gameDir.Equals(other.GameDir()) : base.Equals(other);
+        }
+
         public override bool Equals(object obj)
         {
-            var other = obj as KSP;
-            return other != null ? gameDir.Equals(other.GameDir()) : base.Equals(obj);
+            return Equals(obj as KSP);
         }
 
         public override int GetHashCode()
--- a/Cmdline/Action/Import.cs
+++ b/Cmdline/Action/Import.cs
@@ -46,13 +46,15 @@ namespace CKAN.CmdLine
                 {
                     log.InfoFormat("Importing {0} files", toImport.Count);
                     List<string>    toInstall = new List<string>();
+                    RegistryManager regMgr = RegistryManager.Instance(ksp);
                     ModuleInstaller inst      = ModuleInstaller.GetInstance(ksp, manager.Cache, user);
-                    inst.ImportFiles(toImport, user, mod => toInstall.Add(mod.identifier), !opts.Headless);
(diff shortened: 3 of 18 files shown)


Cmdline/Action/Install.cs
```diff
@@ -127,7 +127,8 @@ public int RunCommand(CKAN.KSP ksp, object raw_options)
                 install_ops.without_enforce_consistency = true;
             }
 
-            IRegistryQuerier registry = RegistryManager.Instance(ksp).registry;
+            RegistryManager regMgr = RegistryManager.Instance(ksp);
+            IRegistryQuerier registry = regMgr.registry;
```
DasSkelett: This registry object is unused, I think you can remove it again.

Core/KSP.cs
```diff
@@ -490,10 +490,14 @@ public override string ToString()
             return "KSP Install: " + gameDir;
         }
 
+        public bool Equals(KSP other)
+        {
+            return other != null ? gameDir.Equals(other.GameDir()) : base.Equals(other);
```
DasSkelett: This can be simplified, no?
If `other` is `null`, the last part could only possibly be true if `this` is also `null`,
and a `NullReferenceException` should've already been thrown by the runtime when trying to call `.Equals()` on it.

So you could also write
`return other != null && gameDir.Equals(other.GameDir())`

HebaruSan: OK, you convinced me. I wanted to be careful with this because the `base.Equals` call looked complicated, but it makes sense to simplify it.

DasSkelett (approved):
Good to see this weird bug finally being resolved. Nice investigation work!

Katikar525:
I am experiencing this bug with version 1.34 after using the auto-update and conducting A-B tests (I wasn't aware that was an issue) I updated CKAN to 1.36.4 and have tried restarting CKAN, deleting the CKAN folder in both directories, and downloading and installing an entirely new version, but nothing fixes the problem. What is the process for resetting the registry and forcing it to recognize the installed mods?

image

HebaruSan:
Hi @Katikar525,  sorry you're having problems. But it's not this one; the bug you're commenting on was fixed over six years ago. I would recommend submitting a new issue rather than commenting on an old pull request.

PR #2970, opened 2020-01-24
Stage modules with hardcoded game versions

HebaruSan:
## Motivation

Currently we have 29 netkans with `x_netkan_staging` enabled because the KSP version compatibility info is hard coded in the netkan, and hence might be out of date anytime a new version is released. This has been a bit tedious to set manually for each such netkan, and it is pretty much the same thing every time, hence it would be nice to automate.

## Changes

Now a new `StagingTransformer` does this automatically. It runs first so it can see the raw netkan data without interference by other transformers, and if it finds `ksp_version`, `ksp_version_min`, or `ksp_version_max`, it enables staging and sets the reason to a string explaining what keys were found, their values, and how to proceed. This way, **any** netkan with hard coded compatibility info will automatically be staged every time, and we can remove the staging properties from the netkans that have it.


techman83:
That's really neat!

PR #2971, opened 2020-01-24
Edit modpack before export

### LINKED ISSUE #1375: Priority for mods
A priority list, to let you set what mods are most important to you.  So you can mark a mod as highest priority, and when you make a mod pack link or whatever (haven't tried yet with ckan) to share your pack, it shows the other users the priority as well.  I'll show my old spreadsheet as an example (half not updated but the priority is intact)
https://docs.google.com/spreadsheets/d/1PCbSwgSAueii_4LpiTorgmkuDpO61Vuo-uvRqhICBVo/edit

HebaruSan:
## Background

Currently the File menu has an option called "Export installed mods..." that brings up a save window with several options:

image

image

Two of the options are for .ckan files, with the "modpack" option including mod versions and the favorites/favourites option omitting them. Also one of them treats the mods as dependencies and the other as recommendations; which is which? I don't know and neither do you, because that UI doesn't mention this!

The generated metadata of the .ckan options looks something like this:

```json
{
        "kind": "metapackage",
        "abstract": "A list of modules installed on the ksp-1.8.1 KSP instance",
        "name": "installed-ksp-1.8.1",
        "license": "unknown",
        "version": "2020.01.24.09.25.48",
        "identifier": "installed-ksp-1.8.1",
        "spec_version": "v1.6",
        "recommends": [
                {
                        "name": "ActionGroupManager"
                },
                {
                        "name": "AdjustableModPanel"
                },
                {
                        "name": "AGExt"
                },
                {
                        "name": "AllAboard"
                },
                {
                        "name": "xScienceContinued"
                },
                {
                        "name": "ToolbarController"
                },
                {
                        "name": "ClickThroughBlocker"
                },
                {
                        "name": "ModuleManager"
                }
        ]
}
```

## Motivation

This is kind of confusing and limited.

- What if I want to change whether mod versions are saved without changing whether they're depends/recommends?
- What if I want some mods to be depends, others to be recommends, etc.?
- What if I want to edit the other fields that are saved to my modpack file?
- What if I don't even notice the file type dropdown and assume there's only one option for exporting?
- My user tester felt that "Export installed mods..." sounded like you would be saving *the mods themselves* somehow (imagine a big ZIP of your whole GameData, I guess) rather than just a text listing of them

## Changes

Now the old menu option is split and replaced:

image

"Save installed mod list..."  opens the save prompt with the non-.ckan options available:

image

"Export modpack..." opens a new Edit Modpack tab with controls to edit the stuff that we save to .ckan files:

image

- You can click one of the mods in the central listview to see its mod info at the right
- Then you can click one of the buttons at the bottom left to move it to another section (by default they're all in Depends), with "Ignored" mods being left out of the saved file completely
- Multi-select works (shift- and ctrl-click), so you don't have to click many many times when editing large modpacks
- The "Save mod versions" checkbox determines whether mods' versions are saved to the relationships
- The Export button brings up a save prompt to save your changes to a file and close the tab

To achieve this:

- `RegistryManager.GenerateModpack` creates a `CkanModule` corresponding to the metadata that goes into a .ckan file by default
- A new `EditModpack` control handles editing a `CkanModule` and saving it to disk
- `CkanModule.ToJson` generates a nice JSON string, with the order of properties updated to match `NetKAN.Transformers.PropertySortTransformer`

Fixes #1375.

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -329,7 +329,7 @@
         "identifier" : {
             "description" : "Unique identifier for this mod.",
             "type"        : "string",
-            "pattern"     : "^[A-Za-z0-9-]*$"
+            "pattern"     : "^[A-Za-z0-9][A-Za-z0-9-]+$"
         },
         "version" : {
             "description" : "A version string",
--- a/Core/Exporters/DelimeterSeparatedValueExporter.cs
+++ b/Core/Exporters/DelimeterSeparatedValueExporter.cs
@@ -4,23 +4,23 @@ using System.Linq;
 
 namespace CKAN.Exporters
 {
-    public sealed class DelimeterSeperatedValueExporter : IExporter
+    public sealed class DelimeterSeparatedValueExporter : IExporter
     {
         private const string WritePattern = "{1}{0}{2}{0}{3}{0}{4}{0}{5}" +
                                             "{0}{6}{0}{7}{0}{8}{0}{9}{0}{10}" +
                                             "{0}{11}{0}{12}{0}{13}{0}{14}{0}{15}" +
                                             "{0}{16}{0}{17}{0}{18}";
-        private readonly string _delimter;
+        private readonly string _delimeter;
 
-        public DelimeterSeperatedValueExporter(Delimter delimter)
+        public DelimeterSeparatedValueExporter(Delimeter delimeter)
         {
-            switch (delimter)
+            switch (delimeter)
             {
-                case Delimter.Comma:
-                    _delimter = ",";
+                case Delimeter.Comma:
+                    _delimeter = ",";
                     break;
-                case Delimter.Tab:
-                    _delimter = "\t";
+                case Delimeter.Tab:
+                    _delimeter = "\t";
                     break;
                 default:
                     throw new ArgumentOutOfRangeException();
@@ -32,7 +32,7 @@ namespace CKAN.Exporters
             using (var writer = new StreamWriter(stream))
             {
                 writer.WriteLine(WritePattern,
-                    _delimter,
+                    _delimeter,
                     "identifier",
                     "version",
                     "name",
@@ -57,7 +57,7 @@ namespace CKAN.Exporters
                 foreach (var mod in registry.InstalledModules.OrderBy(i => i.Module.name))
                 {
                     writer.WriteLine(WritePattern,
-                        _delimter,
+                        _delimeter,
                         mod.Module.identifier,
                         mod.Module.version,
                         QuoteIfNecessary(mod.Module.name),
@@ -139,7 +139,7 @@ namespace CKAN.Exporters
 
         private string QuoteIfNecessary(string value)
         {
-            if (value != null && value.IndexOf(_delimter, StringComparison.Ordinal) >= 0)
+            if (value != null && value.IndexOf(_delimeter, StringComparison.Ordinal) >= 0)
             {
                 return "\"" + value + "\"";
             }
@@ -149,12 +149,11 @@ namespace CKAN.Exporters
             }
         }
 
-        public enum Delimter
+        public enum Delimeter
         {
             Comma,
             Tab
         }
-
 
     }
 }
--- a/Core/Exporters/Exporter.cs
+++ b/Core/Exporters/Exporter.cs
@@ -30,10 +30,10 @@ namespace CKAN.Exporters
                     _exporter = new BbCodeExporter();
                     break;
                 case ExportFileType.Csv:
-                    _exporter = new DelimeterSeperatedValueExporter(DelimeterSeperatedValueExporter.Delimter.Comma);
+                    _exporter = new DelimeterSeparatedValueExporter(DelimeterSeparatedValueExporter.Delimeter.Comma);
                     break;
                 case ExportFileType.Tsv:
-                    _exporter = new DelimeterSeperatedValueExporter(DelimeterSeperatedValueExporter.Delimter.Tab);
+                    _exporter = new DelimeterSeparatedValueExporter(DelimeterSeparatedValueExporter.Delimeter.Tab);
                     break;
                 default:
                     throw new ArgumentOutOfRangeException();
--- a/Core/Registry/RegistryManager.cs
+++ b/Core/Registry/RegistryManager.cs
@@ -270,18 +270,6 @@ namespace CKAN
             }
         }
 
-        /// <summary>
-        /// Returns the currently installed modules in json format suitable for outputting to a ckan file.
-        /// Defaults to using depends and with version numbers.
-        /// </summary>
-        /// <param name="recommmends">If the json should use a recommends field instead of depends</param>
-        /// <param name="with_versions">If version numbers should be included</param>
-        /// <returns>String containing a valid ckan file</returns>
-        public string CurrentInstallAsCKAN(bool recommmends, bool with_versions)
-        {
-            return SerializeCurrentInstall(recommmends, with_versions);
-        }
-
         private void Load()
         {
             // Our registry needs to know our KSP install when upgrading from older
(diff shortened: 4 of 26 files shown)


DasSkelett:
A few things:
* The vertical scroll bar is buggy on Mono. It is only shown (but not usable) as long as no mod is selected. As soon as a mod is selected, it becomes invisible with glitchy texture artifacts:
grafik
I guess the scollbar needs to be explicit enabled in the code? Setting `this.RelationshipsListView.Scrollable = true` doesn't work though.

* This made me look how we are handling it for the recommendation UserCcontrol, and well, there's no scrollbar at all (simple to test with the recommendation audit option). I think we could need one there too. I suspect the solution will be the same for the two controls.

* 1) File > Export Modpack
  2) File > Audit Recommendations.
  3) Cancel on the recommendation audit tab
  4) Experience this nice crash:
```
/usr/bin/mono /GIT/CKAN/_build/out/CKAN-GUI/Debug/bin/CKAN-GUI.exe
System.ArgumentOutOfRangeException: ControlCollection does not have that many controls
Parameter name: index
Actual value was 2.
  at System.Windows.Forms.Control+ControlCollection.get_Item (System.Int32 index) [0x00027] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 
  at System.Windows.Forms.TabControl.GetTab (System.Int32 index) [0x00006] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 
  at System.Windows.Forms.TabControl.GetTabRect (System.Int32 index) [0x00000] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 
  at (wrapper remoting-invoke-with-check) System.Windows.Forms.TabControl.GetTabRect(int)
  at System.Windows.Forms.Theming.Default.TabControlPainter.Draw (System.Drawing.Graphics dc, System.Drawing.Rectangle area, System.Windows.Forms.TabControl tab) [0x000df] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 
  at System.Windows.Forms.ThemeWin32Classic.DrawTabControl (System.Drawing.Graphics dc, System.Drawing.Rectangle area, System.Windows.Forms.TabControl tab) [0x0000a] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 
  at System.Windows.Forms.TabControl.Draw (System.Drawing.Graphics dc, System.Drawing.Rectangle clip) [0x00005] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 
  at System.Windows.Forms.TabControl.OnPaintInternal (System.Windows.Forms.PaintEventArgs pe) [0x00017] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 
  at System.Windows.Forms.Control.WmPaint (System.Windows.Forms.Message& m) [0x0006d] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 
  at System.Windows.Forms.Control.WndProc (System.Windows.Forms.Message& m) [0x001a4] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 
  at System.Windows.Forms.TabControl.WndProc (System.Windows.Forms.Message& m) [0x00057] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 
  at System.Windows.Forms.Control+ControlWindowTarget.OnMessage (System.Windows.Forms.Message& m) [0x00000] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 
  at System.Windows.Forms.Control+ControlNativeWindow.WndProc (System.Windows.Forms.Message& m) [0x0000b] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 
  at System.Windows.Forms.NativeWindow.WndProc (System.IntPtr hWnd, System.Windows.Forms.Msg msg, System.IntPtr wParam, System.IntPtr lParam) [0x00085] in <1ad3b9c5e8484f3cb4aaaf843739b786>:0 

Process finished with exit code 0.
```
Warning! Do not attach debugger while doing this, it makes your DE freeze :slightly_frowning_face: 

I think the solution for this could be as simple as disabling the top menu bar while the modpack view is enabled, like we do during mod installation. "Direct" tab switching is already disabled anyway.
Also consider the same for the recommendation control?

HebaruSan:
Pushed a fix for the crash, I agree with disabling the menu since it's done in other places.
I'll reboot in an hour or two to check out Mono.

HebaruSan:
Looks like the Mono scrollbar problem relates to threading; added some `Util.Invoke` calls to make sure we're doing GUI ops on the right thread. This seemed to clear up the EditModpack problems for me, but the recommendations tab still has some glitchiness that I could not eliminate.

DasSkelett:
Does it make sense to allow setting the KSP version min/max when the mod versions are saved? In the end it's these mod versions that restrict it, and setting the min/max too restrictive (which can happen easily when you don't check the compatibility data for each mod) can prevent the modpack from being installable.

Also I wouldn't set any <Relationship>RadioButton if mods of different relationship groups are selected. It feels like that's what other UIs would do too.

I couldn't find a way to incorporate this into the switch statement, I tried something where `kinds.Select(k => k.Name).ToList()` would be switched and each case would be something like this `{"DependsGroup"}`, so multiple entries in one list would fall under default.
Instead I could get it to work with an enclosing if/else:
```diff
diff --git a/GUI/Controls/EditModpack.cs b/GUI/Controls/EditModpack.cs
index e08be9f2..18502692 100644
--- a/GUI/Controls/EditModpack.cs
+++ b/GUI/Controls/EditModpack.cs
@@ -104,7 +104,7 @@ private void LoadRelationships(IRegistryQuerier registry)
             GroupToRelationships.Add(RecommendationsGroup, module.recommends);
             GroupToRelationships.Add(SuggestionsGroup,     module.suggests);
             GroupToRelationships.Add(IgnoredGroup,         ignored);
-            
+
             RelationshipsListView_ItemSelectionChanged(null, null);
         }
 
@@ -187,26 +187,37 @@ private void RelationshipsListView_ItemSelectionChanged(Object sender, ListViewI
                 .Select(lvi => lvi.Group)
                 .Distinct()
                 .ToList();
-            switch (kinds.FirstOrDefault()?.Name)
-            {
-                case "DependsGroup":
-                    DependsRadioButton.Checked = true;
-                    break;
-                case "RecommendationsGroup":
-                    RecommendsRadioButton.Checked = true;
-                    break;
-                case "SuggestionsGroup":
-                    SuggestsRadioButton.Checked = true;
-                    break;
-                case "IgnoredGroup":
-                    IgnoreRadioButton.Checked = true;
-                    break;
-                default:
-                    DependsRadioButton.Checked    = false;
-                    RecommendsRadioButton.Checked = false;
-                    SuggestsRadioButton.Checked   = false;
-                    IgnoreRadioButton.Checked     = false;
-                    break;
+
+            if (kinds.Count > 1)
+            {
+                DependsRadioButton.Checked    = false;
+                RecommendsRadioButton.Checked = false;
+                SuggestsRadioButton.Checked   = false;
+                IgnoreRadioButton.Checked     = false;
+            }
+            else
+            {
+                switch (kinds.FirstOrDefault()?.Name)
+                {
+                    case "DependsGroup":
+                        DependsRadioButton.Checked = true;
+                        break;
+                    case "RecommendationsGroup":
+                        RecommendsRadioButton.Checked = true;
+                        break;
+                    case "SuggestionsGroup":
+                        SuggestsRadioButton.Checked = true;
+                        break;
+                    case "IgnoredGroup":
+                        IgnoreRadioButton.Checked = true;
+                        break;
+                    default:
+                        DependsRadioButton.Checked    = false;
+                        RecommendsRadioButton.Checked = false;
+                        SuggestsRadioButton.Checked   = false;
+                        IgnoreRadioButton.Checked     = false;
+                        break;
+                }
             }
             if (RelationshipsListView.SelectedItems.Count > 0)
             {
```

And a third one: it would be neat if after you switch the relationship type of a mod, the focus would return to the module list. It's kind of annoying that you have to click on it first so you can scroll again.
Maybe also focus the list after opening the tab.

HebaruSan:
> Does it make sense to allow setting the KSP version min/max when the mod versions are saved?

Well, some mod versions have wide compatibility ranges. I might want to release a modpack for 1.7.3 where the constituent mods work on 1.5 through 1.7. I'd prefer to allow an overly obsessive user to shoot him- or herself in the foot rather than preventing something that might be needed only rarely in a special circumstance.

> Also I wouldn't set any RadioButton if mods of different relationship groups are selected. It feels like that's what other UIs would do too.

OK, I'll change that.

> And a third one: it would be neat if after you switch the relationship type of a mod, the focus would return to the module list. It's kind of annoying that you have to click on it first so you can scroll again.
> Maybe also focus the list after opening the tab.

Excellent points, will definitely work on that.

DasSkelett:
> > Does it make sense to allow setting the KSP version min/max when the mod versions are saved?
> 
> Well, some mod versions have wide compatibility ranges. I might want to release a modpack for 1.7.3 where the constituent mods work on 1.5 through 1.7. I'd prefer to allow an overly obsessive user to shoot him- or herself in the foot rather than preventing something that might be needed only rarely in a special circumstance.

Good reason.

Will the changes in `CKANModule` trigger a big re-indexation wave again?
It looks like we are fine again on the infra side with https://github.com/KSP-CKAN/NetKAN-Infra/pull/125, but we maybe should check with @techman83 if the credits have already replenished enough.

Otherwise I'm done reviewing the code, looks good. Will finish up the translation now.

HebaruSan:
> Will the changes in CKANModule trigger a big re-indexation wave again?

No; for better or worse, NetKAN does not use `CkanModule` to generate metadata.

PR #2976, opened 2020-01-28
Downgrade to building on Mono 5.20

HebaruSan:
## Problem

The latest release (1.26.8) on Windows:

```
Unhandled Exception: System.ArgumentException: Object of type 'System.Globalization.CalendarId[]' cannot be converted to type 'System.Int32[]'.
   at System.RuntimeType.TryChangeType(Object value, Binder binder, CultureInfo culture, Boolean needsSpecialCast)
   at System.RuntimeType.CheckValue(Object value, Binder binder, CultureInfo culture, BindingFlags invokeAttr)
   at System.Reflection.RtFieldInfo.UnsafeSetValue(Object obj, Object value, BindingFlags invokeAttr, Binder binder, CultureInfo culture)
   at System.Runtime.Serialization.FormatterServices.SerializationSetValue(MemberInfo fi, Object target, Object value)
   at System.Runtime.Serialization.ObjectManager.CompleteObject(ObjectHolder holder, Boolean bObjectFullyComplete)
   at System.Runtime.Serialization.ObjectManager.DoNewlyRegisteredObjectFixups(ObjectHolder holder)
   at System.Runtime.Serialization.ObjectManager.RegisterObject(Object obj, Int64 objectID, SerializationInfo info, Int64 idOfContainingObj, MemberInfo member, Int32[] arrayIndex)
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.RegisterObject(Object obj, ParseRecord pr, ParseRecord objectPr, Boolean bIsString)
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.ParseObjectEnd(ParseRecord pr)
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.Parse(ParseRecord pr)
   at System.Runtime.Serialization.Formatters.Binary.__BinaryParser.Run()
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.Deserialize(HeaderHandler handler, __BinaryParser serParser, Boolean fCheck, Boolean isCrossAppDomain, IMethodCallMessage methodCallMessage)
   at System.Runtime.Serialization.Formatters.Binary.BinaryFormatter.Deserialize(Stream serializationStream, HeaderHandler handler, Boolean fCheck, Boolean isCrossAppDomain, IMethodCallMessage methodCallMessage)
   at System.Resources.ResourceReader.DeserializeObject(Int32 typeIndex)
[...]

PR #2985, opened 2020-02-04
Update Cake to 0.37.0

DasSkelett:
## Problem
You currently can't build with Cake on Mono 6.8, due to a complex issue I don't want to explain here, you can read it up in https://github.com/cake-build/cake/issues/2695 and all the other issues that are linked there :P
Cake 0.37.0 fixed this.

## Changes
This PR updates the Cake dependency from 0.36.0 to 0.37.0, now that it is released. This allows us to build CKAN using the build script on Mono 6.8 again.
(I guess Kate also automatically removed a BOM)

Note that I don't change back the Mono versions that we're using for the NetCore build on Travis. There's still the other problem, which makes Mono 6 compiled binaries not work on Windows or Mono 5, see #2976. Although we're not using that for anything release-related, let's play it conservatively for now.

I tested the compilation on Mono 6.8 and the exe runs.
We should also test compilation on Windows and running a Mono-compiled exe on Windows before merging, so we cover all the issues we encountered the last days.
I'm going to do this myself tomorrow, too.

### DIFF
--- a/packages.config
+++ b/packages.config
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Cake" version="0.36.0" />
+  <package id="Cake" version="0.37.0" />
 </packages>


DasSkelett:
Windows-compiled binaries work on Windows, so do Mono5-compiled ones.
Unfortunately, Mono6-compiled binaries still don't work on Windows, but I didn't expect that.

DasSkelett:
Rebase on current master@10c2887a

PR #2995, opened 2020-02-08
Don't report an unknown error if it is known

DasSkelett:
## Problem
When there's an exception thrown during the installation of mods, we catch the usual ones and log an explanation message on the WaitTabPage.
However in `PostInstallMods()` which is called when the background worker completes `InstallMods()`, the message `An unknown error occurred, please try again!` is written to the "log".
This message is actually not true most of the time, because e.g. download errors or inconsistency problems are all catched and a nice explaining error message has already been printed out.

## Changes
Now `PostInstallMods()` distinguishes catched and uncatched errors (the second one will result in `RunWorkerCompletedEventArgs.Error` being set).
It keeps the old message for uncatched errors, but displays a new one for known ones:
```
If the above message indicates a download error, please try again. Otherwise, please open a issue for us to investigate.
```
(Suggestions to change this message are welcome)

This way we don't contradict ourselves, and also encourage the users to report metadata/client problems.

### DIFF
--- a/GUI/Main/MainInstall.cs
+++ b/GUI/Main/MainInstall.cs
@@ -12,8 +12,8 @@ namespace CKAN
     {
         private BackgroundWorker installWorker;
 
-        // used to signal the install worker that the user canceled the install process
-        // this may happen on the recommended/suggested mods dialogs
+        // Used to signal the install worker that the user canceled the install process.
+        // This may happen on the recommended/suggested mods dialogs or during the download.
         private volatile bool installCanceled;
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


DasSkelett:
Easy way to test this PR: Block github.com and archive.org in your hosts file (works on Windows too).

GUI/Main/MainInstall.cs
```diff
@@ -417,16 +417,26 @@ private void PostInstallMods(object sender, RunWorkerCompletedEventArgs e)
                     );
                 }
             }
-            else
+            else if (e.Error != null)
```
politas: Can you reverse this test and swap the then and else around? Always better to avoid a negative test for code readability.

DasSkelett:
> Took me a while to figure out exactly when we fail without `e.Error` set; it's in all the `catch` blocks of `InstallMods` that print an error and `return`. But this makes sense!

Yes indeed. `e.Error` is set by the runtime when an exception inside the worker hasn't been catched.
I was thinking about refactoring the `InstallMods` to not catch the exceptions and let `PostInstallMods` handle it, but I didn't want to mess with it and potentially introduce bugs just for this little output change.

HebaruSan:
That would make more sense, arguably more in line with how `BackgroundWorker` is designed.

PR #2996, opened 2020-02-10
Create skip-releases option for NetKAN

DasSkelett:
## Motivation
#2681 gave us the great option to inflate multiple releases with netkan.exe.
But sometimes you don't want to begin with the latest release, for example if the folder structure has changed in the history of the mod, and you have to adjust the netkan file for the older releases. Or if one release is failing inflation and you have to skip it.
I had to deal with both of these cases in https://github.com/KSP-CKAN/CKAN-meta/pull/1786, so I finally decided to implement this feature.

## Changes
This PR gives netkan.exe a new commandline option, `--skip-releases <int>`.
This commandline option is fed into the `TransformOptions`, which has a new `int? SkipReleases` property.
It is set with the constructor, the queue handler passes `null`.

Each kref source transformer now skips the given number of releases, and starting from this index, the number of releases given with `opts.Releases` are taken.

This feature may be useful for https://github.com/KSP-CKAN/NetKAN-Infra/issues/5, too.

### DIFF
--- a/Netkan/Transformers/CurseTransformer.cs
+++ b/Netkan/Transformers/CurseTransformer.cs
@@ -39,6 +39,10 @@ namespace CKAN.NetKAN.Transformers
                 // Look up our mod on Curse by its Id.
                 var curseMod = _api.GetMod(metadata.Kref.Id);
                 var versions = curseMod.All();
+                if (opts.SkipReleases.HasValue)
+                {
+                    versions = versions.Skip(opts.SkipReleases.Value);
+                }
                 if (opts.Releases.HasValue)
                 {
                     versions = versions.Take(opts.Releases.Value);
--- a/Netkan/Transformers/JenkinsTransformer.cs
+++ b/Netkan/Transformers/JenkinsTransformer.cs
@@ -35,6 +35,10 @@ namespace CKAN.NetKAN.Transformers
                 JenkinsRef jRef = new JenkinsRef(metadata.Kref);
 
                 var versions = _api.GetAllBuilds(jRef, options);
+                if (opts.SkipReleases.HasValue)
+                {
+                    versions = versions.Skip(opts.SkipReleases.Value);
+                }
                 if (opts.Releases.HasValue)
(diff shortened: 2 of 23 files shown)


Netkan/Transformers/CurseTransformer.cs
```diff
@@ -39,6 +39,10 @@ public IEnumerable<Metadata> Transform(Metadata metadata, TransformOptions opts)
                 // Look up our mod on Curse by its Id.
                 var curseMod = _api.GetMod(metadata.Kref.Id);
                 var versions = curseMod.All();
+                if (opts.SkipReleases.HasValue)
+                {
+                    versions = versions.Take(opts.SkipReleases.Value);
```
HebaruSan: Shouldn't this be `Skip` instead of `Take`? If I use something like `--skip-releases 3 --releases 5`, this would reduce the `versions` list to the first three elements before we try to grab 5 of them.

Netkan/Transformers/JenkinsTransformer.cs
```diff
@@ -35,6 +35,10 @@ public IEnumerable<Metadata> Transform(Metadata metadata, TransformOptions opts)
                 JenkinsRef jRef = new JenkinsRef(metadata.Kref);
 
                 var versions = _api.GetAllBuilds(jRef, options);
+                if (opts.SkipReleases.HasValue)
+                {
+                    versions = versions.Take(opts.SkipReleases.Value);
```
HebaruSan: Same comment re: `Skip` vs `Take`.

Netkan/Transformers/CurseTransformer.cs
DasSkelett: Ugh, that's what you get for only runnning it with SpaceDock and writing tests for GitHub...
Yeah, you're right, this has to be `Skip` of course.

DasSkelett: Changed in both cases :+1:

Issue #2902, opened 2019-10-23
[Bug] "Launch KSP" Button not working


hinneLinks:
Background
----------
* Operating System: Ubuntu 18.04.3 LTS
* CKAN Version: v1.26.4
* KSP Version: 1.8.0.2686

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
No

Problem
-------
Yesterday i opened Steam, and that updated my KSP vrom 1.7 to 1.8. Since then the Button "Launch KSP" in CKAN does nothing, KSP wont open.
For every Button-Click i see this in the console (if i start CKAN via terminal):
`Found path: /data/SteamLibrary/steamapps/common/Kerbal Space Program/./KSP.x86_64`
I also tried turning the log on, there for every click on the Button Messages like this appear for each Mod i've installed:
`2019-10-23 19:24:00 DEBUG CKAN.CkanModule - Testing if SCANsat v18.13 is compatible with KSP [Versions: 1.8.0.2686, 1.8, 1.7, 1.6, 1.5]
`
No other logs are created on the Button-Click. No ERROR or WARN Log Messages at all.
Worked with KSP 1.7.3.2594 I've tried to downgrade KSP via Steam (https://steamcommunity.com/sharedfiles/filedetails/?id=1086279994), but then even 1.7 didn't work anymore. Reinstalling CKAN or removing all Mods didn't help either. 

Starting KSP via Terminal or Steam works fine, including the Mods installed.

**Steps to reproduce**
Open CKAN
Click on "Launch KSP"
*nothing happens*

**Expected behavior**
KSP should start

HebaruSan:
What's in your Settings &rarr; KSP command-line?

![image](https://user-images.githubusercontent.com/1559108/67428291-7735ce00-f5a3-11e9-8c88-276712623507.png)

HebaruSan:
Also, does that file exist on your system?

`/data/SteamLibrary/steamapps/common/Kerbal Space Program/./KSP.x86_64`

hinneLinks:
The command line settings where a good hint, mine where:
`./KSP.x86_64 -single-instance -popupwindow`
I only added the `-popupwindow`.
When I remove the `-single-instance`, KSP starts again clicking the "Launch KSP" Button.

So its not directly a CKAN-Problem. It might be an [Unity Problem](https://issuetracker.unity3d.com/issues/linux-segmentation-fault-when-running-a-built-project-with-single-instance-argument).

[...]

PR #3001, opened 2020-02-12
Extend -single-instance fix to 1.9

DasSkelett:
## Problem
#2902, which was fixed for KSP 1.8 in #2931, still applies to KSP 1.9 unfortunately:
```
Set current directory to /HDD/KubuntuSteamGames/steamapps/common/Kerbal Space Program
Found path: /HDD/KubuntuSteamGames/steamapps/common/Kerbal Space Program/KSP.x86_64
Speicherzugriffsfehler (Speicherabzug geschrieben)
```
(Now you know the German translation for a segmentation fault)

## Changes
The version range for which we remove the `-single-instance` command has been extended to include 1.9.

I created a new helper function which allows to create a `KspVersionRange` from two `KspVersion` objects. The resulting version range extends from the lowest possible version of the first argument to the highest possible version of the second argument.
It uses the `KspVersion`->`KspVersionRange` logic from `kspversion.ToVersionRange()`.

### DIFF
--- a/Core/Types/GameComparator/StrictGameComparator.cs
+++ b/Core/Types/GameComparator/StrictGameComparator.cs
@@ -24,13 +24,8 @@ namespace CKAN
             {
                 if (module.ksp_version_min != null && module.ksp_version_max != null)
                 {
-                    var minRange = module.ksp_version_min.ToVersionRange();
-                    var maxRange = module.ksp_version_max.ToVersionRange();
-                    if (minRange.Lower.Value <= maxRange.Upper.Value)
-                    {
-                        moduleRange = new KspVersionRange(minRange.Lower, maxRange.Upper);
-                    }
-                    else
+                    moduleRange = new KspVersionRange(module.ksp_version_min, module.ksp_version_max);
+                    if (moduleRange.Lower.Value > moduleRange.Upper.Value)
                     {
                         log.WarnFormat("{0} is not less or equal to {1}",
                             module.ksp_version_min, module.ksp_version_max);
@@ -39,15 +34,11 @@ namespace CKAN
                 }
                 else if (module.ksp_version_min != null)
                 {
-                    var minRange = module.ksp_version_min.ToVersionRange();
-
-                    moduleRange = new KspVersionRange(minRange.Lower, KspVersionBound.Unbounded);
+                    moduleRange = new KspVersionRange(module.ksp_version_min, KspVersion.Any);
                 }
                 else if (module.ksp_version_max != null)
                 {
-                    var maxRange = module.ksp_version_max.ToVersionRange();
-
-                    moduleRange = new KspVersionRange(KspVersionBound.Unbounded, maxRange.Upper);
+                    moduleRange = new KspVersionRange(KspVersion.Any, module.ksp_version_max);
                 }
             }
             else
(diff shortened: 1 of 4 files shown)


Core/Versioning/KspVersionRange.cs
```diff
@@ -27,6 +27,20 @@ public KspVersionRange(KspVersionBound lower, KspVersionBound upper)
             _string = DeriveString(this);
         }
 
+        public KspVersionRange(KspVersion lower, KspVersion upper)
```
HebaruSan: Could this constructor layer atop `KspVersionRange(KspVersionBound lower, KspVersionBound upper)` ?

```csharp
    : this(lower?.ToVersionRange().Lower, upper?.ToVersionRange().Upper)
```

That way the `DeriveString` logic would be factored out into one common place.

Core/Types/GameComparator/StrictGameComparator.cs
```diff
@@ -39,15 +34,11 @@ public override bool SingleVersionsCompatible(KspVersion gameVersion, CkanModule
                 }
                 else if (module.ksp_version_min != null)
                 {
-                    var minRange = module.ksp_version_min.ToVersionRange();
-
-                    moduleRange = new KspVersionRange(minRange.Lower, KspVersionBound.Unbounded);
+                    moduleRange = new KspVersionRange(module.ksp_version_min, new KspVersion());
```
HebaruSan: Could we re-use the static instance `KspVersion.Any` here instead of making a new one?
(same question for the change below)

GUI/Main/Main.cs
```diff
@@ -959,8 +959,11 @@ private void launchKSPToolStripMenuItem_Click(object sender, EventArgs e)
             // https://issuetracker.unity3d.com/issues/linux-segmentation-fault-when-running-a-built-project-with-single-instance-argument
```
HebaruSan: Comment above this could be updated to say "1.8 and 1.9"

DasSkelett: Went with "to", easier to update for KSP 1.10.

DasSkelett:
Thanks for the suggestions!
I also added a test, while this might be implicitly tested in tests for the `StrictGameComparator`, explicit ones are a good idea.

DasSkelett:
No risk no fun :grin:

HebaruSan:
@DasSkelett we need this for 1.10,

https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1280-dyson/&do=findComment&comment=3835778

DasSkelett:
Oh, thanks! Also: `System.ComponentModel.Win32Exception (0x80004005): Cannot find the specified file` :roll_eyes: 

And `xdg-open` doesn't want to open binaries, so `Utilities.ProcessStartURL` is out:
```
Bei der Datei file:///NAS/games/KubuntuGames/steamapps/common/Kerbal Space Program/./KSP.x86_64 handelt es sich um ein ausführbares Programm. Es wird aus Sicherheitsgründen nicht gestartet.
```
Rough translation:
```
The file "file:///NAS/games/KubuntuGames/steamapps/common/Kerbal Space Program/./KSP.x86_64" is an executable program. It won't be started for security reasons.
```

We'll have to find yet another workaround for this...

DasSkelett:
Well, this is weird. I don't know how I got the Win23Exception to show up earlier, but it's not happening anymore.
So we should be fine just extending the fix.

PR #3002, opened 2020-02-13
Respect installing modules during dependency resolution

### LINKED ISSUE #1909: Replacing dependencies
**CKAN Version:**
v1.20.1-0-gd41fe54

**Operating System:**
Ubuntu GNU/Linux 16.04

Cannot replace dependency: I am trying to replace RSSTextures8192 with RSSTextures4096. Both packages satisfy the same dependency requirements but it seems that I cannot replace them withouth uninstalling first all packages depending on RSSTextures8192.

Install RO with RSSTextures8192 and try to then install RSSTexture4096

RSSTextures4096 v10.4 conflicts with RSSTextures8192 v10.4, can't install both.
Install canceled. Your files have been returned to their initial state.

I would expect CKan to tell me that (as both packages satisfy the same dependency), RSSTexture8192 would be removed and RSSTextures4096 would be installed and ask for confirmation.

DasSkelett:
## Problem
Found out thanks to [this comment](https://forum.kerbalspaceprogram.com/index.php?/topic/159443-12-181-spectra-v124-visual-compilation-10th-february-20/&do=findComment&comment=3739637) on the Spectra forum thread.

Currently when uninstalling dependencies of a mod and installing a new mod that also provides these dependencies, CKAN will remove the dependent mod, even though its dependencies will be satisfied after the process completed, thanks to the new mod.

## Example
Scatterer depends on Scatterer-config and Scatterer-sunflare.
Spectra provides both of these modules.

When you have Scatterer installed with its default config + sunflare and mark Spectra for installation, CKAN correctly reports conflicts.
Then you de-select the default config + sunflare so they get uninstalled.
Even though Spectra provides these modules too (which triggered the TooManyModsProvide conflict), CKAN (both GUI and ConsoleUI) wants to remove Scatterer.

## Cause
This is due to the mods that are about to be installed not being respected in `FindReverseDependencies()` > `FindUnsatisfiedDepends()` > `MatchesAny()`.
Only the currently installed mods are taken into account for the dependency resolution.

## Changes
This commit adds an optional argument to `FindReverseDependencies()`, `IEnumerable<CkanModule> modulesToInstall`.
This list will be added to the list of "currently installed" mods, so `FindUnsatisfiedDepends()` considers them and their provided modules.

`UninstallList` now takes a `IEnumerable<CkanModule> installing` instead of `IEnumerable<string> installing` which it passes on to `FindReverseDependencies()`.
`ConsoleUI/InstallScreen.Run()` and `GUI/Main/MainInstall.InstallMods()` are changed to pass the list of mods that are about to be installed to `UninstallList()`.
`GUI/MainModList.ComputeChangeSetFromModList()` also passes `modules_to_install`.

I've removed the `UninstallList()` override which takes a single string instead of a list of mods to uninstall, because it was only used in one single test.
This test now creates a new `List<string>` on the fly.

Closes #1909

### DIFF
--- a/Core/Registry/Registry.cs
+++ b/Core/Registry/Registry.cs
@@ -1073,10 +1073,10 @@ namespace CKAN
 
                     if (modulesToInstall != null)
                     {
-                        // We added modules to the hypothetical list which are not yet installed,
-                        // but not their dependencies (we don't know them yet).
-                        // Thus the SanityChecker wants to remove them again -> don't let him.
-                        broken = broken.Except(modulesToInstall.Select(m => m.identifier)).ToHashSet();
+                        // Make sure to only report modules as broken if they are actually currently installed.
+                        // This is mainly to remove the modulesToInstall again which we added
+                        // earlier to the hypothetical list.
+                        broken.IntersectWith(origInstalled.Select(m => m.identifier));
                     }
                     // Lazily return each newly found rev dep
                     foreach (string newFound in broken.Except(modulesToRemove))
--- a/Core/Registry/Registry.cs
+++ b/Core/Registry/Registry.cs
@@ -1027,8 +1027,15 @@ namespace CKAN
         /// Finds and returns all modules that could not exist without the listed modules installed, including themselves.
         /// Acts recursively and lazily.
         /// </summary>
+        /// <param name="modulesToRemove">Modules that are about to be removed.</param>
+        /// <param name="modulesToInstall">Optional list of modules that are about to be installed.</param>
+        /// <param name="origInstalled">Modules that are already installed</param>
+        /// <param name="dlls">Installed DLLs</param>
+        /// <param name="dlc">Installed DLCs</param>
+        /// <returns>List of modules whose dependencies are about to be or already removed.</returns>
         internal static IEnumerable<string> FindReverseDependencies(
             IEnumerable<string> modulesToRemove,
+            IEnumerable<CkanModule> modulesToInstall,
             IEnumerable<CkanModule> origInstalled,
             IEnumerable<string> dlls,
             IDictionary<string, UnmanagedModuleVersion> dlc
@@ -1050,6 +1057,12 @@ namespace CKAN
                 {
                     // Make our hypothetical install, and remove the listed modules from it.
                     HashSet<CkanModule> hypothetical = new HashSet<CkanModule>(origInstalled); // Clone because we alter hypothetical.
+                    if (modulesToInstall != null)
+                    {
+                        // Pretend the mods we are going to install are already installed, so that dependencies that will be
+                        // satisfied by a mod that is going to be installed count as satisfied.
+                        hypothetical = hypothetical.Concat(modulesToInstall).ToHashSet();
+                    }
                     hypothetical.RemoveWhere(mod => modulesToRemove.Contains(mod.identifier));
 
                     log.DebugFormat("Started with {0}, removing {1}, and keeping {2}; our dlls are {3}", string.Join(", ", origInstalled), string.Join(", ", modulesToRemove), string.Join(", ", hypothetical), string.Join(", ", dllSet));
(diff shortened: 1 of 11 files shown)


HebaruSan:
I think you can mark this as fixing #1909.

HebaruSan:
I installed Scatterer w/ defaults, then Spectra. The changelog still says it's going to uninstall Scatterer:

image

Is that right?

HebaruSan:
After accepting that changeset, it got stuck waiting forever at this point:

image

DasSkelett:
> I installed Scatterer w/ defaults, then Spectra. The changelog still says it's going to uninstall Scatterer:
> 
> image
> 
> Is that right?

It doesn't happen if you've ModuleManager already installed, interestingly (and that I always had during my testing...)
I'll try to find out what's going on there.

HebaruSan:
Oops, this was also the install where I was partway through the `"kind":"dlc"` project. That might be contributing...

HebaruSan:
I'm seeing some very weird behavior with the conflicts. With all three checked:

image

With the sunflare unchecked:

image

So the default config only conflicts if the sunflare is installed??

HebaruSan:
OK, the freeze happens here:

https://github.com/KSP-CKAN/CKAN/blob/09a98125abe2cb5658c147dade42148fb4bbe4f2/Core/ModuleInstaller.cs#L761

The identifier that it's looking for is "Spectra", which is not yet installed and so doesn't belong in `goners`.

HebaruSan:
Yeah, I think that's the problem, `FindReverseDependencies` is going to tell us that we need to uninstall things that aren't even installed yet. Needs to be smarter around that.

DasSkelett:
I moved the lines where the new modules are added to the installed modules inside the internal `FindReverseDependencies()`, and subtract future modules again before returning them.

Fixed

This should be rather safe to do, because we don't rely on `FindReverseDependencies()` to find out which additional mods have to be installed to satisfy all dependencies, that's the job of `RealtionshipResolver` later on in the process.

I didn't encounter any freezes with this now (just Spectra taking a long time to install due to its size).

HebaruSan:
Nice, it worked perfectly that time! I can go back and forth between Spectra and the default configs at will. This is some great progress!

I'm seeing this in the log though:

```
2547507 [1] ERROR CKAN.RelationshipResolver (null) - Assertion failed: Adding Scatterer-config twice in relationship resolution
```

I'll take a closer look at the code sometime in the next day or so.

Core/Registry/Registry.cs
```diff
@@ -1027,45 +1027,65 @@ public List<string> GetSanityErrors()
...
                     // Find what would break with this configuration.
                     var broken = SanityChecker.FindUnsatisfiedDepends(hypothetical, dllSet, dlc)
                         .Select(x => x.Key.identifier).ToHashSet();
 
+                    if (modulesToInstall != null)
+                    {
+                        // We added modules to the hypothetical list which are not yet installed,
+                        // but not their dependencies (we don't know them yet).
+                        // Thus the SanityChecker wants to remove them again -> don't let him.
+                        broken = broken.Except(modulesToInstall.Select(m => m.identifier)).ToHashSet();
```
HebaruSan: Rather than removing the modules to install, could we take the intersection with `origInstalled` ? That would directly guarantee that `FindReverseDependencies` can only return modules that are already installed.

I keep thinking that there's still a way to make this freeze...

1. Install ModuleManager
2. In one change set:
   1. Install something that depends on something not yet installed that depends on ModuleManager
   2. Remove ModuleManager
3. Should say it's going to remove the intermediate dependency and freeze because it isn't installed yet

In actual testing this works fine, but I'm not sure why that is. I think we may still be leaving ourselves open to this in subtle ways and would like to see stronger protection against it.

DasSkelett: Hmm, I played this through, and it seems to work out:

Assuming A depends on B depends on MM.
`FindUnsatisfiedDepends()` only returns A (which is missing B), because B is not yet included (search for needed dependencies is done alter in the process).
A is removed from `broken` again because it is in `modulesToinstall` (yeah an intersection is cleaner), so broken is now empty.
Thus, `modulesToRemove` is a superset of `broken` and the loop stops.

PR #3005, opened 2020-02-15
Check current dir for portable install

### LINKED ISSUE #2956: [Feature] CLI should auto-detect KSP install based on working directory
I use the CKAN CLI when I'm on my (OSX) laptop.  It works fine but it's quite annoying to have to pass the `--ksp whatever_install` parameter to every call that needs it.  Very often I find myself making the calls from within that KSP install's directory anyway.

Setting a default install works but is annoying, especially when I need to change installs (and as a modder I often find myself with several for testing different things).

So it would be great if the CKAN CLI would auto-detect an install if the current working directory is in that install or a subdirectory.  This would take lower priority than the default install or directly specifying the `--ksp` option.

I think I see how to implement this, just wanted to get buy-in before I started trying to work on it.

HebaruSan:
## Motivation

Normally if you launch CKAN, the "default install" is loaded.

If you install `ckan.exe` inside a game folder and launch it from there, that game folder is loaded instead.

For users who work with multiple installs but run CKAN from a single system install point (fewer copies of `ckan.exe` means fewer updates for a new release), it has been suggested that it would be nice to also check the current directory.

## Changes

Now if you run CKAN from a game dir (currently defined by the "portable" logic as a directory having a `GameData` directory inside of it), we will open it as a "portable" install. This works for CmdLine, ConsoleUI, and GUI.

Fixes #2956.

### DIFF
--- a/Core/KSP.cs
+++ b/Core/KSP.cs
@@ -180,20 +180,31 @@ namespace CKAN
 
         /// <summary>
         /// Returns the path to our portable version of KSP if ckan.exe is in the same
-        /// directory as the game. Otherwise, returns null.
+        /// directory as the game, or if the game is in the current directory.
+        /// Otherwise, returns null.
         /// </summary>
         public static string PortableDir()
         {
+            string curDir = Directory.GetCurrentDirectory();
+
+            log.DebugFormat("Checking if KSP is in my current dir: {0}", curDir);
+
+            if (IsKspDir(curDir))
+            {
+                log.InfoFormat("KSP found at {0}", curDir);
+                return curDir;
+            }
+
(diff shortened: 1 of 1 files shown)


DasSkelett:
Should it prefer the working directory over the exeDir?
Assuming someone has saved the exe in one KSP install and is launching it from another one (maybe via a symlink or PATH), they might might want it to use the instance from the CWD.

HebaruSan:
Yes, I think you're right, I would expect it to use ksp-1.4 in this example:

```bash
/games/ksp-1.3$ cd ../ksp-1.4
/games/ksp-1.4$ ../ksp-1.3/ckan.exe
```

PR #3009, opened 2020-02-19
Don't auto-stage if version is "any"

HebaruSan:
The auto-staging from #2970 has been triggering when the corresponding value is `"any"`:

- KSP-CKAN/CKAN-meta#1790
- KSP-CKAN/CKAN-meta#1797
- KSP-CKAN/CKAN-meta#1839

This isn't aligned with the intentions of this feature; a module compatible with all versions is much less likely to change compatibility than a module with specific compatible versions.

Now we only auto-stage if at least one of the present properties isn't `"any"` or `null`.


techman83:
That seems reasonable to me. It will also reduce the amount of branches we generate by 200. Will be interesting to see the impact on our disk usage.

PR #3025, opened 2020-04-02
Improve modpack identifier validation

HebaruSan:
## Problems

Discord user Silverlight pointed out some issues with Export Modpack:

- The default value for the identifier field can be invalid, since the KSP instance name is used and may contain characters that are not valid for identifiers
- The error message says `Identifier must contain one or more letters, numbers, and dashes, and must start with a letter or number`, which could be interpreted to mean that other characters are allowed as long as one or more of the indicated characters are present. ~~I don't know whether the German translation has this problem.~~ The German translation does not have this problem.

## Cause

The export function always generated invalid identifiers going way back, but only with #2971 did we start validating the string, since the user can now edit it. I think most of my testing was with instances with names that are valid identifiers.

## Changes

- The default identifier string is now generated by stripping any non-alphanumeric characters off the front, then replacing any other non-alphanumeric characters with dashes
- The error message is clarified to indicate that **only** the specified characters are allowed
- Ctrl-A while the modpack list has focus will select all
- Escape while the modpack list has focus will deselect all

Test build:

- [ckan.zip](https://github.com/KSP-CKAN/CKAN/files/4423183/ckan.zip)
  (I think the cmd window will remain open on Windows, just ignore that, it's for debugging)

### DIFF
--- a/Core/Registry/RegistryManager.cs
+++ b/Core/Registry/RegistryManager.cs
@@ -440,7 +440,7 @@ namespace CKAN
             {
                 // v1.18 to allow Unlicense
                 spec_version = new ModuleVersion("v1.18"),
-                identifier   = name,
+                identifier   = Identifier.Sanitize(name),
                 name         = name,
                 @abstract    = $"A list of modules installed on the {kspInstanceName} KSP instance",
                 kind         = "metapackage",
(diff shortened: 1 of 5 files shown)


DasSkelett:
Ugh, I remember encountering this myself, but forgot to mention it in the review :/

> I don't know whether the German translation has this problem.

I caught it in the German translation, it contains a "nur" (=only)

PR #3028, opened 2020-04-04
Consider supports in recommendations screen descriptions

HebaruSan:
## Background

In #2960 we added a "Supported By" section to the GUI recommendations screen. Unlike the recommendation and suggestion sections, which show mods mentioned by the mods chosen for install, this section is based on relationships in which **other** mods mention the chosen mods. This provides better support for certain scenarios:

- If I install a tech tree, I would be able to see which mods integrate with that tech tree
- If I install RPM, I would be able to see mods that provide MFD pages
- If I install TweakScale, I would be able to see part mods that allow scaling

## Problem

The orange marks highlight places in the current UI that incorrectly state or imply that "supported by" mods are recommended or suggested by the chosen mods (because that was how the screen worked after #2744 and the text was not updated):

image

Only the blue highlighted part indicates the status of those mods, and it's pretty terse.

The German translation also has this problem.

## Changes

Now the English UI no longer says that supported-by modules are recommended or suggested by the chosen mods:

- The tab caption now mentions supporting mods explicitly
- The explanatory text mentions supported-by explicitly
- The column header now just says "Related mod" since there are several different relationships
- The checkbox just says "(De-)select all" instead of specifying a relationship
- The "Supported By" header now clarifies "(not endorsed by chosen mods)", to set the expectation that this section is not generated from or owned by the mods the user selected

image

The German UI is updated as well.

### DIFF
--- a/GUI/Controls/ChooseRecommendedMods.resx
+++ b/GUI/Controls/ChooseRecommendedMods.resx
@@ -119,12 +119,12 @@
   </resheader>
   <data name="RecommendedModsCancelButton.Text" xml:space="preserve"><value>Cancel</value></data>
   <data name="RecommendedModsContinueButton.Text" xml:space="preserve"><value>Continue</value></data>
-  <data name="RecommendedModsToggleCheckbox.Text" xml:space="preserve"><value>(De-)select all recommended or suggested mods</value></data>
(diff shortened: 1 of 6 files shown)


DasSkelett:
:+1:. Gonna update the German translation and do a PR to your branch shortly

DasSkelett:
Done in https://github.com/HebaruSan/CKAN/pull/10

PR #3043, opened 2020-04-30
Prompt user to overwrite manually installed files

### LINKED ISSUE #949: Saner handling/overwriting of autodetected mods

### LINKED ISSUE #2186: [Feature] Convert AD mods to CKAN control

### LINKED ISSUE #3029: [Feature] Make mod installation not fail if an unowned identical craft file already exists
Problem
-------
CKAN provides the safety feature to never overwrite files it did not install itself. I believe this is a very good thing, and may have helped me in past years when I restored GameData from backups instead of using CKAN.

It is however possible that craft files, normally provided by a mod, are present before the mod is installed through CKAN. I *believe* this is a side-effect of Steam, which appears to keep the saved crafts in the cloud and "restores" them when you install KSP fresh through Steam. Extremely helpful for custom crafts, but not so helpful for mod-provided craft files, as the mod can now no longer be installed through CKAN without manual user intervention

Suggestions
-----------
While the never-overwrite feature of CKAN should stay, I think it could be improved for specific cases like the .craft files. I think the ideal approach for this is:

1. Identify the situation where a mod-provided file already exists and is not "owned" by CKAN (this part already exists and works as expected/intended)
2. IF the file is a .craft file, handle it as a special case (this could apply to other cloud-saved files, if applicable)
3. Verify that the existing .craft file is an exact duplicate of the to-be-installed .craft file (a checksum like sha256 of both files should prove or disprove that the files are identical)
4. IF the files are identical, skip creating the mod-provided file, as an exact duplicate already exists.
5. IF the files are NOT identical, abort/roll-back as usual, to not overwrite any user-modified files.
6. "Own" the file, as if the file was mod-provided, so mods can be installed without external user intervention.

**Describe alternatives you've considered**
Keep the safety-net as-is, but in the case of craft files in the VAB/SPH directories, inform the user that this can be caused by cloud-syncing. Currently it instructs the user to remove the mod that placed the file, even if no such mod exists.

**Additional context**
The .craft files are provided by mods as "documentation" and/or as example, i.e. for reference. While some users may alter them and save them under the same name, I believe most users use them for reference only, if at all. The checksum would prevent users from losing data in the first case, while most users would benefit from an even smoother mod-installation process when cloud-syncing is also active. 

This improvement would eliminate issues like #2894 . In my case, it was the same mod (ID: Mk2Expansion, 2:1.8.7) and same files, though I immediately identified this as a mod-provided craft by it's name. Since this is a pretty fresh installation, I can't think of any other way that I got the .craft files (and nothing else from the mod) except for Steam's cloud-sync.

HebaruSan:
## Motivation

Most new KSP users start out by installing mods manually and find out about CKAN later, at which point they already have an install with some mods and would like to have CKAN take over.

Currently this user's only option is to manually clean out their game folders and start from scratch with CKAN. Some of the manually installed mods will show up in the GUI as "AD", but CKAN can't remove/install/upgrade them. Other manually installed mods will simply show up as not installed, but error out if you try to install them.

## Changes

Now you can:

1. Select an AD mod
2. Go to the Versions tab
3. Check the box for one of the versions (usually you'd want the most recent)
   (Or instead of steps 1-3, select a manually installed mod that doesn't show up as AD and add it to be installed normally)
   The apply button becomes enabled
4. Click the apply button and continue through the installation
5. Get a new prompt explaining that some files are already present and manually installed:
  image
   Files that are different are highlighted at the top of the list.
6. Click No to cancel and abort the install; or
   Click Yes to replace the files
7. Finish the installation and enjoy your freshly CKAN-installed mod

Similarly, CmdLine's `install` command can overwrite AD mods' files with the same prompt.

The Versions tab is involved because I don't know how to turn the current "AD" text in the checkbox column into an installable checkbox in a clean and clear way. This is kind of a pilot program for some delicate functionality so I'd like to take a cautious approach. If this goes well we can try to build upon it and make the UI better.

Fixes #949.
Fixes #2186.
Fixes #3029.

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -290,13 +290,13 @@ namespace CKAN
             ModuleVersion version = registry.InstalledVersion(module.identifier);
 
             // TODO: This really should be handled by higher-up code.
-            if (version != null)
+            if (version != null && !(version is UnmanagedModuleVersion))
             {
                 User.RaiseMessage("    {0} {1} already installed, skipped", module.identifier, version);
                 return;
             }
 
-            // Find our in the cache if we don't already have it.
+            // Find ZIP in the cache if we don't already have it.
             filename = filename ?? Cache.GetCachedZip(module);
 
             // If we *still* don't have a file, then kraken bitterly.
@@ -348,6 +348,7 @@ namespace CKAN
         /// Installs the module from the zipfile provided.
         /// Returns a list of files installed.
         /// Propagates a BadMetadataKraken if our install metadata is bad.
+        /// Propagates a CancelledActionKraken if the user decides not to overwite unowned files.
         /// Propagates a FileExistsKraken if we were going to overwrite a file.
         /// </summary>
         private IEnumerable<string> InstallModule(CkanModule module, string zip_filename, Registry registry)
@@ -360,6 +361,32 @@ namespace CKAN
 
                 try
                 {
+                    var dll = registry.DllPath(module.identifier);
+                    if (dll != null && !files.Any(f => ksp.ToRelativeGameDir(f.destination) == dll))
+                    {
+                        throw new DllLocationMismatchKraken(dll, $"DLL for module {module.identifier} found at {dll}, but it's not where CKAN would install it. Aborting to prevent multiple copies of the same mod being installed. To install this module, uninstall it manually and try again.");
+                    }
+
+                    // Look for overwritable files if session is interactive
+                    if (!User.Headless)
+                    {
+                        var conflicting = FindConflictingFiles(zipfile, files, registry).Memoize();
+                        if (conflicting.Any())
+                        {
+                            var fileMsg = conflicting
+                                .OrderBy(c => c.Value)
+                                .Aggregate("", (a, b) =>
+                                    $"{a}\r\n- {ksp.ToRelativeGameDir(b.Key.destination)}  ({(b.Value ? "same" : "DIFFERENT")})");
+                            if (User.RaiseYesNoDialog($"Module {module.name} wants to overwrite the following manually installed files:\r\n{fileMsg}\r\n\r\nOverwrite?"))
+                            {
+                                DeleteConflictingFiles(conflicting.Select(f => f.Key));
+                            }
+                            else
+                            {
+                                throw new CancelledActionKraken($"Not overwriting manually installed files, can't install {module.name}.");
+                            }
+                        }
+                    }
                     foreach (InstallableFile file in files)
                     {
                         log.DebugFormat("Copying {0}", file.source.Name);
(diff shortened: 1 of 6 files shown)


techman83:
I don't have the time to familiarise myself to review, but this is really really cool! Excellent work @HebaruSan !

DasSkelett:
Looks like the YesNoDialog needs scrollbars for bigger mods: 😅 
YesNoDialog too big for screen

DasSkelett:
Encountered this through accident:
If you have a mod with a DLL installed manually, but in a different directory than CKAN would do (e.g. ActionGroupsExtended in `GameData/AGExt` instead of `GameData/Diazo/AGExt`), you now can "re-"install it with CKAN. Before, you got an error during installation (`AGExt already installed, skipped`). However, since no files are really conflicting, the existing files stay in GameData, so the mod is now effectively installed twice.

But I know that removing the original files, if they are not conflicting, is basically impossible currently, since we don't know them, we only know the DLL.
For now we could check whether the DLL, if there already is one for this module, is in `conflicting`, and if not, well, either error out or remove only this DLL file.

Got an idea?

---

Somewhat related: Assuming a correct manual installation, there might still be additional files in the subfolders that won't be overwritten by the CKAN-installed version (for example if the AD mod is an older version, or if there are dynamically generated cache/save/config files).
Some of them theoretically should be deleted by CKAN in the replacing process, some of the maybe shouldn't.
Should we try to squeeze in a prompt similar to https://github.com/KSP-CKAN/CKAN/pull/2962, with the highest folder containing a file that got overwritten as "root"?

This would make the process much more complicated than it currently is, and it's going into rabbit holes we wanted to avoid, so I am fine with letting it as is.

HebaruSan:
Scrollbar coming up shortly...

image

And a new error for your mismatched install path scenario...

image

For the extra-files concern, let's keep it simple at first and respond based on whether anyone actually has this problem. We might end up making things complicated unnecessarily.

PR #3045, opened 2020-05-05
NetKAN warnings

### LINKED ISSUE #969: Better health reporting on mod indexing.
This is very much a "grand scheme" ticket, but it wold be really nice to have better reporting on mod indexing. Not just a log from our netkan-bot, but an actual nagios-style display of what's working, and what's not. If we fail to index a mod (eg, because it's missing) then a light goes red. If a mod is indexed but looks weird (eg: the latest version compares to be older than an existing version) we can flip things to a warning.

Such a display would not only allow users to check when a mod was last indexed, and the health of its metadata, but also could alert developers if something goes awry.

My sysadmin side thinks [nagios](https://www.nagios.org/) would be a pretty cool choice here, but I've not used it in anger for many years.

HebaruSan:
## Motivation

See KSP-CKAN/NetKAN-Infra#160, it would be nice to be able to capture non-fatal warnings in netkan.exe to discover opportunities to improve the metadata.

## Changes

- `AvcTransformer` now prints a warning via `Log.Warn` if a version file's `URL` property contains an invalid URL, or if the URL can't be downloaded
- A new `PluginCompatibilityValidator` prints a warning via `Log.Warn` if a module contains a DLL file and claims to be compatible with all future versions, since plugins always break eventually
- A new `VrefValidator` prints a warning via `Log.Warn` if a module contains a version file and has no vref, OR if a module has a vref and no version file
- A new `ModuleManagerDependsValidator` prints a warning via `Log.Warn` if a module's cfg files use ModuleManager syntax but the module doesn't depend on ModuleManager, OR if the module depends on ModuleManager and doesn't use ModuleManager syntax
- A new `QueueAppender` class intercepts `Log.Warn` calls when running in SQS queue mode and collects them for later inclusion in the new `WarningMessages` message attribute

Together, this will give users, the PR validation scripts, and the NetKAN-Infra Indexer access to these warnings to do with as they see fit. A future NetKAN-Infra PR will probably update the Indexer to add them to the status page somehow.

In addition, each intermediate stage of transforming a netkan is now checked by `NetkanValidator`, to catch invalid data in metanetkans or internal ckans.

Fixes #969. ("If a mod is indexed but looks weird we can flip things to a warning." Although note that we already addressed "(eg: the latest version compares to be older than an existing version)" with auto-epoching, see #2824.)

### DIFF
--- a/Netkan/Validators/ModuleManagerDependsValidator.cs
+++ b/Netkan/Validators/ModuleManagerDependsValidator.cs
@@ -55,7 +55,7 @@ namespace CKAN.NetKAN.Validators
         private string[] identifiers = new string[] { "ModuleManager" };
 
         private static readonly Regex moduleManagerRegex = new Regex(
-            @"^\s+[@%]",
+            @"^\s*[@+$\-!%]",
             RegexOptions.Compiled | RegexOptions.Multiline | RegexOptions.Singleline
         );
 
--- a/Netkan/Validators/ModuleManagerDependsValidator.cs
+++ b/Netkan/Validators/ModuleManagerDependsValidator.cs
@@ -23,25 +23,32 @@ namespace CKAN.NetKAN.Validators
 
             JObject    json = metadata.Json();
             CkanModule mod  = CkanModule.FromJson(json.ToString());
-            ZipFile    zip  = new ZipFile(_http.DownloadPackage(
-                metadata.Download,
-                metadata.Identifier,
-                metadata.RemoteTimestamp
-            ));
-
-            bool hasMMsyntax = _moduleService.GetConfigFiles(mod, zip)
-                .Select(cfg => new StreamReader(zip.GetInputStream(cfg.source)).ReadToEnd())
-                .Any(contents => moduleManagerRegex.IsMatch(contents));
-
-            bool dependsOnMM = mod?.depends?.Any(r => r.ContainsAny(identifiers)) ?? false;
-
-            if (hasMMsyntax && !dependsOnMM)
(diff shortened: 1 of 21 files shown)


Netkan/Validators/ModuleManagerDependsValidator.cs
```diff
@@ -0,0 +1,67 @@
...
+                        Log.Warn("ModuleManager dependency may not be needed, no ModuleManager syntax found");
+                    }
+                }
+            }
+        }
+
+        private string[] identifiers = new string[] { "ModuleManager" };
+
+        private static readonly Regex moduleManagerRegex = new Regex(
+            @"^\s+[@%]",
```
DasSkelett, line 58: I would extend the regex a little bit to catch more MM operators (sourced from the [wiki](https://github.com/sarbian/ModuleManager/wiki/Module-Manager-Syntax):
```suggestion
            @"^\s*[@+$\-!%]",
```

HebaruSan: Thanks! That regex was definitely a bit threadbare, I think I meant to return to it and flesh it out but forgot.

HebaruSan:
> if the property is set in a metanetkan or an internal ckan file (as it is for Astrogator, e.g.), a false positive warning will be generated. A few of the existing validators also suffer from this issue

I think I see a nice way to address this. Putting this on hold while I work on that...

HebaruSan:
"Known limitations" section removed because latest commit now validates metanetkans and internal ckans.

PR #3054, opened 2020-05-16
Non-parallel GitHub downloads, one progress bar per download

### LINKED ISSUE #1085: TRAP THE ERRORSSSSSS!!!!!!
It's a huge pain in the ass to be downloading/installing mods, then get an unhandled exception error when some web page that supposedly allows the plugin to download gives a 403 error! Also, I'm getting unhandled errors when faced with contradictory mods that have been selected for install. It needs to be given a tab selection method for choosing between the two.

### LINKED ISSUE #1817: Mods fail to download, fail to error - indefinite wait for interrupt.
CKAN 0.18.0
Windows 7 SP1 x64
On installing a list of mods, partway through downloading, the download stops indefinitely. No error displayed, until clicking 'Cancel', when the following errors were displayed (sequentially).

Failed to download "https://github.com/Crzyrndm/B9-PWings-Fork/releases/download/0.40.7/B9-PWings-Fork.v0.40.7.zip" - error: An exception occurred during a WebClient request.

Failed to download "https://github.com/jrossignol/ContractPack-AnomalySurveyor/releases/download/1.6.0/ContractPack-AnomalySurveyor_1.6.0.zip" - error: The remote server returned an error: (403) Forbidden.

Failed to download "https://github.com/DaMichel/KerbalFlightData/releases/download/R21/KerbalFlightDataR21.zip" - error: The remote server returned an error: (403) Forbidden.

Ideally, under this circumstance it would be preferable to display an error message, rather than hanging indefinitely. 

The list of mods was:

About to install...
- Advanced Jet Engine v2.7.2
- Asteroid Day 1.1.3
- B9 Aerospace Procedural Wings - Fork 1:0.40.7
- Community Tech Tree 1:2.4
- Contract Pack: Anomaly Surveyor 1.6.0
- Filter Extensions - Plugin 2.6.0.1
- Filter Extensions - Stock Filter Rework 2.6.0.1
- Hide Empty Tech Tree Nodes 0.4
- IndicatorLights 0.10
- Kerbal Construction Time 1.3.4
- Kerbal Flight Data R21
- Kerbal NRAP - Procedural Test Weights 1.5.1.1
- Kerbal Inventory System 1.2.12.0
- Near Future Construction 0.6.4(cached)
- Near Future Electrical 0.7.5(cached)
- Near Future Propulsion 0.7.3(cached)
- Near Future Solar 0.6.2(cached)
- Procedural Fairings v3.17
- 6 Seat Mk3 Cockpit 1.1
- Filter Extensions - Default Configuration 2.6.0.1
- Solver Engines plugin v2.3
- MagiCore 1.1.1
- Near Future Electrical Core 0.7.5(cached)
- Near Future Solar Core 0.6.2(cached)

### LINKED ISSUE #2210: CKAN's GitHub downloads are breaking the rules

### LINKED ISSUE #2400: Weird SSL(?) error
**CKAN Version:** 1.24.0
**KSP Version:** 1.3.1
**Operating System:** Debian  9
**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**  nope

Problem
-------
Tried to install Realism Overhaul for KSP1.3.1. 

That's a lot of mods, some of them come in rather large files. CKAN downloaded most files over the course of about 10 minutes, then stalled. I cancelled the install, and started over from the CKAN main menu, doing the same selections as before. Most mods were cached, but the three still outstanding downloads were failed with an certificate error.

Re-importing certificates as suggested in https://github.com/KSP-CKAN/CKAN/wiki/SSL-certificate-errors did not help (ca-certificates-mono is already the newest version, cert-sync has nothing to do).

It strikes me as odd that everything works fine for a while, then throws SSL errors two minutes later. The certificate on the github website is valid until 17-May-2018 (only six weeks to go); I don't think it has been replaced just now. I'm suspecting some load-balancing or abuse prevention mechanism is at work, hence the question mark in the title.

FWIW, here's the full output of the second (failed) download attempt. I'd like to emphasisze once more that all all "cached" files were downloaded just a few minutes before:

```
About to install...

 * Realism Overhaul v12.0.0 (cached)
 * Deadly Reentry Continued v7.6.2 (cached)
 * HangerExtender 3.5.2 (cached)
 * KSP AVC 1.1.6.2 (cached)
 * MechJeb 2 2.7.0.0 (cached)
 * Procedural Fairings v4.0 (cached)
 * Procedural Parts v1.2.14 (cached)
 * PersistentRotation 1.8.4 (cached)
 * KSC Switcher 0.8 (cached)
 * Realism Overhaul Craft Files v12.0.0 (cached)
 * RemoteTech v1.8.9 (cached)
 * TAC Life Support (TACLS) v0.13.6.0 (cached)
 * TextureReplacerReplaced V0.5.4 (cached)
 * Toolbar 1.7.16.5 (cached)
 * Ven's Stock Part Revamp v1.9.6 (cached)
 * DMagic Orbital Science 1.3.11 (cached)
 * Kerbal Attachment System 0.6.3.0 (cached)
 * SCANsat v18.4 (cached)
 * FASA 1:v7.2.2 (cached)
 * SXTContinued 1:0.3.23.1 (cached)
 * Advanced Jet Engine v2.10.0 (cached)
 * Ferram Aerospace Research 3:0.15.9.1 (github.com, 751.9 KB)
 * Solver Engines plugin v3.5 (cached)
 * Kerbal Joint Reinforcement v3.3.3 (cached)
 * RealChute Parachute Systems v1.4.5 (github.com, 2.6 MB)
 * Real Fuels rf-v12.5.0 (cached)
 * Community Resource Pack 0.8.1.0 (cached)
 * RealHeat v4.5 (cached)
 * Real Plume 2:v11.0.0 (cached)
 * SmokeScreen - Extended FX Plugin 2.7.6.1 (cached)
 * Background Resources v0.13.6.0 (cached)
 * Firespitter Core v7.6.0 (github.com, 29.2 MB)
 * Firespitter Resources config v7.6.0 (github.com, 29.2 MB)
 * Retractable Lifting Surface Module 0.1.5 (cached)

Downloading "https://github.com/ferram4/Ferram-Aerospace-Research/releases/download/v0.15.9.1_Liepmann/FAR_0_15_9_1_Liepmann.zip"
Downloading "https://github.com/StupidChris/RealChute/releases/download/v1.4.5/RealChute_v1.4.5.zip"
Downloading "https://github.com/snjo/Firespitter/releases/download/v7.6.0/Firespitter_7.6.0.zip"
Oh no! Our download failed with a certificate error!

Consult this page for help:
	https://github.com/KSP-CKAN/CKAN/wiki/SSL-certificate-errors
Error!
```

HebaruSan:
## Problem

If you try to install lots of mods that you haven't previously downloaded, you'll likely get errors like this (possibly in a different format since we've changed them a bunch recently):

> Failed to download "https://github.com/jrossignol/ContractPack-AnomalySurveyor/releases/download/1.6.0/ContractPack-AnomalySurveyor_1.6.0.zip" - error: The remote server returned an error: (403) Forbidden.

## Cause

GitHub doesn't like it when you try to download multiple things from it at the same time, which CKAN aggressively does. GitHub sometimes responds to such attempts with 403-Forbidden statuses (the precise rules are not publicly known).

The GitHub _API_ works similarly, except it has [defined](https://developer.github.com/v3/#abuse-rate-limits) (and [queryable](https://api.github.com/rate_limit)) throttling rules and explicitly [says not to download from it in parallel](https://developer.github.com/v3/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits):

> Make requests for a single user or client ID serially. Do not make requests for a single user or client ID concurrently.

## Motivation

Since we're planning a pre-release anyway, why not go nuts?

## Changes

Now if multiple GitHub downloads are requested, only one of them will be started, and the rest will wait. Non-GitHub downloads will still be performed in parallel as before. When the current GitHub download finishes, one of the others will be started, and so on until completion.

Now each active download has its own progress bar in GUI, so the user can tell what's happening:

image

Fixes #1817.
Fixes #2210. (URLs haven't been migrated to API format, but that's probably not needed.)
Probably fixes #2400.
Fixes #1085. (The description and conversation went all over the place, so it may not be possible to even define what the issue is about, but the original complaint was about 403 statuses during downloads.)

### DIFF
--- a/GUI/Controls/Wait.cs
+++ b/GUI/Controls/Wait.cs
@@ -1,5 +1,9 @@
 using System;
+using System.Linq;
+using System.Collections.Generic;
+using System.Drawing;
 using System.Windows.Forms;
+using Timer = System.Windows.Forms.Timer;
 
 namespace CKAN
 {
@@ -8,6 +12,7 @@ namespace CKAN
         public Wait()
         {
             InitializeComponent();
+            progressTimer.Tick += (sender, evt) => ReflowProgressBars();
         }
 
         public event Action OnRetry;
@@ -44,10 +49,146 @@ namespace CKAN
             }
         }
 
+        /// <summary>
+        /// React to data received for a module,
+        /// adds or updates a label and progress bar so user can see how each download is going
+        /// </summary>
+        /// <param name="module">The module that is being downloaded</param>
+        /// <param name="remaining">Number of bytes left to download</param>
+        /// <param name="total">Number of bytes in complete download</param>
+        public void SetModuleProgress(CkanModule module, long remaining, long total)
+        {
+            Util.Invoke(this, () =>
+            {
+                if (moduleBars.TryGetValue(module, out ProgressBar pb))
+                {
+                    pb.Value = (int) (100 * (total - remaining) / total);
+                }
+                else
+                {
+                    var rowTop = TopPanel.Height - padding;
+                    var newLb = new Label()
+                    {
+                        AutoSize = true,
+                        Location = new Point(2 * padding, rowTop),
+                        Size     = new Size(labelWidth, progressHeight),
+                        Text     = string.Format(
+                                Properties.Resources.MainChangesetHostSize,
+                                module.name,
+                                module.version,
+                                module.download.Host ?? "",
+                                CkanModule.FmtSize(module.download_size)),
+                    };
+                    moduleLabels.Add(module, newLb);
+                    var newPb = new ProgressBar()
+                    {
+                        Anchor   = AnchorStyles.Top | AnchorStyles.Left | AnchorStyles.Right,
+                        Location = new Point(labelWidth + 3 * padding, rowTop),
+                        Size     = new Size(TopPanel.Width - labelWidth - 5 * padding, progressHeight),
+                        Minimum  = 0,
+                        Maximum  = 100,
+                        Value    = (int) (100 * (total - remaining) / total),
+                        Style    = ProgressBarStyle.Continuous,
+                    };
+                    moduleBars.Add(module, newPb);
+                }
+                progressTimer.Start();
+            });
+        }
+
+        private const int padding        = 5;
+        private const int labelWidth     = 400;
+        private const int progressHeight = 20;
+        private const int emptyHeight    = 85;
+
+        private Dictionary<CkanModule, Label>       moduleLabels = new Dictionary<CkanModule, Label>();
+        private Dictionary<CkanModule, ProgressBar> moduleBars   = new Dictionary<CkanModule, ProgressBar>();
+        private Timer progressTimer = new Timer() { Interval = 3000 };
+
+        /// <summary>
+        /// Add new progress bars and remove completed ones (100%) in a single scheduled pass,
+        /// so they don't constantly flicker and jump around.
+        /// </summary>
+        private void ReflowProgressBars()
+        {
+            Util.Invoke(this, () =>
+            {
+                int rowTop = emptyHeight - padding;
+                var theBars = moduleBars.OrderBy(kvp => kvp.Value.Top).ToList();
+                foreach (var kvp in theBars)
+                {
+                    if (kvp.Value.Value == 100)
+                    {
+                        // Finished, remove in this pass
+                        TopPanel.Controls.Remove(moduleLabels[kvp.Key]);
+                        TopPanel.Controls.Remove(kvp.Value);
+                        // rowTop is unchanged, so next row will replace this one
+                    }
+                    else if (!TopPanel.Controls.Contains(kvp.Value))
+                    {
+                        // Just started, add it in this pass
+                        moduleLabels[kvp.Key].Top = rowTop;
+                        kvp.Value.Top = rowTop;
+                        TopPanel.Controls.Add(moduleLabels[kvp.Key]);
+                        TopPanel.Controls.Add(kvp.Value);
+                        rowTop += progressHeight + padding;
+                    }
+                    else
+                    {
+                        // Not finished, already displayed, just make sure it's in the right position
+                        moduleLabels[kvp.Key].Top = rowTop;
+                        kvp.Value.Top = rowTop;
+                        rowTop += progressHeight + padding;
+                    }
+                }
+                // Make room for everything that's still visible
+                TopPanel.Height = rowTop + padding;
+
+                var removedModules = moduleBars
+                    .Where(kvp => !TopPanel.Controls.Contains(kvp.Value))
+                    .Select(kvp => kvp.Key)
+                    .ToList();
+                foreach (var module in removedModules)
+                {
+                    moduleLabels.Remove(module);
+                    moduleBars.Remove(module);
+                }
+            });
+        }
+
+        /// <summary>
+        /// React to completion of all downloads,
+        /// removes all the module progress bars since we don't need them anymore
+        /// </summary>
+        public void DownloadsComplete()
+        {
+            ClearModuleBars();
+            progressTimer.Stop();
+        }
+
+        private void ClearModuleBars()
+        {
+            Util.Invoke(this, () =>
+            {
+                foreach (var kvp in moduleLabels)
+                {
+                    TopPanel.Controls.Remove(kvp.Value);
+                }
+                foreach (var kvp in moduleBars)
+                {
+                    TopPanel.Controls.Remove(kvp.Value);
+                }
+                moduleLabels.Clear();
+                moduleBars.Clear();
+                TopPanel.Height = emptyHeight;
+            });
+        }
+
         public void Reset(bool cancelable)
         {
             Util.Invoke(this, () =>
             {
+                ClearModuleBars();
                 ProgressValue = DialogProgressBar.Minimum;
                 ProgressIndeterminate = true;
                 RetryCurrentActionButton.Enabled = false;
--- a/Core/Net/IDownloader.cs
+++ b/Core/Net/IDownloader.cs
@@ -1,3 +1,4 @@
+using System;
 ﻿using System.Collections.Generic;
 
 namespace CKAN
@@ -12,6 +13,16 @@ namespace CKAN
         /// Blocks until the downloads are complete, cancelled, or errored.
         /// </summary>
         void DownloadModules(IEnumerable<CkanModule> modules);
+
+        /// <summary>
+        /// Raised when data arrives for a module
+        /// </summary>
+        event Action<CkanModule, long, long> Progress;
+
+        /// <summary>
+        /// Raised when a batch of downloads is all done
+        /// </summary>
+        event Action AllComplete;
 
         /// <summary>
         /// Cancel any running downloads.
--- a/Core/Net/NetAsyncDownloader.cs
+++ b/Core/Net/NetAsyncDownloader.cs
@@ -17,14 +17,10 @@ namespace CKAN
     /// </summary>
     public class NetAsyncDownloader
     {
-
-        public IUser User { get; set; }
-
         // Private utility class for tracking downloads
         private class NetAsyncDownloaderDownloadPart
         {
-            public Uri url;
-            public Uri fallbackUrl;
+            public readonly Net.DownloadTarget target;
             public DateTime lastProgressUpdateTime;
             public string path;
             public long bytesLeft;
@@ -42,8 +38,7 @@ namespace CKAN
 
             public NetAsyncDownloaderDownloadPart(Net.DownloadTarget target, string path = null)
             {
-                this.url = target.url;
-                this.fallbackUrl = target.fallbackUrl;
+                this.target = target;
                 this.mimeType = target.mimeType;
                 this.triedFallback = false;
                 this.path = path ?? Path.GetTempFileName();
@@ -77,10 +72,10 @@ namespace CKAN
 
                 // Check whether to use an auth token for this host
                 string token;
-                if (ServiceLocator.Container.Resolve<IConfiguration>().TryGetAuthToken(this.url.Host, out token)
+                if (ServiceLocator.Container.Resolve<IConfiguration>().TryGetAuthToken(this.target.url.Host, out token)
                         && !string.IsNullOrEmpty(token))
                 {
-                    log.InfoFormat("Using auth token for {0}", this.url.Host);
+                    log.InfoFormat("Using auth token for {0}", this.target.url.Host);
                     // Send our auth token to the GitHub API (or whoever else needs one)
                     agent.Headers.Add("Authorization", $"token {token}");
                 }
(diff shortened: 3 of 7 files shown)


HebaruSan:
In latest commit, we create a 3-second timer to remove completed progress bars and add new ones, so the display won't fill up with full bars, but it also won't be constantly bouncing around.

DasSkelett:
After the timer patche, I get this exception from time to time:
```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
  at System.Collections.ArrayList+ArrayListEnumeratorSimple.MoveNext () [0x00013] in <b2b5eee6b3734e9586bbc3ab32e40bc7>:0
  at System.Windows.Forms.XplatUIX11.NextTimeout (System.Collections.ArrayList timers, System.DateTime now) [0x0003a] in <8637a6a20d8848c3b0482a80613e13f8>:0
  at System.Windows.Forms.XplatUIX11.UpdateMessageQueue (System.Windows.Forms.XEventQueue queue, System.Boolean allowIdle) [0x0009f] in <8637a6a20d8848c3b0482a80613e13f8>:0
  at System.Windows.Forms.XplatUIX11.UpdateMessageQueue (System.Windows.Forms.XEventQueue queue) [0x00000] in <8637a6a20d8848c3b0482a80613e13f8>:0
  at System.Windows.Forms.XplatUIX11.GetMessage (System.Object queue_id, System.Windows.Forms.MSG& msg, System.IntPtr handle, System.Int32 wFilterMin, System.Int32 wFilterMax) [0x0001c] in <8637a6a20d8848c3b0482a80613e13f8>:0
  at System.Windows.Forms.XplatUI.GetMessage (System.Object queue_id, System.Windows.Forms.MSG& msg, System.IntPtr hWnd, System.Int32 wFilterMin, System.Int32 wFilterMax) [0x00000] in <8637a6a20d8848c3b0482a80613e13f8>:0
  at System.Windows.Forms.Application.RunLoop (System.Boolean Modal, System.Windows.Forms.ApplicationContext context) [0x0034e] in <8637a6a20d8848c3b0482a80613e13f8>:0
  at System.Windows.Forms.Application.Run (System.Windows.Forms.ApplicationContext context) [0x00011] in <8637a6a20d8848c3b0482a80613e13f8>:0
  at System.Windows.Forms.Application.Run (System.Windows.Forms.Form mainForm) [0x00006] in <8637a6a20d8848c3b0482a80613e13f8>:0
  at CKAN.Main..ctor (System.String[] cmdlineArgs, CKAN.KSPManager mgr, System.Boolean showConsole) [0x002bc] in /home/kilian/git/CKAN/GUI/Main/Main.cs:164
  at at (wrapper remoting-invoke-with-check) CKAN.Main..ctor(string[],CKAN.KSPManager,bool)
  at CKAN.GUI.Main_ (System.String[] args, CKAN.KSPManager manager, System.Boolean showConsole) [0x00047] in /git/CKAN/GUI/Program.cs:35
  at CKAN.GUI.Main (System.String[] args) [0x00001] in /git/CKAN/GUI/Program.cs:15
```

Unfortunately the stack trace doesn't really help locating the cause. But I suspect it's one of the two foreach loops. Maybe I can track it down further.

GUI/Controls/Wait.cs
```diff
@@ -44,10 +58,137 @@ public bool ProgressIndeterminate
...
+                        Style    = ProgressBarStyle.Continuous,
+                    };
+                    moduleBars.Add(module, newPb);
+                }
+                progressTimer.Start();
+            });
+        }
+
+        /// <summary>
+        // Add new progress bars and remove completed ones (100%) in a single scheduled pass,
```
DasSkelett, line 109: One `/` missing here. My IDE complained about a broken docstring :P

HebaruSan:
Added a `ToList` to the first loop to maybe avoid that enumerator problem.

DasSkelett:
Just wanted to ask if this is intentional or not, but I like it:
Progress bars for downloads that complete during the first three seconds are not shown since the timer also waits 3 seconds for the first trigger.

But I think that's quite nice, because you probably aren't interested in the progress bars of downloads which complete that fast, only for ones that take longer.

HebaruSan:
> Progress bars for downloads that complete during the first three seconds are not shown since the timer also waits 3 seconds for the first trigger.

Partially intentional. :)

To show all the downloads in the first 3 seconds, we would have to restore the code that displays the progress bars on the fly, and the annoying flickering would be back (even if just for 3 seconds). And the first 3 seconds aren't quite special—if a download both starts _and_ ends in _any_ interval, it won't get a progress bar.

PR #3058, opened 2020-05-19
Don't save registry when opening settings

DasSkelett:
## Motivation
I recently noticed that opening the settings window is pretty slow from time to time. When profiling it, I found out that we save the registry on settings startup, and it seems to be the culprit (the serialization, to be precise).

## Problem
In `RefreshReposListBox>()`, the registry is saved after the we sort the list of repositories in the registry to how they are sorted in the GUI.
This is also done when the settings are just opened. In this case there are no repositories listed in the GUI (and the user couldn't have changed the sorting yet), so the whole operation is not doing anything.
However we still save the registry afterwards, which takes a long time.

## Changes
THe costly operation is put behind an if-clause. When `_sortedRepos.Count == 0`, we just pull the  list of repositories from the registry (needed to build the list for the GUI later), but jump over all the sorting and saving stuff.

### DIFF
--- a/GUI/Dialogs/SettingsDialog.cs
+++ b/GUI/Dialogs/SettingsDialog.cs
@@ -41,7 +41,7 @@ namespace CKAN
 
         public void UpdateDialog()
         {
-            RefreshReposListBox();
+            RefreshReposListBox(false);
             RefreshAuthTokensListBox();
             UpdateLanguageSelectionComboBox();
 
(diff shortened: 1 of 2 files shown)


DasSkelett:
Added another commit that skips saving the registry in `KSP.ScanGameData()` if nothing changes, that is when `dllChanged` and `dlcChanged` are `false`.

Should help growing the amount of files in the `KSP/CKAN/history` folder slower.

GUI/Dialogs/SettingsDialog.cs
```diff
@@ -77,25 +77,35 @@ private void UpdateRefreshRate()
...
-            // Give the Repository the priority it
-            // currently has in the gui
-            for (int i = 0; i < _sortedRepos.Count; i++)
+            var manager = RegistryManager.Instance(Main.Instance.CurrentInstance);
+            var registry = manager.registry;
+
+            if (_sortedRepos.Count == 0)
             {
-                _sortedRepos[i].priority = i;
+                // No reporitories or just launched the settings window.
```
HebaruSan, line 85:
```suggestion
                // No repositories or just launched the settings window.
```

GUI/Dialogs/SettingsDialog.cs
```diff
@@ -77,25 +77,35 @@ private void UpdateRefreshRate()
 
         private void RefreshReposListBox()
         {
-            // Give the Repository the priority it
-            // currently has in the gui
-            for (int i = 0; i < _sortedRepos.Count; i++)
+            var manager = RegistryManager.Instance(Main.Instance.CurrentInstance);
+            var registry = manager.registry;
+
+            if (_sortedRepos.Count == 0)
```
HebaruSan, line 83: It seems we are asking this function to handle two different tasks (loading, saving) and making it guess which one to perform, whereas the functions that call it definitely know (`SettingsDialog_Load` &rarr; `UpdateDialog` &rarr; `RefreshReposListBox` definitely doesn't need to save, but `DeleteRepoButton_Click` &rarr; `RefreshReposListBox` always would need to save).

What do you think of splitting this into two functions and only calling the one we need? Or maybe using a parameter to communicate what it should do?

GUI/Dialogs/SettingsDialog.cs
DasSkelett: :roll_eyes: Thanks!

GUI/Dialogs/SettingsDialog.cs
DasSkelett: Good point. I went with the parameter option, because it felt cleaner than separate functions (which would've needed some duplicated lines), and was easier with the changes I already had.

HebaruSan:
This is great. I think we can get it to be instantaneous with the changes in DasSkelett/CKAN#8.

Issue #2916, opened 2019-11-08
[Proposal] Release Date


techman83:
## Problem

Knowing when there are new mod releases has been a long standing request, #1155 was opened in Mid 2015! Historically this has been challenging for two reasons:

1. Not all of sources for metadata contain this information.
2. One of the early design philosophies behind Netkan was that anyone could produce a `ckan` from a `netkan` in exactly the same manner as our bots. This had the side effect of it being stateless, which has meant (especially with the old infrastructure) that populating a field with a current time  stamp would mean it would always change.

Since #2789, we have a number of ways we can tackle this and a solution should be straight forward!

### Spec

Add 'release_date' to the schema, that complies with the following:

- Field assumed to be UTC
- Full date / time
- ISO 8601 compliant (2019-11-07T01:29:59+00:00)

Example:

```json
{
    "spec_version": 1,
    "comment": "flagmod",
    "identifier": "DogeCoinFlag",
    "name": "Dogecoin Flag",
    "release_date": "2015-01-05T01:56:00+00:00",
    "abstract": "Such flag. Very currency. Wow.",
    "description": "Adorn your craft with your favourite cryptocurrency. To the mün!",
    "author": "daviddwk",
    "license": "CC-BY",
    "resources": {
        "homepage": "https://www.reddit.com/r/dogecoin/comments/1tdlgg/i_made_a_more_accurate_dogecoin_and_a_ksp_flag/",
        "repository": "https://github.com/pjf/DogeCoinFlag"
    },
    "version": "v1.02",
    "ksp_version": "any",
    "download": "https://github.com/pjf/DogeCoinFlag/releases/download/v1.02/DogeCoinFlag-1.02.zip",
    "download_size": 53359,
    "download_hash": {
        "sha1": "BFB78381B5565E1AC7E9B2EB9C65B76EF7F6DF71",
        "sha256": "CF70CE1F988F908FB9CF641C1E47CDA2A30D5AD951A4811A5C0C5D30F6FD3BA9"
    },
    "download_content_type": "application/zip",
    "x_generated_by": "netkan"
}
```

### NetKAN

It is desirable to keep Netkan as the source of the truth. It persists the ability for anyone to produce metadata that matches exactly what the bot produces.

As a consequence
- It would be responsible for populating the `release_date` field
  - This would enable us to initially use the system timestamp, but also enhance our transformers to retrieve this information from the relevant APIs where available
[...]

PR #2859, opened 2019-08-28
Purge stale cache entries for SpaceDock

HebaruSan:
This is #2337, but for SpaceDock.

Needs KSP-SpaceDock/SpaceDock#195 to be merged to SpaceDock's production. Currently it's still being tested on the alpha server.

PR #3059, opened 2020-05-21
Generate release_date property in netkan

HebaruSan:
## Motivation

A longstanding request is to display when a mod was released, see #1155 and #2916.

This PR begins the process by capturing these values in current metadata where possible.

## Changes

- Now the schema and spec are updated to support an optional `release_date` property in ISO 8601 format
- Now mods on SpaceDock, GitHub, and Jenkins will have a `release_date` property generated automatically based on API data
- Other mods (there are 15 of them) will not have the property set, to indicate that we don't know the release date
- `CkanModule` now has a `release_date` property that future changes can use to display the value to the user
- The `GitHubApi` was attempting to parse the date properties returned by the GitHub API. However the JSON parsing library already parsed them into dates! The `ToString` call produced an intermediate format without timezone (rather than the original raw string as expected), so the TZ was lost. Now we use the date as provided by the parser.
- The `StagingTransformer` will no longer stage modules where `ksp_version_min` is set to some prior version and `ksp_version_max` isn't set at all. This should help with the server IO load (staging is costly), as well as reduce the maintenance burden for mass metadata sweeps like this (lots of PRs to review).
- The client now uses the release date to purge stale downloads just as Netkan has been doing since #2337 and #2859, which will ensure that users are installing the correct version of a replaced module

I plan to work on the UI side after this is merged (which will re-index almost all mods in a big sweep), so I can have plenty of test data.

### Notes

- We have been generating a release date value internally for some time now for purposes of cache purging (see #2337 and #2859), but they were populated into `x_netkan_asset_updated` and therefore stripped out by `StripNetkanMetadataTransformer`. Now that `release_date` is in the schema, we populate it instead and don't strip it.
- Jenkins returns a timestamp representing milliseconds since 1970-01-01, and the default JSON `UnixDateTimeConverter` only supports seconds, so we have to roll our own converter.
- `RelationshipDescriptor` and `ResourcesDescriptor` are split out into their own files to reduce clutter in `CkanModule.cs`
- SpaceDock's API currently returns mod creation dates in local time, and KSP-SpaceDock/SpaceDock#265 is fixing this to send UTC. Ideally we would wait until after that change is in production to merge this, to avoid multiple mass-update passes of SpaceDock hosted mods.


DasSkelett:
You are referencing the release date of "the mod" in the spec and schema, but isn't it more like the release date of "the mod version"/"the mod release"?
I think we should clarify this a bit.

HebaruSan:
Whoa, the schema says this:

https://github.com/KSP-CKAN/CKAN/blob/c451a204b9c50e7554889f15c05ac684155528af/CKAN.schema#L90

"Drop the leading v" is not something we've ever done. I'll go ahead and remove that now...

HebaruSan:
> You are referencing the release date of "the mod" in the spec and schema, but isn't it more like the release date of "the mod version"/"the mod release"?
> I think we should clarify this a bit.

Most of the existing text uses the same phrasing, but `download` in Spec.md says: "the described version of the mod"
Will switch to that...

techman83:
I'm very much in favour of this implementation, keeps our source of truth consistent and means we get our date/times from upstream.

HebaruSan:
Planning to improve on #3031 so PRs like KSP-CKAN/CKAN-meta#1940 don't get submitted...

HebaruSan:
OK, I think that's done in the latest commit.

HebaruSan:
KSP-SpaceDock/SpaceDock#265 is in production! SpaceDock's timestamps have time zones!

PR #3060, opened 2020-05-23
Support fallback URLs for netkan validate-ckan

### LINKED ISSUE #2938: [Bug] Epoching old mods breaks archive.org fallback URLs
## Background

* Operating System: Windows, Linux
* CKAN Version: 1.26.6
* KSP Version: 1.8.1

:poop:

## Problem

**Describe the bug**
While cleaning up old version epochs in KSP-CKAN/CKAN-meta#1574, I realized that epoching a historical version breaks its automatic archive.org fallback from #2284. There were a few in that PR that are only available on archive,org since they've been removed from GitHub, and they failed to download when the validation scripts attempted to install them. Checking https://archive.org/details/kspckanmods , the mods *were* there, just without an epoch.

**Steps to reproduce**
* Add a netkan for a mod with a redistributable license
* Allow multiple versions of this mod to be indexed; its releases will be added to archive.org automatically
* Remove the original downloads for some of the older versions; the CKAN client will still be able to download them from archive.org 
* Edit the .ckan files in CKAN-meta to change the mods' epochs
* Now the fallback doesn't work

**Expected behavior**
Somehow find the original uploads to archive.org 

## Suggestions

1. Strip epochs when archiving; archive.org would have only the pure upstream version numbers
2. Run a conversion on the currently archived mods to remove epochs
3. Strip epochs out when the download fallback code runs

We would need to figure out a migration strategy; clients from before and after step 3 would expect different fallback URLs, so steps 1 and 2 would need to run at a time that would be minimally disruptive to users.

HebaruSan:
## Problem

If you submit a CKAN-meta pull request for a module that currently relies on its archive.org fallback URL for download, the pull request validation will fail, even though the module can be installed by the client.

## Cause

Netkan uses a downloading service separate from the client's, which does not know about fallback URLs.

## Changes

- Now `Metadata` provides a `FallbackDownload` property that returns an archive.org `Uri` if we have enough info to do so and `null` otherwise.
- Now `CachingHttpService` takes `Metadata` as input when downloading instead of a `Uri`, identifier, and timestamp separately.
- Now if a download fails, `CachingHttpService` attempts to get `FallbackDownload` if it is set, and propagates the exception otherwise.

This will allow `netkan.exe --validate-ckan` to access fallback URLs and pass validation for modules relying on them.

NOTE: I just realized I forgot to check for redistributable licenses. I'll start that and remove the "In Progress" label when done.
NOTE: Also, the cache entries should be stored according to the original URL, not the fallback. Consider this pending as well.

Fixes #2938.

### DIFF
--- a/Netkan/Services/CachingHttpService.cs
+++ b/Netkan/Services/CachingHttpService.cs
@@ -36,13 +36,17 @@ namespace CKAN.NetKAN.Services
                 }
                 else
                 {
-                    return DownloadPackage(fallback, metadata.Identifier, metadata.RemoteTimestamp);
+                    return DownloadPackage(fallback, metadata.Identifier, metadata.RemoteTimestamp, metadata.Download);
                 }
             }
         }
 
-        private string DownloadPackage(Uri url, string identifier, DateTime? updated)
+        private string DownloadPackage(Uri url, string identifier, DateTime? updated, Uri primaryUrl = null)
         {
+            if (primaryUrl == null)
+            {
+                primaryUrl = url;
+            }
             if (_overwriteCache && !_requestedURLs.Contains(url))
             {
                 // Discard cached file if command line says so,
@@ -52,7 +56,7 @@ namespace CKAN.NetKAN.Services
 
             _requestedURLs.Add(url);
 
-            var cachedFile = _cache.GetCachedFilename(url, updated);
+            var cachedFile = _cache.GetCachedFilename(primaryUrl, updated);
 
             if (!string.IsNullOrWhiteSpace(cachedFile))
             {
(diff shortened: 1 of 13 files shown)


Netkan/Services/CachingHttpService.cs
```diff
@@ -20,8 +21,32 @@ public CachingHttpService(NetFileCache cache, bool overwrite = false)
...
+            catch (Exception exc)
+            {
+                var fallback = metadata.FallbackDownload;
+                if (fallback == null)
+                {
+                    throw;
+                }
+                else
+                {
+                    return DownloadPackage(fallback, metadata.Identifier, metadata.RemoteTimestamp, metadata.Download);
```
DasSkelett, line 39: A log statement would probably be handy here, so that you can see that the fallback is used.
I would go for a `Log.Info` at least.

HebaruSan: Good idea, borrowed the one from Core.

PR #3080, opened 2020-06-15
Cache hashes on disk

HebaruSan:
## Motivation

#2940 sped up the Inflator by caching hashes to RAM, but this only helps on the second-and-later passes; the first pass after redeployment still takes a long time.

## Changes

Now when we store a ZIP to the cache, we store two additional files next to it, with the same name plus `.sha1` and `.sha256` appended:

- FFA1EF94-ContractsWindowPlus-9.4.zip
- FFA1EF94-ContractsWindowPlus-9.4.zip.sha1
- FFA1EF94-ContractsWindowPlus-9.4.zip.sha256

When we need to retrieve the hash, we check for these files in between checking our in-memory cache and computing the hashes based on the ZIPs.

This should speed up the first passes of the Inflator.

### DIFF
--- a/Core/Net/NetFileCache.cs
+++ b/Core/Net/NetFileCache.cs
@@ -401,10 +401,12 @@ namespace CKAN
                 DirectoryInfo legDir = new DirectoryInfo(legacyDir);
                 files = files.Union(legDir.EnumerateFiles());
             }
-            return files
-                // Require 8 digit hex prefix followed by dash; any else was not put there by CKAN
-                .Where(fi => cacheFileRegex.IsMatch(fi.Name))
-                .ToList();
+            return files.Where(fi =>
+                    // Require 8 digit hex prefix followed by dash; any else was not put there by CKAN
+                    cacheFileRegex.IsMatch(fi.Name)
+                    // Treat the hash files as companions of the main files, not their own entries
+                    && !fi.Name.EndsWith(".sha1") && !fi.Name.EndsWith(".sha256")
+                ).ToList();
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


DasSkelett:
Can we avoid creating all these files in `/tmp` when downloading a module via ckan.exe?
```
/tmp$ ls *.sha*
tmp134e4009.tmp.sha1    tmp1389927a.tmp.sha1    tmp3d8efbf0.tmp.sha1
tmp42ff984d.tmp.sha1    tmp523f2d0f.tmp.sha1    tmp5b9a79d6.tmp.sha1
tmp7345b525.tmp.sha1    tmp134e4009.tmp.sha256  tmp1389927a.tmp.sha256
tmp3d8efbf0.tmp.sha256  tmp42ff984d.tmp.sha256  tmp523f2d0f.tmp.sha256
tmp5b9a79d6.tmp.sha256  tmp7345b525.tmp.sha256
```
If we also want to use them in ckan.exe we should move them together with the zip files, else not create them at all.

First option would need adjustments in `NetFileCache.Store()` where we move the zips.
Second option could be implemented by passing a bool `createHashFiles` through `NetModuleCache.GetFileHashSha*()` and `NetFileCache.GetFileHashSha*()` , and set it to `False` in `NetModuleCache.Store()`.

HebaruSan:
Interestingly, these files are not needed for ckan.exe; its cache is guarded by the hash checks in `Store`, so it assumes that anything already in the cache is valid. Only netkan.exe really needs to keep retrieving the hashes over and over. So the latest commit stops generating the hash files outside the cache folder.

PR #3089, opened 2020-06-28
Don't run workflow steps that require secrets on forks

DasSkelett:
## Problem
Since #3085 we run CKAN's CI with GitHub Actions. An advantage of this is that they also run in forked repositories automatically, without the need to set up an external service.

However, this also leads the workflows trying to execute all the deployment steps and the Discord notification, which doesn't make sense in forked repos and will ultimately fail since the secrets needed to run these steps aren't available.

## Changes
This PR adds `if`s to all the steps that need secrets to work. These `if`s check whether the needed secrets are set, and if not, skip the step.

Also fixed one network test failing since #3084.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,12 +47,12 @@ jobs:
       run: xvfb-run ./build test+only --configuration=${{ matrix.configuration }} --where="Category!=FlakyNetwork"
 
     - name: Send Discord Notification
-      if: failure()
       env:
         JOB_STATUS: ${{ job.status }}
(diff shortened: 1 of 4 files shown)


DasSkelett:
Umpf, looks like you can't access the `secrets` context in the `if`. Maybe I can setup the env first and use the env vars for the check?

DasSkelett:
Okay, confirmed skipping works: https://github.com/DasSkelett/CKAN/runs/815009204
Let's hope not-skipping works just as well in this repo.

PR #3090, opened 2020-06-28
Be more specific in GUI config parse error

HebaruSan:
## Problem

Apparently this message is confusing:

image

image

## Changes

Now 'it' is replaced with the file name and 'the folder' is replaced with the folder path. Hopefully this gives folks a clearer idea of what to do. I _think_ I got the change right for German, but maybe a "zum" needs to become a "zur" or something.

> CKAN.Kraken: Error trying to parse "/home/user/.local/share/Steam/steamapps/common/Kerbal Space Program/CKAN/GUIConfig.xml": Data at the root level is invalid. Line 2, position 1.. Try to move GUIConfig.xml out of /home/user/.local/share/Steam/steamapps/common/Kerbal Space Program/CKAN and restart CKAN.

### DIFF
--- a/GUI/Model/GUIConfiguration.cs
+++ b/GUI/Model/GUIConfiguration.cs
@@ -149,7 +149,10 @@ namespace CKAN
                         throw;
                     }
 
-                    string message = string.Format(Properties.Resources.ConfigurationParseError, path, additionalErrorData);
+                    var fi = new FileInfo(path);
+                    string message = string.Format(
+                        Properties.Resources.ConfigurationParseError,
+                        path, additionalErrorData, fi.Name, fi.DirectoryName);
                     throw new Kraken(message);
                 }
             }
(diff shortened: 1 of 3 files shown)


DasSkelett:
> Data at the root level is invalid. Line 2, position 1..

Wonder if it makes sense to remove the duplicate period, or if that will result in other error messages having none :D

HebaruSan:
How about a line break?

> $ _build/ckan.exe 
1294 [1] ERROR CKAN.ErrorDialog (null) - Unhandled exception:
CKAN.Kraken: Error trying to parse "/home/paul/.local/share/Steam/steamapps/common/Kerbal Space Program/CKAN/GUIConfig.xml": Data at the root level is invalid. Line 2, position 1.
> Try to move GUIConfig.xml out of /home/paul/.local/share/Steam/steamapps/common/Kerbal Space Program/CKAN and restart CKAN.

DasSkelett:
> Not sure how that would translate to Deutsch though.

This will work, since I plan to change the structure of German message to match the English one. :+1:

PR #3091, opened 2020-06-28
Run workflows in Mono Docker containers

DasSkelett:
## Motivation
The CI builds still take quite a long time to finish, since installing Mono takes around 10 minutes.
@techman83 had the excellent idea of running the builds in Docker containers from the images published [here](https://hub.docker.com/_/mono/?tab=tags&page=1).
Downloading the images should be way faster than installing and pre-compiling it.

## Changes
There exists a `container` keyword for workflow jobs: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer
If I understood the documentation correctly, it makes all steps of this job run in the specified container after pulling it from the Docker hub.
Let's see if it works.

Also realized that I messed up the Discord notifications in #3089 since the env var is named differently.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,19 +17,14 @@ jobs:
         mono: ['5.20', '6.4', '6.6', '6.8']
         configuration: [Debug, Release]
 
+    container:
+      image: mono:${{ matrix.mono }}
+
     steps:
     - uses: actions/checkout@v2
 
-    - name: Install Mono
-      run: |
-        sudo apt-get remove mono-complete --auto-remove --purge -y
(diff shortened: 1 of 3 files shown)


DasSkelett:
Aside from the fact that I forgot to remove the mono installation step, it looks like it correctly pulled the image. Also I think we are root by default in the container.

DasSkelett:
Well, that was fast :open_mouth:

DasSkelett:
> The workflow is not valid. .github/workflows/deploy.yml (Line: 15, Col: 14): Unrecognized named-value: 'env'. Located at position 1 within expression: env.RELEASE_MONO_VERSION

Almost...

DasSkelett:
Hmm, I could either specify the mono version directly in the `image:` line, since it's the only place we use it,
or create a matrix with a single item.

techman83:
This is really neat! That makes the builds way faster and less prone to issues with installing all the mono packages across so many versions.

PR #3096, opened 2020-07-01
Release date column

### LINKED ISSUE #1155: column for date mod updated

### LINKED ISSUE #1766: Add "Updated" filter type

### LINKED ISSUE #2916: [Proposal] Release Date

HebaruSan:
## Motivation

One of my earliest memories of chatting on the CKAN thread is suggesting that a release date column shouldn't be that hard to add:

- https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1272-chandrasekhar/page/14/&tab=comments#comment-3076850

This is a pretty popular request and would allow users to see which mods have been updated recently. Here's another time it came up:

- https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1272-chandrasekhar/page/72/&tab=comments#comment-3624808

## Background

After #3059 the Netkan Inflator has collected a full round of release date metadata from GitHub and Jenkins, and SpaceDock is in progress pending the next run of the webhooks only pass. The only thing left is the UI.

## Changes

Now a new sortable "Release date" column appears in GUI, so users can tell how recently mods were updated.
I attempted a German translation with Google Translate.

Fixes #1155.
Fixes #1766.
Closes #2916.

### DIFF
--- a/GUI/Controls/ManageMods.Designer.cs
+++ b/GUI/Controls/ManageMods.Designer.cs
@@ -63,6 +63,7 @@
             this.LatestVersion = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.KSPCompatibility = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.SizeCol = new System.Windows.Forms.DataGridViewTextBoxColumn();
+            this.ReleaseDate = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.InstallDate = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.DownloadCount = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.Description = new System.Windows.Forms.DataGridViewTextBoxColumn();
@@ -295,6 +296,7 @@
             this.LatestVersion,
             this.KSPCompatibility,
             this.SizeCol,
+            this.ReleaseDate,
             this.InstallDate,
             this.DownloadCount,
             this.Description});
@@ -398,6 +400,14 @@
             this.SizeCol.SortMode = System.Windows.Forms.DataGridViewColumnSortMode.Programmatic;
             this.SizeCol.DefaultCellStyle.Alignment = System.Windows.Forms.DataGridViewContentAlignment.MiddleRight;
             resources.ApplyResources(this.SizeCol, "SizeCol");
+            //
+            // ReleaseDate
+            //
+            this.ReleaseDate.Name = "ReleaseDate";
+            this.ReleaseDate.ReadOnly = true;
+            this.ReleaseDate.SortMode = System.Windows.Forms.DataGridViewColumnSortMode.Programmatic;
+            this.ReleaseDate.Width = 140;
+            resources.ApplyResources(this.ReleaseDate, "ReleaseDate");
             //
             // InstallDate
             //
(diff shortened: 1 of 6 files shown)


techman83:
This is really awesome, thanks for all your hard work and persistence with this @HebaruSan! (Added an edit for the proposal for the new field).

PR #3099, opened 2020-07-05
Stop release Action if build of dmg, deb, or rpm fails

HebaruSan:
## Problem

Initial builds of v1.28.0-PRE1 failed in the step that deploys the RPM file, because RPM doesn't allow dashes in version names.

However, the RPM build step itself succeeded when it should have failed. The deploy only failed because the RPM file didn't exist.

## Cause

`StartProcess` returns the failed exit code, but `.Does(() => ...)` simply ignores it. Cake expects us to throw exceptions when builds fail.

```cake
    .Does(() => StartProcess("make",
        new ProcessSettings { WorkingDirectory = "rpm" }));
```

## Changes

- Now we throw an exception if our `make` commands return an exit status other than 0, which will stop everything
- For a small performance gain, we no longer re-build ckan.exe every time for the dmg, deb, and rpm tasks; instead we build once and then use the same artifact for all three, using the `--exclusive` flag from #3093.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,12 +19,17 @@ jobs:
       - name: Installing runtime dependencies
         run: apt-get install -y libcurl4-openssl-dev xvfb
 
+      - name: Build ckan.exe
+        run: ./build --configuration=Release
+      - name: Run tests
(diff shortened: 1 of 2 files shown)


techman83:
That all makes sense to me.

PR #3107, opened 2020-07-09
Init Curl at netkan.exe and ckan.exe startup

### LINKED ISSUE #1441: NetKAN generates Curl Warning
I'm not sure when this started occurring, as I generated relevant errors for the error parsing and now they don't work!

```
791 [1] WARN CKAN.Curl (null) - Curl environment not pre-initialised, performing non-threadsafe init.
```

HebaruSan:
## Problem

image

## Cause

`Curl.CreateEasy` requires that `Curl.Init()` be called; if it hasn't been, it will be at that point. Supposedly `Curl.Init()` isn't thread-safe, so `CreateEasy` will print a warning about this since it can't guarantee that it's not being executed in some complicated threading scenario.

Netkan falls afoul of this by not calling `Curl.Init()` till it's needed.

## Changes

Now we init Curl at netkan and ckan startup, which isn't in a complicated threading scenario, so the message won't appear anymore.

Fixes #1441.

### DIFF
--- a/Cmdline/Main.cs
+++ b/Cmdline/Main.cs
@@ -55,6 +55,9 @@ namespace CKAN.CmdLine
             // This is on by default in .NET 4.6, but not in 4.5.
             ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12;
 
+            // Make sure Curl is all set up
+            Curl.Init();
+
             try
             {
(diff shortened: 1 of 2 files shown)


techman83:
Ok, this has bugged me for 5 years! Thanks for fixing it!!! :tada:

PR #3108, opened 2020-07-10
Cleanup project, update builds, fixes

Olympic1:
## Problem

I wanted to prepare CKAN for .NET 5.0, and some other stuff I'm working on, but I coudn't open the project in VS.

## Cause

Not sure what exactly, but I traced it back to CKAN.Core. Something was probably malformed in the `.csproj` file. I let VS recreate the file and now I can open it again.

## Changes

* Enabled deterministic builds (better security)
* Upgraded Cake to 0.38.4
* Upgraded NuGet to 5.6.0
* Upgraded Cake tools and addins
* Added vars for build versions (now we only change 2 values instead of 15)
* Fixed `.csproj` files
* Removed bootstrap packages from GUI and consoleUI (we don't use an installer)
* Fixed error with ILRepack @HebaruSan ([Issue](https://github.com/gluck/il-repack/issues/242#issuecomment-557738710)) - [(Culprit)](https://github.com/KSP-CKAN/CKAN/pull/3108/files?diff=unified&w=1#diff-6175e11f0503605833d89347a8051fb1L3-L4)
* Migrated NuGet packages from packages.config to PackageReference ([More info](https://docs.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference))
* Fixed links for reporting issues
* Removed duplicate images from `zh-CN.resx` file (400 kB saved)

**NOTES:**
* The PackageReferences will create `obj` folders outside `_build` but currently you can't change their output path.
* I upgraded netkan.exe to use CommandLineParser v2.8.0 but the CmdLine still uses the old one. I'm currently updating it but have to rewrite some stuff.


HebaruSan:
Huh, this builds fine in a `mono:latest` (6.8.0) image.

```
$ docker pull mono:latest
$ docker run -it --rm --entrypoint bash mono:latest
# apt update && apt install git
# git clone https://github.com/KSP-CKAN/CKAN.git
# cd CKAN/
# git remote add Olympic1 https://github.com/Olympic1/CKAN.git
# git fetch Olympic1
# git checkout -b update/builds Olympic1/update/builds
# ./build
```

No idea why the automated builds say they're missing reference assemblies.

HebaruSan:
Something in this created a bunch of `obj` folders outside of the `_build` directory:

```
$ ls -d */obj
Cmdline/obj  ConsoleUI/obj  Core/obj  GUI/obj  Netkan/obj  Tests/obj
```

This broke the compile when I switched to another branch. Should get those back into `_build`.

Olympic1:
> Something in this created a bunch of `obj` folders outside of the `_build` directory:

Those are the NuGet PackageReferences, I was looking into getting that into the build folder but that isn't possible atm

Olympic1:
.NET 4.8 support was added in mono 6.6.0

Netkan/Program.cs
```diff
@@ -30,21 +31,59 @@ public static class Program
...
 
-                // If we see the --version flag, then display our build info
-                // and exit.
-                if (Options.Version)
-                {
-                    Console.WriteLine(Meta.GetVersion(VersionFormat.Full));
-                    return ExitOk;
-                }
+                // Make sure Curl is all set up
+                Curl.Init();
```
HebaruSan, line 86: This is in here twice now, looks like a problem with conflict resolution.

Netkan/Program.cs
```diff
@@ -30,21 +31,59 @@ public static class Program
...
+                    Console.WriteLine(Meta.GetVersion(VersionFormat.Full));
+                }
+                else
+                {
+                    HelpText ht = HelpText.AutoBuild(parser, h =>
+                    {
+                        h.AddDashesToOption = true;
+                        h.AutoHelp = false;
+                        h.Heading = $"NetKAN {Meta.GetVersion(VersionFormat.Full)}";
+                        h.Copyright = $"Copyright � 2014-{DateTime.Now.Year}";
```
HebaruSan, line 54:
```suggestion
                        h.Copyright = $"Copyright © 2014-{DateTime.Now.Year}";
```

I think that's not a copyright symbol.

Netkan/Program.cs
HebaruSan, line 86:
```suggestion
```

Already present below, looks like a conflict resolution problem.

Netkan/CmdLineOptions.cs
```diff
@@ -49,11 +49,7 @@ internal class CmdLineOptions
         [Option("validate-ckan", HelpText = "Name of .ckan file to check for errors")]
         public string ValidateCkan { get; set; }
 
-        [Option("version", HelpText = "Display the netkan version number and exit")]
-        public bool Version { get; set; }
-
-        // TODO: How do we mark this as required?
-        [ValueOption(0)]
+        [Value(0, Required = true, HelpText = "Name of .ckan file to process.")]
```
HebaruSan, line 52: Will setting this to required break the bot? It runs it just with `--queues`, `--github-token`, and `--cachedir`.

https://github.com/KSP-CKAN/CKAN/blob/ba54c7972d05f6abe8efcbc787122a39acf0e6ef/Dockerfile.netkan#L7-L8

Netkan/Program.cs
Olympic1: Are you using the latest changes, for me this shows up as an addition.

HebaruSan: Yes, this is duplicated in the latest changes.

DasSkelett:
Can you squash these commits before merge? Rebase commits à la "Merge branch 'master' into ..." are always a bit confusing in the commit history.

Issue #3112, opened 2020-07-13
[Bug] netkan.exe no longer able to successfully use IAM 


techman83:
Background
----------
Noted nothing had been checked in 15 hours when I was curious to see if the GitHub [outage](https://www.githubstatus.com/incidents/j597fw8kv04c?utm_ts=1594627693) was impacting indexing.

Problem
-------
I am trying to get some more info, this applies specifically to IAM via EC2, testing locally the latest image is able to use token auth just fine.
```
754 [1] INFO CKAN.KSPManager (null) - Loading KSP instances from registry
790 [1] INFO CKAN.NetKAN.Processors.Inflator (null) - Using user-supplied cache at ckan_cache
32747 [1] FATAL CKAN.NetKAN.Program (null) - Unable to get IAM security credentials from EC2 Instance Metadata Service.
```
It does line up with a merge that update the AWSSDK. One thing that is making it awkward is we have to use 'Shell Form` for our [entrypoint](https://docs.docker.com/engine/reference/builder/#shell-form-entrypoint-example), making it quite difficult to pass in a debug flag via the task definition.

Debug output doesn't tell us a whole lot more
```
650 [1] DEBUG CKAN.NetKAN.Processors.QueueHandler (null) - Initializing SQS queue handler
681 [1] DEBUG CKAN.NetKAN.Processors.Inflator (null) - Initializing inflator
861 [1] INFO CKAN.KSPManager (null) - Loading KSP instances from registry
881 [1] INFO CKAN.NetKAN.Processors.Inflator (null) - Using main CKAN meta-cache at /home/netkan/.local/share/CKAN/downloads
1302 [1] DEBUG CKAN.NetKAN.Processors.QueueHandler (null) - Looking up URL for queue Inbound.fifo
32828 [1] FATAL CKAN.NetKAN.Program (null) - Unable to get IAM security credentials from EC2 Instance Metadata Service.
32830 [1] FATAL CKAN.NetKAN.Program (null) -   at Amazon.Runtime.DefaultInstanceProfileAWSCredentials.FetchCredentials () [0x00013] in <7bf78a1a90b146f1b4841c36916192cf>:0 
  at Amazon.Runtime.DefaultInstanceProfileAWSCredentials.GetCredentials () [0x0001b] in <7bf78a1a90b146f1b4841c36916192cf>:0 
[...]

PR #3114, opened 2020-07-14
Safer build updates

HebaruSan:
This is #3108, minus:

- The AWSSDK upgrade, because it broke the Inflator (see #3112)
- The switch from `packages.config` to `PackageReference`s, because it created build files outside the `_build` folder and broke builds when switching from one branch to another
- The SharpZipLib update, because it gave me some errors about `'ZipFile': type used in a using statement must be implicitly convertible to 'System.IDisposable'.` :middle_finger: 
- The CommandLineParser upgrade, because it created an inconsistency in which version of the library we're using (Netkan was upgraded, CmdLine wasn't; let's hold off till everything is ready)

Now some things are upgraded but everything should still work. The original motivation of being able to open Core in Visual Studio is probably satisfied, as most of those changes are preserved.

@techman83 can you please try building this and see if it works in the Inflator?


techman83:
Looks to be working!
```
447 [1] DEBUG CKAN.NetKAN.Processors.QueueHandler (null) - Initializing SQS queue handler
489 [1] DEBUG CKAN.NetKAN.Processors.Inflator (null) - Initializing inflator
707 [1] INFO CKAN.KSPManager (null) - Loading KSP instances from registry
749 [1] INFO CKAN.NetKAN.Processors.Inflator (null) - Using user-supplied cache at /home/netkan/ckan_cache
1169 [1] DEBUG CKAN.NetKAN.Processors.QueueHandler (null) - Looking up URL for queue Inbound.fifo
1319 [1] DEBUG CKAN.NetKAN.Processors.QueueHandler (null) - Looking up URL for queue Outbound.fifo
1329 [1] DEBUG CKAN.NetKAN.Processors.QueueHandler (null) - Queue URLs: <>
1330 [1] DEBUG CKAN.NetKAN.Processors.QueueHandler (null) - Looking for messages from <>
```

PR #3118, opened 2020-07-16
Purge CurlSharp

### LINKED ISSUE #3095: [Bug] Unable to load DLL 'libcurl'
Background
----------
* Operating System: Windows 10 1909 18363.900
* CKAN Version: 1.27.2
* KSP Version: 1.9.1.2788

No

Problem
-------
**Describe the bug**
CKAN errors with "Unable to load DLL 'libcurl'"

**Steps to reproduce**
* Clean install KSP (empty directory, including CKAN, verify files on Steam)
* Download latest ckan.exe
* Run it.
* Close it after it complains about registry locked and run it agian...
* See error

**Expected behavior**
Updater runs

error

```
Error in auto-update:
Unable to load DLL 'libcurl': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
```

### LINKED ISSUE #3115: [Bug] after update to dyson Ckan give a C++ runtime error "attempt to load c run library library incorrectly"
Background
----------
* Operating System: Win10
* CKAN Version: 1,28.0

* KSP Version: 1.10.0

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** No

Problem
-------
**Describe the bug**
Error message pop ups in screen ask you to click ok, when you do so ask you to select ksp version after you do so ckan starts

**Steps to reproduce**
* Go to... Ckan icon
* Click on...ckan icon
* See error

**Expected behavior**
* ckan start

**Screenshots (if applicable)** atttached
Ckan error

**CKAN error code (if applicable):**none
```
Unhandled Exception: System.BadImageFormatException: An attempt was made to load a program with an incorrect format. (Exception from HRESULT: 0x8007000B)
   at CurlSharp.NativeMethods.curl_global_init(Int32 flags)
   at CurlSharp.Curl.GlobalInit(CurlInitFlag flags)
   at CKAN.Curl.Init()
```

HebaruSan:
## Problem

Some Windows users are seeing this on launch:

screenshot

```
Unhandled Exception: System.BadImageFormatException: An attempt was made to load a program with an incorrect format. (Exception from HRESULT: 0x8007000B)
   at CurlSharp.NativeMethods.curl_global_init(Int32 flags)
   at CurlSharp.Curl.GlobalInit(CurlInitFlag flags)
   at CKAN.Curl.Init()
```

## Causes

In #3107, we started trying to initialize CurlSharp at startup to avoid thread-unsafe initialization, because the warning message about it was chronically spamming the Discord notifications channel for the NetKAN bot. Apparently CurlSharp needs to load the Visual C++ runtime to do its work, but some users have multiple copies of it in their `PATH`, which confuses Windows if it's a 32-bit copy:

- https://gis.stackexchange.com/a/96109
- https://kb.froglogic.com/squish/qt/troubleshoot/r6034-application-made-attempt-load-c-runtime-library-incorrectly/#cause-4---stray-msvc-runtime-dll-files
- https://support.esri.com/en/technical-article/000013127

(The misleading error message blaming the application for this is just one of dozens of ways Windows could do better here.)

Any call to `Curl.Init()` was supposed to catch and ignore all exceptions, but this wasn't done with the new calls in #3107 (because I didn't know about this requirement/assumption, because that's life with code that relies on exceptions for important behavior). Instead, the missing DLL exception was caught and ignored (which probably causes other problems by preventing the fallback from being skipped during downloads on Windows), and this BadImageFormatException is uncaught.

## Background

- https://github.com/KSP-CKAN/CKAN/wiki/libcurl

According to @techman83, the CurlSharp fallback was added the first time a new TLS version requirement broke everything. (The second time was #2293, with which CurlSharp was memorably unhelpful, probably because it's not for Windows.)

The fallback is no longer needed (because .NET and Mono both support the latest TLS) and is an obstacle to refactoring the network code. Any change has to consider the possibility of multiple connections being attempted for the same URL with exceptions thrown, caught, re-thrown, etc. It can be difficult even to identify where a given change should go, and to avoid breaking something.

## Workaround

@Dominiquini [figured out](https://github.com/KSP-CKAN/CKAN/issues/3115#issuecomment-659634251) that clearing the `PATH` environment variable helps!

```cmd
C:\Windows\System32\cmd.exe /c "SET PATH= && START /D ^"C:\Program Files (x86)\CKAN" ckan.exe"
```

(Details may vary, not guaranteed effective on any particular Windows system, use at your own risk.)

## Changes

Now all the libcurl stuff is removed. We try to download a URL once with WebClient, and if that doesn't work, too bad.

As well, all projects now have this set for all builds:

```xml
    <Prefer32Bit>false</Prefer32Bit>
```

Previously it was only set for Debug builds, which may have played a role in how these DLLs were loaded.

Fixes #3115.
Fixes #3095.


.github/workflows/build.yml
```diff
@@ -66,7 +66,7 @@ jobs:
 
     steps:
       - name: Install runtime dependencies
-        run: sudo apt-get update && sudo apt-get install -y libcurl4-openssl-dev
+        run: sudo apt-get update
```
techman83, line 69: We probably don't need this step at all. `update` only refreshes the package lists.

Dockerfile.netkan
```diff
@@ -1,6 +1,6 @@
 FROM mono:5.20.1
 RUN useradd -ms /bin/bash netkan
-RUN apt-get update && apt-get install -y libcurl4-openssl-dev
+RUN apt-get update
```
techman83, line 3: Like previous comment :slightly_smiling_face:

PR #3127, opened 2020-07-27
Cache build tools

HebaruSan:
- Now all build actions cache `_build/tools`
  - NuGet
  - NUnit
  - ILRepack
- Now all build actions cache `_build/cake`
  - Cake.SemVer
  - semver
  - Cake.Docker
- Now the release build step caches `_build/tools`, `_build/cake`, `_build/lib/nuget`, and `~/.nuget/packages`
- Indenting in the yml files is more consistent

This should allow quicker builds with fewer downloads.


HebaruSan:
Hmm, that last commit had no effect. The NetCore builds should be caching the new folders now, but they didn't...

PR #3139, opened 2020-08-11
Don't fail modpack export when having unindexed mods installed

### LINKED ISSUE #3138: [Bug] Exporting a modpack using CKAN 1.27.2 on Debian 4.19.118-2 locks up CKAN
Background
----------
* Operating System: Debian 4.19.118-2
* CKAN Version: 1.27.2
* KSP Version: 1.9.1.2788

* Mono version: 5.18.0.240

Some minor cfg file nudges, nothing earth shaking

Problem
-------
When attempting to export a modpack, the tab comes up and I can edit the name, identifier, and description, but not the version dropdowns,  The window where the list of mods is supposed to appear does not populate and no buttons do anything other than reflecting a click by changing color momentarily.  One cannot cancel out as it behaves as all other buttons.  Using the OS window widget to close CKAN does not exit, though it does change color so is reflecting the click.  I have to kill the mono process to get CKAN to end, then delete the lock file (hope that is what I should do and all I should do)

**Steps to reproduce**
* Go to...Export Modpack
* Click on...Cancel, or Export, or X-window close widget
* See error

**Expected behavior**
Populates modlist, exports, doesn't hang

```
PLACE CODE HERE !
```

DasSkelett:
## Problem
When you have mods or mod versions installed that are not in the list of available modules ("DarkKAN modules"), the modpack export tab freezes.
This is caused by `registry.LatestAvailable()` in `AddGroup()` throwing an `ModuleNotFoundKraken` for unknown mods when trying to fetch the abstract for each installed mod.

## Changes
Now we search the registry for an `InstalledModule` with the given identifier, and use its abstract property. We should always have an `InstalledModule` for each module we're handling in the export logic, and if for some reason we haven't, it shouldn't break anything.

Fixes #3138 
...maybe, hard to say without the registry or any information about installed mods, but it solved it for me.
@mgalyean if you can still reproduce the problem and want to try a debug build of ckan.exe with this fix implemented, let me know and I can upload one. That way we could confirm that it does indeed solve your problem too.

### DIFF
--- a/Core/Registry/IRegistryQuerier.cs
+++ b/Core/Registry/IRegistryQuerier.cs
@@ -116,7 +116,7 @@ namespace CKAN
         /// This includes DLLs, which will have a version type of `DllVersion`.
         /// This includes Provides if set, which will have a version of `ProvidesVersion`.
         /// </summary>
-        Dictionary<string, ModuleVersion> Installed(bool include_provides = true);
+        Dictionary<string, ModuleVersion> Installed(bool withProvides = true, bool withDLLs = true);
 
         /// <summary>
         /// Returns the InstalledModule, or null if it is not installed.
--- a/Core/Registry/Registry.cs
+++ b/Core/Registry/Registry.cs
@@ -904,20 +904,23 @@ namespace CKAN
         /// <summary>
         /// <see cref = "IRegistryQuerier.Installed" />
         /// </summary>
-        public Dictionary<string, ModuleVersion> Installed(bool withProvides = true)
+        public Dictionary<string, ModuleVersion> Installed(bool withProvides = true, bool withDLLs = true)
         {
             var installed = new Dictionary<string, ModuleVersion>();
 
-            // Index our DLLs, as much as we dislike them.
-            foreach (var dllinfo in installed_dlls)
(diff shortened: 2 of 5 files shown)


HebaruSan:
> ("DarkKAN modules")

Is that a thing? Like dark matter, dark energy, and the dark web? :grin:

HebaruSan:
One thought, the typical user won't be able to install "DarkKAN modules" without adding a new metadata repo (or downloading a .ckan file). Does it make sense to include them in exported modpacks? I guess it could, since maybe the modpack creator will include instructions on how to add a repo. But we might want to display a warning about that somewhere.

HebaruSan:
Now that I think about it, we have a possibly sensible thing to do with these: Put them in the Ignored group by default. That way the user can just click Export to generate a modpack that will work for other people, but if he or she really wants to have weird custom modules listed in the modpack, the option is there to do that manually.

HebaruSan:
Submitted DasSkelett/CKAN#9 to try to do that...

HebaruSan (approved):
Admittedly I'm biased, but these changes look fantastic to me. :wink:

PR #3149, opened 2020-09-05
Escape '&' in mod list, consistent path separators in exception

DasSkelett:
## Problems
Ampersands (`&`) in mod names or abstracts in the mod list are currently interpreted as hint to create automatic keyboard shortcuts (visualized through underscores), instead of printing them as literal.
See for example the mod `DumpAndBurn`, with the name `Dump&Burn!`:
unescaped-ampersand

Error messages of krakens thrown while trying to load the `GUIConfig.xml` have a mix of `/` and `\` as path seperators on Windows, like:
```
Unhandled Exception:
CKAN.Kraken: Error trying to parse "c:/program files(x86)/steam/SteamApps/common/Kerbal Space Program/CKAN\GUIConfig.xml": Root element missing.
```

## Changes
The texts in the name, author and abstract are now escaped using `&&` in `ModList.MakeRow()`. I didn't spot another location where they would need escaping (tested recommendation tab and modpack export tab).

The exception about broken `GUIConfig.xml` files now gets the path string from a `FileInfo` object, which should always be formatted consistently and correctly for the user's OS.

### DIFF
--- a/GUI/Controls/ManageMods.Designer.cs
+++ b/GUI/Controls/ManageMods.Designer.cs
@@ -232,16 +232,16 @@
             this.FilterAllButton.Size = new System.Drawing.Size(307, 30);
             this.FilterAllButton.Click += new System.EventHandler(this.FilterAllButton_Click);
             resources.ApplyResources(this.FilterAllButton, "FilterAllButton");
-            // 
+            //
             // FilterTagsToolButton
-            // 
+            //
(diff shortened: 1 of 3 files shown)


HebaruSan (approved):

HebaruSan:
In Mono, one spot is wrong and one is right:

image

So it might not be possible to make the grid work correctly for everyone, but the mod info label is probably fixable.

PR #3156, opened 2020-09-23
Remove invalid zip files in NetKAN

DasSkelett:
## Problem
Invalid zip files that fail the `ZipValid` check fill up the bot's disk over time. The tempfiles but never get removed or moved into the cache, thus filling the disk with every indexer run.
In contrast, files failing to download at all get removed immediately, and files failing inflation later on are still stay in the cache and don't get redownloaded for the next try.

## Changes
Simply remove the file if it fails the `ZipValid` check.

This has a downside when using netkan manually, if one would want to investigate the zip file after netkan fails (I remember [my comment](https://github.com/KSP-CKAN/CKAN/pull/2903#issuecomment-546433290)...). Now you have to download the file manually again to check it out.
I tried to think of an easy solution, like keeping the file if `--debug` is set, but as of right now we don't save that anywhere globally accessible in netkan.
I think it's worth it though, and not every mod is as big as HumanStuff, so redownloading the file manually isn't much of a deal usually.

### DIFF
--- a/Netkan/Services/CachingHttpService.cs
+++ b/Netkan/Services/CachingHttpService.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 using log4net;
 using CKAN.NetKAN.Model;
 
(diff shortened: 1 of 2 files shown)


HebaruSan (approved):
Here's hoping that restarting the container will also purge the temp folder.

PR #3165, opened 2020-10-02
ConsoleUI fixes for DLC

### LINKED ISSUE #3140: NullReferenceException on ctrl-F in ConsoleUI
Hello , when i try to search mod using ctrl + F and start digits CKAN crash with this screen
Schermata 2020-08-15 alle 20 34 37

HebaruSan:
## Problems

- If you have a DLC installed, and there is no `"kind": "dlc"` module for it, ConsoleUI throws an exception when searching
- ConsoleUI thinks it can install and uninstall DLC (you can add them to the change set), but if you try it, an exception is thrown
- The mod info page for a DLC shows an exception where it would normally say where the mod is hosted
- The `store` and `steamstore` resources aren't shown anywhere in ConsoleUI

## Causes

- The `CkanModule.Searchable*` properties are null because we create a fake DLC CkanModule without calling `CalculateSearchables`
- The "+"/"-" key binds and screen tips were not DLC-aware
- `CkanModule.download` is null for DLC, but `ModInfoScreen.HostedOn` did not allow for this

## Changes

- Now we call `CalculateSearchables` when we make a fake DLC CkanModule
- Now ConsoleUI won't let you try to install or uninstall DLC
- Now the hosted on area will tell you where you can buy the DLC
- Now the mod info screen's hamburger menu contains the store and steamstore links for DLC

Fixes #3140.

### DIFF
--- a/ConsoleUI/ModInfoScreen.cs
+++ b/ConsoleUI/ModInfoScreen.cs
@@ -104,10 +104,13 @@ namespace CKAN.ConsoleUI {
             AddBinding(Keys.Escape, (object sender) => false);
 
             AddTip("Ctrl+D", "Download",
-                () => !manager.Cache.IsMaybeCachedZip(mod)
+                () => !manager.Cache.IsMaybeCachedZip(mod) && !mod.IsDLC
             );
             AddBinding(Keys.CtrlD, (object sender) => {
-                Download();
+                if (!mod.IsDLC)
+                {
+                    Download();
+                }
                 return true;
             });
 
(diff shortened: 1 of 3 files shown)


DasSkelett:
It still offers to mark as auto-installed (F8), or to download it in the details view (CTRL-D). Both keys obviously throw exceptions.

image
image

HebaruSan:
Nice catches! :+1: 
Should be fixed in latest commit.

DasSkelett (approved):
Yep, looks good now! Thanks!

PR #3166, opened 2020-10-03
Retry failed web requests, fix install stanza error

HebaruSan:
## Problems

- Since #3074, a mod with a bad install stanza says, "Sequence contains no elements" instead of explaining that the stanza didn't match any files
- Even if the previous point is fixed, a mod using the default install stanza will throw a null reference error if it doesn't match any files
- Warnings about fetching remote version files don't print the remote URL (which would have been convenient for the following point)
- Prior to the recent flood of SpaceDock "The request was canceled" spam, after #3118, the netkan-bot channel was plagued with these errors:
  image
  This is mostly harmless, _unless_ a mod author is actually using the remote version file to update compatibility after release, in which case it can cause CKAN's metadata to temporarily revert to the at-release state.

## Causes

- The code to find `shortestMatch` threw an exception if there were no matches, when it should have defaulted to null
- `CkanModule.DescribeInstallStanzas` assumed `install` would not be null
- ksp-avc.cybutek.net and ksp.spacetux.net are flaky servers, probably because they're run by individuals rather than big companies with cloud resources. They seem to fail in about 0.1% of requests; prior to #3118, we fell back to CurlSharp in these cases, but after that PR the fallback is removed.

## Changes

- Now errors like `Module contains no files matching: find="Cosmogator"` are back
- Now `CkanModule.DescribeInstallStanzas` knows how default install stanzas work
- Now remote version file fetch errors print the URL
- Now if a URL fails for non-protocol reasons (i.e., if it's a network failure rather than a HTTP 404), we try again immediately, then after waiting 0.1s, then 0.2s, then 0.3s. If it still doesn't work, we finally give up and fail as before.

### DIFF
--- a/Core/Net/Net.cs
+++ b/Core/Net/Net.cs
@@ -5,10 +5,11 @@ using System.Linq;
 using System.Net;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Threading;
 using Autofac;
 using ChinhDo.Transactions.FileManager;
-using CKAN.Configuration;
 using log4net;
+using CKAN.Configuration;
 
 namespace CKAN
 {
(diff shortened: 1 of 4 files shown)


DasSkelett (approved):
Nice, thanks! Can't wait for the bot to stop spamming all the download errors.

HebaruSan:
Yeah, I may un-mute the netkan-bot channel soon!

PR #3167, opened 2020-10-04
Search timer for Linux

### LINKED ISSUE #3120: [Bug] letters aren't in the typing order in the search box
Background
----------
* Operating System: Ubuntu 20.04
* CKAN Version: 1.28.0

Problem
-------
As I type in the search textbox ckan starts filtering mods, for some reason letters ends in a different order than how I typed them. Example, if I type quickly "kerbal", in the search box I see "kbelra". If there are no mods because none match the filter then additional input doesn't end mixed up.
I have to type one letter at a time and wait for the list of mods to update before writing the next letter for get what I want.

This is an issue I been having for a while in previous versions as well, is not particular to 1.28.0 and probably is not noticeable in fast computers.

**Steps to reproduce**
* type any word qucikly in the search box
* check if the text in the search box is correct.

HebaruSan:
## Background

GUI search currently works slightly differently per platform: On Mac, it waits 700 milliseconds after you stop typing to update the filters, but on Windows and Linux it updates immediately after each keystroke.

## Problems

On Linux:

- Sometimes (exact conditions unclear) if you type a search quickly, the characters appear out-of-order in the search box, which obviously searches for something other than what you intended
- Even if you don't suffer from the above problem (I had seen it happen previously but was not able to reproduce it while working on this fix), the UI updates are kind of slow and unreliable; there's an impression of repeated flickering and the cursor moves before the characters appear.

## Cause

The `DataGridView` updates more slowly on Mono than on Windows, and it happens on the GUI thread, so updating it immediately after each keystroke inhibits the text box from responding. (I tried performing all of CKAN's own search logic in a background thread and then only applying the changes in the GUI thread, but it made no noticeable difference.)

## Changes

- Now the search timer is enabled for all Mono clients (Linux **and** Mac) rather than just Mac
- The timer is changed from 700 milliseconds to 500 milliseconds to make it seem somewhat more responsive
- Pressing Enter triggers the search immediately without delay
- Clearing the search box (e.g. with Escape) triggers the search immediately without delay

Fixes #3120.

### DIFF
--- a/GUI/Controls/EditModSearch.cs
+++ b/GUI/Controls/EditModSearch.cs
@@ -189,6 +189,13 @@ namespace CKAN
             switch (e.KeyCode)
             {
                 case Keys.Enter:
+                    // Bypass the timer for immediate update
+                    e.Handled = true;
+                    e.SuppressKeyPress = true;
+                    filterTimer?.Stop();
+                    TriggerSearch();
+                    break;
+
                 case Keys.Up:
                 case Keys.Down:
                 case Keys.PageUp:
@@ -204,9 +211,9 @@ namespace CKAN
 
         private void TriggerSearchOrTimer()
         {
-            if (Platform.IsMac)
+            if (Platform.IsMono && !string.IsNullOrEmpty(FilterCombinedTextBox.Text))
             {
-                // Delay updating to improve typing performance on OS X.
+                // Delay updating to improve typing performance on OS X and Linux
                 RunFilterUpdateTimer();
             }
             else
@@ -227,7 +234,7 @@ namespace CKAN
             {
                 filterTimer = new Timer();
                 filterTimer.Tick += OnFilterUpdateTimer;
-                filterTimer.Interval = 700;
+                filterTimer.Interval = 500;
                 filterTimer.Start();
             }
             else


GUI/Controls/EditModSearch.cs
```diff
@@ -189,6 +189,12 @@ private void FilterTextBox_KeyDown(object sender, KeyEventArgs e)
             switch (e.KeyCode)
             {
                 case Keys.Enter:
+                    // Bypass the timer for immediate update
+                    e.Handled = true;
```
DasSkelett, line 193: Apparently Windows plays an annoying warning sound when pressing enter in single line text boxes. Adding `e.SuppressKeyPress = true` avoids this. (https://stackoverflow.com/a/16350929)

DasSkelett (approved):
I didn't test on Linux, however I did test on Windows after removing the OS check of the if clause. Works fine, but would be nice if we could also get rid of the warn ding mentioned above.

GUI/Controls/EditModSearch.cs
HebaruSan: Huh, I must have that sound disabled in my Windows somehow. Still, I see no harm in adding this. Force-pushed.

PR #3194, opened 2020-11-12
Upgrade DLL build to .NET 5

DasSkelett:
## Motivation
Mostly because I already have the changes and because @HebaruSan said I should do a PR :D
Also [.NET 5 has been released yesterday](https://devblogs.microsoft.com/dotnet/announcing-net-5-0/) as "the first release in [Microsoft's] .NET unification journey."

The Core DLL isn't really used anywhere on its own, especially in its .NET (Core) form, so this shouldn't break anything either.

One day we want to create a new GUI based on MAUI, so it's not a bad idea to keep the Core up to date.

## Changes
* Changed the `netstandard2.0` `TargetFramework` to `net5.0` in `CKAN-core.csproj`. The `net45` one is untouched.
* Changed the `netcoreapp3.1` `TargetFramework` to `net5.0` in `Tests.csproj`. The `net45` one is untouched.
* Put our own `IEnumerable.TohashSet()` implementation between `#if NET45 ... #endif`, since .NET 5 (and even .NET Framework 4.7.2) comes with a native implementation. We'll have to update this if/when we ever change our Framework build away from 4.5.
* Related, changed a line in the test that called `TohashSet()` via the full namespace path as static funtion, passing the object, to calling it on the object.
* Upgraded Cake to v0.38.5 which has improved .NET 5 support. I've also just seen that there is [Cake v1.0.0-rc1 available now](https://cakebuild.net/blog/#cake-v1-0-0-rc1-released), but I haven't tested it yet.
* Upgraded the cake build script to target `net5.0` for .NET (Core) builds.
* And changed our GitHub workflow to install .NET 5 prior to running our netcore build script.

### DIFF
--- a/Core/Configuration/JsonConfiguration.cs
+++ b/Core/Configuration/JsonConfiguration.cs
@@ -370,7 +370,7 @@ namespace CKAN.Configuration
                     // Ensure the directory exists
                     new FileInfo(configFile).Directory.Create();
 
-#if !NETSTANDARD
+#if !NETCOREAPP
                     // If we are not running on .NET Standard, try to migrate from the real registry
                     if (Win32RegistryConfiguration.DoesRegistryConfigurationExist())
                     {
--- a/Core/Configuration/JsonConfiguration.cs
+++ b/Core/Configuration/JsonConfiguration.cs
@@ -371,7 +371,7 @@ namespace CKAN.Configuration
                     new FileInfo(configFile).Directory.Create();
 
 #if !NETCOREAPP
-                    // If we are not running on .NET Standard, try to migrate from the real registry
+                    // If we are not running on .NET Core, try to migrate from the real registry
                     if (Win32RegistryConfiguration.DoesRegistryConfigurationExist())
                     {
                         Migrate();
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,7 +79,7 @@ jobs:
       - name: Setup .NET Core
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: '3.1.301'
+          dotnet-version: '5.0.x'
       - name: Restore cache for _build/tools
         uses: actions/cache@v1
         with:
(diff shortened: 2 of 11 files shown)


Core/Configuration/JsonConfiguration.cs
```diff
@@ -370,7 +370,7 @@ private void LoadConfig()
                     // Ensure the directory exists
                     new FileInfo(configFile).Directory.Create();
 
-#if !NETSTANDARD
+#if !NETCOREAPP
                     // If we are not running on .NET Standard, try to migrate from the real registry
```
HebaruSan, line 374: Should this comment about .NET Standard be updated now that the condition is no longer `!NETSTANDARD`?

HebaruSan (changes requested):
There are a bunch of warnings like this in the PR's code listing:

image

I guess they have a point. We could just not catch those exceptions.

DasSkelett:
> There are a bunch of warnings like this in the PR's code listing:
> 
> I guess they have a point. We could just not catch those exceptions.

Removed the try-catch blocks, and instead moved the explaining messages into proper code documentation. There's only this one left, something something Code Access Security. I don't really know that, so I have to do some reading to find out what it's for and how to migrate that to .NET Core.

HebaruSan:
I would say nuke it, but the FileSystemWatcher documentation still mentions it:

- https://docs.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher?view=net-5.0

> This class contains a link demand and an inheritance demand at the class level that applies to all members. A SecurityException is thrown when either the immediate caller or the derived class does not have full-trust permission. For details about security demands, see Link Demands.

I don't want to have to deal with SecurityExceptions unexpectedly. Let's leave this as-is; hopefully, if it runs on a runtime that supports CAS it will work OK, and if it runs on a runtime that doesn't support CAS no exception will be thrown regardless.

HebaruSan (approved):
:canoe:

PR #3180, opened 2020-10-24
Allow installation into Ships/Script

HebaruSan:
## Problems

- kOS stores scripts in the `Ships/Script` folder under the game root. Currently CKAN doesn't allow this, and KSP-CKAN/NetKAN#7967 is in stasis because it needs to install kOS scripts to that folder.
- `ShipSaveSplicer` is being installed into `GameData/veSplicer` because it starts with `Ships`

## Changes

- Now we allow installation into `Ships/Script`, including creation of new directories.
  **Since this requires an update of the schema, the next release of CKAN will need to be v1.29.0 if this is merged.**
- Now we require a path to start with `Ships/` with a slash to trigger that truncation, so `ShipSaveSplicer` will be installed correctly.
  Fixes #3187.

PR #3196, opened 2020-11-14
Fix error when removing file from GameRoot

### LINKED ISSUE #3195: mod won't remove
Background
----------
* Operating System: Windows 8.1
* CKAN Version: 1.29.0
* KSP Version: 1.9.1.2788

I removed the mod folder for the mod and deleted it

Problem
-------
**Describe the bug**
so I have enjoyed using ckan for playing modded ksp for about 6 months and I can't get a mod to remove
when I go to remove the mod which is Advanced fly by wire (windows) 1.8.3.2 it says /is already absolute error during installation

**Steps to reproduce**
* Go to... advanced fly by wire install it
* Click on... remove
* See error /is already absolute error during installation

**Expected behavior**
shouldn't remove the mod and give an error message

Screenshot (1)

```
PLACE CODE HERE!
```/is already absolute

### PR DESCRIPTION
**Pull request by HebaruSan**
## Problem

If you install a mod that installs a file to GameRoot (e.g. [AdvancedFlyByWire](https://github.com/KSP-CKAN/CKAN-meta/blob/master/AdvancedFlyByWire/AdvancedFlyByWire-1.8.3.2.ckan)), and then remove it, the removal fails:

```
About to remove:

 * Advanced Fly-By-Wire (Windows) 1.8.3.2
 * Toolbar Controller 1:0.1.9.4
 * ClickThrough Blocker 1:0.1.10.14
 * Zero MiniAVC 1:1.1.0.1
Removing AdvancedFlyByWire...
/ is already absolute
Error during installation!
```

## Cause

In #3180 we fixed `KSPPathUtils.NormalizePath` to no longer incorrectly convert `"/"` to `""`, because those are two different directories, and normalization is supposed to generate a standardized format of the **same** directory.

That change indirectly confused this loop:

https://github.com/KSP-CKAN/CKAN/blob/a494e1dafb5ed2f19307bdc6180a163c6f5ef9ed/Core/ModuleInstaller.cs#L900-L908

For a file in GameRoot, the absolute path to the game folder is included in the `directories` parameter of `ModuleInstaller.AddParentDirectories`, which results in `relativeHead` containing the empty string. Then in the first iteration of the loop, `buildPath` is set to `/`, and `ToAbsolute` calls `NormalizePath` on it, after which it is still `/`, so `ToAbsolute` throws because that path is already absolute.

## Considered and not done

We _could_ change `ToAbsolute` to use some method other than `NormalizePath` to massage its inputs. I think it would be better to continue to uncover places in Core where paths are treated sloppily and fix them. `ToAbsolute("/", "/myroot/")` _should_ throw an exception if `ToAbsolute("/afile", "/myroot/")` would.

## Changes

Now if `relativeHead` is empty, that means there are no parent folders to check, so we skip that loop.

Fixes #3195.

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -775,7 +775,7 @@ namespace CKAN
                         }
                         else
                         {
-                            // Add this files' directory to the list for deletion if it isn't already there.
+                            // Add this file's directory to the list for deletion if it isn't already there.
                             // Helps clean up directories when modules are uninstalled out of dependency order
                             // Since we check for directory contents when deleting, this should purge empty
                             // dirs, making less ModuleManager headaches for people.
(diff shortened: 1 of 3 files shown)


DasSkelett:
Do you think this is worth adding tests for? `ModuleInstallerDirTest.cs` already has some for `AddParentDirectories()`. I feel like this is something we could accidentally break again by changing code only related over multiple corners.

HebaruSan:
Yes, I agree that's a good idea.

HebaruSan:
OK, added a test for that, as well as for `Ships/Script` since there's already a framework for it.

DasSkelett (approved):
Great, let's release our users from the install-but-never-uninstall-again trap :smile:

PR #3175, opened 2020-10-13
Search dropdown

HebaruSan:
## Background / motivation

In #3041, we replaced the 3 old search fields at the top with 1 at the bottom, which can be expanded to to reveal 8 detail-level fields:

 

Initially during development I tried to keep the search field at the top, but I had trouble figuring out how to create a dropdown in WinForms that floats over the grid with editable controls in it, and it wasn't clear whether that was important. An expand/collapse (in the style of the above screenshots) on the top was jarring because it caused the top of the grid to jump up and down. Moving it to the bottom allowed those changes to be finished up and released with less risk of bugs, and I found examples of other applications that do just that:

 

We included this change in [the v1.28.0-PRE1 pre-release](https://github.com/KSP-CKAN/CKAN/releases/tag/v1.28.0-PRE1) and didn't hear a peep, but after it was fully released, multiple users expressed a preference for having the search at the top (see #3163).

## Changes

Now the search is moved back to the top:

image

If you expand it, the other fields appear in a dropdown:

image

This is not easy to do in WinForms (there's not one simple switch to flip to turn it on; rather you have to tweak many poorly-documented settings in concert in just the right way), so careful testing is needed and quirks are likely to be found. In effect, the dropdown is a new top-level/standalone window with special flags set, with which the main window communicates from the outside.

Fixes #3163.

PR #3198, opened 2020-11-16
Search UI fixes

HebaruSan:
## Problem

I force-pushed some Windows UI fixes for #3175, but when code review was finished, I mistakenly merged the original Linux branch instead, resulting in the release of some annoyances that users might eventually complain about.

This was why GitHub did not automatically close that PR or the issue it fixed upon merge. I thought it was due to something else. Lesson learned, listen to GitHub!

## Changes

Luckily I still had the Windows branch sitting around.

- The search dropdown button is flat style instead of 3D to better fit in to the other UI elements
- The dropdown window is no longer always-on-top
- Pressing Esc in the dropdown will first clear the focused text box, if not empty, otherwise close the dropdown
- The labels in the search dropdown are 10 pixels wider to give enough space for all the text


DasSkelett (approved):
Didn't test on Windows now, but the changes do match the ones from the other PR, so I guess it's fine :+1:

Issue #3185, opened 2020-10-31
Path does not correspond to a KSP folder (macOS Catalina)


I-need-help140:
I am on the latest version of macOS Catalina, and I am trying to use Ckan on my mac. I have been playing KSP for three years but on a PC and my PC recently broke and I have been trying to play on my MacBook. Whenever I try to put in my file directory, this error shows up *Path does not correspond to a KSP folder!* 
<img width="558" alt="Screen Shot 2020-10-31 at 10 43 04 AM" src="https://user-images.githubusercontent.com/73739262/97776167-35504800-1b66-11eb-81f4-b504566bc44e.png">

I will appreciate any help.

HebaruSan:
Try `/Applications/Kerbal Space Program` instead.

I-need-help140:
Yes,  that worked. Thank you. 
<img width="558" alt="Screen Shot 2020-10-31 at 7 24 12 PM" src="https://user-images.githubusercontent.com/73739262/97786862-af0c2400-1bae-11eb-8dd9-640a018d2f6b.png">

I-need-help140:
Could you also help me on how to use this version of CKAN. What commands I have to put in. I have very basic experience with *Terminal*

HebaruSan:
The bottom of the window lists the actions available on each screen. They take the form of the name of the keys to press and then a short description of what that keystroke does. For example, to quit CKAN from the above screen, you could press Ctrl+Q.

I-need-help140:
Okay, so how would I list all currently available mods and install them? 
I'm not sure if this is right but is it: *control+ R/F10* then select the mod and. *Enter*?

HebaruSan:
If you are still on the above screen, then "Enter - Select" is how to satisfy the "Select a game instance:" prompt.

I-need-help140:
<img width="558" alt="Screen Shot 2020-10-31 at 9 33 32 PM" src="https://user-images.githubusercontent.com/73739262/97789470-087d4e80-1bc1-11eb-9239-f5d4c6844f0c.png">
Is this right?

HebaruSan:
Yes.

I-need-help140:
Thank you so much for your help and I'm sorry for my inexperience with macOS causing you trouble.

HebaruSan:
You're welcome, and it's no trouble.

Issue #3047, opened 2020-05-07
"Path does not correspond o a KSP folder" on MACOS 10.15.5


Hazuka:
I have a big problem whili initiated the CKAN nogui on macos 10.15
I cant follow the path of my KSP folder (on Application/Kerbal\ Space\ Program)

Any ideas ?

https://imgur.com/sWBnblW
<img width="682" alt="Capture d’écran 2020-05-07 à 20 31 47" src="https://user-images.githubusercontent.com/17553127/81331864-8513f200-90a2-11ea-92fd-b17f8d648c33.png">

Thank you`

HebaruSan:
What if you remove the back slashes? I don't think they're required (but I would have to check).

Hazuka:
Sorry i cant find a way to add the labels xD

Hazuka:
> What if you remove the back slashes? I don't think they're required (but I would have to check).

Omg its that...
Thank you xD

HebaruSan:
Glad you got it working!

HebaruSan:
> Sorry i cant find a way to add the labels xD

GitHub only allows dev team members to manipulate labels (in general, not just for CKAN). It's not super important, just a convenient way to organize things.

I-need-help140:
Could you help? I do not understand what to do, I tried. Application/Kerbal Space Program and Application Kerbal Program.

PR #3199, opened 2020-11-16
Streamline Mac onboarding

HebaruSan:
## Motivation

Mac users have been struggling with how to start using Console UI, see #3047, #3185, and recent discussions in the macos support channel in Discord. On close inspection, not all the on-screen prompts are as clear or helpful as they could be.

## Problems

If you start ConsoleUI with no instances and you don't have KSP installed in Steam, you get the KSP instance list screen:

select a game instance

- There's not much of a hint of what to do (the right answer is to press <kbd>a</kbd> to add an instance.)
- "Select a game instance" is incomplete on a screen where you can also add instances

When adding an instance:

enter a value

- The fields have the default `<Enter a value>` ghost text, which could be more helpful
- Not all users may necessarily know what a path is, and if they do, they might not be sure about what format to use (macOS uses `:` as a path separator in some contexts and `/` in others)

Multiple users have asked for help with this exact screenshot:

does not correspond

- The header says "Edit KSP Instance", but we're adding
- I think they got here by dragging and dropping their KSP folder into Terminal, which "helpfully" adds backslashes on the incorrect assumption that the string will be used by `bash`
- I think `/Applications/Kerbal Space Program` is where the KSP Store installer installs KSP

## Changes

- Now if you're on Mac, you don't have any instances, and your game is located in `/Applications/Kerbal Space Program`, we detect it automatically like we can with Steam
- New more inclusive help label on the instance list screen:
  image
- Now if you don't have any instances and we can't detect the game automatically, Console UI will jump to the add instance screen instead of the empty list

When adding an instance:

image

- Add instance title changed from "Edit KSP Instance" to "Add KSP Instance"
- Field ghost text now gives more info about what the fields are for
  - Also done for the add repo screen, which was the only other spot where we weren't customizing the ghost text
- A help label under the path field shows `<Program Files>\Kerbal Space Program` or `/Applications/Kerbal Space Program` in your OS's path format, so you can figure out what to enter even if it's different from the default
- If you drag and drop `/Applications/Kerbal\ Space\ Program` onto the path field, we remove the backslashes when you try to save. Note that backslashes _in general_ are not removed; **only** the ones in the specific string `Kerbal\ Space\ Program` because this is common and guaranteed wrong

### DIFF
--- a/ConsoleUI/KSPAddScreen.cs
+++ b/ConsoleUI/KSPAddScreen.cs
@@ -1,3 +1,7 @@
+using System;
+using System.IO;
+using CKAN.ConsoleUI.Toolkit;
+
 namespace CKAN.ConsoleUI {
 
     /// <summary>
@@ -9,7 +13,14 @@ namespace CKAN.ConsoleUI {
         /// Initialize the Screen
         /// </summary>
         /// <param name="mgr">KSP manager containing the instances</param>
-        public KSPAddScreen(KSPManager mgr) : base(mgr) { }
+        public KSPAddScreen(KSPManager mgr) : base(mgr)
+        {
+            AddObject(new ConsoleLabel(
+                labelWidth, pathRow + 1, -1,
+                () => $"Example: {examplePath}",
+                null, () => ConsoleTheme.Current.DimLabelFg
+            ));
+        }
 
         /// <summary>
         /// Return whether the fields are valid.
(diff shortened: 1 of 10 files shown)


ConsoleUI/KSPAddScreen.cs
```diff
@@ -9,7 +13,14 @@ public class KSPAddScreen : KSPScreen {
         /// Initialize the Screen
         /// </summary>
         /// <param name="mgr">KSP manager containing the instances</param>
-        public KSPAddScreen(KSPManager mgr) : base(mgr) { }
+        public KSPAddScreen(KSPManager mgr) : base(mgr)
+        {
+            AddObject(new ConsoleLabel(
+                labelWidth, pathRow + 1, -1,
+                () => $"Example: {Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles), "Kerbal Space Program")}",
```
DasSkelett, line 20: Apparently `Environment.SpecialFolder.ProgramFiles` is an empty string on Linux:
```
Example: Kerbal Space Program
```
We can probably expect Linux users to know where their stuff is installed (or at least know where to find it), but it looks kinda weird like this. Maybe we can fall back to the standard steam path or so?

HebaruSan: Yeah, it doesn't look like Mono was particularly careful in its implementation of these special folders. Some of them make sense and others seem like whatever random choice they felt like making at the time.

https://github.com/mono/mono/blob/09307e6411ace79740781df7a34f8fe09ad98f9f/mcs/class/corlib/System/Environment.cs#L621-L747

Now if only the `??` operator could handle empty strings...

HebaruSan: Alright, now if Program Files isn't defined, it should fall back to `~/.local/share`.

DasSkelett (approved):
Yup, does a good job at pointing the users in the right directions now.

Issue #3193, opened 2020-11-10
Can't install Kopernicus Bleeding Edge


DarkEnergy9999:
Problem
whenever i click on Kopernicus Bleeding Edge Beta the apply changes button stays gray, but when i click any other mod it goes green.
**Describe the bug**

**Steps to reproduce**
* Go to...
* Click on...
* See error

**Expected behavior**

**Screenshots (if applicable)**
![screenshot](https://media.discordapp.net/attachments/765935261271719937/775830559279677440/Screenshot_2020-11-10_161244.png?width=408&height=427)

**CKAN error code (if applicable):**

```
PLACE CODE HERE !
```

HebaruSan:
There's probably a conflict. Can you screenshot the whole window? The message explaining the conflict would appear in the lower left.

DarkEnergy9999:
> There's probably a conflict. Can you screenshot the whole window? The message explaining the conflict would appear in the lower left.

yes i will put a pic of the whole screen
![Screenshot 2020-11-10 164654](https://user-images.githubusercontent.com/72878452/98737564-5c7ff400-2374-11eb-9dc0-3d3f9b7318bb.png)

HebaruSan:
Yeah, this is due to how @R-T-B decided to structure the BE releases; it replaces ModularFlightIntegrator itself rather than creating a separate build for that mod. Uninstall ModularFlightIntegrator and try again.

![image](https://user-images.githubusercontent.com/1559108/98737837-70275c80-236c-11eb-9290-2b7cab4815d3.png)

DarkEnergy9999:
ok thank you that worked

R-T-B:
@HansSebaru, I would be happy to stop doing this if ModularFlightIntegrator is now marked as 1.10.1 ready...  is it, per chance?

HebaruSan:
Doesn't look like it:

- https://ksp.sarbian.com/jenkins/job/ModularFlightIntegrator/lastSuccessfulBuild/artifact/ModularFlightIntegrator/ModularFlightIntegrator.version

R-T-B:
Well the 1.9.1 version is what we use actually.  It works fine in 1.10.1.  Could we get the metadata updated manually pethaps for that reason?

HebaruSan:
@R-T-B better to go upstream when possible.

@sarbian any chance you could update this version file to reflect apparent compatibility with KSP 1.10?

[...]

PR #3200, opened 2020-11-16
Fix dependency resolution in mod upgrades

DasSkelett:
# Problem
Upgrading mods that previously provided a module and depend on it in the next version currently fails:
```
About to upgrade...

Downloading "https://github.com/R-T-B/Kopernicus/releases/download/UBE-release-46/KopernicusBE_1101_Release46.zip"
Module "Kopernicus Bleeding Edge Beta - DEV RELEASE" successfully installed
* Kopernicus-BE UBEE_1101_46 missing dependency ModularFlightIntegrator 1.2.7.0 or later
Error during installation!
ModularFlightIntegrator 1.2.7.0 conflicts with Kopernicus-BE UBEE_1101_45
```
In this case, Kopernicus-BE v1.10.1-45 provided ModularFlightIntegrator, but Kopernicus-BE v1.10.1-46 no longer does so and has it as dependency instead (see #3193). CKAN should automatically pull in MFI to satisfy the new dependency.

## Causes
1) `ModuleInstaller.Upgrade(IEnumerable<string> identifiers, ...)` doesn't tell the newly created `RelationshipResolver` that it also removes the old modules before adding the new ones. So the dependency is wrongly satisfied by the old module.
2) `RelationshipResolver(IEnumerable<string> modulesToInstall, IEnumerable<string> modulesToRemove, ...)` tries to find matching `CkanModule`s for all the identifiers in `modulesToInstall` and `modulesToRemove`. It does so by calling `TranslateModule()` -> `FromIDandVersion()`, that ultimately returns the latest compatible module.
This is fine for `modulesToInstall`, however for `modulesToRemove` we only want search through the currently installed modules (we can only remove modules that are actually installed after all).
3) `RelationshipResolver.ResolveStanza()` checks whether dependencies of the new modules are already satisfied by the installed modules, however it does so by querying the registry, which might have outdated information. Thus it incorrectly assumes the dependencies are satisfied (by the old mod versions). Only the resolvers own `installed_modules` knows about the modules that are going to be removed.

## Changes

1) `ModuleInstaller.Upgrade()` now tells the relationship resolver which modules it's about remove by passing the same identifier list for `modulesToRemove` as it passes for `modulesToInstall`.
2) `RelationshipResolver()` now uses `registry.InstalledModule()` to convert identifiers to `CkanModule`s for `modulesToRemove`. We can use the registry here since we're just in the process of creating the resolver and `installed_modules` is only declared and populated a few steps later.
3) `RelationshipResolver.ResolveStanza()` now looks into the resolvers own `installed_modules` instead of querying the potentially outdated registry. `installed_modules` is already reduced by the modules that are about to be removed, thus the resolver has to find other mods that can satisfy the .dependencies

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -967,7 +967,16 @@ namespace CKAN
         /// </summary>
         public void Upgrade(IEnumerable<string> identifiers, IDownloader netAsyncDownloader, ref HashSet<string> possibleConfigOnlyDirs, RegistryManager registry_manager, bool enforceConsistency = true)
         {
-            var resolver = new RelationshipResolver(identifiers.ToList(), null, RelationshipResolver.DependsOnlyOpts(), registry_manager.registry, ksp.VersionCriteria());
+            // When upgrading, we are removing these mods first and install them again afterwards (but in different versions).
+            // So the list of identifiers of modulesToRemove and modulesToInstall is the same,
+            // RelationshipResolver take care of finding the right CkanModule for each identifier.
+            List<string> identifierList = identifiers.ToList();
+            var resolver = new RelationshipResolver(
+                identifierList,
+                identifierList,
+                RelationshipResolver.DependsOnlyOpts(),
+                registry_manager.registry, ksp.VersionCriteria()
+            );
             Upgrade(resolver.ModList(), netAsyncDownloader, ref possibleConfigOnlyDirs, registry_manager, enforceConsistency);
         }
 
@@ -980,7 +989,7 @@ namespace CKAN
         {
             modules = modules.Memoize();
             User.RaiseMessage("About to upgrade...\r\n");
-            
+
             if (resolveRelationships)
             {
                 var resolver = new RelationshipResolver(modules, null, RelationshipResolver.DependsOnlyOpts(), registry_manager.registry, ksp.VersionCriteria());
--- a/Core/Relationships/RelationshipResolver.cs
+++ b/Core/Relationships/RelationshipResolver.cs
@@ -106,9 +106,9 @@ namespace CKAN
         /// <summary>
         /// Creates a new Relationship resolver.
         /// </summary>
-        /// <param name="options"><see cref="RelationshipResolverOptions"/></param>
-        /// <param name="registry">The registry to use</param>
-        /// <param name="kspversion">The version of the install that the registry corresponds to</param>
+        /// <param name="options">Options for the RelationshipResolver</param>
+        /// <param name="registry">CKAN registry object for current game instance</param>
+        /// <param name="kspversion">The current KSP version criteria to consider</param>
         public RelationshipResolver(RelationshipResolverOptions options, IRegistryQuerier registry, KspVersionCriteria kspversion)
         {
             this.registry = registry;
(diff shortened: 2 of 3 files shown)


Core/ModuleInstaller.cs
```diff
@@ -967,7 +967,12 @@ public void AddRemove(ref HashSet<string> possibleConfigOnlyDirs, RegistryManage
         /// </summary>
         public void Upgrade(IEnumerable<string> identifiers, IDownloader netAsyncDownloader, ref HashSet<string> possibleConfigOnlyDirs, RegistryManager registry_manager, bool enforceConsistency = true)
         {
-            var resolver = new RelationshipResolver(identifiers.ToList(), null, RelationshipResolver.DependsOnlyOpts(), registry_manager.registry, ksp.VersionCriteria());
+            var resolver = new RelationshipResolver(
+                identifiers.ToList(),
+                identifiers.ToList(),
```
HebaruSan, line 972: This looks really weird, but after some thought it's clearly right. Can we add a `//` comment explaining what we're doing?

Also it might be nice to call `.ToList()` once and store it in a `var`.

Core/Relationships/RelationshipResolver.cs
```diff
@@ -124,7 +124,9 @@ public RelationshipResolver(RelationshipResolverOptions options, IRegistryQuerie
         }
 
         /// <summary>
-        /// Attempts to convert the module_names to ckan modules via  CkanModule.FromIDandVersion and then calls RelationshipResolver.ctor(IEnumerable{CkanModule}, Registry, KSPVersion)"/>
+        /// Attempts to convert the identifiers to CkanModules and then calls RelationshipResolver.ctor(IEnumerable{CkanModule}, IEnumerable{CkanModule}, Registry, KSPVersion)"/>
+        /// For modulesToInstall CkanModule.FromIDandVersion is used, which tries to find the latest compatible version,
+        /// for modulesToRemove Registry.InstalledModule is used to get the currently installed version.
         /// </summary>
         /// <param name="modulesToInstall">Identifiers of modules to install</param>
```
HebaruSan, line 131: Elements of `modulesToInstall` aren't just identifiers, they can also be `identifier=version`. That's why `FromIDAndVersion` exists. We should have added `///<param>` tags explaining this in `FromIDandVersion` a long time ago. Can we now?

Core/Relationships/RelationshipResolver.cs
```diff
@@ -135,7 +137,8 @@ public RelationshipResolver(RelationshipResolverOptions options, IRegistryQuerie
             KspVersionCriteria kspversion) :
                 this(
                     modulesToInstall?.Select(name => TranslateModule(name, options, registry, kspversion)),
-                    modulesToRemove?.Select(name => TranslateModule(name, options, registry, kspversion)),
+                    modulesToRemove?.Where(name => registry.InstalledModule(name) != null)
```
HebaruSan, line 140: Maybe we should strip off a `=version` suffix in case the user passes it?

Core/ModuleInstaller.cs
```diff
@@ -967,7 +967,16 @@ public void AddRemove(ref HashSet<string> possibleConfigOnlyDirs, RegistryManage
...
         public void Upgrade(IEnumerable<string> identifiers, IDownloader netAsyncDownloader, ref HashSet<string> possibleConfigOnlyDirs, RegistryManager registry_manager, bool enforceConsistency = true)
         {
-            var resolver = new RelationshipResolver(identifiers.ToList(), null, RelationshipResolver.DependsOnlyOpts(), registry_manager.registry, ksp.VersionCriteria());
+            // When upgrading, we are removing these mods first and install them again afterwards (but in different versions).
+            // So the list of identifiers of modulesToRemove and modulesToInstall is the same,
+            // RelationshipResolver take care of finding the right CkanModule for each identifier.
+            List<string> identifierList = identifiers.ToList();
+            var resolver = new RelationshipResolver(
+                identifierList,
+                identifierList,
```
HebaruSan, line 976: Maybe we should strip off a `=version` suffix in case the user passes it?

Core/Relationships/RelationshipResolver.cs
DasSkelett: Tried to imrpove the code docs situation a bit. Feel free to push some changes if something doesn't make sense or you're able to formulate it in a more understandable way.

Core/ModuleInstaller.cs
HebaruSan: Never mind, somehow this already works just fine as-is...

```
PS C:\Users\user\github\CKAN> .\_build\ckan.exe install Kopernicus-BE=UBEE_1101_45
About to install...

 * Kopernicus Bleeding Edge Beta - DEV RELEASE UBEE_1101_45 (cached)
 * Module Manager 4.1.4 (cached)


Continue? [Y/n]


Installing mod "Kopernicus-BE UBEE_1101_45"
Installing mod "ModuleManager 4.1.4"
Updating registry
Committing filesystem changes
Rescanning GameData
Done!

PS C:\Users\user\github\CKAN> .\_build\ckan.exe upgrade Kopernicus-BE=UBEE_1101_44

Upgrading modules...

About to upgrade...

Downloading "https://github.com/R-T-B/Kopernicus/releases/download/UBE-release-44/KopernicusBE_1101_Release44.zip"
0 B/sec - downloading - 0 B left - 100%           6366 [1] WARN CKAN.ModuleInstaller (null) - Downgrading Kopernicus-BE from UBEE_1101_45 to UBEE_1101_44

Removing "CKAN.InstalledModule"
Installing "Kopernicus-BE UBEE_1101_44"
Updating registry
Committing filesystem changes
Done!
Done!

PS C:\Users\user\github\CKAN> .\_build\ckan.exe upgrade Kopernicus-BE=UBEE_1101_46

Upgrading modules...

About to upgrade...

Installing previously uninstalled mod ModularFlightIntegrator

Removing "CKAN.InstalledModule"
Installing "Kopernicus-BE UBEE_1101_46"
Installing "ModularFlightIntegrator 1.2.7.0"
Updating registry
Committing filesystem changes
Done!
Done!
```

HebaruSan (approved):
:guitar:

Core/ModuleInstaller.cs
DasSkelett: Oh, I added the sanitization inside `RelationshipResolver.ctor` instead of `Upgrade()`, to also fix it for other possible routes.

PR #3203, opened 2020-11-19
Sign debian releases for SecureAPT

techman83:
Currently you would need to allow insecure to install packages from these repos. Signing them will mean these will work without warnings.
```
Get:23 https://ksp-ckan.s3-us-west-2.amazonaws.com/deb nightly Release [129 B]                                       
Ign:24 https://ksp-ckan.s3-us-west-2.amazonaws.com/deb nightly Release.gpg                    
Reading package lists... Done                       
E: The repository 'https://ksp-ckan.s3-us-west-2.amazonaws.com/deb nightly Release' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

Instructions for adding the key + repo
```
curl -sS https://raw.githubusercontent.com/KSP-CKAN/CKAN/master/debian/repo.pub | sudo apt-key add - 
echo "deb https://ksp-ckan.s3-us-west-2.amazonaws.com/deb nightly main" | sudo tee /etc/apt/sources.list.d/ksp-ckan.list
```

Further info at https://wiki.debian.org/SecureApt


HebaruSan:
Don't we need to protect the .deb file somehow? So far this just signs the Release file.

https://wiki.debian.org/SecureApt#Secure_apt_groundwork:_checksums

techman83:
I believe the releases file contains checksums, but we could go about signing the deb files as well if this works out ok

HebaruSan:
Right, it's supposed to for apt-secure, but we haven't added that yet.

https://ksp-ckan.s3-us-west-2.amazonaws.com/deb/dists/nightly/Release

```
Components: main
Architectures: all
Description: Packages of the latest releases of the KSP-CKAN official client
Version: 1.29.1
```

techman83:
Aha, that might need some tweaking then

HebaruSan:
(Is Debian really still using MD5sums for this? I thought all the cool kids upgraded to SHA1 or SHA256 years ago.)

techman83:
> APT was originally designed around MD5 but people have since managed to construct collisions and so support for newer hash functions has been added.

Pretty sure it'll take SHA-256

HebaruSan:
The Packages.gz file already has the MD5, SHA1, and SHA256 for the .deb, so I think we just need to add checksums for that file into the Release file.

techman83:
Ok we already genearte a `Packages.gz`, so what we're missing is a checksum for that file in there.

techman83:
I didn't see your comment :P

HebaruSan (approved):
Let's try it. :wrench:

HebaruSan:
@techman83 can you pls add the key to the Action config?

```
Run echo -e "***" | gpg --batch --import
  echo -e "***" | gpg --batch --import
  gpg --list-secret-keys --keyid-format LONG
  ./build deb-sign --configuration=Release --exclusive
  shell: sh -e {0}
  env:
    DEBIAN_PRIVATE_KEY: ***
gpg: directory '/github/home/.gnupg' created
gpg: keybox '/github/home/.gnupg/pubring.kbx' created
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
Error: Process completed with exit code 2.
```

HebaruSan:
I think `${{ env.DEBIAN_PRIVATE_KEY }}` isn't defined and should be `$DEBIAN_PRIVATE_KEY` instead within the context of the bash commands.

HebaruSan:
Same.

```
Run echo -e "$DEBIAN_PRIVATE_KEY" | gpg --batch --import
  echo -e "$DEBIAN_PRIVATE_KEY" | gpg --batch --import
  gpg --list-secret-keys --keyid-format LONG
  ./build deb-sign --configuration=Release --exclusive
  shell: sh -e {0}
  env:
    DEBIAN_PRIVATE_KEY: ***
gpg: directory '/github/home/.gnupg' created
gpg: keybox '/github/home/.gnupg/pubring.kbx' created
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
Error: Process completed with exit code 2.
```

DasSkelett:
I think it's a problem with line breaks getting lost. If I do
```
PK=`gpg --export-secret-keys --armor <key-id>` && echo -e $PK
```
the output is missing the line breaks and
```
PK=`gpg --export-secret-keys --armor <key-id>` && echo -e $PK | gpg --batch --import
```
throws the error we see here too.

HebaruSan:
@DasSkelett what if you try `echo -e "$PK"` with quotes, as in the yml file? This seems to preserve the line breaks in my testing.

DasSkelett:
Oh yeah, that does indeed keep them and makes the import work. So it's something else...

PR #3205, opened 2020-11-21
Allow sort by multiple columns

HebaruSan:
## Motivation

Request from Discord:

image

## Changes

Now the main mod list in GUI supports multi-column sorting. If you shift-click a column header, instead of replacing the previous sort, it is added as a secondary sort.

Note that the previous saved sort setting is replaced with a new, multi-column one, without a migration. This means the value of the old setting will be lost on upgrade, and users will be presented with the default view of the grid sorted by mod name. Users can restore their old setting by clicking the column header.


DasSkelett (approved):
Yup, nice addition. Thanks!

PR #3206, opened 2020-11-22
Suppress confirmation prompt for GUI upgrades

HebaruSan:
## Problem

image

## Cause

#3204 added a yes/no prompt in `ModuleInstaller.Upgrade`, but I thought GUI only used `UninstallList` and `InstallList`. No, it does call `Upgrade`.

## Changes

Now the confirmation prompt is suppressed.

- Now GUI passes `ConfirmPrompt=False`
  - This is only available on the `CkanModule`-based version of `Upgrade`, and GUI previously used the `string`-based version, so now we've switched to the other


DasSkelett (approved):
Looks good, glad you caught that one before the next release :smile:

PR #3214, opened 2020-11-27
Manage repos with separate name and URL fields

HebaruSan:
## Problems

The repo and auth token sections of the GUI Settings use pipe characters to fake multiple columns:

image

The popup to add a repo:

image

- has title "Settings"
- is enormous compared to how much space its content requires
- isn't resizable
- also uses pipe characters to fake multiple columns in the list
- requires the user to type pipe characters in a special format to enter a custom repo
- has no label for its text field
- has the text field, cancel, and OK buttons inside the "Official repositories" frame even though these controls are not related to official repositories

## Changes

image

- The repo table in Settings is a ListView with two columns
- The auth token table in Settings is a ListView with two columns

image

- Title is now "Add Repository"
- The blank space in the repo list is reduced
- The form is resiable
- The list is now a ListView with two columns
- The text box is split into one for name, one for URL, each with a label, so the user doesn't have to type pipes
- The text box and buttons are moved outside the "Official repositories" frame
- The "OK" button is now relabeled to the more descriptive "Add" and is the same width as the other button, and they have keyboard shortcuts

### DIFF
--- a/GUI/Dialogs/NewRepoDialog.Designer.cs
+++ b/GUI/Dialogs/NewRepoDialog.Designer.cs
@@ -31,74 +31,138 @@
             this.components = new System.ComponentModel.Container();
             System.ComponentModel.ComponentResourceManager resources = new SingleAssemblyComponentResourceManager(typeof(NewRepoDialog));
             this.RepositoryGroupBox = new System.Windows.Forms.GroupBox();
+            this.RepoNameLabel = new System.Windows.Forms.Label();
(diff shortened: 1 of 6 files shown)


DasSkelett:
I've just pushed a small fix for the auth token section, we used to disable the `DeleteRepoButton` instead of the `DeleteAuthTokenButton` there.

DasSkelett (approved):
Improved usability :tada:

PR #3216, opened 2020-11-29
Publish keyring for APT repo

HebaruSan:
## Background

#3203 added signing of our APT repo's contents, so users can install our .deb files with confidence that they're authentic (or something). This requires the user to import a new `debian/repo.pub` key file into APT's trusted keyring. The installation instructions currently pinned to the linux channel in Discord are:

```
curl -sS https://raw.githubusercontent.com/KSP-CKAN/CKAN/master/debian/repo.pub | sudo apt-key add - 
echo "deb https://ksp-ckan.s3-us-west-2.amazonaws.com/deb nightly main" | sudo tee /etc/apt/sources.list.d/ksp-ckan.list
sudo apt-get update
sudo apt-get install ckan
```

I'm looking at this in more depth now to prepare the release notes and instructions for the next CKAN release.

## Problem

Running `apt-key` prints this:

```
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
```

`man apt-key` confirms the deprecation:

```
COMMANDS
       add filename (deprecated)
           Add a new key to the list of trusted keys. The key is read from the
           filename given with the parameter filename or if the filename is -
           from standard input.

           It is critical that keys added manually via apt-key are verified to
           belong to the owner of the repositories they claim to be for
           otherwise the apt-secure(8) infrastructure is completely
           undermined.

           Note: Instead of using this command a keyring should be placed
           directly in the /etc/apt/trusted.gpg.d/ directory with a
           descriptive name and either "gpg" or "asc" as file extension.
```

Apparently Debian in their infinite wisdom decided to deprecate the `apt-key` command, because sometimes `gpg` isn't available (it went from required to optional in some Debian release). It's not super clear.

- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=851774
- https://blog.sleeplessbeastie.eu/2018/08/08/how-to-download-public-key-used-to-verify-gnupg-signature-for-the-repository/

So in order to give the best instructions in our release notes, we need to replace this step with something else. The recommendation appears to be to put our own standalone keyring file into `/etc/apt/trusted.gpg.d`, since this doesn't require `gpg` to process the keys.

## Changes

I imported our key into a fresh keyring file in `debian/ksp-ckan.gpg` using the old method, and now we will store this on the public source code repo alongside `debian/repo.pub`. This way the user can download this file directly into `/etc/apt/trusted.gpg.d/ksp-ckan.gpg` without needing to use `apt-key` at all, and our release notes will not have to reference deprecated commands. The new instructions will be:

```
sudo curl -sS -o /etc/apt/trusted.gpg.d/ksp-ckan.gpg https://raw.githubusercontent.com/KSP-CKAN/CKAN/master/debian/ksp-ckan.gpg
sudo apt-add-repository -y deb https://ksp-ckan.s3-us-west-2.amazonaws.com/deb stable main
sudo apt install ckan
```

Note that these commands will not work until **after** this PR is merged, because the file won't exist on the main `master` branch, and because we won't have a `stable` build until the release is created. This should work in the meantime:

```
sudo curl -sS -o /etc/apt/trusted.gpg.d/ksp-ckan.gpg https://raw.githubusercontent.com/HebaruSan/CKAN/fix/deb-keyring/debian/ksp-ckan.gpg
sudo apt-add-repository -y deb https://ksp-ckan.s3-us-west-2.amazonaws.com/deb nightly main
sudo apt install ckan
```


techman83 (approved):
Ok, yes that seems a little wild, but I'm all for not using deprecated instructions.

PR #3218, opened 2020-12-01
Make Mono 6 builds work on Windows

HebaruSan:
## Background

Before #2964, we built releases on Mono 5.16, then that PR changed it to Mono 6.6. This broke things (see below), so as 
a workaround, #2976 changed it back to Mono 5.20 in a panic. Here we attempt to address the root cause.

## Problem

When we build on Mono 6, running the GUI on Windows throws an exception at startup:

```
Unhandled Exception: System.ArgumentException: Object of type 'System.Globalization.CalendarId[]' cannot be converted to type 'System.Int32[]'.
   at System.RuntimeType.TryChangeType(Object value, Binder binder, CultureInfo culture, Boolean needsSpecialCast)
   at System.RuntimeType.CheckValue(Object value, Binder binder, CultureInfo culture, BindingFlags invokeAttr)
   at System.Reflection.RtFieldInfo.UnsafeSetValue(Object obj, Object value, BindingFlags invokeAttr, Binder binder, CultureInfo culture)
   at System.Runtime.Serialization.FormatterServices.SerializationSetValue(MemberInfo fi, Object target, Object value)
   at System.Runtime.Serialization.ObjectManager.CompleteObject(ObjectHolder holder, Boolean bObjectFullyComplete)
   at System.Runtime.Serialization.ObjectManager.DoNewlyRegisteredObjectFixups(ObjectHolder holder)
   at System.Runtime.Serialization.ObjectManager.RegisterObject(Object obj, Int64 objectID, SerializationInfo info, Int64 idOfContainingObj, MemberInfo member, Int32[] arrayIndex)
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.RegisterObject(Object obj, ParseRecord pr, ParseRecord objectPr, Boolean bIsString)
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.ParseObjectEnd(ParseRecord pr)
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.Parse(ParseRecord pr)
   at System.Runtime.Serialization.Formatters.Binary.__BinaryParser.Run()
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.Deserialize(HeaderHandler handler, __BinaryParser serParser, Boolean fCheck, Boolean isCrossAppDomain, IMethodCallMessage methodCallMessage)
[...]

PR #3219, opened 2020-12-01
Add test for Control.Language property resource

HebaruSan:
## Background

#3218 fixed a problem with Mono 6 generating builds that don't work on Windows, but there's a risk that Visual Studio will create the problem again by setting the `$this.Language` property of a `.resx` file. If this happens, the current maintainers may not remember or may never have known that this is a problem, and a broken release could happen.

## Changes

Now a new Test checks that `$this.Language` is not set in any known resource file, which will prevent this problem from recurring in the future. To achieve this, `SingleAssemblyComponentResourceManager` is marked as `internal` and the Tests project is given internals-visibility into the GUI project.

### DIFF
--- a/GUI/Properties/AssemblyInfo.cs
+++ b/GUI/Properties/AssemblyInfo.cs
@@ -1,7 +1,9 @@
 ﻿using System.Resources;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 
 [assembly: AssemblyTitle("CKAN-GUI")]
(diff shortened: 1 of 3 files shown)


HebaruSan:
Planning to move the outer loop to `[TestCase()]`, stay tuned...

DasSkelett (approved):
Yup, should reduce the risk of sneaking it back in. Thanks!

PR #3220, opened 2020-12-02
Allow label to pin installed mod version

HebaruSan:
## Motivation

Sometimes a user doesn't want to upgrade an upgradeable mod for the foreseeable future:

image

image

image

## Changes

Now a new "Don't upgrade" checkbox is added to the Edit Labels form:

image

 If checked, mods with this label will not be upgraded by the "Add available updates" button (though the Update column will still appear to let users know an update exists and isn't being installed, and the user can still choose to upgrade manually):

image

A user installing CKAN for the first time will have a default label "Held" created with this feature enabled; existing users will have to create their own labels.

### DIFF
--- a/GUI/Controls/ManageMods.cs
+++ b/GUI/Controls/ManageMods.cs
@@ -1,11 +1,8 @@
 using System;
-using System.IO;
 using System.Linq;
 using System.Drawing;
-using System.Diagnostics;
 using System.Collections.ObjectModel;
 using System.Collections.Generic;
-using System.Threading;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using log4net;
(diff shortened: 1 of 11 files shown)


HebaruSan:
Pushed some small layout fixes for the form, basically the controls are moved/expanded down to fill the blank space at the bottom. Let me know if you care about updating the screenshots, otherwise I'll leave them as-is out of laziness.

DasSkelett:
Ugh, for some reason I can't enter Umlauts in any textbox on Mono. Works fine on Windows of course.
(Want to translate "Held" with "Zurückgehalten")

Let's do some debugging...

HebaruSan:
This worked for me (set a compose key, then type compose+shift+' followed by u):

- https://help.gnome.org/users/gnome-help/stable/tips-specialchars.html.en

DasSkelett:
Does work in other applications, but unfortunately not in CKAN (nor the [notepad sample](https://github.com/mono/mono/blob/master/mcs/class/System.Windows.Forms/samples/notepad.cs) from Mono).
Going to switch from Mono stable to nightly again and see if it makes a difference.

DasSkelett (approved):
I'm going to ignore my umlaut problem for the time being, it isn't related to this PR anyways, and maybe it's just some borkiness of my system.

The changes of this PR work fine. Maybe we can also find a solution for our  ConsoleUI users someday.

HebaruSan:
Oops, just meant to re-base but forgot to pull first, need to re-add the German...

PR #2177, opened 2017-11-07
Create text UI inspired by Turbo Vision

HebaruSan:
# CKAN Console UI

There has been discussion lately of replacing CKAN's current WinForms GUI with a GTK# one (#1840), as well as of enhancing the CmdLine interface (#2161, #2174). While this pull request doesn't do either of those things, I think it's related and may satisfy some of the needs that inspired those requests.

This pull request adds a new cross-platform UI modeled after the pre-Windows text interfaces of the late 1980s and early 1990s such as Borland's Turbo Vision and IBM's Common User Access, influenced by the Debian package management application "aptitude", with a few modern tweaks. I hope it can provide value in the niche between a full-on GUI and somewhat cumbersome command line tools.

## Status: **Open for more feedback, but probably ready to merge**

I think this project is pretty complete; I've addressed all bugs that I was able to find and added all the reasonable features I could think of. However, my knowledge of CKAN is not as deep or thorough as others', so there's a good chance I missed something. Please help me find:

- Unhandled exceptions/crashes and other bugs
- Missing features
- Better names for anything and everything
- Colors that are hard to read
- Bad layout
- How to make useful tests
- ~~Whether the key bindings work on MacOS X~~
- Etc.

I developed this on Linux and also tested in a Windows VM.
I do not have a Mac available for testing.

## New-ish features

- Import downloads option (related to #1788)
- After registry update, Y/N prompt to show newly compatible mods if any (related to #1766 and #1923)
- Search of `~dkeyword` shows reverse dependencies (related to #1100, #1494, and KSP-CKAN/NetKAN#5646)
- Instead of a forced registry auto-update, shows the approximate time since last update so user can press ctrl-R if needed (related to #2099)
- Shows where a download is hosted on the mod info screen (related to #2062)
- Shows when a mod was installed on the mod info screen (related to #954 and #1933)
- Can view the mod info screen for optional dependencies (related to #858 and #1868)
- Audit recommendations option to browse optional dependencies of already installed mods (related to #1764)
- Text search looks at identifier, name, and abstract (related to #1313)

## Dependencies

There are no third party dependencies; the whole thing is built directly on top of `Console.BackgroundColor`, `Console.SetCursorPosition`, etc., using classes contained in the `CKAN\ConsoleUI\Toolkit` folder (all written by me and kosher to release under CKAN's license). This has several advantages:

- No risk that a library we depend on will disappear or cease to be supported
- No risk of being affected by bugs we can't fix in a third party package
[...]

PR #3224, opened 2020-12-05
April Fools' Day prank for macOS

HebaruSan:
## Background

On macOS 10.15 and later, the GUI doesn't work anymore (see #2272), so we fall back to ConsoleUI, which is based on DOS user interfaces circa 1990 (see #2177), the high point of PC-vs-Mac rivalry. It amuses me to no end to see Mac users using the style of program most emblematic of their vendor's big enemy (and some of them have even said they like it!).

However, we also have a fully functional `ckan prompt` command emulating command line user interfaces circa 1980 (see #2273), which almost nobody knows about. It might be fun to have a little Easter egg that shows it off. Or this might not be a good idea :shrug:.

## Changes

Now if you double click the CKAN.app bundle while your system clock indicates a date of April 1st of any year, your Terminal will run `ckan prompt` instead of `ckan consoleui`, dropping you into a simple command line environment with commands for all of CKAN's functionality.

To avoid being overly cruel, `ckan prompt` now prints an April Fools' Day message on that day (as a clue to why it's different) and recommends the `consoleui` command (to get Mac users back into their familiar DOS-like environment easily). At worst it should be a minor annoyance for 24 hours per year, and at best some users might discover they like the command line environment.

### DIFF
--- a/Cmdline/Action/Prompt.cs
+++ b/Cmdline/Action/Prompt.cs
@@ -13,6 +13,19 @@ namespace CKAN.CmdLine
         public int RunCommand(KSPManager manager, object raw_options)
         {
             CommonOptions opts = raw_options as CommonOptions;
+            // Print an intro if not in headless mode
+            if (!(opts?.Headless ?? false))
+            {
+                Console.WriteLine("Welcome to CKAN!");
+                Console.WriteLine("");
+                if (DateTime.Now.Month == 4 && DateTime.Now.Day == 1)
+                {
+                    Console.WriteLine("Happy April Fools' Day! You may want to try the consoleui command.");
+                    Console.WriteLine("");
+                }
+                Console.WriteLine("To get help, type help and press enter.");
+                Console.WriteLine("");
+            }
             bool done = false;
             while (!done)
             {
--- a/macosx/CKAN
+++ b/macosx/CKAN
@@ -67,9 +67,15 @@ else
     # Default to consoleui on Catalina and later if they double click
     if ! supports_32_bit && [[ "$@" == "" ]]
     then
+        COMMAND=consoleui
+        if [[ $(date +'%m%d') == 0401 ]]
+        then
+            # Prank Mac users on April 1
+            COMMAND=prompt
+        fi
         osascript <<END
 tell application "Terminal"
-    do script "cd \"`pwd`\"; \"$MONO\" \"$ASSEMBLY\" consoleui"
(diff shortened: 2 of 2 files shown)


peterb12:
Please...no.

DasSkelett:
If I'm honest, I don't like April fools that break users' workflows. Users _will_ get confused, and most of them _won't_ know what to do. Even with an explanatory message, people likely won't see and/or read it, or don't know what to do. This message probably doesn't help someone who has no idea how to use a terminal or our prompt.

> You may want to try the consoleui command.

»How do you "try a command"? How do you try "the consoleui" command specifically? What _is_ "the consoleui command" and what happens if I "try it"? I just want to install mods.«

And just look at the mess in the r/KerbalSpaceProgram Discord today, because Sarbian thinks it's a good idea to additionally show the nyan cats, that are normally **the** sign of a broken installation (broken/outdated DLLs, wrong install location of ModuleManager, wrong Kopernicus DLL...), on the Japanese Cat Day (February 22nd) and April 1st.
Dozens of users thinking their (Kopernicus) installation somehow broke without doing anything, being confused, asking for help, even thinking they got a virus...

And that's just showing some arbitrary symbol, without changing anything, without needing any user intervention.
Now what happens if we actually do an April fool that "breaks" CKAN, and stops it from working as it did the day before? I'm really not looking forward to it.

I'm not per se against April fools, but they should be done carefully IMHO. Nothing that breaks users' workflows and the way CKAN behaves normally. _Maybe_ a ConsoleUI theme change, though even there I can see "CKAN looks different, what should I do?" support requests.

HebaruSan:
Yeah, that's fair; I'm happy to be talked out of this. Chalk it up to:

Ian

What do you think about keeping just the intro text part, so `ckan prompt` is a bit more responsive than just printing a prompt?

https://github.com/KSP-CKAN/CKAN/blob/98b634e2b77fa6894a13d5917128db0460450288/Cmdline/Action/Prompt.cs#L16-L20
https://github.com/KSP-CKAN/CKAN/blob/98b634e2b77fa6894a13d5917128db0460450288/Cmdline/Action/Prompt.cs#L26-L28

DasSkelett:
> What do you think about keeping just the intro text part, so `ckan prompt` is a bit more responsive than just printing a prompt?

I'm in on that!

HebaruSan:
OK, will start a separate branch/PR that doesn't say "April Fools'" for that...

PR #3226, opened 2020-12-05
Dark theme for ConsoleUI

HebaruSan:
## Motivation

Currently ConsoleUI only supports one visual style, based on TuboVision.

image

Some users may prefer a less bright display.

## Changes

Now ConsoleUI supports a `--theme` parameter accepting options `default` (the default) and `dark`. If you choose `dark`, a monochrome green-on-black style is used:

image
image

To achieve this, `Toolkit.ConsoleTheme` is changed from a singleton to a parameter that is passed to any spot that needs to draw things. A `ConsoleTheme.Themes` dictionary holds the defined themes, and new logic in `ConsoleCKAN` looks up the theme based on the `--theme` parameter and passes it to the application code.

### DIFF
--- a/Cmdline/Main.cs
+++ b/Cmdline/Main.cs
@@ -177,7 +177,7 @@ namespace CKAN.CmdLine
                         return Gui(manager, (GuiOptions)options, args);
 
                     case "consoleui":
-                        return ConsoleUi(manager, options, args);
+                        return ConsoleUi(manager, (ConsoleUIOptions)options, args);
 
                     case "prompt":
                         return new Prompt().RunCommand(manager, cmdline.options);
@@ -277,11 +277,13 @@ namespace CKAN.CmdLine
             return Exit.OK;
         }
 
-        private static int ConsoleUi(GameInstanceManager manager, CommonOptions opts, string[] args)
+        private static int ConsoleUi(GameInstanceManager manager, ConsoleUIOptions opts, string[] args)
         {
             // Debug/verbose output just messes up the screen
             LogManager.GetRepository().Threshold = Level.Warn;
-            return CKAN.ConsoleUI.ConsoleUI.Main_(args, manager, opts.Debug);
+            return CKAN.ConsoleUI.ConsoleUI.Main_(args, manager,
+                opts.Theme ?? Environment.GetEnvironmentVariable("CKAN_CONSOLEUI_THEME") ?? "default",
+                opts.Debug);
         }
 
         private static int Version(IUser user)
--- a/Cmdline/Options.cs
+++ b/Cmdline/Options.cs
@@ -477,7 +477,11 @@ namespace CKAN.CmdLine
         public bool ShowConsole { get; set; }
     }
 
-    internal class ConsoleUIOptions : InstanceSpecificOptions { }
+    internal class ConsoleUIOptions : InstanceSpecificOptions
+    {
+        [Option("theme", HelpText = "Name of color scheme to use, falls back to environment variable CKAN_CONSOLEUI_THEME")]
+        public string Theme { get; set; }
+    }
 
     internal class UpdateOptions : InstanceSpecificOptions
     {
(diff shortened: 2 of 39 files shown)


techman83:
Ok this is really cool!

DasSkelett:
<details>
<summary>Haha, took me some tries to select a terminal theme that works</summary>

My default theme, not very dark:
bright-dark-theme

Second try: 
invisible-dark-theme

Third time's a charm:
dark-dark-theme

</details>

DasSkelett:
Found an unrelated bug with the Console UI while testing this. Happens when leaving the path filed empty when adding a new game instance:
```
Unhandled Exception:
System.ArgumentException: The specified path is not of a legal form (empty).
at System.IO.Path.InsecureGetFullPath (System.String path) [0x00025] in <9969b53fa7824abc872802c31d2374b0>:0
at System.IO.Path.GetFullPath (System.String path) [0x00000] in <9969b53fa7824abc872802c31d2374b0>:0
at System.IO.DirectoryInfo..ctor (System.String path) [0x00006] in <9969b53fa7824abc872802c31d2374b0>:0
at (wrapper remoting-invoke-with-check) System.IO.DirectoryInfo..ctor(string)
at CKAN.ConsoleUI.GameInstanceScreen.pathValid () [0x00031] in <30624ca0a0a44ac58f64c7b4f4f2cf39>:0
at CKAN.ConsoleUI.GameInstanceAddScreen.Valid () [0x00009] in <30624ca0a0a44ac58f64c7b4f4f2cf39>:0
at CKAN.ConsoleUI.GameInstanceScreen.<.ctor>b__0_0 (System.Object sender, CKAN.ConsoleUI.Toolkit.ConsoleTheme theme) [0x00001] in <30624ca0a0a44ac58f64c7b4f4f2cf39>:0            
```
Fix should be as easy as this:
```diff
diff --git a/ConsoleUI/GameInstanceScreen.cs b/ConsoleUI/GameInstanceScreen.cs
index d9f55bc4..555a0b1e 100644
--- a/ConsoleUI/GameInstanceScreen.cs
+++ b/ConsoleUI/GameInstanceScreen.cs
@@ -94,6 +94,12 @@ protected bool nameValid()
         /// </summary>
         protected bool pathValid()
         {
+            if (string.IsNullOrEmpty(path.Value)) {
+                // Complain about empty path
+                RaiseError("Path cannot be empty!");
+                SetFocus(path);
+                return false;
+            }
             if (Platform.IsMac) {
                 // Handle default path dragged-and-dropped onto Mac's Terminal
                 path.Value = path.Value.Replace("Kerbal\\ Space\\ Program", "Kerbal Space Program");
```
Do you want to add it to this PR? Otherwise I can do a separate one.


Otherwise this looks good. I thought about adding a way to make the theme choice permanent so you don't have to add the argument yourself every time, and to help those launching the Console UI via the system menu entry from #3052. A new field in `config.json`?

HebaruSan:
Interestingly, that exception doesn't conform to the .NET documentation; `ArgumentException` is only supposed to be for invalid characters:

https://docs.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.-ctor?view=netframework-4.7.2#exceptions

I'm thinking this should catch all exceptions and treat them as failure...

HebaruSan:
For a persistent setting, ConsoleUI currently has none of its own, and I'd prefer to keep it that way.
How about an environment variable?

DasSkelett:
Yeah, after writing the comment it came to me that we shouldn't put UI-specific settings into the core's `config.json`.

Environment variables are a good idea. I kinda hoped that CommandlineParser would support them already, but doesn't look like it. There's an open PR for it though: https://github.com/commandlineparser/commandline/pull/698

HebaruSan:
> I kinda hoped that CommandlineParser would support them already, but doesn't look like it.

Same, this made me miss Python's `click`.

HebaruSan:
Hold on, the "no such theme" message should print the valid options. Gimme a minute to add that...

DasSkelett (approved):
Works like a charm and looks very nice, even on Windows!

PR #3234, opened 2020-12-10
Metadata GitHub Actions Container

techman83:
Deprecating or Jenkin's CI infrastructure has been on the road map for some time. To support this path we can build a container with the ckan/netkan tools for use during GitHub Actions.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -61,7 +61,7 @@ jobs:
       - name: Run tests
         run: xvfb-run ./build test+only --configuration=Release --where="Category!=FlakyNetwork"
 
-      - name: Generate Docker image and publish to Hub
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -72,6 +72,14 @@ jobs:
         run: |
           echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
           ./build docker-inflator --exclusive
+      - name: Generate inflator Docker image and publish to Hub
```
HebaruSan, line 75:
```suggestion
      - name: Generate metadata tester Docker image and publish to Hub
```

techman83: Whoops!

HebaruSan (approved):
Let's try it out.

PR #3235, opened 2020-12-10
Build Inflator on latest Mono

HebaruSan:
As of #3218 we know what the problem was with our Mono 6 builds: a serialization problem that only affected GUI and is now fixed.

Now we build the Inflator on `mono:latest` since it is safe to do so.


techman83 (approved):
Yep, I agree.

PR #3239, opened 2020-12-17
Format installation errors for GitHub Actions in headless mode

HebaruSan:
## Problem

If metadata pull request validation (see KSP-CKAN/xKAN-meta_testing#61) fails at the installation stage, errors are printed but aren't recognized as errors by the GitHub workflow framework, so they're not highlighted or pinned to the affected file:

- https://github.com/KSP-CKAN/CKAN-meta/runs/1572305201

image

## Cause

Error messages are just passed to `IUser.RaiseMessage`, despite reporting error conditions, which in turn just outputs via `System.Console.WriteLine`. GitHub actions require special formatting to recognize error messages.

## Changes

- Now one call per error is made to `RaiseError`, the one that's most likely to be helpful if pinned to a file
- Now if we're in `Headless` mode, `RaiseError` prints via `log.ErrorFormat`, and also replaces all line break sequences with `%0A` so multi-line messages can be handled properly
- Some other minor cleanup like using `DependencyNotSatisfiedKraken.Message` since it's nicer than the message we had previously

As far as I could tell we do not have a warning-level message in this code currently, which is good because `IUser` doesn't have a designated way to handle them.

### DIFF
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -143,14 +143,7 @@ namespace CKAN.CmdLine
                 }
                 catch (DependencyNotSatisfiedKraken ex)
                 {
-                    if (ex.version == null)
-                    {
-                        user.RaiseMessage("{0} requires {1} but it is not listed in the index, or not available for your version of KSP.", ex.parent, ex.module);
-                    }
-                    else
-                    {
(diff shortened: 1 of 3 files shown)


DasSkelett:
We should probably do the changes from `Cmdline/ConsoleUser.cs` to `Netkan/ConsoleUser.cs`, too.

HebaruSan:
We can do that, but it looks like all we use that object for is to pass it to `KSPManager`.

DasSkelett (approved):
Thanks, let's give it a try!

PR #3243, opened 2020-12-20
Modpack usability fixes

HebaruSan:
## Background

@kerbal-propulsion-laboratory (ezee on Discord) presented us with the idea of a metapackage in a custom repo. It works, with quirks.

## Problems

- The Download button in mod info is enabled for metapackages
- The right click menu download option is enabled for metapackages
- Exporting a modpack doesn't populate the author and release date
- The changeset tab has only two options, Clear and Apply, neither of which lets you go back to the modlist with the same changeset intact
- The column headers on the changeset tab sometimes disappear on Mono
- The "release status" and "replaced by" fields in mod info are always shown even when they're not set, cluttering the display
- The version of the current Breaking Ground expansion is not detected, so its other metadata (such as the store links) isn't available
- In Mono, the mod info panel flickers noticeably when you select a mod row
- The versions tab of mod info is slow to refresh when clicking Scatterer and frequently says that compatible modules are incompatible
- If you install a large module hosted on a server that doesn't send the download size:
  ```
  41176 [1] ERROR CKAN.ErrorDialog (null) - Unhandled exception:
  System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentOutOfRangeException: '-61800' is not a valid value for 'Value'. 'Value' should be between 'Minimum' and 'Maximum'
  Parameter name: Value
    at System.Windows.Forms.ProgressBar.set_Value (System.Int32 value) [0x00027] in <47c199caaa3c4c6c9c48a7c6ebb3fd29>:0 
    at (wrapper remoting-invoke-with-check) System.Windows.Forms.ProgressBar.set_Value(int)
    at CKAN.Wait+<>c__DisplayClass13_0.<SetModuleProgress>b__0 () [0x00161] in <3cb683943e2a481fbd7ca14981128ab6>:0 
    at (wrapper managed-to-native) System.Reflection.RuntimeMethodInfo.InternalInvoke(System.Reflection.RuntimeMethodInfo,object,object[],System.Exception&)
  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in <9f0df102fe6e4cfea29d2e46f585d8a5>:0 
  ```
- The "Automatically sort by 'Update'-column when clicking 'Add available updates'" setting overwrites your current sort, so if you chose a complex multi-column sort, it's gone after you update

## Causes

- `<KSP>/GameData/SquadExpansion/Serenity/readme.txt` now has whitespace after the version string, but the standard DLC detector doesn't allow this
- Mono's `DataGridView` has some quirks in how it fires its `SelectionChanged` event:
  - Usually it fires twice—once to deselect the previously selected row (with an empty selection) and once to select the newly selected row (with it as the selection)—which caused mod info to be collapsed and expanded rapidly every time the selection changes
  - Occasionally it fires multiple times in a row with the same selection
    - If Mod info was told to show the same mod twice in a row, it would perform all the same visual updates all over again, ending up in the same state but with extra flickering
- The tag list in mod info uses a `FlowLayoutPanel`, which was updated without turning off its layout, exacerbating flickering caused by repeated rapid updates
- The versions tab of mod info has its own logic for relationships which is slow and inaccurate when virtual modules are involved
- The per-module progress bars from #3054 don't clamp the calculated percentage to a minimum or maximum value, in effect assuming that the server will always send the file size, but it's a big negative value if that isn't the case
- We treat the update-sort as if it was the same as clicking a column header

## Changes

- Now the Download button in mod info is disabled for metapackages and the label explains there's no download
- Now the right click menu download option is disabled for metapackages
- Now the export modpack screen has an author field, defaulting to your system username (use commas to separate multiple authors)
- Now exporting a modpack sets the release date to the current time
- Now the changeset tab has a Back button which returns to the modlist without clearing the changeset
- Now the workaround from #2685 is applied to the changeset tab, preserving the column headers
- Now the "release status" and "replaced by" fields in mod info are hidden when they're not set
- Now the standard DLC detector allows whitespace after the version string
- Now we only fire `ManageMods_OnSelectedModuleChanged` if the selected module is not `null`, which will suppress the superfluous collapse/expand events
- Now mod info doesn't update itself if it's already showing the info for the given mod, which will reduce flickering
- Now we call tag list's `FlowLayoutPanel.SuspendLayout` before we update the tags and `TableLayoutPanel.ResumeLayout` after, which will also reduce flickering
- Now the versions tab of mod info uses `ModuleInstaller.CanInstall`, which is now `public`, and which is faster and more accurate, so the compatibility shown is correct
- Now the per-module progress bars handle missing server download size info without crashing
- Now the update-sort feature adds the Update column to the beginning of the sorted columns list, with the previous setting retained as secondary sorts. Once updates are installed, everything in the column will be `-`, and the previous sort restores itself by default.

### DIFF
--- a/Core/DLC/StandardDlcDetectorBase.cs
+++ b/Core/DLC/StandardDlcDetectorBase.cs
@@ -26,7 +26,7 @@ namespace CKAN.DLC
         private IGame game;
 
         private static readonly Regex VersionPattern = new Regex(
-            @"^Version\s+(?<version>\S+)$",
+            @"^Version\s+(?<version>\S+)",
             RegexOptions.Compiled | RegexOptions.IgnoreCase
         );
 
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -1315,17 +1315,16 @@ namespace CKAN
         /// Determine whether there is any way to install the given set of mods.
         /// Handles virtual dependencies, including recursively.
         /// </summary>
-        /// <param name="registry">Registry of instance into which we want to install</param>
-        /// <param name="versionCriteria">Compatible versions of instance</param>
         /// <param name="opts">Installer options</param>
         /// <param name="toInstall">Mods we want to install</param>
+        /// <param name="registry">Registry of instance into which we want to install</param>
         /// <returns>
         /// True if it's possible to install these mods, false otherwise
         /// </returns>
-        private bool CanInstall(
+        public bool CanInstall(
             RelationshipResolverOptions opts,
             List<CkanModule>            toInstall,
-            Registry                    registry
+            IRegistryQuerier            registry
         )
         {
             string request = toInstall.Select(m => m.identifier).Aggregate((a, b) => $"{a}, {b}");
--- a/Core/Registry/RegistryManager.cs
+++ b/Core/Registry/RegistryManager.cs
@@ -441,11 +441,13 @@ namespace CKAN
                 spec_version = new ModuleVersion("v1.18"),
                 identifier   = Identifier.Sanitize(name),
                 name         = name,
+                author       = new List<string>() { System.Environment.UserName },
                 @abstract    = $"A list of modules installed on the {kspInstanceName} KSP instance",
                 kind         = "metapackage",
                 version      = new ModuleVersion(DateTime.UtcNow.ToString("yyyy.MM.dd.hh.mm.ss")),
                 license      = new List<License>() { new License("unknown") },
                 download_content_type = "application/zip",
+                release_date = DateTime.Now,
             };
 
             List<RelationshipDescriptor> mods = registry.Installed(false, false)
--- a/GUI/Controls/AllModVersions.cs
+++ b/GUI/Controls/AllModVersions.cs
@@ -49,38 +49,25 @@ namespace CKAN
             }
         }
 
-        private bool installable(CkanModule module)
+        private bool installable(ModuleInstaller installer, CkanModule module, IRegistryQuerier registry)
         {
-            var currentInstance = Main.Instance.Manager.CurrentInstance;
-            var version = currentInstance.VersionCriteria();
-            if (!module.IsCompatibleKSP(version))
-            {
-                return false;
-            }
-            try
-            {
-                RelationshipResolver resolver = new RelationshipResolver(
-                    new CkanModule[] { module },
-                    null,
-                    new RelationshipResolverOptions()
-                    {
-                        with_recommends = false,
-                        without_toomanyprovides_kraken = true,
-                    },
-                    RegistryManager.Instance(currentInstance).registry,
-                    version
-                );
-            }
-            catch (Exception)
-            {
-                return false;
-            }
-            return true;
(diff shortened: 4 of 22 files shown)


DasSkelett:
> The per-module progress bars from #3054 don't clamp the calculated percentage to a minimum or maximum value, in effect assuming that download_size will always be set (but it's optional according to the spec), but it's a big negative value if download_size is null

Hm, I'm missing how `download_size` affects the progress bar. `remaining` and `total` come from the WebClient agent (through several layers of encapsulation).
I guess `download_size` somehow gets written to `total`? Where, and why doesn't it throw a `DivideByZeroException` if `download_size` and `total` are `0`?

HebaruSan:
> Hm, I'm missing how `download_size` affects the progress bar. `remaining` and `total` come from the WebClient agent (through several layers of encapsulation).
> I guess `download_size` somehow gets written to `total`? Where[...]?

`download_size` is fed into the system here:

https://github.com/KSP-CKAN/CKAN/blob/cd33a56370ad737eb4c5badb353769f4081f70ef/Core/Net/NetAsyncModulesDownloader.cs#L76

... filling the `size` parameter of `DownloadTarget`:

https://github.com/KSP-CKAN/CKAN/blob/cd33a56370ad737eb4c5badb353769f4081f70ef/Core/Net/Net.cs#L151

`NetAsyncDownloaderDownloadPart` then copies that to its `size` and `bytesLeft` properties:

https://github.com/KSP-CKAN/CKAN/blob/cd33a56370ad737eb4c5badb353769f4081f70ef/Core/Net/NetAsyncDownloader.cs#L45

You're right that those properties should be overwritten with values from `WebClient` here, though.

https://github.com/KSP-CKAN/CKAN/blob/cd33a56370ad737eb4c5badb353769f4081f70ef/Core/Net/NetAsyncDownloader.cs#L345-L346

Now I'm confused.

> and why doesn't it throw a `DivideByZeroException` if `download_size` and `total` are `0`?

Good question; I don't remember. I'll have another look at all this...

HebaruSan:
Following the logical chain and [a possibly relevant Stackoverflow question](https://stackoverflow.com/questions/40763313/c-sharp-webclient-totalbytestoreceive-is-returning-1-always), I think `DownloadProgressChangedEventArgs.TotalBytesToReceive` must have been -1 due to the server not sending it. But now I am not able to find such a server (SpaceDock and GitHub both send it). I'll edit the description to remove `download_size` since that looks like an incorrect explanation.

I wish I could remember which mod I was trying to install when that exception was thrown. Maybe Kopernicus_BE?

DasSkelett:
Maybe Sarbian's Jenkins? But he doesn't have any "big" mods there.
In any case, your change should fix it regardless of why, when and how it occurred.
Just pushed a commit with the German translations for the new labels, feel free to squash.

DasSkelett (approved):
Other changes look fine as well. More QoL enhancements :tada:

PR #3251, opened 2020-12-29
Validate renamed files by their new names in Netkan

HebaruSan:
## Problem

This from KSP-CKAN/NetKAN#8275 is a false positive:

image

And there are probably several related false negatives (i.e. missing errors that should be there).

## Cause

When Netkan looks for files with a particular suffix in a ZIP, it looks at the source filename in the ZIP, not the destination filename that will be installed. But what matters at run time is the destination file name, and the above PR renames many .txt files to .cfg. Currently they're not checked for MM syntax.

## Changes

Now Netkan looks at the destination file name when listing files from a ZIP by suffix, so files that are renamed from .txt to .cfg will be checked for MM syntax.

This will probably have some fun conflicts with #3223. Will rebase that one later.

I'm planning to merge quickly this so I can try revalidating that PR.

### DIFF
--- a/Netkan/Services/IModuleService.cs
+++ b/Netkan/Services/IModuleService.cs
@@ -11,8 +11,8 @@ namespace CKAN.NetKAN.Services
         JObject GetInternalCkan(string filePath);
         bool HasInstallableFiles(CkanModule module, string filePath);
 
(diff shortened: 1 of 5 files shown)


DasSkelett (commented):
Just inspecting the code without any testing, it looks fine.

PR #3259, opened 2021-01-06
Store remote version file URL in metadata resources

HebaruSan:
## Motivation

When investigating issues with mods' compatiiblity, the one piece of info that isn't readily available on http://status.ksp-ckan.space/ is the remote version file URL, which is important because it contains the latest compatibility data that is used for inflation. To get this, we have to download the ZIP and read the internal version file.

## Changes

Now `resources.remote-avc` is populated with the remote version file URL when available and displayed in GUI. I believe this will also flow through to the status page with no further changes, see KSP-CKAN/NetKAN-status#9.

The spec and schema are updated to support this new property, but I don't think we need to bump the minor version of the next CKAN release for this.

```json
    "resources": {
        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/155998-*",
        "repository": "https://github.com/HebaruSan/Astrogator",
        "bugtracker": "https://github.com/HebaruSan/Astrogator/issues",
        "remote-avc": "https://raw.githubusercontent.com/HebaruSan/Astrogator/master/Astrogator.version"
    },
```

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -220,6 +220,11 @@
                     "type"        : "string",
                     "format"      : "uri"
                 },
+                "remote-avc" : {
+                    "description" : "Mod's remote hosted version file",
+                    "type"        : "string",
+                    "format"      : "uri"
+                },
                 "store": {
                     "description" : "Purchase DLC here",
                     "type"        : "string",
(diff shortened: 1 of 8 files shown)


DasSkelett (approved):
Tried my best finding a suitable translation for "Remote version file", not that easy :D
Probably not very useful for the average user, but all the more for us, and maybe some modders.

Issue #2863, opened 2019-09-06
[Feature] KSP2 support plans


HebaruSan:
## Motivation

KSP2 has been announced:

- https://forum.kerbalspaceprogram.com/index.php?/topic/187313-we-are-excited-about-kerbal-space-program-2/
- https://forum.kerbalspaceprogram.com/index.php?/topic/187315-kerbal-space-program-2-master-post/

This raises the question of how CKAN should handle the sequel. I'll assume for now that we want to index mods for it and will be able to do so. (If we decide we want nothing to do with KSP2, or if we can't due to some hypothetical exclusive deal with Twitch, then that's obviously far simpler.)

This is a WIP issue to track discussion and hash out a concrete plan. Please reply with further thoughts, considerations, etc.

## Questions

- [x] Will mods still work the same way? Unzip files into a folder like GameData?
- [x] How can we determine the installed version? Is buildID.txt still there?
- [x] [Will KSP1 ever release a 2.0 version within its numbering scheme?](https://forum.kerbalspaceprogram.com/index.php?/topic/187315-kerbal-space-program-2-master-post/&do=findComment&comment=3667823)
  - 1.9 will be followed by 1.10
    https://forum.kerbalspaceprogram.com/index.php?/topic/193978-kerbal-space-program-esa-110-update-sharedhorizons/
  - 1.12.3 is the final update of KSP1

## Issues

- [x] Shouldn't have to download 20+ MB of metadata for KSP1 that you'll never use (see #1975)

## Tasks

Client updates to support the new game:

- [x] Version detector
- [x] Game folder paths

Infrastructure:

- [x] Split NetKAN metadata into `NetKAN/KSP1` and `KetKAN/KSP2` subdirs?
- [x] Split CKAN-meta somehow so you only download the metadata you need

politas:
We already had plans to split CKAN-meta up to lower the download burden. I'd like to see a system where there is a separate metadata file for each major point version, downloaded according to KSP version compatibility settings and merged as required. That would make supporting KSP 2 pretty seamless.

techman83:
There is an unmerged branch in the old bot where I came up with some logic for this. That was extendable based on some JSON config. It will need a slight re-think, but I reckon we could come up with something with the new architecture (though we've got a bit of a backlog to clean up and get merged in).

HebaruSan:
New timeline from the devs:

[...]

PR #3264, opened 2021-01-13
Preliminary support for Factorio

HebaruSan:
## Motivation

In #3223 we tried to separate KSP1-specific logic from generic core logic in preparation for KSP2, see #2863.

However, just having tried to make those changes doesn't exercise the code robustly the way that an actual additional game would. The more bugs we find before KSP2 comes out, the better.

## Background

This PR builds upon the fixes in #3308.

The [CFAN project](https://github.com/trakos/CFAN) is a fork of CKAN that was customized for Factorio. It was started quite a long time ago, and it was last updated around the time TLS1.2 became mandatory on GitHub (early 2018).

CFAN unfortunately defined its own `.cfan` file format rather than using `.ckan`, but it kind of makes sense because Factorio itself defines a fairly robust core mod metadata format that all its mods apparently use, so CFAN piggybacked on that and added some of its own fields.

## Changes

- Now the elements of CFAN that are specific to Factorio are added to a new `Core/Games/Factorio.cs` class, which is added to the list of known games. This means if you click to add a game instance to CKAN, a Factorio folder can now be selected and will work. We use CFAN's metadata repository to populate Factorio instances.
- The repo update code now scans for `.cfan` files and translates them to `CkanModule` format.

No changes are made to Netkan yet.

### DIFF
--- a/Core/Games/Factorio.cs
+++ b/Core/Games/Factorio.cs
@@ -1,0 +1,399 @@
+using System;
+using System.Linq;
+using System.IO;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Autofac;
+using log4net;
+using Newtonsoft.Json;
+using CKAN.GameVersionProviders;
+using CKAN.Versioning;
+
+namespace CKAN.Games
+{
+    public class Factorio : IGame
+    {
+        public string ShortName => "Factorio";
+
+        public bool GameInFolder(DirectoryInfo where)
+        {
+            return Directory.Exists(Path.Combine(where.FullName, "data"))
+                && Directory.Exists(Path.Combine(where.FullName, "data", "base"))
+                && File.Exists(Path.Combine(where.FullName, "data", "base", "info.json"));
+        }
+
+        /// <summary>
+        /// Finds the Steam KSP path. Returns null if the folder cannot be located.
+        /// </summary>
+        /// <returns>The KSP path.</returns>
+        public string SteamPath()
+        {
+            // Attempt to get the Steam path.
+            string steamPath = CKANPathUtils.SteamPath();
+
+            if (steamPath == null)
+            {
+                return null;
+            }
+
+            // Default steam library
+            string installPath = FactorioDirectory(steamPath);
+            if (installPath != null)
+            {
+                return installPath;
+            }
+
+            // Attempt to find through config file
+            string configPath = Path.Combine(steamPath, "config", "config.vdf");
+            if (File.Exists(configPath))
+            {
+                log.InfoFormat("Found Steam config file at {0}", configPath);
+                StreamReader reader = new StreamReader(configPath);
+                string line;
+                while ((line = reader.ReadLine()) != null)
+                {
+                    // Found Steam library
+                    if (line.Contains("BaseInstallFolder"))
+                    {
+                        // This assumes config file is valid, we just skip it if it looks funny.
+                        string[] split_line = line.Split('"');
+
+                        if (split_line.Length > 3)
+                        {
+                            log.DebugFormat("Found a Steam Libary Location at {0}", split_line[3]);
+
+                            installPath = FactorioDirectory(split_line[3]);
+                            if (installPath != null)
+                            {
+                                log.InfoFormat("Found a Factorio install at {0}", installPath);
+                                return installPath;
+                            }
+                        }
+                    }
+                }
+            }
+
+            // Could not locate the folder.
+            return null;
+        }
+
+        /// <summary>
+        /// Get the default non-Steam path to KSP on macOS
+        /// </summary>
+        /// <returns>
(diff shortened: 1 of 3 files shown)


HebaruSan:
@trakos, is it correct that CFAN's metadata is no longer being updated? The compatible game versions look outdated by a few releases.

trakos:
@HebaruSan yes, it's true, there were some changes in factorio's mod portal APIs and netfan updates stopped working. Since no-one reported it, I've assumed users moved on to using game's built-in mod installation (it improved in the meantime).

Anyway, I could probably get it fixed if you want to test ckan's multi-game capability on it.

trakos:
I fixed a couple things in netfan, it's non-empty now, though there are still some things lacking for now.

Core/Games/Factorio.cs
```diff
@@ -0,0 +1,398 @@
...
+        public FactorioAggregatorData aggregatorData;
+
+        public CkanModule ToCkan()
+        {
+            if (downloadUrls.Length < 1)
+            {
+                // No modpacks, thanks
+                return null;
+            }
+            var module = new CkanModule()
```
DasSkelett, line 294: The empty `CkanModule()` constructor isn't supposed to be used outside of tests (and JSON deserialization), according to its doc string:
https://github.com/KSP-CKAN/CKAN/blob/ae3dd4443cf6d693896565aa3bc0aa1ad3e71b41/Core/Types/CkanModule.cs#L269-L279

It's missing health checks, and doesn't populate the `Searchable*` fields.

Should we create a new constructor which takes all the (required) fields as individual parameters instead?

HebaruSan: Heh, nobody told me that for the DLC stuff, apparently. :grimacing: 

https://github.com/KSP-CKAN/CKAN/blob/ae3dd4443cf6d693896565aa3bc0aa1ad3e71b41/Core/Registry/Registry.cs#L881-L893

On the bright side, we can just call `CalculateSearchables`.

Core/Registry/RegistryManager.cs
```diff
@@ -314,14 +314,14 @@ private void LoadOrCreate()
...
             }
-
             AscertainDefaultRepo();
         }
 
         private void Create()
         {
             log.InfoFormat("Creating new CKAN registry at {0}", path);
             registry = Registry.Empty();
+            AscertainDefaultRepo();
```
DasSkelett, line 324: Was there a case where the the repositories didn't get populated? It's also run at the end of `loadOrCreate()`, which is the only place we call `Create()`.

Core/Games/Factorio.cs
DasSkelett: Oh, whoops :laughing: 
IMHO we should still at least change that documentation for the empty constructor then.
But preferably we should create a new constructor, which does all the health checks and runs `CalculateSearchables()` at the end, and make the empty constructor private.

Core/Registry/RegistryManager.cs
HebaruSan: I think I wanted to _move_ it into `Create` since that would make more sense and so that we would not be calling `Save()` with missing repo info. Then I had to re-add it back to `LoadOrCreate` due to some wrinkle.

Core/Games/Factorio.cs
HebaruSan: Agreed, will do.

Also did it for modpacks: :grimacing: 

https://github.com/KSP-CKAN/CKAN/blob/ae3dd4443cf6d693896565aa3bc0aa1ad3e71b41/Core/Registry/RegistryManager.cs#L438-L451

HebaruSan: > But preferably we should
> - [x] create a new constructor,
> - [x] which does all the health checks
> - [x] and runs `CalculateSearchables()` at the end,
>
> and make the empty constructor private.

Done except for making the empty constructor private, because I don't feel like fighting tests or deserialization code. We just won't use it in those other places anymore.

Core/Registry/RegistryManager.cs
DasSkelett: Okay. Guess there's no harm in keeping it twice.

Core/Games/Factorio.cs
DasSkelett: Thank you very much! Feels much better now :)

HebaruSan:
Gonna split the fixes into a separate PR, then hopefully we can look at releasing v1.30.0...

DasSkelett:
The first commit looks pretty good. Got a bit caught up in all the other stuff.
I'll give it one last test run, then you can merge that commit in separately, if you want without extra PR.

HebaruSan:
I'll create the new PR. That way it will be easier to see what is and is not in the master branch and why.

HebaruSan:
Let's not do this. KSP 2 will be out soon anyway.

PR #3265, opened 2021-01-13
Better version overrides in Netkan

HebaruSan:
## Problems

See KSP-CKAN/NetKAN#8291; if mod authors only track compatibility on SpaceDock (i.e. don't use a KSP-AVC version file), then the handy prompt to update compatibility when a new version is released just ends up changing the single compatible game version, so that for example compatibility with 1.11 is gained while compatibility with 1.10 is lost, when they may be expecting the total compatibility with prior versions to be remembered over the long term. All of @KSPSnark's mods have been subject to this forever, see KSP-CKAN/NetKAN#6629 and KSP-CKAN/NetKAN#6639 for the last time we tried wrestling with it.

The most natural thing is to add this:

```json
    "x_netkan_override": [ {
        "version": "1.10",
        "override": {
            "ksp_version_min": "1.8"
        }
    } ],
```

But that currently errors out:

```
2194 [1] FATAL CKAN.NetKAN.Program (null) - ksp_version mixed with ksp_version_(min|max)
```

## Cause

SpaceDock's compatibility goes into `ksp_version` and nowhere else. The override would need to delete that property, because it's not valid to mix `ksp_version_min` with `ksp_version`, but then the useful info it contained would be lost, and we would be back to 100% manual maintenance for compatibility of such mods rather than pulling author-generated data from SpaceDock.

## Background

Our AVC processing code has some pretty good logic for combining and overriding version ranges (which is part of why version files are so useful). Essentially it calculates an overall minimum and overall maximum from two input ranges, and then figures out the right metadata to represent the resulting range. This logic has usefulness for more than just AVC handling.

## Changes

- Now `AvcTransformer.ApplyVersions` is moved to `ModuleService` (except for the handling of `x_netkan_trust_version_file` which only makes sense for AVC), and calling code is updated to use the new location
- Now `VersionedOverrideTransformer` uses special logic to handle overrides of the compatibility properties; instead of just setting them blindly (and often generating invalid metadata), it passes them to `ModuleService.ApplyVersions`

After these changes, the above override works as intended:

```json
    "ksp_version_min": "1.8",
    "ksp_version_max": "1.11.0",
```

This will allow us to fix the compatibility of @KSPSnark's mods with just some slight additional manual maintenance, while still taking advantage of the author-generated data from SpaceDock.

This is the first part of fixing KSP-CKAN/NetKAN#8291 (the next part will be updating the netkans for the affected mods).

### DIFF
--- a/Netkan/Services/ModuleService.cs
+++ b/Netkan/Services/ModuleService.cs
@@ -8,6 +8,7 @@ using ICSharpCode.SharpZipLib.Zip;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
+using CKAN.Versioning;
 using CKAN.Extensions;
 using CKAN.NetKAN.Sources.Avc;
 using CKAN.Games;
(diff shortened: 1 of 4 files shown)


HebaruSan:
One test error is because I cut too much out of `AvcKrefTransformer`, the rest are because more functions are being called on mocked objects (so, make-work to fix tests rather than catching problems with the real code).

HebaruSan:
`ApplyVersions` is going to have to become static and essentially called as a global variable to avoid messing up this Moq stuff...

DasSkelett (approved):
Reasonable improvement, gives us more power over compatibility data while still making use of the author-provided data. Thanks!

HebaruSan:
#3263 looks to have worked properly.

HebaruSan:
HotSpot broke because it has many overrides that match and assumes they will overwrite each other, not expand compatibility as with AVC (and this PR):

- https://github.com/KSP-CKAN/CKAN-meta/commit/9878758f5ce528578ffaff1baf227a01d50af791
- https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/HotSpot.netkan

Will probably send a PR for that metanetkan to make the overrides only match their individual ranges.

Issue #3241, opened 2020-12-18
[Bug] NullReferenceException on spacebar with empty filter


Hydroxa:
Background
----------
Operating System: Windows 10
CKAN Version: 1.29.2
KSP Version: 1.11.0

No manual changes to GameData

Problem
-------
A NullReferenceException is thrown in ManageMods.cs when the space bar is pressed on ModList, though I'm not sure on which line. I would suspect the cause be something around line 595, where ModGrid.CurrentCell is referenced..

The error occurred while on an empty Upgradeable filter, as this means there is no selected entry to reference, so ModGrid.CurrentCell is likely null. This would mean that, upon pressing the space bar, it would try to reference a null object, thus throwing the named exception.

**Steps to reproduce**
Go to the mod manager, and open a filter that makes the list empty
Press the space bar
Get NullReferenceException

HebaruSan:
Sounds like you're pretty close to a fix. Any interest in submitting a pull request? The CKAN project welcomes contributions.

Hydroxa:
I'll look into it. Despite having GitHub for a little while now, I'm still learning about its features haha

HebaruSan:
Hi @Hydroxa, how's it going? I saw that you made a branch with a fix for this, do you need any assistance in submitting it?

Hydroxa:
I had some issues when trying to build the solution, so I couldn't test the fix :/ The solution was unable to find any dependencies, despite other solutions succeeding in this

But yeah, I did also try to push it here in case the issue was just with my computer and couldn't really work out how ^^;

HebaruSan:
Well, doing a quick checkout and compile on my system, it looks OK so far. There are some build instructions over here that might help (substitute `.\build.ps1` for `./build` on Windows):

- https://github.com/KSP-CKAN/CKAN/blob/master/doc/building.md
- https://github.com/KSP-CKAN/CKAN/wiki/Build-Environment
- https://github.com/KSP-CKAN/CKAN/wiki/New-Developers%27-Guide

If you click here, it should take you to the pull request creation flow, where you can enter a description of the change and then submit it:

- https://github.com/Hydroxa/CKAN/pull/new/new_branch

That gives us a place to do code review and, once any issues are figured out, eventually merge it.

PR #3266, opened 2021-01-14
Added null check in ManageMods.cs to prevent crash on an empty filter

Hydroxa:
Fixed KSP-CKAN/CKAN#3241 where application would err when using certain keyboard controls on an empty filter

### DIFF
--- a/GUI/Controls/ManageMods.cs
+++ b/GUI/Controls/ManageMods.cs
@@ -580,19 +580,21 @@ namespace CKAN
             {
                 case Keys.Home:
                     // First row.
-                    ModGrid.CurrentCell = ModGrid.Rows[0].Cells[SelectableColumnIndex()];
+                    if (ModGrid.Rows.Count > 0) //Handles for empty filters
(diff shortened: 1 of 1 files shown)


HebaruSan:
This change looks great, thanks!

<kbd>Home</kbd> and <kbd>End</kbd> with an empty filter also crash GUI in the same function (they reference `ModGrid.Rows[0]` and `ModGrid.Rows[ModGrid.Rows.Count - 1]`, which throws `System.ArgumentOutOfRangeException`). I think it would make sense to fix those here as well.

Hydroxa:
> 
> 
> This change looks great, thanks!
> 
> Home and End with an empty filter also crash GUI in the same function (they reference `ModGrid.Rows[0]` and `ModGrid.Rows[ModGrid.Rows.Count - 1]`, which throws `System.ArgumentOutOfRangeException`). I think it would make sense to fix those here as well.

Just did those now. I hadn't seen them, but these conditions should fix

HebaruSan (approved):
This looks great, thanks for contributing!

PR #3271, opened 2021-01-18
Visual cues for incompatibility reasons

HebaruSan:
## Motivation

We continue to get questions about why a mod is shown as incompatible when the user expects it to be compatible. The answer is that there is a conflict or an unsatisfied dependency, but this is generally not obvious.

## Changes

Now if a mod is incompatible because of the KSP versions, `<!>` is added to its Versions tab, which shows the range of compatible versions:

image

And if a mod is incompatible because of a conflict or an unsatisfied dependency, `<!>` is added to its Relationships tab, which shows those relationships:

image

Ideally these would have been big red stop sign icons instead, but Mono's `TabControl` does not display icons (it _makes room_ for them if you set `TabPage.ImageIndex` or `TabPage.ImageKey`, but they're never _drawn_ :roll_eyes:).

Priorities in choosing these characters:

- They should be wide enough to get attention but not _too_ wide (I found parentheses and square brackets too narrow)
- They should imply an alert or attention condition (hence exclamation mark)
- They should have a slight visual resemblance to a stop sign (angle brackets almost form a hexagon, which is almost an octagon)
- They should not crash Mono (as 🛑 does)

### DIFF
--- a/GUI/Controls/ModInfo.cs
+++ b/GUI/Controls/ModInfo.cs
@@ -2,6 +2,7 @@ using System;
 using System.Linq;
 using System.Collections.Generic;
 using System.Drawing;
+using System.ComponentModel;
 using System.Windows.Forms;
 using System.IO;
 using CKAN.Versioning;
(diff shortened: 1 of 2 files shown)


DasSkelett:
What do you think about opening the relevant tab as well? Should increase the chance to draw the user's attention, might be a bit disruptive, though.

HebaruSan:
I thought of that but wasn't sure about it. I'm open to it if you think it's a good idea, though.

DasSkelett:
Hm, I am not fully confident either. In that case, let's leave it out for now, and try whether the angle brackets that almost form a hexagon, which is almost an octagon already suffice to help users find out what's up.

DasSkelett (approved):
Works as advertised!

PR #3272, opened 2021-01-18
language: fr-FR

### LINKED ISSUE #3270: [question] How can I help to translate CKAN?
Hi, I would like to help and translate CKAN in French, but don't know how.
I followed what the wiki says: I forked the repository and went into the localization folder. At this point, I'm instructed to copy the English folder, rename the files accordingly and start translating. However, I ran into an issue: there are only four files there, which, it seems, do not contain anything translatable. In other languages' folder, there are 27 files, with an actual complete translation. 
The thing is, I can only translate from English, as my knowledge is limited.
I suppose the English files can be generated in some way, but I do not know how.
So, the actual (probably annoying) questions:
- do you think it is possible to translate CKAN with little knowledge in coding?
- is there a way for me to obtain those files?
- if not, do you think you could send them to me in some way?
- did I miss something in the wiki?
- do you like chocolate?

vinix38:
French localization

Fixes #3270.


HebaruSan:
Nice! I just pushed a few additional changes to wrap up the code side of things; you can retrieve them locally with a `git pull` command or equivalent in whatever git tool you're using, in case you want to make more edits.

Here's a test build so you can try out your changes:

- [ckan.zip](https://github.com/KSP-CKAN/CKAN/files/5831802/ckan.zip)

HebaruSan:
A couple of spots to investigate, might be missed translations or Core strings unavailable for translation or some technical mix-up...

image

image

image

image

image

image

vinix38:
> some layout needs reworking

That's what I expected, I will try to adjust box sizes

> Maybe these are English-French cognates?

exactly, in this one case, it is intended

> A couple of spots to investigate

that's what I'm going to do :)
thanks for the quick response guys

vinix38:
I checked the Labels translations:
image
As you can see, "favorites" and "hidden" are translated, but do not show up as such in the GUI

Otherwise, I corrected some oupsies, shortened some sentences and changed the buttons a tiny bit so that everything (should) fit. 
Please, could you provide another test build to check if the visuals are correct? (I'll learn how to do that some day, hopefully)

HebaruSan:
> As you can see, "favorites" and "hidden" are translated, but do not show up as such in the GUI

Oh sorry, I didn't mean to highlight those; they are generated when they first created and saved to disk, since they're user-editable, so of course I still have the English ones. Rather, in that screenshot I was wondering about "Labels", since I don't know whether that is also a cognate. It can be hard to choose two terms that mean Tags and Labels, I know we struggled with that for German.

New build:

[ckan.zip](https://github.com/KSP-CKAN/CKAN/files/5832382/ckan.zip)

vinix38:
Indeed, "Labels" was an artistic (if I daresay) choice: "étiquette" is often preferred, but Label is also used, I've seen it in French laws for example. But could have been an oopsie from me as well

> New build:

Thanks :)

HebaruSan:
These both have the same alt hotkey. I don't think WinForms usually handles that very well. Is there a common alternative for one of these?

image

vinix38:
I changed "Ajouter" into "Rajouter" to avoid any confusion.
In which dialogs does the alt-key play a role? Is there a way to define alt-keys without changing the actual word?
I did not take that into account when translating, but it shouldn't be too hard to fix I think.

HebaruSan:
Yes, it's a keyboard accessibility thing; in WinForms, putting a & in a button caption causes the next character to be underlined, and if the user presses Alt and that character, it's treated as clicking that button. But if two different buttons have the same hotkey, I think it just switches focus between them if I recall correctly.

vinix38:
Okay, I think everything is translated correctly. There shouldn't be anything catastrophic.
I would say this pull request is ready for review/merge.
By the way, the artefact build is really helpful and a lot faster, thanks :)

HebaruSan (approved):
Cool! Looks good to me. Thanks for contributing this!

PR #3276, opened 2021-01-24
Refactor merge script into classes, don't require rebase

HebaruSan:
## Background

In #3263 we created a Python script to merge pull requests more easily.

## Motivation

The script is very procedurally-oriented; it's one long function containing a sequence of complex steps that aren't always clearly related to one another, and not much structure or abstraction. This makes it harder to isolate one particular operation and check that it works properly.

## Problem

The script seems unable to merge branches that aren't rebased on current master, even if the conflicts are trivial. It is inconvenient to have to rebase every pull request before merging.

## Cause

`repo.index.merge_tree` doesn't do well with conflicts, but it's what you'll find if you go looking for ways to merge with GitPython. I think it is intended to be a low-level tool for apps that intend to re-implement parts of the `git merge` command themselves differently. That's not us; we just want the regular merge command.

## Changes

- New class `CkanRepo` represents the repo on disk and handles common operations such as checking if the master is up to date or editing files or knowing where the change log is
- New class `CkanPullRequest` represents a pull request and handles other operations such as getting the change log entry and commit message and doing the actual merge
- The merge steps now use `repo.git.merge`, which handles trivial conflicts better, and `repo.git.commit`, which is needed when the merge is started with `repo.git.merge`. These are simple wrappers around the `git merge` and `git commit` commands, so we can be more sure they'll be similar to the old commands.
- Also added automatic generation of the change log entry badge based on the labels; it translates them from label to badge, and if only one is recognized, it uses that, otherwise it uses `Multiple` for multiple and `UNKNOWN` for none, to be filled in by the user

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ All notable changes to this project will be documented in this file.
 ### Internal
 
 - [GUI] Make Mono 6 builds work on Windows (#3218, #3219 by: HebaruSan; reviewed: DasSkelett)
-- [Cmdline] Format installation errors for GitHub Actions in headless mode (#3239 by: HebaruSan; reviewed: DasSkelett)
+- [CLI] Format installation errors for GitHub Actions in headless mode (#3239 by: HebaruSan; reviewed: DasSkelett)
 - [Netkan] Fix Netkan timezones again (#3246 by: HebaruSan; reviewed: DasSkelett)
 - [Netkan] Better version overrides in Netkan (#3265 by: HebaruSan; reviewed: DasSkelett)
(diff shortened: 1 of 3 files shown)


DasSkelett:
One thing I didn't catch in #3263, the packages are called `gitpython` and `pygithub`. Amended a change renaming them and reordering the dependencies alphabetically.

Since `repo.git.merge` is just a wrapper, it should also handle deleted files correctly now, I guess. There was no open PR with deleted files to test.

DasSkelett (approved):
Nice, makes it more readable, too.

PR #3279, opened 2021-01-29
 Support indexing mulitple release assets on GitHub 

DasSkelett:
## Motivation
Kopernicus is a complex mod usually breaking between KSP releases. It used to only support one "primary" KSP version and some older ones in a separate [backports repository](https://github.com/Kopernicus/Kopernicus-Backport/releases).

Now that @R-T-B has taken over development, he introduced a new build system which allows him to easily support multiple KSP versions at the same time in the same repo.
Preferably he would want to release all assets/zips for every supported KSP version in a single release tag, like he does for his [bleeding edge repo](https://github.com/R-T-B/Kopernicus/releases), where the metadata is currently [maintained by hand](https://github.com/R-T-B/CKAN-meta-dev-Kopernicus_BE).

However, NetKAN currently doesn't support this. With `/ckan/github` $krefs, it always takes the module version from the release tag. Even multiple .netkan files with different `asset_match` wouldn't work, since they'd all end up with the same version.

## Concept
Now there is a new optional `version_from_asset` extension to the GitHub $kref.
It is in the format `/version_from_asset/:version_regexp`, where the `version_regexp` is matched against all release assets. All matching release assets will be inflated individually and result in a new metadata module.
The `version_regexp` must provide a named capturing group called `version`, which will be used to retrieve the module version for each asset.

`asset_match` and `version_from_asset` are mutually exclusive. `version_from_asset` also provides the filtering aspect of `asset_match` if needed, because it just skips unmatched assets.

To allow filling in the release tag into the final version string, `x_netkan_version_edit` now supports a special `${tag}` placeholder, which will be replaced by the release tag.

## Changes
There is a new `GithubReleaseAsset` class which holds a few asset-specific properties. `GithubRelease` has a new field, `IEnumerable<GithubReleaseAsset> Assets`, and lost asset-specific ones.
`GithubApi` creates the `GithubRelease` with the list of all `GithubReleaseAsset` now.

`GithubRef`'s regex is extended to allow the `version_from_asset` directive.

`GithubTransformer.Transform()` decides based on whether a `version_from_asset` regex is given or not how to inflate the metadata. If we have a regex, we try to get the version out of it, and pass it to `TransformOne()` together with the asset.

`GithubTransformer.TransformOne()` also takes a `GithubReleaseAsset` and `String version` as argument, from which it takes parts of the needed metadata.
Furthermore it adds a entry for the release tag to the metadata under `x_netkan_version_pieces`.

`VersionEditTransformer` reads the dict from `x_netkan_version_pieces` to replace any special vars.

The spec is also updated. Suggestions welcome if you can put it in better words.

I'm also not really found fond of calling it `version_from_asset`. It misses the fact that it can result in multiple .ckans.
`multi_asset_match` misses the fact that it gets the version from the file name.
`multi_asset_match_with_version_from_asset` is a bit long...

## Testing
Take [Kopernicus.netkan](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/Kopernicus.netkan), switch the kref to
```json
"$kref"         : "#/ckan/github/R-T-B/Kopernicus/version_from_asset/^KopernicusBE_(?<version>.+)\\.zip",
```
optionally add 
```json
"x_netkan_version_edit": {
        "find": "(?<ksp_version>.+)_Release(?<build>[0-9]+)",
        "replace": "release-${ksp_version}-${build}",
        "strict": true
    },
```json
or a different `replace` to test the tag replacement
```
        "replace": "${tag}-${ksp_version}",
```
to make the version format match the old ones.

Inflating `Kopernicus.netkan` should result in 4 .ckan files.

### DIFF
--- a/Netkan/Sources/Github/GithubRelease.cs
+++ b/Netkan/Sources/Github/GithubRelease.cs
@@ -1,21 +1,20 @@
 using System;
+using System.Collections.Generic;
 using CKAN.Versioning;
 
 namespace CKAN.NetKAN.Sources.Github
 {
     public sealed class GithubRelease
     {
-        public string           Author       { get; }
-        public ModuleVersion    Version      { get; }
-        public Uri              Download     { get; }
-        public DateTime?        AssetUpdated { get; }
+        public string        Author { get; }
+        public ModuleVersion Tag    { get; }
+        public List<GithubReleaseAsset> Assets { get; }
 
-        public GithubRelease(string author, ModuleVersion version, Uri download, DateTime? updated)
+        public GithubRelease(string author, ModuleVersion tag, List<GithubReleaseAsset> assets)
         {
             Author       = author;
-            Version      = version;
-            Download     = download;
-            AssetUpdated = updated;
+            Tag          = tag;
+            Assets       = assets;
         }
     }
 }
--- a/Netkan/Sources/Github/GithubReleaseAsset.cs
+++ b/Netkan/Sources/Github/GithubReleaseAsset.cs
@@ -1,0 +1,20 @@
+using System;
+using System.Collections.Generic;
+using CKAN.Versioning;
+
+namespace CKAN.NetKAN.Sources.Github
+{
+    public sealed class GithubReleaseAsset
+    {
+        public String        Name     { get; }
+        public Uri           Download { get; }
+        public DateTime?     Updated  { get; }
+
+        public GithubReleaseAsset(string name, Uri download, DateTime? updated)
+        {
+            Name     = name;
+            Download = download;
+            Updated  = updated;
+        }
+    }
+}
--- a/Netkan/CKAN-netkan.csproj
+++ b/Netkan/CKAN-netkan.csproj
@@ -85,6 +85,7 @@
     <Compile Include="Sources\Github\GithubApi.cs" />
     <Compile Include="Sources\Github\GithubRef.cs" />
     <Compile Include="Sources\Github\GithubRelease.cs" />
+    <Compile Include="Sources\Github\GithubReleaseAsset.cs" />
     <Compile Include="Sources\Github\GithubRepo.cs" />
     <Compile Include="Sources\Github\IGithubApi.cs" />
     <Compile Include="Sources\Jenkins\IJenkinsApi.cs" />
--- a/Netkan/Model/Metadata.cs
+++ b/Netkan/Model/Metadata.cs
@@ -1,5 +1,5 @@
 ﻿using System;
-﻿using System.Linq;
+using System.Linq;
 using CKAN.Versioning;
 using Newtonsoft.Json.Linq;
 
--- a/Netkan/Sources/Github/GithubApi.cs
+++ b/Netkan/Sources/Github/GithubApi.cs
@@ -2,7 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
-﻿using System.Text.RegularExpressions;
+using System.Text.RegularExpressions;
 using log4net;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
@@ -80,31 +80,36 @@ namespace CKAN.NetKAN.Sources.Github
                         var version = new ModuleVersion((string)release["tag_name"]);
                         var author = (string)release["author"]["login"];
 
-                        Uri       download = null;
-                        DateTime? updated  = null;
+                        List<GithubReleaseAsset> allAssets;
 
                         if (reference.UseSourceArchive)
                         {
                             Log.Debug("Using GitHub source archive");
-                            download = new Uri((string)release["zipball_url"]);
-                            updated = (DateTime)release["published_at"];
+                            Uri download = new Uri((string)release["zipball_url"]);
+                            DateTime?  updated = (DateTime)release["published_at"];
+                            allAssets = new List<GithubReleaseAsset> { new GithubReleaseAsset(version.ToString(), download, updated) };
                         }
                         else
                         {
+                            allAssets = new List<GithubReleaseAsset>();
                             var assets = (JArray)release["assets"];
 
                             foreach (var asset in assets.Where(asset => reference.Filter.IsMatch((string)asset["name"])))
                             {
                                 Log.DebugFormat("Using GitHub asset: {0}", asset["name"]);
-                                download = new Uri((string)asset["browser_download_url"]);
-                                updated = (DateTime)asset["updated_at"];
-                                break;
+                                Uri download = new Uri((string)asset["browser_download_url"]);
+                                DateTime? updated = (DateTime)asset["updated_at"];
+
+                                allAssets.Add(new GithubReleaseAsset(
+                                        (string)asset["name"], download, updated
+                                    )
+                                );
                             }
                         }
 
-                        if (download != null)
+                        if (allAssets.Any())
                         {
-                            yield return new GithubRelease(author, version, download, updated);
+                            yield return new GithubRelease(author, version, allAssets);
                         }
                     }
                 }
@@ -116,13 +121,13 @@ namespace CKAN.NetKAN.Sources.Github
                 }
             }
         }
-        
+
         public List<GithubUser> getOrgMembers(GithubUser organization)
         {
             return JsonConvert.DeserializeObject<List<GithubUser>>(
                 Call($"orgs/{organization.Login}/public_members")
             );
-        }        
+        }
 
         /// <summary>
         /// Download a URL via the GitHubAPI.
--- a/Netkan/Sources/Github/GithubRef.cs
+++ b/Netkan/Sources/Github/GithubRef.cs
@@ -6,7 +6,7 @@ namespace CKAN.NetKAN.Sources.Github
     internal sealed class GithubRef : RemoteRef
     {
         private static readonly Regex Pattern = new Regex(
-            @"^(?<account>[^/]+)/(?<project>[^/]+)(?:/asset_match/(?<filter>.+))?$",
+            @"^(?<account>[^/]+)/(?<project>[^/]+)(?:(/asset_match/(?<filter>.+))|(/version_from_asset/(?<versionFromAsset>.+)))?$",
             RegexOptions.Compiled
         );
 
@@ -14,6 +14,7 @@ namespace CKAN.NetKAN.Sources.Github
         public string Project { get; private set; }
         public string Repository { get; private set; }
         public Regex Filter { get; private set; }
+        public Regex VersionFromAsset { get; private set; }
         public bool UseSourceArchive { get; private set; }
         public bool UsePrerelease { get; private set; }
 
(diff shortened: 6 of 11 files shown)

### CODE REVIEW

File: Netkan/Sources/Github/GithubReleaseAsset.cs [hunk 1]
```diff
@@ -0,0 +1,22 @@
...
+using CKAN.Versioning;
+
+namespace CKAN.NetKAN.Sources.Github
+{
+    public sealed class GithubReleaseAsset
+    {
+        public String        Name         { get; }
+        public ModuleVersion Version      { get; set; }
+        public Uri           Download     { get; }
+        public DateTime?     AssetUpdated { get; }
```
HebaruSan, line 12: Maybe time to rename this to `Updated`, since it's now defined as part of an asset?

Netkan/Sources/Github/GithubRelease.cs
```diff
@@ -1,21 +1,20 @@
...
 {
     public sealed class GithubRelease
     {
-        public string           Author       { get; }
-        public ModuleVersion    Version      { get; }
-        public Uri              Download     { get; }
-        public DateTime?        AssetUpdated { get; }
+        public string        Author { get; }
+        public ModuleVersion Tag    { get; }
+        public IEnumerable<GithubReleaseAsset> Assets { get; }
```
HebaruSan, line 11: Any objections to making this a `List`? `IEnumerable`'s potential for lazy evaluation makes me nervous when it's in a class.

Netkan/Sources/Github/GithubReleaseAsset.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.Collections.Generic;
+using CKAN.Versioning;
+
+namespace CKAN.NetKAN.Sources.Github
+{
+    public sealed class GithubReleaseAsset
+    {
+        public String        Name         { get; }
+        public ModuleVersion Version      { get; set; }
```
HebaruSan, line 10: > I'm not entirely happy about the `json.SafeAdd("version", ghAsset.Version?.ToString() ?? ghRelease.Tag.ToString());` line, ideas welcome.

My suggestion is to remove the `Version` property from the asset class, since it isn't an integral part of the upstream data. Then add a `Version` parameter to `GithubTransformer.TransformOne`; each of the two calls to `TransformOne` in `Transform` would know which value to use, either `new ModuleVersion(extractedVersion.Value)` or `rel.Tag.ToString()`.

Netkan/Sources/Github/GithubRelease.cs
DasSkelett: Nope, fine with that. Good idea, given that we've had some interesting experiences with `IEnuerables` in netkan. Only need to update a few arrays to lists.

Netkan/Sources/Github/GithubReleaseAsset.cs
DasSkelett: That sounds good, we'd get rid of two ugly-ish places of code at the same time.

HebaruSan (changes requested):
This looks great! My only real concern is that other mod authors will be upset we didn't do it earlier. :grimacing: 

Thinking about how it might be used and abused, I can imagine a mod author wanting to combine the tag and a substring from the asset. I.e.:

```yaml
Tag: 1.2.3 (purely mod-version string unrelated to game version)
Assets:
  - Name: ForKSP-1.7.3.zip (just game version)
  - Name: ForKSP-1.10.x.zip (just game version)
```

Is that a possibility here? If not, could we achieve it?

Netkan/Sources/Github/GithubReleaseAsset.cs
DasSkelett: Turns out this also fixed another bug (at least in the test) we didn't see yet.

I initially changed the expected value of the second assertion in `Transform_SkipReleases_SkipsCorrectly`, because it returned the version from the filename.
However it shouldn't have done that, since its kref didn't have the new `version_from_asset`.

Turns out, the objects returned by the `apiMockUp` don't get regenerated for every test, but are permanent. So with `Transform_MultipleAssets_TransformsAll` running beforehand, and through it `Transform()` writing to `asset.Version`, it wasn't `null` anymore and that `json.SafeAdd()` line used it instead of the `ghRelease.Tag`.

Reminder to myself that I shouldn't just "fix" tests without thinking about why they are failing...

DasSkelett:
> Thinking about how it might be used and abused, I can imagine a mod author wanting to combine the tag and a substring from the asset.

Yeah, also thought of that and that it could be useful.
Currently it's not possible, we only take what's inside the `version` capture group and that's it.

We'd need to declare "special" variables (like `tag`) that we would fill in into the version string.
E.g. if the version string contains `${tag}` we would replace it with the tag.
Let me have a think how exactly we could achieve this, maybe with some regex magic.

HebaruSan:
```
  --releases           (Default: 1) Number of releases to inflate, or 'all'

  --skip-releases      (Default: 0) Number of releases to skip / index of 
                       release to inflate.
```

Should the numbers passed to these options relate to the number of _tags_ we process, or the number of _modules_ we ultimately inflate?

DasSkelett:
> ```
>   --releases           (Default: 1) Number of releases to inflate, or 'all'
> 
>   --skip-releases      (Default: 0) Number of releases to skip / index of 
>                        release to inflate.
> ```
> 
> Should the numbers passed to these options relate to the number of _tags_ we process, or the number of _modules_ we ultimately inflate?

Currently they dictate the number of processed respectively skipped tags. Given the original use case for those flags, that is backfilling missing/skipped releases, I'd say they should keep that behaviour.
If we want an option to limit the total inflated modules, we should introduce a new option for that. Not sure where we'd need that currently though, since it is undefined which asset(s) you'd get.

We could clarify the description of those options though, maybe by replacing "releases" with "release tags"?

HebaruSan:
> Given the original use case for those flags, that is backfilling missing/skipped releases, I'd say they should keep that behaviour.

OK, I'm convinced. The descriptions are probably fine as they are.

DasSkelett:
So regarding using the tag, there's the problem that the current .netkan-provided regex is only used as a "find" regex, i.e. one to get information with. To bring in the tag, we need another place where we can specify how to format the new version and where to fill in the tag (and possibly other data in the future) inside the string, like a "replace" regex.

Yup, this sounds a lot like our `x_netkan_version_edit`.

Some approaches that came to my mind:
1) Use the `x_netkan_version_edit` `replace` string by allowing to add a `${tag}` (or `${x_tag}` to avoid conflicts with user-provided capture groups), which we replace then.
However there's a caveat: `VersionEditTransformer` has no idea of any GitHub tags or anything else, given that it happens way later in the inflation process and is source independent. It's basically "stateless" in relation to other transformers.
We would need to introduce a way to submit a dict of "additional replacements" the `VersionEditTransformer` should execute, tied to each module. Temporarily encode it in the `metadata` variable?

2) Still use `x_netkan_version_edit` `replace` string, but read and edit it in-place in the `GithubTransformer` (maybe via a static helper function in `ModuleService` to make it available to other transformers as well). When `VersionEditTransformer` is executed, the `${tag}` / `${x_tag}` is already replaced with the actual tag.

3) Change `/version_from_asset/:version_regexp` to `/version_from_asset/:version_regexp[/:replace]` which allows a replacement format/string to be specified directly in the kref, basically duplicating the a bunch of logic from `VersionEditTransformer`. Least favourite solution.

4) Any additional ideas you come up with.

HebaruSan:
> 1. `VersionEditTransformer` has no idea of any GitHub tags or anything else, given that it happens way later in the inflation process and is source independent. It's basically "stateless" in relation to other transformers.
>    We would need to introduce a way to submit a dict of "additional replacements" the `VersionEditTransformer` should execute, tied to each module. Temporarily encode it in the `metadata` variable?

Sounds like a possible job for `TransformOptions` (defined in `ITransformer.cs`, available to `GithubTransformer.Transform` in its `opts` parameter)? Maybe it could have a `Dictionary<string, string> VersionPieces` which transformers could populate with keys and values, and then `VersionEditTransformer` could pull those in?

HebaruSan:
OK, it's _possible_ to use `TransformOptions`, but it doesn't turn out very well. The signature of `ITransformer.Transform` needs to change to return `Tuple<Metadata, TransformOptions>` or similar, which becomes a mess for typical code and downright complicated for your case. Instead, how about:

> 1. Temporarily encode it in the `metadata` variable?

I think that's excusable because `Metadata` is intended to hold the state of the inflation as it proceeds, and we already have a pattern of creating new instances when anything changes.

DasSkelett:
That works like a charm!
```
1931 [1] WARN CKAN.NetKAN.Transformers.VersionEditTransformer (null) - Replacing ${tag} with UBE-release-71 in ${tag}-1.10.1
1933 [1] WARN CKAN.NetKAN.Transformers.VersionEditTransformer (null) - Replacing ${tag} with UBE-release-71 in ${tag}-1.11.1
1933 [1] WARN CKAN.NetKAN.Transformers.VersionEditTransformer (null) - Replacing ${tag} with UBE-release-71 in ${tag}-1.8.1
1933 [1] WARN CKAN.NetKAN.Transformers.VersionEditTransformer (null) - Replacing ${tag} with UBE-release-71 in ${tag}-1.9.1
1933 [1] WARN CKAN.NetKAN.Transformers.VersionEditTransformer (null) - Replacing ${tag} with UBE-release-70 in ${tag}-1.10.1
1933 [1] WARN CKAN.NetKAN.Transformers.VersionEditTransformer (null) - Replacing ${tag} with UBE-release-70 in ${tag}-1.11.1
1933 [1] WARN CKAN.NetKAN.Transformers.VersionEditTransformer (null) - Replacing ${tag} with UBE-release-70 in ${tag}-1.8.1
1933 [1] WARN CKAN.NetKAN.Transformers.VersionEditTransformer (null) - Replacing ${tag} with UBE-release-70 in ${tag}-1.9.1
```
```
$ ll -n Kopernicus*.ckan
-rw-rw-r-- 1 1000 1000 1,8K Jan 30 05:13 Kopernicus-2-UBE-release-70-1.10.1.ckan
-rw-rw-r-- 1 1000 1000 1,8K Jan 30 05:13 Kopernicus-2-UBE-release-70-1.11.1.ckan
-rw-rw-r-- 1 1000 1000 1,8K Jan 30 05:13 Kopernicus-2-UBE-release-70-1.8.1.ckan
-rw-rw-r-- 1 1000 1000 1,8K Jan 30 05:13 Kopernicus-2-UBE-release-70-1.9.1.ckan
-rw-rw-r-- 1 1000 1000 1,8K Jan 30 05:13 Kopernicus-2-UBE-release-71-1.10.1.ckan
-rw-rw-r-- 1 1000 1000 1,8K Jan 30 05:13 Kopernicus-2-UBE-release-71-1.11.1.ckan
-rw-rw-r-- 1 1000 1000 1,8K Jan 30 05:13 Kopernicus-2-UBE-release-71-1.8.1.ckan
-rw-rw-r-- 1 1000 1000 1,8K Jan 30 05:13 Kopernicus-2-UBE-release-71-1.9.1.ckan
```
with
```json
    "$kref"         : "#/ckan/github/R-T-B/Kopernicus/version_from_asset/^KopernicusBE_(?<version>.+)\\.zip",
    "x_netkan_version_edit": {
        "find": "(?<ksp_version>.+)_Release(?<build>[0-9]+)",
        "replace": "${tag}-${ksp_version}",
        "strict": true
    },
```

HebaruSan (approved):
I'm pretty happy. Now we can think about how to announce this to mod authors that might benefit. :grimacing:

PR #3285, opened 2021-01-30
Fixes for French

vinix38:
Just some extra spelling fixes, nothing big


HebaruSan:
For the eventual release announcement, is this correct?

> Bonjour! Je m'appelle Hebarusan et CKAN parle maintenant français!

HebaruSan (approved):
I can't vouch for the spelling or grammar, but the XML syntax is valid, and we can continue making changes as needed.
Thanks for the contribution!

vinix38:
Your French announcement is correct :)

Just a thought, should the "bug report" buttons mention the fact that GitHub issues must be written in English?

HebaruSan:
Well, if someone doesn't speak English, we have [options for trying to establish communications anyway](https://translate.google.com/). I would rather have a challenging exchange that results in fixing a bug than send such a person away.

PR #3297, opened 2021-02-19
Fix checkbox sorting

HebaruSan:
## Problem

If you have an AD mod and sort by the Installed column, the AD mod will be shown after all the non-installed compatible mods. This is bad because the installed mods should be listed together at the top.

Originally the sort order of a checkbox column was:

- Checked checkboxes
- Text not containing "-"
- Unchecked checkboxes
- Text containing "-"

## Cause

In #3205 the checkbox column sort logic was rewritten and the original logic was lost; it is now:

- Checked checkboxes
- Unchecked checkboxes
- Text of any kind, regardless of content

## Changes

Now the sort order for a checkbox column is restored to what it was.

### DIFF
--- a/GUI/Controls/ManageMods.cs
+++ b/GUI/Controls/ManageMods.cs
@@ -1246,6 +1246,15 @@ namespace CKAN
             return CompareColumn(a, b, ModName);
         }
 
+        /// <summary>
+        /// Compare two rows based on one of their columns
+        /// </summary>
+        /// <param name="a">First row</param>
+        /// <param name="b">Second row</param>
+        /// <param name="col">The column to compare</param>
+        /// <returns>
+        /// -1 if a&lt;b, 1 if a&gt;b, 0 if a==b
+        /// </returns>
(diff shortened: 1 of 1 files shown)


GUI/Controls/ManageMods.cs
```diff
@@ -1246,6 +1246,15 @@ private int CompareRows(DataGridViewRow a, DataGridViewRow b)
...
         }
 
+        /// <summary>
+        /// Compare two rows based on one of their columns
+        /// </summary>
+        /// <param name="a">First row</param>
+        /// <param name="b">Second row</param>
+        /// <param name="col">The column to compare</param>
+        /// <returns>
+        /// -1 if a&lt;b, 1 if b&gt;a, 0 if a==b
```
DasSkelett, line 1256:
```suggestion
        /// -1 if a&lt;b, 1 if b&lt;a, 0 if a==b
```

...I think

HebaruSan: Oops, yes, was only supposed to swap one of the operator or the operands ordering...

DasSkelett (approved):
Looks good, and sorts correctly!

PR #3304, opened 2021-02-23
Don't warn about incompatible DLCs, fix conflict highlighting with DLC installed

### LINKED ISSUE #3269: [Feature] Suppress incompatibility warnings for already installed mods if version compatibility has changed accordingly in the meantime
CKAN 1.29.2, KSP 1.10.1

Example: this mod version was installed while the compatibility information was set to KSP 1.11.0 only, it was later changed to 1.9-1.11

popup
despite seeing the correct version information
screenshot

### LINKED ISSUE #3302: [Bug] Warns about incompatible modules which are actually DLC
Background
----------
* Operating System: Windows 10 Home 20H2 build 19042.804
* CKAN Version: 1.29.2
* KSP Version: 1.11.1.3066

No

Problem
-------
CKAN complains some installed modules are incompatible and warns might not be safe to launch KSP. (MakingHistory-DLC 1.11.0 (KSP 1.11.0)). This does NOT happen on a clean install of CKAN with no downloaded/cached, or installed modules. Only after a module has been downloaded and installed does it complain, and removing the module does not change this constant popup. No actual incompatibility has been found.

**Steps to reproduce**
Fresh install of CKAN and KSP with all DLC.
Install module(s).
Attempt to launch KSP from CKAN
Warning window appears, only listed item is the Making History DLC.

**Expected behavior**
CKAN should not erroneously warn about incompatible modules which aren't even modules, its an official DLC. 

EDIT: screenshot of pop-up as requested
ckan

HebaruSan:
## Problems

popup

Similarly, this can appear for a mod that was incompatible when you installed it but has since become compatible (because it's on SpaceDock and someone put in a `ksp_version_min` override).

Also if you have any DLC installed and you click to install two conflicting mods, they aren't highlighted red.

## Cause

SQUAD didn't update the MH version number correctly (flashbacks to the extra whitespace for the other DLC in #3243). Current `GameData/SquadExpansion/MakingHistory/readme.txt`:

```
*****************************************************************************************************
      __ _______ ____     __  ___      __   _               __  ___      __ 
     / //_/ ___// __ \   /  |/  /___ _/ /__(_)___  ___  _  / / / (_)____/ / ____ 
    / ,<  \__ \/ /_/ /  / /|_/ / __ `/ //_/ / __ \/ __ `/ / /_/ / / ___/ __/ __ \/ ___/ / / /
   / /| |___/ / ____/  / /  / / /_/ / ,< / / / / / /_/ / / __  / (__ ) /_/  /_/ / /  / /_/ /
  /_/ |_/____/_/      /_/  /_/\__,_/_/|_/_/_/ /_/\__, / /_/ /_/_/____/\__/\____/_/   \__, /  
                                                /____/                              /____/ 
 
*****************************************************************************************************

Thank you for downloading the Making History Expansion for Kerbal Space Program!

Version 1.11.0
```
...
```
ChangeLog:
=================================== v1.11.1 - Requires KSP v1.11.1 ================================
+++ Bugfixes
* Fix the servicemodules shielding antennae on launch when the shrouds are turned off.
* Fix KAL being unselectable in action pane in the mission builder.
* Fix the missing Robotics icon category in the action pane.
* Fix surface attachment issues with the LV-Tx87 Bobcat.

```

Since the `Version 1.11.0` line is what CKAN uses to detect DLC versions, CKAN thinks everyone using MH on 1.11.1 has an incompatible DLC.

When selecting conflicting rows, `ModuleIsDLCKraken` is thrown here, because the resolver is being told to install DLC:

https://github.com/KSP-CKAN/CKAN/blob/b748670c20db29ea366af3095837e2ffe0cbd7b7/GUI/Model/ModList.cs#L483-L488

## Changes

- Now the incompatible modules warning excludes DLCs.
- Now the incompatible modules warning excludes mods with compatible metadata in the available modules list but incompatible metadata in the installed modules list.
- Now the main mod list grid conflict detector doesn't pass installed modules to the resolver, since it gets them from the registry itself anyway. This means that DLC will never be in its to-install list.

Fixes #3302. Fixes #3269.

### DIFF
--- a/Core/Registry/InstalledModule.cs
+++ b/Core/Registry/InstalledModule.cs
@@ -157,7 +157,8 @@ namespace CKAN
         /// <summary>
         /// Ensures all files for this module have relative paths.
         /// Called when upgrading registry versions. Should be a no-op
-        /// if called on newer registries.</summary>
+        /// if called on newer registries.
+        /// </summary>
         public void Renormalise(GameInstance ksp)
         {
             var normalised_installed_files = new Dictionary<string, InstalledModuleFile>();
--- a/Core/Registry/Registry.cs
+++ b/Core/Registry/Registry.cs
@@ -128,7 +128,9 @@ namespace CKAN
         public IEnumerable<InstalledModule> IncompatibleInstalled(GameVersionCriteria crit)
         {
             return installed_modules.Values
-                .Where(im => !im.Module.IsCompatibleKSP(crit));
+                .Where(im => !im.Module.IsCompatibleKSP(crit)
+                    && !(GetModuleByVersion(im.identifier, im.Module.version)?.IsCompatibleKSP(crit)
+                        ?? false));
         }
 
         #region Registry Upgrades
--- a/GUI/Main/Main.cs
+++ b/GUI/Main/Main.cs
@@ -639,7 +639,8 @@ namespace CKAN
                 return;
 
             var registry = RegistryManager.Instance(CurrentInstance).registry;
-            var incomp   = registry.IncompatibleInstalled(CurrentInstance.VersionCriteria());
+            var incomp   = registry.IncompatibleInstalled(CurrentInstance.VersionCriteria())
+                .Where(m => !m.Module.IsDLC).ToList();
             if (incomp.Any())
             {
                 // Warn that it might not be safe to run Game with incompatible modules installed
--- a/GUI/Model/ModList.cs
+++ b/GUI/Model/ModList.cs
@@ -431,8 +431,8 @@ namespace CKAN
         public static Dictionary<GUIMod, string> ComputeConflictsFromModList(IRegistryQuerier registry,
             IEnumerable<ModChange> change_set, GameVersionCriteria ksp_version)
         {
-            var modules_to_install = new HashSet<string>();
-            var modules_to_remove = new HashSet<string>();
+            var modules_to_install = new HashSet<CkanModule>();
+            var modules_to_remove  = new HashSet<CkanModule>();
             var options = new RelationshipResolverOptions
             {
                 without_toomanyprovides_kraken = true,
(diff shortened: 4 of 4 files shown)


HebaruSan:
@SliceofLie, if you want to try this out, there's a test build at the bottom of this page:

https://github.com/KSP-CKAN/CKAN/actions/runs/593969600

DasSkelett:
Hm, why is `installed` even contained in `modules_to_check`->`modulesToInstall`? The RelationshipResolver already considers them when checking for conflicts:
https://github.com/KSP-CKAN/CKAN/blob/d3624b9278f10ff10759aae5444dfcedf17b398f/Core/Relationships/RelationshipResolver.cs#L250-L251

HebaruSan:
I suspect that our assumptions about handling of installed modules changed over time in the resolver, and `ModList.ComputeConflictsFromModList` wasn't updated to reflect those changes. My goal was to find the smallest change to minimize the risk of unexpectedly breaking something else, but we can try something more drastic if you think it's a good idea.

DasSkelett:
I think the impact of a slightly misbehaving `ComputeConflictsFromModList` would be limited.

But testing it everything seems to work as expected, conflicts with already installed modules get shown, as well as conflicts between two modules in the changeset about to be installed.
The RelationshipResolver of the actual install flow seems to work the same, with only the new modules in `moduleToInstall`.

So we should be fine removing `installed` from `modules_to_check` here.

HebaruSan:
OK, will do. There's also some unnecessary conversion from `CkanModule` to identifier back to `CkanModule` here that could cause its own problems down the line, so might as well clean that up too.

HebaruSan:
Done; now this looks very similar a block in `MainInstall.cs`, maybe we can consolidate them at some point...

DasSkelett (approved):
Nice!

PR #3307, opened 2021-02-27
Fixes for GUI and .ckan-installed modules

DasSkelett:
### GUI
* If you cancel an installation too late and can't be stopped anymore and finishes successfully, the changeset doesn't get cleared.
* Clearing the changeset on the changeset tab doesn't remove incompatible mods checked using the `AllModVersiions` tab. Neither does it remove mods queued for removal because they've been checked as auto-installed without any dependent mods.
* A Discord user from the r/KerbalSpaceProgram server who tends to find every bug and glitch in CKAN, likes to double-click the "Accept" button in the changeset tab. This throws an exception:
```
System.InvalidOperationException: This BackgroundWorker is currently busy and cannot run multiple tasks concurrently.
   at System.ComponentModel.BackgroundWorker.RunWorkerAsync(Object argument)
   at CKAN.Changeset.ConfirmChangesButton_Click(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnClick(EventArgs e)
   ```
* If you install a mod from a .ckan file which isn't known to the registry (identifier-wise), the "Versions" tab doesn't update when you select it
* Furthermore, for those mods the dependencies tab always has the new STOP sign from #3271
* If you downsize the GUI, some buttons of the menu strip disappear if they don't fit in the window.

### ConsoleUI
* When loading a new instance the modlist is empty at first (or only contains the DLCs). This tends to confuse users. The GUI does automatically refresh the registry in this case.
* If you launch the un-repacked `CKAN-ConsoleUI.exe` directly (e.g. via IDE), you get an exception because the `themename` passed to `ConsoleCKAN()` is `null`.

### Core
* CKAN offers to upgrade modules even if it would violate another module's relationships (but the installation fortunately fails); related it shows all versions in the `VersionsListView` as installable.
* While downloading a huge set of mods I got a NullReferenceException in `NetAsyncDownloader.FileProgressReport()`, for the `t` variable in the `queuedDownloads` foreach-loop. Unfortunately KDE crashed some time later and I didn't save the stacktrace to disk yet.

### GUI
* We don't run `UpdateModsList` after the install process has been cancelled, thus the changeset doesn't get cleared. We can't run it immediately, because it would close the WaitTabPage and remove the useful log messages.
* `ClearChangeSet()` doesn't detect changes for incompatible mods since `IsInstallChecked` is `false` for them. `GetModChanges()` find it by comparing `GUIMod.SelectedMod` with `GUIMod.InstalledMod.Module`.
* And it doesn't know about the auto-installed modules because toggling the chexkbox isn't an action itself that gets added to the changeset – it's applied instantly by setting `InstalledMod.AutoInstalled`.
* If you manage to hit the "Accept" button again before the GUI switches to the WaitTabPage, it tries to call `installWorker.RunWorkerAsync()` a second time, throwing the exception.
* `AllModVersions` uses `registry.AvailableByIdentifier(value.Identifier)` to populate the `VersionListView`. If the registry can't find anything, we just return, even without clearing the existing version items from the previous mod.
* `GUIMod.IsIncompatible` is always `true` for unknown modules because `registry.LatestAvailable()` can't find anything.
* >  By default, menu items are not displayed if they cannot fit within the available space.

  https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.menustrip.canoverflow?view=net-5.0#System_Windows_Forms_MenuStrip_CanOverflow

### Core
* We don't tell the RelationshipResolver the module version that gets removed, only the new one to be installed. Thus it considers dependencies still satisfied by the old version.
* I guess if a new download gets picked out of the queue while we're in the loop adding all the sizes together, the `DownloadTarget` can become `null`.

### GUI
* Add OK button to wait tab page; it is enabled if the install process failed or the user cancelled too late. It gives time to read the log messages, but ensures `UpdateModsList()` runs at some point. The tabController stays locked and the top menu deactivated until the user hits "OK". This means we also need to unlock the tabController when cancelling earlier in the installation process like when selecting the recommendations/suggestions or provides, because we never show the FailWaitDialog in this case.
* `ClearChangeSet()` also compares `SelectedMod` and `InstalledMod` now, and does the reverse of `GetModChanges()`, i.e. setting `SelectedMod` back to `InstalledMod` if they differ.
* Additionally, it deduces the previous state of now-marked-as-auto-installed modules by checking if they're in the changeset with `SelectionReason.NoLongerUsed`, and if so, sets it to false again. Doesn't catch all mods whose checkbox got changed, but it should suffice for most cases, to clear the changeset.
* If `installWorker` is already running, don't run it a second time. This was easier than trying to disable the button and finding all the right places where we'd need to enable it aǵain.
* If the registry doesn't know a mod, we still add the `CkanModule` associated with the `GUIModule`  to the list, if it exists.
* The `GUIMod` constructor that takes an `InstalledModule` as arg now overwrites `IsIncompatible` based on the compatibility of the installed module.
* Make `menuStrip2` overflowable. Instead of just not drawing the buttons, there's now a small dropdown to list those that didn't fit anymore. Doesn't work on Mono (it behaves just like before), but on Windows:
  

### ConsoleUI
* Update the registry if the loaded instance doesn't contain any modules, just like the GUI does. Should make initial setup easier for new users.
* If the `themeName` passed to `ConsoleCKAN` is `null` it falls pack to `default`.

### Core
* The two `RelationshipResolver` instantiations in `ModuleInstaller.CanInstall()` and `IRegistryQuerierHelper.HasUpdate()` now pass the currently installed module (if any) as `moduleToRemove`, just like the resolver for the actual install process does.
* Do a quick `null` check in the two foreach-loops in `NetAsyncDownloader.FileProgressReport()` and skip if `t` is `null`.

### DIFF
--- a/ConsoleUI/ModListScreen.cs
+++ b/ConsoleUI/ModListScreen.cs
@@ -17,7 +17,7 @@ namespace CKAN.ConsoleUI {
         /// </summary>
         /// <param name="mgr">Game instance manager object containing the current instance</param>
         /// <param name="dbg">True if debug options should be available, false otherwise</param>
-        public ModListScreen(GameInstanceManager mgr, bool dbg)
+        public ModListScreen(GameInstanceManager mgr, bool dbg, ConsoleTheme _theme)
         {
             debug    = dbg;
             manager  = mgr;
@@ -25,7 +25,7 @@ namespace CKAN.ConsoleUI {
 
             moduleList = new ConsoleListBox<CkanModule>(
                 1, 4, -1, -2,
-                GetAllMods(),
+                GetAllMods(_theme),
                 new List<ConsoleListBoxColumn<CkanModule>>() {
                     new ConsoleListBoxColumn<CkanModule>() {
                         Header   = "",
@@ -205,7 +205,7 @@ namespace CKAN.ConsoleUI {
                 }
                 return true;
             });
-            
+
             moduleList.AddTip("F8", "Mark auto-installed",
                 () => moduleList.Selection != null && !moduleList.Selection.IsDLC
                     && (!registry.InstalledModule(moduleList.Selection.identifier)?.AutoInstalled ?? false)
@@ -265,7 +265,7 @@ namespace CKAN.ConsoleUI {
                 null,
                 new ConsoleMenuOption("Refresh mod list", "F5, Ctrl+R",
                     "Refresh the list of mods",
-                    true, UpdateRegistry),
+                    true, (ConsoleTheme th) => UpdateRegistry(th)),
                 new ConsoleMenuOption("Upgrade all",          "Ctrl+U",
                     "Mark all available updates for installation",
                     true, UpgradeAll, null, null, HasAnyUpgradeable()),
@@ -328,7 +328,7 @@ namespace CKAN.ConsoleUI {
         private bool ImportDownloads(ConsoleTheme theme)
         {
             DownloadImportDialog.ImportDownloads(theme, manager.CurrentInstance, manager.Cache, plan);
-            RefreshList();
+            RefreshList(theme);
             return true;
         }
 
@@ -393,7 +393,7 @@ namespace CKAN.ConsoleUI {
                         }
                     }
                     if (needRefresh) {
-                        RefreshList();
+                        RefreshList(theme);
                     }
                 } else {
                     RaiseError("Installed mods have no unsatisfied recommendations or suggestions.");
@@ -414,7 +414,7 @@ namespace CKAN.ConsoleUI {
             return (DateTime.Now - File.GetLastWriteTime(filename)).Days;
         }
 
-        private bool UpdateRegistry(ConsoleTheme theme)
+        private bool UpdateRegistry(ConsoleTheme theme, bool showNewModsPrompt = true)
         {
             ProgressScreen ps = new ProgressScreen("Updating Registry", "Checking for updates");
             LaunchSubScreen(theme, ps, (ConsoleTheme th) => {
@@ -447,11 +447,11 @@ namespace CKAN.ConsoleUI {
                     }
                 }
             });
-            if (recent.Count > 0 && RaiseYesNoDialog(newModPrompt(recent.Count))) {
+            if (showNewModsPrompt && recent.Count > 0 && RaiseYesNoDialog(newModPrompt(recent.Count))) {
                 searchBox.Clear();
                 moduleList.FilterString = searchBox.Value = "~n";
             }
-            RefreshList();
+            RefreshList(theme);
             return true;
         }
 
@@ -481,7 +481,7 @@ namespace CKAN.ConsoleUI {
             if (!prevInst.Equals(manager.CurrentInstance)) {
                 plan.Reset();
                 registry = RegistryManager.Instance(manager.CurrentInstance).registry;
-                RefreshList();
+                RefreshList(theme);
             }
             return true;
         }
@@ -492,17 +492,23 @@ namespace CKAN.ConsoleUI {
             return true;
         }
 
-        private void RefreshList()
+        private void RefreshList(ConsoleTheme theme)
         {
-            moduleList.SetData(GetAllMods(true));
+            // In the constructor this is called while moduleList is being populated, just do nothing in this case.
+            // ModListScreen -> moduleList = (GetAllMods ...) -> UpdateRegistry -> RefreshList
+            moduleList?.SetData(GetAllMods(theme,true));
         }
 
         private List<CkanModule> allMods = null;
 
-        private List<CkanModule> GetAllMods(bool force = false)
+        private List<CkanModule> GetAllMods(ConsoleTheme theme, bool force = false)
         {
             ScanForMods();
             if (allMods == null || force) {
+                if (!registry?.HasAnyAvailable() ?? false)
+                {
+                    UpdateRegistry(theme, false);
+                }
                 allMods = new List<CkanModule>(registry.CompatibleModules(manager.CurrentInstance.VersionCriteria()));
                 foreach (InstalledModule im in registry.InstalledModules) {
                     CkanModule m = null;
@@ -546,7 +552,7 @@ namespace CKAN.ConsoleUI {
         private bool ApplyChanges(ConsoleTheme theme)
         {
             LaunchSubScreen(theme, new InstallScreen(manager, plan, debug));
-            RefreshList();
+            RefreshList(theme);
             return true;
         }
 
@@ -598,9 +604,9 @@ namespace CKAN.ConsoleUI {
             return total;
         }
 
-        private GameInstanceManager       manager;
-        private IRegistryQuerier registry;
-        private bool             debug;
+        private GameInstanceManager manager;
+        private Registry            registry;
+        private bool                debug;
 
         private ConsoleField               searchBox;
         private ConsoleListBox<CkanModule> moduleList;
@@ -839,7 +845,7 @@ namespace CKAN.ConsoleUI {
         /// This mod is installed and not upgradeable or planned to be removed
         /// </summary>
         Installed,
-        
+
         /// <summary>
         /// Like Installed, but can be auto-removed if depending mods are removed
         /// </summary>
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -1347,7 +1347,7 @@ namespace CKAN
             {
                 RelationshipResolver resolver = new RelationshipResolver(
                     toInstall,
-                    null,
+                    toInstall.Select(m => registry.InstalledModule(m.identifier)?.Module).Where(m => m != null),
                     opts, registry, ksp.VersionCriteria()
                 );
 
--- a/Core/Registry/IRegistryQuerier.cs
+++ b/Core/Registry/IRegistryQuerier.cs
@@ -194,7 +194,7 @@ namespace CKAN
             {
                 RelationshipResolver resolver = new RelationshipResolver(
                     new CkanModule[] { newest_version },
-                    null,
+                    new CkanModule[] { querier.InstalledModule(identifier).Module },
                     new RelationshipResolverOptions()
                     {
                         with_recommends = false,
--- a/GUI/Controls/AllModVersions.cs
+++ b/GUI/Controls/AllModVersions.cs
@@ -122,6 +122,11 @@ namespace CKAN
                         visibleGuiModule.PropertyChanged += visibleGuiModule_PropertyChanged;
                     }
                 }
+                VersionsListView.Items.Clear();
+                if (value == null)
+                {
+                    return;
+                }
 
                 // Get all the data; can put this in bg if slow
                 GameInstance currentInstance = Main.Instance.Manager.CurrentInstance;
@@ -139,16 +144,28 @@ namespace CKAN
                 }
                 catch (ModuleNotFoundKraken)
                 {
-                    // No versions to be shown, abort and hope an auto refresh happens
+                    // Identifier unknown to registry, maybe installed from local .ckan
+                    allAvailableVersions = new Dictionary<CkanModule, bool>();
+                }
+
+                // Take the module associated with GUIMod, if any, and append it to the list if it's not already there.
+                var installedModule = value.InstalledMod?.Module;
+                if (installedModule != null && !allAvailableVersions.ContainsKey(installedModule))
+                {
+                    allAvailableVersions.Add(installedModule, installable(installer, installedModule, registry));
+                }
+
+                if (!allAvailableVersions.Any())
+                {
                     return;
                 }
+
                 ModuleVersion installedVersion = registry.InstalledVersion(value.Identifier);
 
                 // Update UI; must be in fg
                 ignoreItemCheck = true;
                 bool latestCompatibleVersionAlreadyFound = false;
 
-                VersionsListView.Items.Clear();
                 // Only show checkboxes for non-DLC modules
                 VersionsListView.CheckBoxes = !value.ToModule().IsDLC;
 
--- a/GUI/Dialogs/ManageGameInstancesDialog.cs
+++ b/GUI/Dialogs/ManageGameInstancesDialog.cs
@@ -19,7 +19,7 @@ namespace CKAN
             Filter           = GameFolderFilter(Main.Instance.Manager),
             Multiselect      = false
         };
-        
+
         /// <summary>
         /// Generate filter string for OpenFileDialog
         /// </summary>
@@ -213,7 +213,7 @@ namespace CKAN
 
         private void OpenDirectoryMenuItem_Click(object sender, EventArgs e)
         {
-            string path = GameInstancesListView.SelectedItems[0].SubItems[2].Text;
+            string path = _manager.Instances[(string) GameInstancesListView.SelectedItems[0].Tag].GameDir();
 
             if (!Directory.Exists(path))
             {
--- a/GUI/Main/MainChangeset.cs
+++ b/GUI/Main/MainChangeset.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Linq;
 using System.Collections.Generic;
 using System.Windows.Forms;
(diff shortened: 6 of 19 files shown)

### CODE REVIEW

**Comment by HebaruSan** [comment]
Heh, you really love your custom .ckan files, don't you? :laughing: 
Good thing, as I practically never use them, so I'll issues related to them.

**Comment by HebaruSan** [comment]
(Wow; the example code in [the documentation of IsBusy], instead of noting that it may be needed to avoid the world's dumbest `System.InvalidOperationException` exception, instead shows how to waste CPU cycles with a tight polling loop!)

[the documentation of IsBusy]: https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.backgroundworker.isbusy?view=netframework-4.7.2

**Comment by DasSkelett** [comment]
> Heh, you really love your custom .ckan files, don't you? laughing
> Good thing, as I practically never use them, so I'll miss issues related to them.

Guess so :smile:
I also use them when indexing new modules to make sure they get installed correctly, all dependencies work out, give a second look to make sure I didn't forget anything and so on.
And it's useful to have a dummy module sitting around which you can quickly edit to simulate some relationship edge cases or whatever, comes in handy every now and then.

File: GUI/Main/MainChangeset.cs [hunk 1]
```diff
@@ -37,6 +37,8 @@ private void Changeset_OnConfirmChanges()
             // Using the changeset passed in can cause issues with versions.
             // An example is Mechjeb for FAR at 25/06/2015 with a 1.0.2 install.
             // TODO Work out why this is.
+            if (installWorker.IsBusy)
+                return;
```
HebaruSan, line 41: Hmm, I'm not completely sold on using `IsBusy` for this. Technically it still leaves us vulnerable to a race condition if we confirm that `IsBusy` is false and then another thread causes it to become true before we call `RunWorkerAsync`. What if we catch and ignore `InvalidOperationException`?

HebaruSan: Alternately, we could disable the button after the user clicks it.

DasSkelett: > Hmm, I'm not completely sold on using IsBusy for this. Technically it still leaves us vulnerable to a race condition if we confirm that IsBusy is false and then another thread causes it to become true before we call RunWorkerAsync.

What thread apart from the GUI thread could start it though, realistically?

> What if we catch and ignore InvalidOperationException?
Alternately, we could disable the button after the user clicks it.

In this case I prefer the catch solution. Disabling the button was my first attempt, but it didn't work out out at all, there are too many places where we hide and show and change the changeset tab. Sometimes the button needs to be re-enabled without the changeset changing (after failed installations), but not too early so it actually works, but also re-enable when the changeset changes...

GUI/Dialogs/ManageGameInstancesDialog.cs
```diff
@@ -213,7 +213,7 @@ private void GameInstancesListView_Click(object sender, MouseEventArgs e)
 
         private void OpenDirectoryMenuItem_Click(object sender, EventArgs e)
         {
-            string path = GameInstancesListView.SelectedItems[0].SubItems[2].Text;
+            string path = GameInstancesListView.SelectedItems[0].SubItems[3].Text;
```
HebaruSan, line 216: Oops! :facepalm:

GUI/Main/MainWait.cs
```diff
@@ -87,5 +87,14 @@ public void Wait_OnCancel()
                 cancelCallback();
             }
         }
+
+        public void Wait_OnOk()
+        {
+            ManageMods.UpdateModsList(null);
```
HebaruSan, line 93: In principle, this tab is supposed to be generic and usable by any task that needs to show progress, for example repository updates, auto-updating CKAN, and downloading files from the right click menu, so it shouldn't make assumptions about which tasks it is displaying. Is there a place more specific to the mod list where we could put the `UpdateModsList` call?

GUI/Dialogs/ManageGameInstancesDialog.cs
DasSkelett: Can't blame you for that one, hardcoding it like that is guaranteed to break at some point.
Thinking about it, how about recalculating it like this? No dependency on column count or ordering.
```
string path = _manager.Instances[(string) GameInstancesListView.SelectedItems[0].Tag].GameDir();
```

GUI/Main/MainWait.cs
DasSkelett: Hm, maybe we can make that another callback (like `Wait_OnCancel()` above), and (un)register `UpdateModsList()` to it during `PostInstallMods` or somewhere else. Need to try that out.

GUI/Dialogs/ManageGameInstancesDialog.cs
HebaruSan: I like that idea. :+1:

ConsoleUI/ModListScreen.cs
```diff
@@ -303,6 +303,15 @@ public ModListScreen(GameInstanceManager mgr, bool dbg)
             mainMenu = new ConsolePopupMenu(opts);
         }
 
+        public override void Run(ConsoleTheme theme, Action<ConsoleTheme> process = null)
+        {
+            if (!((Registry) registry)?.HasAnyAvailable() ?? false)
```
HebaruSan, line 308:
```suggestion
            if (!registry.HasAnyAvailable())
```

Reading this line makes me think that `registry` has the wrong type. There isn't much benefit from using `IRegistryQuerier`, especially since `Registry` inherits from it. And the object that we retrieve initially is a `Registry` (line 24). Trying it out locally, I think we can change line 616 to store a `Registry` and everything will still work. Does it look OK to you?

Core/ModuleInstaller.cs
```diff
@@ -1347,7 +1347,7 @@ IRegistryQuerier            registry
             {
                 RelationshipResolver resolver = new RelationshipResolver(
                     toInstall,
-                    null,
+                    toInstall.Select(m => registry.InstalledModule(m.identifier)?.Module).Where(m => m != null),
```
HebaruSan, line 1350: Why do we want to simulate uninstalling everything here?

GUI/Controls/AllModVersions.cs
```diff
@@ -139,8 +139,18 @@ public GUIMod SelectedModule
...
+                    var module = value.ToModule();
+                    if (module == null)
+                    {
+                        VersionsListView.Items.Clear();
+                        return;
+                    }
+                    allAvailableVersions = new Dictionary<CkanModule, bool>()
+                    {
+                        { module, installable(installer, module, registry) }
+                    };
```
HebaruSan, line 153: It seems like the installed module should also be included in the list even if other versions of the mod are found (for example, if I wanted to test `PlanningNode-v0.1.3.ckan` before it is released). Can we move this logic outside the `catch` block to do that?

ConsoleUI/ModListScreen.cs
```diff
@@ -303,6 +303,15 @@ public ModListScreen(GameInstanceManager mgr, bool dbg)
...
+            {
+                UpdateRegistry(theme);
```
HebaruSan, line 310: Could we do this in `GetAllMods` instead? Overriding `Run` just to add some extra initialization makes me slightly uncomfortable, I fear we'll end up adding more and more logic here when there might be better places for it.

HebaruSan:
> thus the changeset doesn't get cleared

I noticed that the change set clearing logic also doesn't work for forced installs of incompatible modules, since `ManageMods.ClearChangeSet` is implemented in terms of the install column checkboxes, which are "-" for those rows, but every time I try to solve this I experience a GUI-related mind fog. Do you see a fix?

GUI/Main/MainChangeset.cs
HebaruSan: Another possibility, tell the button to handle double clicks separately:

- https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.controlstyles?view=netframework-4.7.2#System_Windows_Forms_ControlStyles_StandardDoubleClick
- https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.doubleclick?view=netframework-4.7.2
- https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.button.mousedoubleclick?view=netframework-4.7.2

image

DasSkelett: Wouldn't this still leave room for the exception to throw, if the user clicks slow enough that it isn't registered as double-click, but too fast for the GUI to switch to the new tab?

ConsoleUI/ModListScreen.cs
DasSkelett: The `IRegistryQuerier` is kinda worthless right now anyway. I'm currently trying to implement a new "distributed" registry, which splits the local instance data like installed mods and the (theoretically) global data like available modules into different files, to hopefully reduce the chance for corruptions, and an interface only for read methods is pretty much useless since we need to cast it to `Registry` pretty much everywhere. An `IRegistry` would be much easier to handle.

But that's another topic, for here and now just storing it as `Registry` sounds reasonable :+1:

HebaruSan: > splits the local instance data like installed mods and the (theoretically) global data like available modules into different files

Nice, that has been on my to-do list for years now!

GUI/Controls/AllModVersions.cs
DasSkelett: Doing that,good suggestion.
What should we do if the dict already contains a module with this version?
If we replace the existing module, you won't see updated compat data in case it changes in the repository.
If we don't replace it, the tab would show the compat data of the repository even if you installed it from .ckan and it differs. You might also come to the conclusion that the install somehow didn't work because the compat data didn't update.

I don't see a way to reliably differentiate between outdated data due to repo changes and "newer" data from local .ckan.
I'd opt for not replacing to display the repo data since this is what the average user might encounter and benefit from. Installing from local .ckans is something probably only we do, and we can deal with that.

HebaruSan: > What should we do if the dict already contains a module with this version?

I think I'm fine with whatever you choose there. This is a pretty rare case to begin with. One option not considered so far is to list both if we find that they're different, so for example you might see that there is a 1.5.0 version that is compatible with 1.11 but the 1.5.0 that you have installed is only compatible with 1.10. That could be pretty confusing though.

DasSkelett: > One option not considered so far is to list both if we find that they're different

I did try that already, but a dict can only have unique keys, and `CkanModule`s are considered equal if the `identifier` and `version` matches...

I'm going with my preferred option then, so users at least see the "correct" version range in case it changed since they installed it.

ConsoleUI/ModListScreen.cs
DasSkelett: Ah, that looks like a good place! Did find that one when I looked through the files :smile:

GUI/Dialogs/ManageGameInstancesDialog.cs
DasSkelett: Oops, that should've been up one line...

DasSkelett:
> I noticed that the change set clearing logic also doesn't work for forced installs of incompatible modules, since `ManageMods.ClearChangeSet` is implemented in terms of the install column checkboxes, which are "-" for those rows, but every time I try to solve this I experience a GUI-related mind fog. Do you see a fix?

I might have found a way to achieve this. `ClearChangeSet` is now comparing `GUIMod.SelectedMod` and `GUIMod.InstalledMod`, and if they differ it set the former to the later.

* If an incompatible mod has been marked for installation, `SelectedMod` is not `null` whereas `InstalledMod` still is. Setting `SelectedMod` to `null` again removes it from the changeset again and unchecks the entry in `AllModVersions` (` SelectedMod` -> OnPropertyChanged()` -> `visibleGuiModule_PropertyChanged()` -> `UpdateSelection()`)
* Up/-downgrading an already force-installed module will also reset `SelectedMod` to the installed module.
* Unchecking an already force-installed module is handled correctly by the existing if case (though in theory the new else case should handle it just fine as well).

I also gave modules queued for removal because they have been marked as auto installed a stab. I've seen users checking this checkbox without knowing what it does, then despairing because CKAN wants to remove their mods.
Now if the changeset contains the mod with `SelectionReason.NoLongerUsed()`, we uncheck the box again if the changeset gets cleared. It doesn't necessarily reset it for all mods, if toggling it for a previous mod already removed it from the changeset.

This doesn't catch modules that still have dependent mods, but it's impossible to know their previous state. And they shouldn't be much of a problem anyway, since they aren't about to be removed.

HebaruSan (approved):
Nice! Good stuff all around.

Core/Registry/IRegistryQuerier.cs
```diff
@@ -194,7 +194,7 @@ public static bool HasUpdate(this IRegistryQuerier querier, string identifier, G
             {
                 RelationshipResolver resolver = new RelationshipResolver(
                     new CkanModule[] { newest_version },
-                    null,
+                    new CkanModule[] { querier.InstalledModule(identifier).Module },
```
DasSkelett, line 197: This one broke upgrading AD mods, `InstalledModule()` is `null` for those, thus we get an exception which caught below and returns `false`.

The easy fix is to pass no array if it's a DLL. However since DLLs are considered to satisfy _all_ dependencies, it's still possible that an upgrade would fail if the new version violated a versioned dependency.

The cleaner way would be to track installed DLLs separately in `RelationshipResolver`, and give the constructor a new option to tell the DLLs that are about to removed.

HebaruSan: > This one broke upgrading AD mods

Maybe we need some tests for that. It seems a bit brittle at the moment.

HebaruSan: By pure luck, adding a `?` also works, because `RelationshipResolver` is conservative with nulls:

```csharp
                    new CkanModule[] { querier.InstalledModule(identifier)?.Module },
```

> The cleaner way would be to track installed DLLs separately in `RelationshipResolver`, and give the constructor a new option to tell the DLLs that are about to removed.

I wonder whether AD mods should have their own `InstalledModule`, possibly with a `CkanModule` with `"kind": "dll"`?

DasSkelett: > I wonder whether AD mods should have their own `InstalledModule`, possibly with a `CkanModule` with `"kind": "dll"`?

Interesting idea. I'm a bit anxious that things might break if they expect `InstalledModule()`  to only return CKAN-managed modules. And if `InstalledModule()` returns modules that aren't in `InstalledModules`, more things might get confused (or would we also include DLLs in `InstalledModules`?)
And remembering all the bugs we had to fix because we didn't filter out `"kind": "dlc"`...

HebaruSan: > And remembering all the bugs we had to fix because we didn't filter out `"kind": "dlc"`...

Yes, it's definitely a high risk change. For now, I think I'm going to try writing that test and doing the easy fix.

DasSkelett: And I'm currently trying out tracking DLLs separately in the RelationshipResolver, so far it brought up some more places where we missed to pass all the data we should've to the resolver.
But since it basically includes the easy fix anyway, I think it's good that you do it as a separate PR, especially if it contains some tests that help catching any problems with more complex changes.

HebaruSan: > especially if it contains some tests that help catching any problems with more complex changes

Heh, I was just going to do the one simple test of `HasUpdate` with an AD mod, but now I'm looking for a few more...

> * CKAN offers to upgrade modules even if it would violate another module's relationships

Do you remember which relationships these were? Did another installed module have a `depends` with `version`?

DasSkelett: > Do you remember which relationships these were? Did another installed module have a `depends` with `version`?

Yup, BeyondHome with Parallax: https://github.com/KSP-CKAN/NetKAN/pull/8378
And this one is also interesting since I've read from a lot of people having Parallax installed manually, sometimes because they didn't know of the downgrade functionality. So that got me into the whole AD and upgrade with unsatisfied dependency rabbit hole.

PR #3308, opened 2021-03-05
Multi-game fixes

HebaruSan:
## Motivation

In #3223 we tried to separate KSP1-specific logic from generic core logic in preparation for KSP2, see #2863.

However, just having tried to make those changes doesn't exercise the code robustly the way that an actual additional game would. The more bugs we find before KSP2 comes out, the better.

## Background

#3264 started adding Factorio support and in the process found bugs and things that were missed relating to multi-game support. Eventually those were split into a separate commit, and now that commit is in this pull request.

## Changes

- Two places in `ModuleInstallDescriptor` still had KSP1-specific logic; both are now moved to `IGame` (`ReservedPaths` and `CreateableDirs`)
- Some more KSP1-specific code is moved to `Core/Games/KerbalSpaceProgram`
- The convention for matching all patch levels of a minor version in Factorio seems to be to set the maximum to MAJOR.MINOR.MAXINT. This caused an overflow in CKAN, so now we round MAXINT to Undefined
- `GameInstance.Scan` now silently aborts if the game's default mod folder doesn't exist


DasSkelett (approved):
Thanks a lot for the patience! Looks very good :heavy_check_mark:

PR #3313, opened 2021-03-06
Update dialog was too small for French localization

vinix38:
Since this dialog only shows up on updating, I had no way to check if it worked.
When updating to 1.30, I noticed that the "not now" button wasn't big enough.
I made it bigger, and I shortened the text, but I don't know how to test my changes.

### DIFF
--- a/GUI/Dialogs/AskUserForAutoUpdatesDialog.Designer.cs
+++ b/GUI/Dialogs/AskUserForAutoUpdatesDialog.Designer.cs
@@ -29,19 +29,18 @@
         private void InitializeComponent()
         {
             System.ComponentModel.ComponentResourceManager resources = new SingleAssemblyComponentResourceManager(typeof(AskUserForAutoUpdatesDialog));
-            this.label1 = new System.Windows.Forms.Label();
+            this.autoCheckLabel = new System.Windows.Forms.Label();
             this.YesButton = new System.Windows.Forms.Button();
-            this.button1 = new System.Windows.Forms.Button();
+            this.NoButton = new System.Windows.Forms.Button();
             this.SuspendLayout();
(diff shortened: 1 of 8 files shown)


HebaruSan:
Cool, we can fake an update with some temporary code hacks, I'll post a screenshot of these changes when I get a chance...

HebaruSan:
Just for completeness I turned on all the startup prompts that I could find:

image

image

image

I'll see if we can turn on word wrap for that first one...

HebaruSan:
With just a little finagling (commit incoming shortly):

image

HebaruSan:
Also updated the names of the controls because `button1` is a poor counterpart to `YesButton`.

HebaruSan:
> I made it bigger, and I shortened the text, but I don't know how to test my changes.

@vinix38, from that description it sounds like you meant to widen this button:

https://github.com/KSP-CKAN/CKAN/blob/25fa82d4ad41857b22b57c6731492f614159dcbf/GUI/Dialogs/NewUpdateDialog.Designer.cs#L85-L93

Would you still want to do that?

vinix38:
>from that description it sounds like you meant to widen this button:
You are right,, I modified the wrong dialog layout (oupsi...)

image
I did not mean to modify this one, the previous size was probably alright, you can revert my changes I think

image
This was the one I aimed to fix. The new translation of "not now" seems to be enough, no need to change the button size then

HebaruSan:
OK, I'll revert the wide "Non" button, but I do think the other buttons could stand to be wider (and the same visual style as the ones on the other popups). Incoming with next commit:

image

image

HebaruSan (approved):
Thanks for fixing this!

PR #3315, opened 2021-03-07
Fix AD mod upgrading, add tests, and fix all warnings

HebaruSan:
## Problem

In #3190, we allowed AD mods to be upgraded, replacing them with a CKAN-managed module.

In v1.30.0, AD mods again are no longer treated as upgradeable.

## Causes

`InstalledModule` returns `null` for AD mods, and after #3307 we attempt to access properties of that return value, which throws a null reference exception. `HasUpdate` catches this exception and returns `false`.

https://github.com/KSP-CKAN/CKAN/blob/6ec7270e2fe92845b662ce2ae5d14132a5429590/Core/Registry/IRegistryQuerier.cs#L197

## Changes

Now we pass `null` for the modules to remove if `InstalledModule` returns `null`. In future we would like to tell `RelationshipResolver` to check what will happen if we remove the DLL, but for now this will at least bring the upgrade checkboxes back.

Two tests are added to make it harder to break this accidentally:

- Make sure that if another installed mod depends on a specific version of another mod, we do not attempt to upgrade that other mod (the original fix from #3307)
- Make sure that an AD mod is treated as upgradeable when there's a compatible module with the same identifier (this fix)

### Side fixes

The `Tests/GUI` folder is reorganized slightly, with two files moved to a new `Model` subfolder to match where their tested code lives.

Fixed warnings during build:

- >  ModListScreen.cs(20,78): warning CS1573: Parameter '_theme' has no matching param tag in the XML comment for 'ModListScreen.ModListScreen(GameInstanceManager, bool, ConsoleTheme)' (but other parameters do)`

  Added documentation of this parameter

- > SingleAssemblyComponentResourceManager.cs(22,43): warning CS0618: 'ResourceManager.ResourceSets' is obsolete: 'call InternalGetResourceSet instead'
  > SingleAssemblyComponentResourceManager.cs(46,36): warning CS0618: 'ResourceManager.ResourceSets' is obsolete: 'call InternalGetResourceSet instead'
  > SingleAssemblyResourceManager.cs(20,43): warning CS0618: 'ResourceManager.ResourceSets' is obsolete: 'call InternalGetResourceSet instead'
  > SingleAssemblyResourceManager.cs(44,36): warning CS0618: 'ResourceManager.ResourceSets' is obsolete: 'call InternalGetResourceSet instead'

  We were using this `protected Hashtable` as a cache for our `ResourceSet`s because that's what what I found on StackExchange, but the base class no longer uses this object, hence this warning. Now we use a `private Dictionary` instead for the same purpose.

Fixed warnings from tests:

- > 6443 [NonParallelWorker] WARN CKAN.NetKAN.Transformers.GithubTransformer (null) - Multiple assets found for  1.2 without `version_from_asset`

  Reworked `GithubTransformerTests` slightly so the multi-asset release is only processed by code that's prepared for it

- > 6609 [NonParallelWorker] WARN CKAN.NetKAN.Validators.TagsValidator (null) - Tags not found, see https://github.com/KSP-CKAN/CKAN/wiki/Suggested-Tags
  > 6641 [NonParallelWorker] WARN CKAN.NetKAN.Validators.TagsValidator (null) - Tags not found, see https://github.com/KSP-CKAN/CKAN/wiki/Suggested-Tags

  Added tags to the test data in `CkanValidatorTests`


DasSkelett (approved):
Looking very good, thanks!
(Wait with the merge until #3319 finished deployment :grin:)

Issue #3309, opened 2021-03-06
[Bug] Update not working with the deb file repo


blazux:
Background
----------
* Operating System:  Debian 10
* CKAN Version: Latest

Problem
-------
apt upgrade fails to update to the latest CKAN version

Problem with the deb file maybe ?

```
The following packages will be upgraded:
  ckan
1 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
Need to get 775 kB of archives.
After this operation, 0 B of additional disk space will be used.
Do you want to continue? [Y/n] Y
Get:1 https://ksp-ckan.s3-us-west-2.amazonaws.com/deb stable/main amd64 ckan all 1.30.0.21065 [775 kB]
Err:1 https://ksp-ckan.s3-us-west-2.amazonaws.com/deb stable/main amd64 ckan all 1.30.0.21065
  File has unexpected size (774308 != 774780). Mirror sync in progress? [IP: 52.218.184.25 443]
  Hashes of expected file:
   - SHA256:20bd6bbfe1132e3656e696f4e77bce8a5e0ae7e605563814b2f149ace96dbd0b
   - SHA1:24241c129cf0b4478706b430362568df35d55cd5 [weak]
   - MD5Sum:3bd7a5896c00e81b1bcd1ebe19c34909 [weak]
   - Filesize:774780 [weak]
E: Failed to fetch https://ksp-ckan.s3-us-west-2.amazonaws.com/deb/pool/ckan_1.30.0.21065_all.deb  File has unexpected size (774308 != 774780). Mirror sync in progress? [IP: 52.218.184.25 443]
   Hashes of expected file:
    - SHA256:20bd6bbfe1132e3656e696f4e77bce8a5e0ae7e605563814b2f149ace96dbd0b
    - SHA1:24241c129cf0b4478706b430362568df35d55cd5 [weak]
    - MD5Sum:3bd7a5896c00e81b1bcd1ebe19c34909 [weak]
    - Filesize:774780 [weak]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?

```

DasSkelett:
Interesting, I had this happen to me a few days ago, but right now it is fine, for both nightly and stable.
But looking at the files, it _should_ also happen to me:
`deb/dists/stable/main/binary-amd64/Packages.gz` contains
```
Version: 1.30.0.21065
Filename: pool/ckan_1.30.0.21065_all.deb
Size: 774780
[...]

PR #3317, opened 2021-03-07
Suppress nightly debs with even builds, push release debs to s3 first

HebaruSan:
## Problem

Upgrading the stable .deb to v1.30.0 isn't working:

```
The following packages will be upgraded:
  ckan
1 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
Need to get 775 kB of archives.
After this operation, 0 B of additional disk space will be used.
Do you want to continue? [Y/n] Y
Get:1 https://ksp-ckan.s3-us-west-2.amazonaws.com/deb stable/main amd64 ckan all 1.30.0.21065 [775 kB]
Err:1 https://ksp-ckan.s3-us-west-2.amazonaws.com/deb stable/main amd64 ckan all 1.30.0.21065
  File has unexpected size (774308 != 774780). Mirror sync in progress? [IP: 52.218.184.25 443]
  Hashes of expected file:
   - SHA256:20bd6bbfe1132e3656e696f4e77bce8a5e0ae7e605563814b2f149ace96dbd0b
   - SHA1:24241c129cf0b4478706b430362568df35d55cd5 [weak]
   - MD5Sum:3bd7a5896c00e81b1bcd1ebe19c34909 [weak]
   - Filesize:774780 [weak]
E: Failed to fetch https://ksp-ckan.s3-us-west-2.amazonaws.com/deb/pool/ckan_1.30.0.21065_all.deb  File has unexpected size (774308 != 774780). Mirror sync in progress? [IP: 52.218.184.25 443]
   Hashes of expected file:
    - SHA256:20bd6bbfe1132e3656e696f4e77bce8a5e0ae7e605563814b2f149ace96dbd0b
    - SHA1:24241c129cf0b4478706b430362568df35d55cd5 [weak]
    - MD5Sum:3bd7a5896c00e81b1bcd1ebe19c34909 [weak]
    - Filesize:774780 [weak]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

Unrelated, the caching steps in our cache work flow have decided that they don't want to run in release work flows:

screenshot

## Cause

https://ksp-ckan.s3-us-west-2.amazonaws.com/deb/pool/ckan_1.30.0.21065_all.deb has 774308 bytes, but 
https://ksp-ckan.s3-us-west-2.amazonaws.com/deb/dists/stable/main/binary-amd64/Packages.gz says it should be 774780.

This happened because the deploy work flow to update the nightly build ran for v1.30.0 when 0376e0332f3815a0fc0e23329204bdab00b5dff5 was pushed, 2 minutes before the release work flow to update the stable build, so we ended up with the Packages.gz file from the stable build (of course) but the deb file from the nightly build.

(Something similar happened when #3304 and #3300 were merged, but we're not trying to fix that here.)

Trying to fix it after the fact by re-running the work flow didn't work, because it tried to upload assets to the release first, which GitHub rightfully didn't like.

## Changes

- Now the release work flow updates S3 before GitHub, so we can try re-doing the .deb if this happens again
- Now the deploy work flow only uploads deb files to S3 if the version ends in an odd digit, because even build numbers are for stable releases
- Now the caching steps are removed from the release work flow

Fixes the architectural side of #3309, since now only the release step would have updated S3 for a release. The current files for v1.30.0 won't be fixed by this, though, so we'll have to figure out a way to do that to close that issue.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,7 +2,8 @@ name: Deploy CKAN + NetKAN
 
 on:
   push:
-    branches: [master]
+    branches:
+      - master
 
 jobs:
   deploy:
@@ -14,6 +15,17 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Check version
+        id: check_version
+        run: |
+          VERSION=$(egrep '^\s*\#\#\s+v.*$' CHANGELOG.md | head -1 | sed -e 's/^\s*\#\#\s\+v//' -e 's/-.*$//')
+          if [[ $VERSION =~ [13579]$ ]]
+          then
+            echo '::set-output name=odd_build::true'
+          else
+            echo '::set-output name=odd_build::false'
+          fi
+
       - name: Installing build dependencies
         run: apt-get update && apt-get install -y git make sed gzip fakeroot lintian dpkg-dev
       - name: Installing runtime dependencies
@@ -22,6 +34,7 @@ jobs:
         run: |
           curl -fsSL https://get.docker.com -o get-docker.sh
           sh get-docker.sh
+
       - name: Restore cache for _build/tools
         uses: actions/cache@v1
         with:
@@ -45,10 +58,12 @@ jobs:
 
       - name: Build ckan.exe and netkan.exe
         run: ./build --configuration=Release
+
       - name: Build deb
         env:
           CODENAME: nightly
         run: ./build deb --configuration=Release --exclusive
+        if: ${{ steps.check_version.outputs.odd_build }}
       - name: Sign deb release
         env:
           CODENAME: nightly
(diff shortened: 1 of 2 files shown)


DasSkelett (approved):
> The current files for v1.30.0 won't be fixed by this, though, so we'll have to figure out a way to do that to close that issue.

For a moment I thought I might be able to do this manually via this AWS console, however I realized that building a new deb would likely also result in a different file with different hash and size, so I'd have to regenerate the Packages.gz and thus the (In)Release file, which is signed...

HebaruSan:
Ugh, I guess it doesn't run bash?

image

HebaruSan:
[`shell: bash`](https://github.com/KSP-CKAN/CKAN/commit/76f656b99ef8120511053e3bd099a8ec86a7f29f) fixed it:

image

So once again [GitHub's documentation lies](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell):

image

Bash is not the default shell on non-WIndows platforms. If it was, specifying `shell: bash` would have no effect.

HebaruSan:
Notes for the future, `echo '::set-output name=odd_build::false'` sets `odd_build` to the **string** "false", not the bool `false`. Non-null strings are treated as `true` by the work flow engine. So this didn't quite work.

Fixed in bf0d3a6e6dccdf98b1bf652892094470b3bb42a9.

PR #3323, opened 2021-03-10
Multiple search boxes with OR logic in GUI

### LINKED ISSUE #3071: [Feature] Filter Compatible & Installed
Problem
-------
Installed mods hidden when using filters that exclude them.

Suggestions
-----------
Ability to show installed mods regardless of the filter set, for example showing Installed mods with the Compatible filter regardless of if the mod is compatible or not.

### LINKED ISSUE #3301: Checkbox to (de)select all mods covers column header
Background
----------
* CKAN Version: 1.29.2
* KSP Version: 1.11.1.3066

nope
ckan bug

image

Problem
-------
**Describe the bug**

**Steps to reproduce**
* Go to...
* Click on...
* See error

**Expected behavior**
Not to cover the text

File attached

```
PLACE CODE HERE !
```

HebaruSan:
## Motivations

- The built in filters (Compatible, Installed, Cached, Newly Compatible, Upgradeable, Replaceable) cannot be combined with one another, via either AND- or OR-logic
- You can only filter by one tag or label at a time, via the filter dropdown
- The search box only supports AND-based searches
- You can only search for one author, language, depends, recommends, suggests, or conflicts at a time (within each field, not across them)
- The install-all checkbox covers some of the Inst column header, which looks worse on Windows, and its function is not fully clear

## Changes

image

- Now multiple values are supported for Author, Language, and the relationship searches
- Now the search dropdown supports searching by tags and labels
- A new tri-state toggle control is added for each of the previously built in filters: Compatible, Installed, Cached, Newly Compatible, Upgradeable, Replaceable
- Now if you choose a filter from the filter dropdown, the corresponding search will populate into the search fields
- Now the filter dropdown's caption always says "Filters" instead of changing based on the current filter, because you can see the currently active filter in the search boxes, and because many of the filters you can now construct are too complex to fit into the caption of the button
- A new "+" button next to the search dropdown can be clicked to add a new blank search box, the results of which are OR'd with any other searches, so we are finally able to combine Installed with Compatible (or anything else). Clearing a non-empty search deletes it.
- Your current searches are saved when you close the GUI and restored when you restart it, falling back to the active filter from previous versions the first time you open this version
- Now the install-all checkbox has a tooltip and is moved from the Inst header cell to directly above it, moving with the grid as searches are added and removed, and the header caption is restored to Installed to fill the newly restored space.

Addresses post-close comments in #2736 about wanting to search both name and description with OR-logic.
Fixes #3071.
Fixes #3301.


Olympic1:
I'm expecting questions about what ✓ , ✓✗ and ✗ mean. Can you make the options more descriptive, or (if possible) make a legend.

HebaruSan:
> I'm expecting questions about what ✓ , ✓✗ and ✗ mean.

Really? Huh, I thought they were pretty obvious.

> Can you make the options more descriptive, or (if possible) make a legend.

How about tooltips? They say "Yes", "Yes or No", and "No".

Olympic1:
> Really? Huh, I thought they were pretty obvious.

Just thinking ahead.

> How about tooltips? They say "Yes", "Yes or No", and "No".

I like this idea.

HebaruSan:
> > How about tooltips? They say "Yes", "Yes or No", and "No".
> 
> I like this idea.

Cool, already done:

https://github.com/KSP-CKAN/CKAN/blob/512fa7b512254137ffbc10788302f72f183ef65c/GUI/Controls/TriStateToggle.cs#L27

https://github.com/KSP-CKAN/CKAN/blob/512fa7b512254137ffbc10788302f72f183ef65c/GUI/Controls/TriStateToggle.cs#L38

https://github.com/KSP-CKAN/CKAN/blob/512fa7b512254137ffbc10788302f72f183ef65c/GUI/Controls/TriStateToggle.cs#L48

https://github.com/KSP-CKAN/CKAN/blob/512fa7b512254137ffbc10788302f72f183ef65c/GUI/Properties/Resources.resx#L402-L404

Olympic1 (approved):

HebaruSan:
Gonna hold off on merging so @DasSkelett can check out the i18n changes.

DasSkelett:
Hm, the symbols aren't rendered on my system:
image
Is this a font thing?

Should we put an "OR" to the left of all additional search boxes, so people now they're chained with OR? Even the tooltip only says "combine" and doesn't give a hint how they are combined.

Once you remove the first text box (ESC or the cross), CTRL+F to focus the new highest box doesn't work anymore.

HebaruSan:
> Hm, the symbols aren't rendered on my system:
> image
> Is this a font thing?

Ouch, yes, it's a font thing. Not sure why I thought I could get away with using Unicode this time. Will look for alternatives.

> Should we put an "OR" to the left of all additional search boxes, so people now they're chained with OR? Even the tooltip only says "combine" and doesn't give a hint how they are combined.

I thought about that, but in casual usage people sometimes swap the meanings of "and" and "or", because rather than evaluating a Boolean expression they're thinking of manipulating sets (which is just as valid, really): "I want to see compatible mods AND installed mods."

I thought that users would have to learn it by experimentation anyway, so better to say something vague like "combine" rather than give them a possibly confusing/false starting point.

> Once you remove the first text box (ESC or the cross), CTRL+F to focus the new highest box doesn't work anymore.

Good catch, we need to set `ActiveControl` on remove...

DasSkelett:
> Ouch, yes, it's a font thing. Not sure why I thought I could get away with using Unicode this time. I guess I'll need to make icons.

I'm surprised that my font has problems displaying Unicode though, I can't remember any problems anywhere else. I'm going to play around a bit and see if/how I can fix it on my system.
But if I already have problems, there's a decent chance it affects other Mono users as well.

> I thought about that, but in casual usage people sometimes swap the meanings of "and" and "or", because rather than evaluating a Boolean expression they're thinking of manipulating sets (which is just as valid, really): "I want to see compatible mods AND installed mods."
I thought that users would have to learn it by experimentation anyway, so better to say something vague like "combine" rather than give them a possibly confusing/false starting point.

I can see your point there. But I'm not really happy leaving our users on their own having to find out how the software works through experimentation. Maybe I can think of another way to make it clear how the searches interact with each other.

HebaruSan:
> But if I already have problems, there's a decent chance it affects other Mono users as well.

And who knows about Windows. Force pushed some new icons and updated button backcolor behavior.

> Once you remove the first text box (ESC or the cross), CTRL+F to focus the new highest box doesn't work anymore.

Is this better for you now? Now updating `ActiveControl` on remove.

HebaruSan:
Need a few more pixels of space for the labels on Windows...

image

HebaruSan:
> I can see your point there. But I'm not really happy leaving our users on their own having to find out how the software works through experimentation. Maybe I can think of another way to make it clear how the searches interact with each other.

OK, it's a bit more visually busy, but if it helps users to understand how it works, then that's fine:

image

DasSkelett:
Working on some fixes, need some time navigating through all those `SetSearches()`, `Filter()`, `ApplySearches()`...

DasSkelett:
Force-pushing a simple rebase without any code changes so I can do a PR to your branch which also includes some adjustments to the pt-BR localization.

DasSkelett (approved):
Okay, appears to be stable now in my testing. Let's hope people appreciate all the new search combination possibilities!

PR #3325, opened 2021-03-12
Time out after 10 seconds on text downloads in NetKAN

DasSkelett:
## Problem
Our bot slows down a lot when one of the sites we get our metadata from go down.
This frequently happens with ksp.spacetux.net, ksp-avc.cybutek.net, or right now it is taniwha.org, where the mods if taniwha are hosted on.

With taniwha.org, inflation currently takes 7 minutes per module with kref pointing to that domain:
image

Preferably, a single website being down shouldn't cause our entire infra to slow down massively.

## Changes
Introduce a new `TimeoutWebClient` (taken from https://stackoverflow.com/a/3052637)
It sets the timeout of the underlying `WebRequest` class, which makes requests fail after a certain time period.

It defaults to `Timeout.Infinite`, which should match the behavior of the default `WebClient` (where infinite seems to mean ~2 minutes, or until a stateful firewall closes the connection).

We set a timeout of 10 seconds for `DownloadText()`, only for requests done via NetKAN code.
CKAN users with slower or unstable network connections aren't affected, and neither are mod zip downloads, which can of course take much longer than 10 seconds for bigger zips.
`DownloadText()` is used for API requests and version file downloads in NetKAN.

Since `DownloadText()` retries 3 times before calling it off, this means it takes around 30 seconds for the inflation to fail.
This seems to vary a lot in my testing with taniwha mods, maybe because the domain fails to resolve at all, so things go crazy. So I'm not able to reproduce the 7 minutes it takes the bot to fail, but I think this depends on the behavior of the DNS server used or something.


Core/Net/Net.cs
```diff
@@ -194,11 +194,11 @@ public static void DownloadWithProgress(ICollection<DownloadTarget> downloadTarg
             }.DownloadAndWait(downloadTargets);
         }
 
-        public static string DownloadText(Uri url, string authToken = "", string mimeType = null)
+        public static string DownloadText(Uri url, string authToken = "", string mimeType = null, int timeout = Timeout.Infinite)
```
HebaruSan, line 197: Let's document the units of the new param (milliseconds) in a `/// <param>` comment.

HebaruSan: I think the current default is 100 seconds, not infinite:

https://docs.microsoft.com/en-us/dotnet/api/system.net.webrequest.timeout?view=netframework-4.7.2

> The default value is defined by the descendant class.

https://docs.microsoft.com/en-us/dotnet/api/system.net.httpwebrequest.timeout?view=netframework-4.7.2#System_Net_HttpWebRequest_Timeout

> The default value is 100,000 milliseconds (100 seconds).

Not sure how to reconcile that with multi-minute delays in indexing. Retries, maybe? 

But we should consider whether changing it from 100 seconds to infinite is desirable.

HebaruSan: This sounds suspicious, I wonder why they felt the need to spell this out:

> A Domain Name System (DNS) query may take up to 15 seconds to return or time out. If your request contains a host name that requires resolution and you set Timeout to a value less than 15 seconds, it may take 15 seconds or more before a WebException is thrown to indicate a timeout on your request.

DasSkelett: > I think the current default is 100 seconds, not infinite

Huh, I was sure I've read Infinite. Anyway, that gets us closer to the 7 minutes we see with the bot, 3*100s = 5m

> But we should consider whether changing it from 100 seconds to infinite is desirable.

Yeah I think I'm going to change it to a default of 100s then.

> This sounds suspicious, I wonder why they felt the need to spell this out

I think it means that the timeout is not applied to the DNS resolution. So if it's that one that fails / times out, the request only fails after 15 seconds, not already 10 seconds.
If you want, we can also set the timeout to 15 seconds, but I don't think it's necessary. If It's not DNS that fails, we can still get the 10s timeout.

HebaruSan (approved):
Feel free to merge, let's see how it goes. :hourglass_flowing_sand:

DasSkelett:
Looks like it worked, it spends much less time on CustomBulkheadProfiles, Earlybird and ExtraPlanetaryLaunchpads:
CustomBulkheadProfiles
Earlybird
ExtraPlanetaryLaunchpads

And so far no new time out reported.

linuxgurugamer:
Just an FYI, I've updated my ddclient time to run every 5 minutes, hopefully that will catch any unexpected DHCP changes quickly

PR #3326, opened 2021-03-17
Netkan errors and warnings for Harmony bundlers

HebaruSan:
## Background

See KSP-CKAN/NetKAN#8288, we are attempting to prevent mods that bundle conflicting versions of Harmony from clashing with each other.

### Harmony 1

Before that PR, all Harmony-based mods in CKAN bundled Harmony 1, and CKAN installed them as-is. They may continue to do so, but only if they are appropriately flagged in the metadata:

```json
    "provides": [ "Harmony1" ],
    "x_netkan_staging": true,
    "x_netkan_staging_reason": "Bundles Harmony. Please check the DLL's version, and if it's 2 or later, remove the provides and add a depends on Harmony2.",
```

A mod with this metadata will be treated as incompatible with `Harmony2`, and each time it releases a new version, it will be staged for manual review so we can check whether the author has upgraded to Harmony 2.

### Harmony 2

Mods that use Harmony 2 will share a common dependency on the new `Harmony2` module, and any bundled copies they may include will be filtered out:

```json
    "depends": [
        { "name": "Harmony2" }
    ],
    "install": [ {
        "filter_regexp": [ "0Harmony.*\\.dll" ]
    } ]
```

## Problem

The remaining gap in this strategy is that any mod author might decide to add Harmony to an existing mod at any time, and we would be none the wiser. There may even be some existing ones that we didn't catch in the first round of updates.

## Changes

Now if a mod installs a DLL with "Harmony" in its name and it _doesn't_ provide Harmony1, the inflator emits an error, to prevent any newly bundled files from causing conflicts in user installs. These must then be inspected manually to determine which version of Harmony is bundled, and then the appropriate solution will be applied.

Now if a mod _does_ provide Harmony1 but doesn't seem to bundle it, the inflator emits a warning.

The `Harmony2` module is exempted from these checks.

### DIFF
--- a/Netkan/CKAN-netkan.csproj
+++ b/Netkan/CKAN-netkan.csproj
@@ -127,6 +127,7 @@
     <Compile Include="Validators\CraftsInShipsValidator.cs" />
     <Compile Include="Validators\DownloadVersionValidator.cs" />
     <Compile Include="Validators\HasIdentifierValidator.cs" />
+    <Compile Include="Validators\HarmonyValidator.cs" />
     <Compile Include="Validators\InstallsFilesValidator.cs" />
     <Compile Include="Validators\InstallValidator.cs" />
     <Compile Include="Validators\IsCkanModuleValidator.cs" />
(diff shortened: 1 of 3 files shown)


DasSkelett (approved):
Works like a charm, thanks! Now we are prepared for all eventualities.

Issue #221, opened 2014-10-31
Failure on zipfiles created by some unix utilities


pjf:
We seem to have [this bug](http://community.sharpdevelop.net/forums/t/21758.aspx) which is different from [this other bug](http://community.sharpdevelop.net/forums/t/11466.aspx) which is already fixed in the SharpZipLib version we're using.

Reported by @malahx in KSP-CKAN/CKAN-Meta#33.

I don't suppose C# supports monkey-patching, does it? Because we could monkey-patch this away easily enough. :)

PR #3329, opened 2021-03-21
Switch from outdated SharpZipLib.Patched to newer SharpZipLib

DasSkelett:
## Motivation
https://github.com/KSP-CKAN/NetKAN/issues/8425 and https://github.com/KSP-CKAN/NetKAN/issues/8315 show some weird errors when trying to extract certain zips with special characters on Windows systems with CJK locales.
_Maybe_ this is a bug of `SharpZipLib`, and _maybe_ it is already fixed in a newer versions. It's a total stab in the dark, but worth a try. I've asked affected users to try a debug build with an updated version of the package, we should know when we get an answer.

In any case, there have been some improvements to that library in the releases since, the usual performance enhancements, and support for BZip2 compression.

We have been using a "Patched" version of this library, which has been deprecated now. Its [NuGet page](https://www.nuget.org/packages/ICSharpCode.SharpZipLib.Patched) doesn't give any details on what it's patching, but it seems to be related to:
* https://github.com/KSP-CKAN/CKAN/issues/221
* https://github.com/KSP-CKAN/CKAN/pull/222
* https://github.com/pjf/SharpZipLib/commit/8ec3e47283b2c201474f731a52b8c752c5127e2e
The fix is applied upstream since v1.0.0-rc1:
* https://github.com/icsharpcode/SharpZipLib/pull/53

## Changes
Switch to the upstream SharpZipLib, v1.3.1. I haven't seen any breaking changes, and mod installation and inflation still works perfectly fine in my testing.

Add `venv/` folders to `.gitignore` - I have one inside `bin/` for the Python scripts there.

Add `exitstatus` to `bin/requirements.txt`. For some reason I didn't catch the missing dependency during the PRs for the `ckan-merge-pr.py`.

(Gonna give this the "Enhancement" flair for the time being and not say it closes the issue https://github.com/KSP-CKAN/NetKAN/issues/8425, since we don't know for sure yet)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ StyleCop.Cache
 *.ncrunchsolution
 *.pyc
 node_modules/
+venv/
 
 # Build results
 [Dd]ebug/
--- a/Core/CKAN-core.csproj
+++ b/Core/CKAN-core.csproj
@@ -38,7 +38,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Autofac" Version="4.9.4" />
-    <PackageReference Include="ICSharpCode.SharpZipLib.Patched" Version="0.86.5.1" />
+    <PackageReference Include="SharpZipLib" Version="1.3.1" />
     <PackageReference Include="log4net" Version="2.0.10" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
   </ItemGroup>
--- a/Netkan/CKAN-netkan.csproj
+++ b/Netkan/CKAN-netkan.csproj
@@ -41,7 +41,7 @@
     <PackageReference Include="AWSSDK.Core" Version="3.3.103.64" />
     <PackageReference Include="AWSSDK.SQS" Version="3.3.102.31" />
     <PackageReference Include="CommandLineParser" Version="1.9.71" />
-    <PackageReference Include="ICSharpCode.SharpZipLib.Patched" Version="0.86.5.1" />
+    <PackageReference Include="SharpZipLib" Version="1.3.1" />
     <PackageReference Include="log4net" Version="2.0.10" />
     <PackageReference Include="Namotion.Reflection" Version="1.0.7" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
--- a/Tests/Tests.csproj
+++ b/Tests/Tests.csproj
@@ -42,7 +42,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Castle.Core" Version="4.4.1" />
-    <PackageReference Include="ICSharpCode.SharpZipLib.Patched" Version="0.86.5.1" />
+    <PackageReference Include="SharpZipLib" Version="1.3.1" />
     <PackageReference Include="log4net" Version="2.0.10" />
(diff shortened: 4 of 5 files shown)


HebaruSan:
> image

Except that _will_ auto-close; all it takes is linking to the issue after the word "closes" or "fixes":

image

DasSkelett:
Ugh, I've paid attention that it doesn't underline it in the PR preview, but apparently it only happens once the PR is created. Let's do some trickery.

HebaruSan (approved):
:game_die:

HebaruSan:
Hmm, this might have caused a new problem:

image

That mod hasn't updated since August, and it doesn't contain any duplicates:

image

Is this version of the lib _less_ able to handle CJK locales?

DasSkelett:
Interestingly this inflates and installs successfully for me, even though I had problems with UTF-8 in #3323 :confused:

HebaruSan:
> Interestingly this inflates and installs successfully for me

Me, too. Maybe the Inflator's instance is missing some locale data?

DasSkelett:
Now this is interesting. Running `mono netkan.exe Wanhu-Common.netkan` inside a `mono:latest` container:
```
root@3f119569761c:/ckan/_build# mono netkan.exe --version
v1.30.1+bf7064fbd8fa

root@3f119569761c:/ckan/_build# mono netkan.exe Wanhu-Common.netkan
5855 [1] FATAL CKAN.NetKAN.Program (null) - https://spacedock.info/mod/2507/LongMarch-10?CZ-10?/download/1.3 is not a valid ZIP file:
Error instep EntryHeader for Wanhu/Wanhu_Parts/Flags/Flag??????.png: Exception during test - 'Name is invalid'
```

I expected to see the "multiple files" error, but instead got the on from https://github.com/KSP-CKAN/NetKAN/issues/8425, but on Linux :exploding_head:

techman83:
Good chance the container is missing some locale data. I doubt it's included.

DasSkelett:
Looks like SharpZipLib changed the default for reading and writing zips _away_ from UTF-8 in version 1.1.0:
https://github.com/icsharpcode/SharpZipLib/issues/591

Setting `ICSharpCode.SharpZipLib.Zip.ZipStrings.UseUnicode = true;` right at the beginning in `Program.cs` solves the above exception and makes `netkan.exe` work in a completely fresh Mono container again :tada:

HebaruSan:
> Setting `ICSharpCode.SharpZipLib.Zip.ZipStrings.UseUnicode = true;` right at the beginning

Is there a non-static equivalent? That's a big ugly in terms of encapsulation.

DasSkelett:
There seems to be an overhaul of this in progress, which also mentions going back to support Unicode by default (for creating new zips judging from the discussion, not sure if and how this would affect reading zips):
https://github.com/icsharpcode/SharpZipLib/pull/592

Otherwise I haven't really digged into it yet, I'm currently trying to write a unit test for this in order to prevent this in the future, and also simpler patch testing.

HebaruSan:
It looks like the SharpZipLib team is moving in a good direction, but it's less clear when those changes will land. I'm comfortable setting the static bool as a short term workaround, and I've subscribed to their repo's release notifications so we can revisit this.

PR #3334, opened 2021-03-28
Reset cache dir to default if creation fails

### LINKED ISSUE #3333: System Access Exception after changing PC username
After changing my name (and with this the username of my PC), CKAN seems to be unable to grasp that this change has occured and gives me an error through the console stating there was a system access exception after trying to reach the C:\Users\[oldusername]\AppData\Local\CKAN\Downloads

*oldusername refers to the username of my PC before changing the name. It's most definitely part of the PCs registries since deleting CKAN and downloading it again does absolutely nothing to fix the issue. I'd love to get some help with how to fix this since it not only makes CKAN unusable for me but also means that my deadname is still present somewhere on my PC despite me having gone through and removed it from the registries once already. Very annoying, plz help
(name removed from screenshot on purpose)
Screenshot (1008)

HebaruSan:
## Problem

If the download cache folder disappears, and the user can't write to its previous location anymore, CKAN refuses to start:

image

## Cause

We try to create the cache path at startup to handle onboarding for new users and users upgrading to the version of CKAN after the cache was unified. `Directory.CreateDirectory` throws if this fails, which mostly makes sense because we can't really operate without a valid download cache.

Note that this happens in the CmdLine startup code before we attempt to load GUI, so improving the error message UI isn't a viable option.

## Changes

Now if the configured cache path doesn't exist and we can't create it, we revert it to the default setting and try again. This will work seamlessly if the only thing that's happened is that the user's name has changed, because the default setting will always contain the current user name and should always be creatable. If some other problem blocks the second `CreateDirectory` attempt, we will still print the exception as before so that problem can be diagnosed.

Also a bunch of references to the Windows registry in function names and comments have been removed to reduce confusion. And some test files and code have been multi-gamed (KSP → GameInstance). And a few public functions are refactored as private.

Fixes #3333.

### DIFF
--- a/Core/GameInstanceManager.cs
+++ b/Core/GameInstanceManager.cs
@@ -64,19 +64,20 @@ namespace CKAN
         {
             User = user;
             Configuration = configuration ?? ServiceLocator.Container.Resolve<IConfiguration>();
-            LoadInstancesFromRegistry();
+            LoadInstances();
+            LoadCacheSettings();
         }
 
         /// <summary>
         /// Returns the preferred KSP instance, or null if none can be found.
         ///
         /// This works by checking to see if we're in a KSP dir first, then the
-        /// registry for an autostart instance, then will try to auto-populate
+        /// config for an autostart instance, then will try to auto-populate
         /// by scanning for the game.
         ///
-        /// This *will not* touch the registry if we find a portable install.
+        /// This *will not* touch the config if we find a portable install.
         ///
-        /// This *will* run KSP instance autodetection if the registry is empty.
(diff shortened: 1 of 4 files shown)


HebaruSan:
Hi @DeNomaly, if you'd like to try a test build of this fix instead of editing the JSON file by hand, you can access it at the bottom of this page:

https://github.com/KSP-CKAN/CKAN/actions/runs/695780269

It should make CKAN switch over to your new name at startup automatically.

DasSkelett (approved):
Looks good and makes sense, thanks!

PR #3340, opened 2021-04-05
PTBR Translation

gsantos9489:
Translation for the *.resx files to Brazilian Portuguese (pt-BR).

If you guys find anything wrong, please contact me.

I'm already a software developer, but I never ventured before on open-source projects.
So, this is my first contribution ever. Hope I can contribute more in the future.

Best Regards

PR #3341, opened 2021-04-08
longer refresh label &  better browse box for FR loc

vinix38:
Refresh label is now better thanks to #3340 , and I also fixed a browse box that was not wide enough. It is a separate pull request because I don't know if you can or how to add commits to an existing pull request


HebaruSan:
Thanks! I just added this to #3340.

https://github.com/KSP-CKAN/CKAN/pull/3340/commits/9f1d2aae7124489d90e83dd7e451b46c409e36aa

PR #3344, opened 2021-04-11
Replace repo-reinst with kraken, handle in UIs

### LINKED ISSUE #1889: Default Exception handling from UpdateAllRepositories confusing
CKAN Version: 1.20.0 

When GUI/MainRepo.cs examines exceptions returned from Core/Net/Repo.cs/UpdateAllRepositories, it has customised messaging for UriFormatException and MissingCertificateKraken. All other exceptions throw a dialog: `Failed to connect to repository. Exception: <details>`. 

Change "Relationship changes now prompt reinstalling (#1730, #1885)" means UpdateAllRepositories is throwing more different exceptions. We need to handle them better, since some are related to relationship changes now, rather than a failure to connect.

### LINKED ISSUE #3335: [Bug] Repo-update reinstall doesn't handle TooManyModsProvideKraken
## Background

If you update the registry and one of your installed mods has changed metadata, you'll be asked whether you want to reinstall it. Newly added dependencies will be pulled in automatically.

```
The following mods have had their metadata changed since last update:

- Buffalo v2.9.0
- WildBlueTools v1.81.5
- ExtraPlanetaryLaunchpads 6.8.3.0
- ClassicStockResources v1.1.4

You should reinstall them in order to preserve consistency with the repository.

Do you wish to reinstall now?
```

## Problem

If the newly added dependencies are virtual modules requiring the user to make a choice among several providing modules, the reinstall fails.

```
Failed to connect to repository. Exception: Too many mods provide Buffalo-PlayMode:
* Buffalo-PlayMode-CRP v2.10.2
* Buffalo-PlayMode-ClassicStock v2.10.2
* WildBlue-PlayMode-Pristine v1.82.1
* WildBlue-PlayMode-Simplified v1.82.1
* WildBlue-PlayMode-USI v1.0.0.2
```

First reported in a comment in KSP-CKAN/NetKAN#8420.

This is relevant now because KSP-CKAN/NetKAN#8275 has restructured the WildBlueTools mods using virtual modules, and users who already had those mods installed need a viable upgrade path.

## Causes

The repo update code just calls `ModuleInstaller.Upgrade` without any UI component for prompting the user, because it's embedded deeply inside Core code as a downstream call from `UpdateAllRepositories`. The exception-based `TooManyModsProvideKraken` mechanism is based on the assumption that UI-level code will be updating a change set and trying again, which can't happen this far inside of core code.

https://github.com/KSP-CKAN/CKAN/blob/bf7064fbd8fab16363b6e0b7e6869a63792738ee/Core/Net/Repo.cs#L189-L248

## Ideas

`TooManyModsProvideKraken` may finally have to go. If it was replaced with a callback mechanism passed in to `ModuleInstaller`'s public methods, then in principle we could make this work _and_ dramatically simplify all of the installation UI code.

HebaruSan:
## Background

When you refresh the registry, if the metadata for one of your installed modules has been changed upstream, CKAN will ask whether you want to reinstall those modules, and if you say Yes, it will reinstall them.

```
The following mods have had their metadata changed since last update:

- Buffalo v2.9.0
- WildBlueTools v1.81.5
- ExtraPlanetaryLaunchpads 6.8.3.0
- ClassicStockResources v1.1.4

You should reinstall them in order to preserve consistency with the repository.

Do you wish to reinstall now?
```

## Problems

If one of your installed modules has been updated to have a new dependency on a virtual module, the reinstall fails:

```
Failed to connect to repository. Exception: Too many mods provide Buffalo-PlayMode:
* Buffalo-PlayMode-CRP v2.10.2
* Buffalo-PlayMode-ClassicStock v2.10.2
* WildBlue-PlayMode-Pristine v1.82.1
* WildBlue-PlayMode-Simplified v1.82.1
* WildBlue-PlayMode-USI v1.0.0.2
```

## Causes

`Repo.HandleModuleChanges` creates its own `ModuleInstaller` to try to do the reinstall.

When a dependency on a virtual module is found, `ModuleInstaller` (via other classes) throws `TooManyModsProvideKraken`. The UI is supposed to catch this, prompt the user, add the module they chose to the change set, and try again. However, this is impossible in the case of a repo-reinstall, because the UI does not know what changes `Repo.HandleModuleChanges` was attempting to make and cannot retry the repo update to fix it.

## Changes

Now instead of initiating a half-hearted attempt at doing a reinstall itself, `Repo.HandleModuleChanges` throws a new `ReinstallModulesKraken` to instruct the UI-specific calling code to reinstall the given modules. Cmdline, ConsoleUI, and GUI are updated to catch this exception and launch into their upgrade flows for the given modules.

Now ConsoleUI and GUI perform upgrades after installs, so the installs (which contain the user's choice of virtual module providers) can satisfy dependencies needed by the upgrades.

As a side fix, Cmdline's `ckan upgrade` command is updated to handle prompting for virtual module providers (in a shared function also now called by the repo update code as needed).

Fixes #1889.
Fixes #3335.

### DIFF
--- a/Cmdline/Action/Upgrade.cs
+++ b/Cmdline/Action/Upgrade.cs
@@ -178,7 +178,7 @@ namespace CKAN.CmdLine
         }
 
         /// <summary>
-        /// Upgrade some modules by their CkanModules
+        /// Upgrade some modules by their identifier and (optional) version
         /// </summary>
         /// <param name="manager">Game instance manager to use</param>
         /// <param name="user">IUser object for output</param>
--- a/Cmdline/Action/Upgrade.cs
+++ b/Cmdline/Action/Upgrade.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 ﻿using System.Collections.Generic;
+using System.Transactions;
 using log4net;
 using CKAN.Versioning;
 
@@ -83,7 +84,6 @@ namespace CKAN.CmdLine
 
             try
             {
-                HashSet<string> possibleConfigOnlyDirs = null;
                 var regMgr = RegistryManager.Instance(ksp);
                 var registry = regMgr.registry;
                 if (options.upgrade_all)
@@ -119,13 +119,12 @@ namespace CKAN.CmdLine
                                 mod.Key);
                         }
                     }
-
-                    ModuleInstaller.GetInstance(ksp, manager.Cache, User).Upgrade(to_upgrade, new NetAsyncModulesDownloader(User, manager.Cache), ref possibleConfigOnlyDirs, regMgr, true, true);
+                    UpgradeModules(manager, User, ksp, true, to_upgrade);
                 }
                 else
                 {
                     Search.AdjustModulesCase(ksp, options.modules);
-                    ModuleInstaller.GetInstance(ksp, manager.Cache, User).Upgrade(options.modules, new NetAsyncModulesDownloader(User, manager.Cache), ref possibleConfigOnlyDirs, regMgr);
+                    UpgradeModules(manager, User, ksp, options.modules);
                 }
                 User.RaiseMessage("");
             }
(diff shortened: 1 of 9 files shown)


Cmdline/Action/Upgrade.cs
```diff
@@ -160,5 +159,80 @@ public int RunCommand(CKAN.GameInstance ksp, object raw_options)
...
+                    installer.Upgrade(identsAndVersions, downloader,
+                        ref possibleConfigOnlyDirs, regMgr, true),
+                m => identsAndVersions.Add(m.identifier)
+            );
+        }
+
+        // System.Action<ref T> isn't allowed
+        private delegate void AttemptUpgradeAction(ModuleInstaller installer, NetAsyncModulesDownloader downloader, RegistryManager regMgr, ref HashSet<string> possibleConfigOnlyDirs);
+
+        private static void UpgradeModules(
```
DasSkelett, line 200: I think this one needs some documenation.
It takes a lot of scrolling back and forth to figure out what the callbacks do and why they're there.
Something that explains the this private method provides the outer logic to reduce code duplication,
the `attemptUpgradeCallback` should call the according `ModuleInstaller.Upgrade()`,
that `addUserChoiceCallback` then should modify the list of mods to install to add the user's choice after a `TooManyModsProvideKraken` was thrown, etc.

GUI/Main/MainRepo.cs
```diff
@@ -97,6 +97,11 @@ private void UpdateRepo(object sender, DoWorkEventArgs e)
             {
                 errorDialog.ShowErrorDialog(ex.ToString());
             }
+            catch (ReinstallModuleKraken)
+            {
+                // Bypass the generic error dialog so the Post can handle this
+                throw;
```
DasSkelett, line 103: I thought to remember that we did this (catch-rethrow) somewhere else previously and it caused compiler warnings.
But I don't see any here, so I guess we're fine?

HebaruSan: Maybe in that other case we were re-throwing _all_ exceptions, which would be equivalent to removing the `catch` statements completely?

DasSkelett: Ah good point, maybe the compiler is smart enough to detect that not handling it would change the code path due to the general `catch (Exception)` below, and this catch-rethrow prevents it for this exception :bulb:

Cmdline/Action/Upgrade.cs
HebaruSan: Does that help?  1bca7e8d0fdc0882c6f51dc19e78be3d826ac90e

DasSkelett: Yup, looks good. Thanks!

DasSkelett (approved):
Nice, works perfectly!

PR #3345, opened 2021-04-11
Tell SharpZipLib to use Unicode when opening zips

### LINKED ISSUE #3342: Name is invalid error for file in TundraExploration, seemingly in CJK locales
Background
----------
* Operating System: windows10 home edition（x64）
* CKAN Version: v1.30.0
* KSP Version: 1.10.1.2939

no

Problem
-------
**Describe the bug**
Can not install "Tundra Exploration - Stockalike Dragon V2 and Falcon 9 3.3.0 (spacedock.info, 130.2 MiB)"

**Steps to reproduce**
When it finishes downloading the mod, an error window will appear.

CCAGG0L${5H0Z)6$CPZAB`T(1)

```
PLACE CODE HERE !
```

DasSkelett:
## Problem
Follow-up to #3329 – we still don't know whether it fixed #3342, but it probably didn't.
SharpZipLib did indeed change stuff regarding zip files with Unicode chars in the path strings, however not the way we wanted (https://github.com/icsharpcode/SharpZipLib/issues/591), instead they are now defaulting to _not_ use Unicode. This makes some mod inflations fail:
image

### A short zip + Unicode introduction according to how I understand it:
**From zip(1):**
> Unicode.
Though the zip standard requires storing paths in an archive using a specific character set, in practice zips have stored paths in archives in whatever the local character set is. This creates problems when an archive is created or updated on a system using one character set and then extracted on another system using a different character set. 
When compiled with Unicode support enabled on platforms that support wide characters, **zip now stores, in addition to the standard local path for backward compatibility, the UTF-8 translation of the path.** This provides a common universal character set for storing paths that allows these paths to be fully extracted on other systems that support Unicode and to match as close as possible on systems that don't.

If a file path string in a zip contains Unicode chars, some zip programs seem to save two versions of this path string: one "standard local path for backward compatibility", and the UTF-8 translation of it.
Now depending on the tool and configuration used to read these zips again, they either try to use the "standard" path or the UTF-8 path, where the standard path seems to be failing on some systems, probably depending on whether you have the matching locales/code pages installed or not. The Unicode path should always work fine.

## Solution
Now we force SharpZipLib to use the Unicode path. This is currently only possible with a global static bool, but things are likely to change again soon (https://github.com/icsharpcode/SharpZipLib/pull/592).
The bool is set to `true` in the `NetFileCache` constructor, which I think is the most straight-forward and reliable place for it, since it is guaranteed to be called whenever and before we use the cache, which is when we are about to try to read zips.

I've also managed to write a unit test for it, thanks to @techman83 providing a small test zip that reproduces the behaviour we see from all the other international zips.
The test isn't worth much at least on my system because the zips somehow are read successfully even without `ZipStrings.UseUnicode = true;`, but in a fresh `mono` container the test would fail without `ZipStrings.UseUnicode = true;`.

I've also marked the `GH1866` tests with a new category `Display` to have an easy way to ignore them, since they require either a real or a mock up display (which I didn't want to bother with in the Docker container).

Hopefully fixes #3342
Likely fixes all the inflation errors with `Name is invalid`, at least inflating them in a mono container works with the fix (and fails without it)


HebaruSan:
> The bool is set to `true` in the `NetFileCache` constructor

What do you think about doing this in `NetFileCache`'s static constructor instead? `ZipValid` is static, so it would be possible to call it without the constructor running. But I think static constructors are called before static members of a class are accessed.

DasSkelett:
Oh, good idea! They are also only called once, which is not required but nice to have here.

HebaruSan (approved):
Let's try it! :flying_saucer:

techman83:
We can probably use xvfb if we need to run headless tests that require a framebuffer. 
```shell
➜  ~ Xvfb :99 &; export DISPLAY=:99; echo $DISPLAY
[1] 24107
:99
```

DasSkelett:
> We can probably use xvfb if we need to run headless tests that require a framebuffer.

We already to so in CI:
https://github.com/KSP-CKAN/CKAN/blob/9c265dc4977032b5e13bb8b71c6cd704e4216d97/.github/workflows/build.yml#L54

But I don't want to install and set up xvfb a hundred times in a short-lived container for testing something completely unrelated.
We aren't stopping to run these tests in CI, no worries.
It's just a simple change for a personal benefit, which other people might find useful as well (especially when testing this PR).

DasSkelett:
It actually worked! :star_struck: 
The inflation errors for the mods are disappearing, and the Tundras got successfully indexed!
* https://github.com/KSP-CKAN/CKAN-meta/commit/f64564bfa0092334200c42c23b40b2dc68dc5209
*  https://github.com/KSP-CKAN/CKAN-meta/commit/52fa16d94324d55bcc08ccc339b685741fb5a3cb

techman83:
Heck yeah! Good work @DasSkelett

PR #3346, opened 2021-04-12
Make Incompatible Mods warning dialog use newlines instead of commas

### LINKED ISSUE #2954: List of incompatible mods is hard to read
When starting a game with a number of mods which, according to CKAN, are not compatible, a dialog is show with all the mods listed in a single line of text.  It's hard to read.

Would be better to show a listbox , or at least a text field with one mod per line

DeltaDizzy:
Closes #2954


DasSkelett (approved):
Thanks, looking good!
Now LGG can install all the outdated mods without getting a headache trying to read the warning message :grin:

PR #3347, opened 2021-04-12
Skip modules with parse errors in deserialization

HebaruSan:
## Problem

1. Refresh the registry with CKAN v1.30.0
2. Close CKAN
3. Launch CKAN v1.29.2 or earlier

```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> CKAN.BadMetadataKraken: The license SomethingWeird is invalid
   at CKAN.License..ctor(String license)
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.RuntimeType.CreateInstanceImpl(BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture, Object[] activationAttributes, StackCrawlMark& stackMark)
   at System.Activator.CreateInstance(Type type, BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture, Object[] activationAttributes)
   at System.Activator.CreateInstance(Type type, Object[] args)
   at CKAN.JsonSimpleStringConverter.ReadJson(JsonReader reader, Type objectType, Object existingValue, JsonSerializer serializer)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.DeserializeConvertable(JsonConverter converter, JsonReader reader, Type objectType, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateList(IList list, JsonReader reader, JsonArrayContract contract, JsonProperty containerProperty, String id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateList(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, Object existingValue, String id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
[...]

Issue #3233, opened 2020-12-09
[Bug] opend ckan, getting this error popup, opend it as admin, still not working.


TheEpicGamerL:
Background
----------
* Operating System: windows 10 64bit
* CKAN Version: 1.29.0 not sure
* KSP Version: 1.10.1

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**

Problem
-------
open ckan, pop up error.

**Steps to reproduce**
click on ckan .exe get error.

**Expected behavior**
ckan opening

**Screenshots (if applicable)**

**CKAN error code (if applicable):**
```
Unhandled exception:
Newtonsoft.Json.JsonReaderException: Unexpected end of content while loading JArray. Path 'available_modules.RP-0.module_version['v1.6'].recommends[21]', line 648536, position 8403726.
   in Newtonsoft.Json.Linq.JContainer.ReadTokenFrom(JsonReader reader, JsonLoadSettings options)
   in Newtonsoft.Json.Linq.JArray.Load(JsonReader reader, JsonLoadSettings settings)
   in Newtonsoft.Json.Linq.JToken.ReadFrom(JsonReader reader, JsonLoadSettings settings)
   in CKAN.JsonRelationshipConverter.ReadJson(JsonReader reader, Type objectType, Object existingValue, JsonSerializer serializer)
   in Newtonsoft.Json.Serialization.JsonSerializerInternalReader.DeserializeConvertable(JsonConverter converter, JsonReader reader, Type objectType, Object existingValue)
   in Newtonsoft.Json.Serialization.JsonSerializerInternalReader.SetPropertyValue(JsonProperty property, JsonConverter propertyConverter, JsonContainerContract containerContract, JsonProperty containerProperty, JsonReader reader, Object target)
   in Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateObject(Object newObject, JsonReader reader, JsonObjectContract contract, JsonProperty member, String id)
   in Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   in Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
[...]

Issue #3229, opened 2020-12-07
CKAN not starting up


andyu11:
I have noticed something has gone wrong with my CKAN when I try to start up I get an error message with the following:

Unhandled exception:
Newtonsoft.Json.JsonReaderException: Unterminated string. Expected delimiter: ". Path 'available_modules.WarpShip.module_version['0.3.2'].resources.repository', line 878236, position 44. 
 
I have tried reinstalling several times with no luck and I think it is on 1.29.0 but I am not 100% sure. I use 64 bit Windows 10. I am sorry if this is has been posted before but I couldn't find anything that works. Can anyone help me out here please? Thanks.

HebaruSan:
Your `<KSP>/CKAN/registry.json` file got corrupted. Unfortunately you'll have to delete it and try again.

andyu11:
Thank you for the help, I will locate and delete the file and I will comment again if it worked or not.

andyu11:
> Your `<KSP>/CKAN/registry.json` file got corrupted. Unfortunately you'll have to delete it and try again.

I could not locate this file, where would it be under?

HebaruSan:
It's in the CKAN folder under your Kerbal Space Program folder.

andyu11:
> It's in the CKAN folder under your Kerbal Space Program folder.

The CKAN folder isn't there anymore for some reason I don't remember if I did anything to it but its not there anymore and I don't know why

HebaruSan:
If you're still getting the `Newtonsoft.Json.JsonReaderException` error, then it's still there somewhere. Unfortunately we can't help you search your disk.

andyu11:
I see, I will try to search my disk and see where it might be located

andyu11:
I couldn't find it, I have no idea why it is still doing it

HebaruSan:
Have you made multiple copies of your KSP folder? Sometimes we get questions from people who are looking at the wrong copy.

andyu11:
I have the steam version and the non steam version so ill check the steam version but I intended for the CKAN to be off the steam version

HebaruSan:
If there is no CKAN folder in the Steam copy, then CKAN does not yet know about that copy.

andyu11:
> If there is no CKAN folder in the Steam copy, then CKAN does not yet know about that copy.

I found it on the steam copy! strange because I meant for the steam one to be pure stock

HebaruSan:
[...]

Issue #3188, opened 2020-11-06
Uncaught exception loading registry: Newtonsoft.Json.JsonReaderException: Unterminated string


Rook-TEK:
Background
----------
* Operating System: Windows 10 64bit
* CKAN Version: v1.29.0
* KSP Version: 1.10.1.2939

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
Don't think so

Problem
-------
**Describe the bug**

when I launch CKAN, I get 2 windows that pop up with error messages. I'll copy paste each block of text below. Any help would be appreciated.

**Screenshots (if applicable)**

**CKAN error code (if applicable):**

```
2884 [1] ERROR CKAN.RegistryManager (null) - Uncaught exception loading registry: Newtonsoft.Json.JsonReaderException: Unterminated string. Expected delimiter: ". Path 'available_modules.USI-FTT.module_version['0.6.1.0'].download', line 873222, position 2265787.
   at Newtonsoft.Json.JsonTextReader.ReadStringIntoBuffer(Char quote)
   at Newtonsoft.Json.JsonTextReader.ParseValue()
   at Newtonsoft.Json.JsonReader.ReadForType(JsonContract contract, Boolean hasConverter)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateObject(Object newObject, JsonReader reader, JsonObjectContract contract, JsonProperty member, String id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateDictionary(IDictionary dictionary, JsonReader reader, JsonDictionaryContract contract, JsonProperty containerProperty, String id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
[...]

PR #3351, opened 2021-04-17
Better recovery when registry.json is corrupted

HebaruSan:
## Problem

If your `CKAN/registry.json` file gets truncated somehow, CKAN won't start.

```
Unhandled exception:
Newtonsoft.Json.JsonSerializationException: Unexpected end when deserializing object. Path 'available_modules.WindowShineTR.module_version['1:2.0'].download_hash.sha256', line 979045, position 0.
```

```
Unhandled exception:
Newtonsoft.Json.JsonReaderException: Unexpected end of content while loading JArray. Path 'available_modules.RP-0.module_version['v1.6'].recommends[21]', line 648536, position 8403726.
   in Newtonsoft.Json.Linq.JContainer.ReadTokenFrom(JsonReader reader, JsonLoadSettings options)
   in Newtonsoft.Json.Linq.JArray.Load(JsonReader reader, JsonLoadSettings settings)
   in Newtonsoft.Json.Linq.JToken.ReadFrom(JsonReader reader, JsonLoadSettings settings)
```

```
Newtonsoft.Json.JsonReaderException: Unterminated string. Expected delimiter: ". Path 'available_modules.WarpShip.module_version['0.3.2'].resources.repository', line 878236, position 44.
```

Examples of this happening in the wild:

#1386, #1540, #1904, #1779, #2123, #2448, #2566, #2588, #2871, #3188, #3229, #3233

(Inspired by #3347.)

## Cause

The deserialization call can throw `JsonSerializationException` or `JsonReaderException` (both inherit from `JsonException`), and we don't handle it.

We currently don't have any clues as to how or why the corruption happens, because the exception can be thrown long after the file is corrupted. By the time the user knows it happened, it's too late to investigate.

## Changes

Now if our attempt to load `registry.json` throws a `JsonException`, `RegistryManager.LoadOrCreate` will catch it and move the file to a new location, `CKAN/registry.json_CORRUPTED_20210417020300`, where that last part is the current date. Then a new registry will be created in its place, to be filled in as usual for a new registry. This way even though the user's data is lost, they will still have a working CKAN, and they may be able to recover some data from the corrupted file.

Ideally we would notify the user that this happened. Unfortunately the UI isn't accessible inside of `RegistryManager` so this is currently not done, but we will log an error in the log system.

### DIFF
--- a/Core/Registry/RegistryManager.cs
+++ b/Core/Registry/RegistryManager.cs
@@ -400,10 +400,8 @@ namespace CKAN
 
             file_transaction.WriteAllText(path, Serialize());
 
-            string sanitizedName = string.Join("", ksp.Name.Split(Path.GetInvalidFileNameChars()));
-
             ExportInstalled(
-                Path.Combine(directoryPath, $"installed-{sanitizedName}.ckan"),
+                Path.Combine(directoryPath, LatestInstalledExportFilename()),
                 false, true
             );
             if (!Directory.Exists(ksp.InstallHistoryDir()))
@@ -411,13 +409,13 @@ namespace CKAN
                 Directory.CreateDirectory(ksp.InstallHistoryDir());
             }
             ExportInstalled(
-                Path.Combine(
-                    ksp.InstallHistoryDir(),
-                    $"installed-{sanitizedName}-{DateTime.Now.ToString("yyyy-MM-dd_HH-mm-ss")}.ckan"
-                ),
+                Path.Combine(ksp.InstallHistoryDir(), HistoricInstalledExportFilename()),
                 false, true
             );
         }
+
+        public string LatestInstalledExportFilename() => $"installed-{ksp.SanitizedName}.ckan";
+        public string HistoricInstalledExportFilename() => $"installed-{ksp.SanitizedName}-{DateTime.Now:yyyy-MM-dd_HH-mm-ss}.ckan";
 
         /// <summary>
         /// Save a custom .ckan file that contains all the currently
--- a/Core/Registry/RegistryManager.cs
+++ b/Core/Registry/RegistryManager.cs
@@ -23,7 +23,6 @@ namespace CKAN
         private FileStream lockfileStream = null;
         private StreamWriter lockfileWriter = null;
 
-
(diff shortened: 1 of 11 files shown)


HebaruSan:
> We currently don't have any clues as to how or why it happens, because the exception can be thrown long after the file is corrupted. By the time the user knows it happened, it's too late to investigate.

It's been a while since it's happened, I wonder if #3250 helped.

Core/Registry/RegistryManager.cs
```diff
@@ -309,6 +309,15 @@ private void LoadOrCreate()
                 Create();
                 Load();
             }
+            catch (JsonException exc)
+            {
+                var corruptPath = path + "_CORRUPTED_" + DateTime.Now.ToString("yyyyMMddHHmmss");
+                log.ErrorFormat("{0} is corrupted, achiving to {1}: {2}",
```
Olympic1, line 315: Do you mean `archiving`?

HebaruSan: Yup!

DasSkelett:
Hm, could we set some sort of (unserialized) flag in the Registry object when it is detected as corrupted and moved, which we later check when the GUI is ready, and display a warning?
Hacky, but better than no note to the user at all, IMHO.

HebaruSan:
> Hm, could we set some sort of (unserialized) flag in the Registry object when it is detected as corrupted and moved, which we later check when the GUI is ready, and display a warning?

Done. I liked `RegistryManager` better as a place for this flag since it is not really related to the active registry object itself, and the `RegistryManager` is the class that discovers the problem in the first place.

DasSkelett:
Pushed some small changes, now we also check if a corruption happened (or better: got detected) during instance switch.
The error message also contains a hint towards the `installed-<instanceName>.ckan` file to help with recovery, for that I moved some string calculations around to make them more accessible.

DasSkelett (approved):
Looks good. Probably a long overdue feature, saves some "complicated" manual steps that were inevitable anyways.

Could we automate the reinstallation of all the previously installed modules next...?

PR #3353, opened 2021-04-24
Update or refresh ConsoleUI mod list after repo or compat changes

### LINKED ISSUE #3352: [Bug] ConsoleUI should refresh mod list after changing compatible versions
Background
----------
* Operating System: macOS
* CKAN Version: v1.30.0
* KSP Version: 1.11.2

Yeah the person who encountered it did, but that's irrelevant.

Problem
-------
**Describe the bug**
I've just spent some time helping a macOS+ConsoleUI user on the r/KerbalSpaceProgram Discord who wasn't able to find ClickThroughBlocker in the mod list, even after adding KSP 1.11 (and a whole bunch more) to the list of compatible versions.
Turns out, the mod list doesn't reload after you change compatible versions, thus changing compatible KSP versions "doesn't have an effect", only after pressing F5 or restarting KSP.

I was able to reproduce this.

**Steps to reproduce**
* Be on KSP 1.11.2
* Open Console UI
* Remove _all_ compatible KSP versions from one instance
* Quit and reopen Console UI
* Search for ClickThroughBlocker in the mod list, it won't show up
* Go to the instance management screen again, and add KSP 1.11 as compatible version
* F2, Enter the instance
* Search for ClickThroughBlocker again, it still won't show up
* Only after you restart the ConsoleUI again, or hit F5, or refresh via the F10 menu it'll show up

**Expected behavior**
It should show up directly when coming back to the mod list of the instance, i.e. the mod list should refresh automatically after changing compatible KSP versions (or active metadata repositories).
That's what a user would expect to happen, and that's what happens in the GUI as well.
At the very least we should show some sort of notice that refreshing the mod list manually is required to take the changes into effect.

HebaruSan:
## Problem

If you edit your current instance's repositories in ConsoleUI, the new module list is not fetched.

If you edit your current instance's compatible versions in ConsoleUI, the mod list is not updated to reflect the new compatibility.

## Cause

We refresh the mod list if you switch instances, but not otherwise.

## Changes

Now if you add or remove a repo, we update the registry upon returning to the mod list.

Now if you change the compatible versions, we refresh the mod list upon returning to it.

Fixes #3352.

### DIFF
--- a/ConsoleUI/ModListScreen.cs
+++ b/ConsoleUI/ModListScreen.cs
@@ -477,14 +477,32 @@ namespace CKAN.ConsoleUI {
         private bool SelectInstall(ConsoleTheme theme)
         {
             GameInstance prevInst = manager.CurrentInstance;
+            var prevRepos = new SortedDictionary<string, Repository>(registry.Repositories);
+            var prevVerCrit = prevInst.VersionCriteria();
             LaunchSubScreen(theme, new GameInstanceListScreen(manager));
-            // Abort if same instance as before
             if (!prevInst.Equals(manager.CurrentInstance)) {
+                // Game instance changed, reset everything
                 plan.Reset();
                 registry = RegistryManager.Instance(manager.CurrentInstance).registry;
                 RefreshList(theme);
+            } else if (!SortedDictionaryEquals(registry.Repositories, prevRepos)) {
+                // Repos changed, need to fetch them
+                UpdateRegistry(theme, false);
+                RefreshList(theme);
+            } else if (!manager.CurrentInstance.VersionCriteria().Equals(prevVerCrit)) {
+                // VersionCriteria changed, need to re-check what is compatible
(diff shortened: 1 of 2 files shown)


DasSkelett:
Not so smart compiler:
image

I don't think we actually need `GetHashCode()`, since we aren't using the repositories as keys of a sorted list/dictionary/hashset anywhere (only as values). But if we want to get rid of the warning, we have to implement it.
But we could keep it a simple one. Since implementations of `GetHashCode()` only have to guarantee that equal objects have the same hash code, but the same hash code doesn't require the objects to be equal, and we don't rely on as few collisions as possible for performance, we could just return `name.ToHashCode()` and be happy.

HebaruSan:
OK, returning `name.GetHashCode()` now.

DasSkelett (approved):
Thanks, works great!

PR #3356, opened 2021-04-28
Make ModuleInstaller a non-singleton

### LINKED ISSUE #333: ModuleInstaller shouldn't be a singleton
This may have made our lives easier in the past, but it now makes writing tests much more difficult, and means it's very tightly coupled to KSPManager.

ModuleInstaller really should be just fine as a regular class; and if it's not, we should be able to find and fix any inconsistencies.

### LINKED ISSUE #3355: [Bug] CKAN can't read zips from custom cache location on Windows
Background
----------
* Operating System: Windows 10  20H2
* CKAN Version: v1.30.0
* KSP Version: v1.10.1

No

Problem
-------
**Describe the bug**
Installations of any mod fails when cache is purged and cache path is changed.

**Steps to reproduce**
* Go to CKAN settings and Purge All cache
* Change the cache folder path to somewhere else
* Try to install any mod
* You would get this error _sometimes_
```
Downloading "https://spacedock.info/mod/1466/FuelEfficientPilots/download/0.2"
Trying to install FuelEfficientPilots 0.2, but it's not downloaded or download is corrupted
Error during installation!
```

**Expected behavior**
Mod would install

HebaruSan:
## Problems

1. Install something
2. Change the cache path
3. Install something else; the download will succeed but the install will fail:
   ```
   Downloading "https://spacedock.info/mod/1466/FuelEfficientPilots/download/0.2"
   Trying to install FuelEfficientPilots 0.2, but it's not downloaded or download is corrupted
   Error during installation!
   ```

## Cause

`ModuleInstaller` is a singleton (per game instance). It stores a reference to the cache object that it is given when created; if the user changes the cache path, that cache object becomes stale, but the installer doesn't know that.

1. Install something; this will create a persistent `ModuleInstaller` instance for your game instance, using the currently configured cache path
2. Change the cache path; this _replaces_ `Manager.Cache` with a new cache object, _but the persistent `ModuleInstaller` instance doesn't know that happened and still has the old object_
3. Install something else; the old `ModuleInstaller` and cache object will be used

## Changes

Now `ModuleInstaller` is a regular class; we create an instance of it when we need one, and when we're done with it, it fades out forever. This way it will always use the current cache object.

Fixes #333.
Fixes #3355.

### DIFF
--- a/Cmdline/Action/Import.cs
+++ b/Cmdline/Action/Import.cs
@@ -47,7 +47,7 @@ namespace CKAN.CmdLine
                     log.InfoFormat("Importing {0} files", toImport.Count);
                     List<string>    toInstall = new List<string>();
                     RegistryManager regMgr = RegistryManager.Instance(ksp);
-                    ModuleInstaller inst      = ModuleInstaller.GetInstance(ksp, manager.Cache, user);
+                    ModuleInstaller inst      = new ModuleInstaller(ksp, manager.Cache, user);
                     inst.ImportFiles(toImport, user, mod => toInstall.Add(mod.identifier), regMgr.registry, !opts.Headless);
                     HashSet<string> possibleConfigOnlyDirs = null;
                     if (toInstall.Count > 0)
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -136,7 +136,7 @@ namespace CKAN.CmdLine
                 try
                 {
                     HashSet<string> possibleConfigOnlyDirs = null;
-                    var installer = ModuleInstaller.GetInstance(ksp, manager.Cache, user);
+                    var installer = new ModuleInstaller(ksp, manager.Cache, user);
                     installer.InstallList(modules, install_ops, regMgr, ref possibleConfigOnlyDirs);
                     user.RaiseMessage("");
                     done = true;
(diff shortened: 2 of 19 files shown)


DasSkelett:
@melon-husk do you want to try out if this fixes your problem as well? Just to make sure that's all that's needed to make it work for you as well.
You can click on the "Checks" tab below the pull request headline, then open the "Artifacts" drop-down on the right, and download the ckan.exe.

melon-husk:
@DasSkelett Trying it out, I'll update you.

melon-husk:
image
@DasSkelett  I cannot run it, I get this error.

HebaruSan:
Hmm, 1,029 KB isn't the right file. Here's a direct link:
https://github.com/KSP-CKAN/CKAN/suites/2599424232/artifacts/56833368

melon-husk:
image
I still got 1,029 KB file, let me try another browser.

HebaruSan:
OK, that's a ZIP file and for whatever reason your PC isn't unzipping it. Are you renaming it without the .zip suffix when you save it to disk?

melon-husk:
It works now ! I can change cache path now.

DasSkelett (approved):
Great! I can confirm that it also fixes my reproduction case, only  single cache dir showing up in the debug output. Code looks good, too!

PR #3362, opened 2021-05-10
Fix installation of metapackages on cmdline

DasSkelett:
## Problem
When trying to install metapackages from cmdline, using `ckan.exe -c metapackage.ckan`, it doesn't actually install any mods:
```
$ mono ckan.exe install --instance "Fake-1.11.2" -c asdf.ckan 
About to install:


Continue? [Y/n] 
Updating registry
Committing filesystem changes
Rescanning GameData
Done!
```
whereas the GUI installs them correctly.

## Cause
In ` InstallList(List<string> modules, ...)`, we are running the `RelationshipResolver` for the identifier list, to get back a list of `CkanModule` to install. Then we filter out all modules installed as dependencies, to make sure the resolver running in `InstallList(ICollection<CkanModule> modules, ...)` can re-detect them as dependencies, so we can mark them as auto-installed.
https://github.com/KSP-CKAN/CKAN/blob/72c4a569ec59ef6998b256ac89e345e83269477a/Core/ModuleInstaller.cs#L108-L113

The problem with metapackages is, that the first RR filters them out of the mod list (we don't want to actually _install_ the metapackage after all). Thus the list of mods passed to `InstallList(ICollection<CkanModule> modules, ...)` is empty.

## Changes
Now we also keep modules whose selection reason is "I'm a dependency|recommendation|suggestion of a metapackage" in the list passed to `InstallList(ICollection<CkanModule> modules, ...)` .
The ` InstallList(List<string> modules, ...)` method is only used in the CmdLine; GUI sends the metapackage as `CkanModule` to `InstallList(ICollection<CkanModule> modules, ...)`, and the ConsoleUI pretty much only deals in `CkanModule`s  (and doesn't have a "Install from .ckan" feature).

`InstallList()` is now also returning early if there are no mods to install, no need to go through the whole transaction and registry saving shebang.

I was considering two other solutions:
b) make `CmdLine.Install.RunCommand())` call the other `InstallList()` passing a list of `CkanModule` instead of identifier+version combos. This might or might not have introduced other issues causing installation of outdated modules (I think we moved to a list of strings some time ago deliberately?)
c) make `RelationshipResolver.Resolve()` mark dependencies of metapackages as `SelectionReason.UserRequested`, not `Depends`. Might have had other side effects as well.

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -108,8 +108,15 @@ namespace CKAN
         public void InstallList(List<string> modules, RelationshipResolverOptions options, RegistryManager registry_manager, ref HashSet<string> possibleConfigOnlyDirs, IDownloader downloader = null)
         {
             var resolver = new RelationshipResolver(modules, null, options, registry_manager.registry, ksp.VersionCriteria());
-            // Only pass the CkanModules of the parameters, so we can tell which are auto
-            InstallList(resolver.ModList().Where(m => resolver.ReasonFor(m) is SelectionReason.UserRequested).ToList(), options, registry_manager, ref possibleConfigOnlyDirs, downloader);
+            // Only pass the CkanModules of the parameters, so we can tell which are auto-installed,
+            // and relationships of metapackages, since metapackages aren't included in the RR modlist.
+            var list = resolver.ModList().Where(
+                m =>
+                {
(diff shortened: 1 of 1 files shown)


DasSkelett:
Good point, changed it to keep all relationships of metapackages.

HebaruSan (approved):
:zap:

PR #3366, opened 2021-05-14
Fix installation of AVP while removing EVE default configs

DasSkelett:
Or: Fix resolution of circular dependencies when conflicts are involved.
Or: Another open heart surgery of the relationship resolver.

## Problem
Have EnvironmentalVisualEnhancements with its default configs EnvironmentalVisualEnhancements-HR installed.
Now deselect EnvironmentalVisualEnhancements-HR and select AVP.

See this double error message:
```
AVP-8kTextures v1.12 depends on AstronomersVisualPack, which is not compatible with the currently installed game version
```

The issue goes down to `MightBeInstallable()`  returning false for any AstronomersVisualPack,
because `LatestAvailableWithProvides()` can't find any provider for `AVP-Textures`,
because they all depend on AstronomersVisualPack again,
which `LatestAvailableWithProvides()` / `AvailableModule.Latest()`  sees as incompatible 
due to the conflict with `EnvironmentalVisualEnhancements-HR`,
which `AvailableModule.Latest()` doesn't know we're uninstalling.

In short: We pass the list of actually installed modules to  `AvailableModule.Latest()` instead of the theoretical one of the `RelationshipResolver`.

Another bug: our invocations of `FindRemovableAutoInstalled()` also try to remove mods that will be re-required by new mods we're about to install, so those mods get scheduled once for removal and once for reinstallation. Not a big problem, but preferably it shouldn't remove them in the first hand.

And: If the install step in `MainInstall.InstallMods()` throws a `TooManyModsProvide` kraken and it gets resolved by the user, it tries to start again from the beginning, re-trying to remove mods that have already been removed in the first round.

## Changes
The whole chain of `LatestAvailableWithProvides()` methods gets a new optional `installed` parameter, next to `toInstall`. If given, it passes that one along to `AvailableModule.Latest()` instead of `Registry.InstalledModules`.
Additionally, `MightBeInstallable()` passes the current source/parent of the stanza we're investigating as `toInstall`  `LatestAvailableWithProvides()`, since it's not yet in the list of installed modules. Basically we assume that the stanza source is compatible when checking its dependencies. I don't think this was needed to fix the above bug, but it should help when force-installing outdated mods.

I've written a unit test that models this relationship constellation. I'm actually quite proud of this one, it works really well. You can cherry-pick / checkout that commit on itself to see the test fail.

To fix `FindRemovableAutoInstalled()` staging removals of dependencies we need for new modules again, we concatenate the list of installed modules with the list of modules we're about to install, wrapped in pseudo-`InstalledModule`s.

And to fix the installation failure when `TooManyModsProvide` is thrown after the removal step, we clear the change lists after finishing them.

### DIFF
--- a/Core/Types/RelationshipDescriptor.cs
+++ b/Core/Types/RelationshipDescriptor.cs
@@ -17,7 +17,10 @@ namespace CKAN
 
         public abstract bool WithinBounds(CkanModule otherModule);
 
-        public abstract List<CkanModule> LatestAvailableWithProvides(IRegistryQuerier registry, GameVersionCriteria crit, IEnumerable<CkanModule> toInstall = null);
+        public abstract List<CkanModule> LatestAvailableWithProvides(
+            IRegistryQuerier registry, GameVersionCriteria crit, IEnumerable<CkanModule> installed = null,
+            IEnumerable<CkanModule> toInstall = null
+        );
 
         public abstract bool Equals(RelationshipDescriptor other);
 
@@ -138,9 +141,12 @@ namespace CKAN
             return false;
         }
 
-        public override List<CkanModule> LatestAvailableWithProvides(IRegistryQuerier registry, GameVersionCriteria crit, IEnumerable<CkanModule> toInstall = null)
+        public override List<CkanModule> LatestAvailableWithProvides(
+            IRegistryQuerier registry, GameVersionCriteria crit, IEnumerable<CkanModule> installed = null,
+            IEnumerable<CkanModule> toInstall = null
+        )
         {
-            return registry.LatestAvailableWithProvides(name, crit, this, toInstall);
+            return registry.LatestAvailableWithProvides(name, crit, this, installed, toInstall);
         }
 
         public override bool Equals(RelationshipDescriptor other)
@@ -225,9 +231,12 @@ namespace CKAN
                 ?? false;
         }
 
-        public override List<CkanModule> LatestAvailableWithProvides(IRegistryQuerier registry, GameVersionCriteria crit, IEnumerable<CkanModule> toInstall = null)
+        public override List<CkanModule> LatestAvailableWithProvides(
+            IRegistryQuerier registry, GameVersionCriteria crit, IEnumerable<CkanModule> installed = null,
+            IEnumerable<CkanModule> toInstall = null
+        )
         {
-            return any_of?.SelectMany(r => r.LatestAvailableWithProvides(registry, crit, toInstall)).ToList();
+            return any_of?.SelectMany(r => r.LatestAvailableWithProvides(registry, crit, installed, toInstall)).ToList();
         }
 
         public override bool Equals(RelationshipDescriptor other)
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -645,7 +645,9 @@ namespace CKAN
                 )).Memoize();
             var goners = revdep.Union(
                     registry_manager.registry.FindRemovableAutoInstalled(
-                        registry_manager.registry.InstalledModules.Where(im => !revdep.Contains(im.identifier)))
+                        registry_manager.registry.InstalledModules
+                            .Where(im => !revdep.Contains(im.identifier))
+                            .Concat(installing?.Select(m => new InstalledModule(null, m, new string[0], false)) ?? new InstalledModule[0]))
                     .Select(im => im.identifier))
                 .ToList();
 
--- a/Core/Registry/IRegistryQuerier.cs
+++ b/Core/Registry/IRegistryQuerier.cs
@@ -61,10 +61,11 @@ namespace CKAN
         /// If no KSP version is provided, the latest module for *any* KSP version is given.
         /// </summary>
         List<CkanModule> LatestAvailableWithProvides(
-            string                  identifier,
-            GameVersionCriteria      ksp_version,
-            RelationshipDescriptor  relationship_descriptor = null,
-            IEnumerable<CkanModule> toInstall               = null
+            string identifier,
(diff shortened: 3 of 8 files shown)


Core/Types/RelationshipDescriptor.cs
```diff
@@ -225,9 +231,12 @@ public override bool WithinBounds(CkanModule otherModule)
...
         }
 
-        public override List<CkanModule> LatestAvailableWithProvides(IRegistryQuerier registry, GameVersionCriteria crit, IEnumerable<CkanModule> toInstall = null)
+        public override List<CkanModule> LatestAvailableWithProvides(
+            IRegistryQuerier registry, GameVersionCriteria crit, IEnumerable<CkanModule> installed = null,
+            IEnumerable<CkanModule> toInstall = null
+        )
         {
-            return any_of?.SelectMany(r => r.LatestAvailableWithProvides(registry, crit, toInstall)).ToList();
+            return any_of?.SelectMany(r => r.LatestAvailableWithProvides(registry, crit, null, toInstall)).ToList();
```
HebaruSan, line 239:
```suggestion
            return any_of?.SelectMany(r => r.LatestAvailableWithProvides(registry, crit, installed, toInstall)).ToList();
```

Looks like this should be `installed` instead of `null`?

HebaruSan:
I'm still seeing a conflict message in the status bar even after I click to uninstall EVE-HR. Should it disappear?

image

Hmm, this even sticks around after completely resetting the change set. I thought we had a way to clear it...

HebaruSan:
1. Install EVE and EVE-HR
2. Check the box for AVP. The apply button remains disabled because of the conflict.
3. Uncheck the box for EVE-HR. The apply button lights up because the conflict is gone.
4. Recheck the box for EVE-HR to go back to what we had in step 2. The conflict should be back, but the apply button remains enabled. Something's not going back to its previous state somehow.

Core/Types/RelationshipDescriptor.cs
DasSkelett: Good catch, a spot I missed in the refactoring.

DasSkelett:
> I'm still seeing a conflict message in the status bar even after I click to uninstall EVE-HR. Should it disappear?

I see, in other cases it does disappear (tested with RealismOverhaul and KerbalJointReinforcementNext). Ugh, no idea where that goes wrong.

>     1. Install EVE and EVE-HR
> 
>     2. Check the box for AVP. The apply button remains disabled because of the conflict.
> 
>     3. Uncheck the box for EVE-HR. The apply button lights up because the conflict is gone.
> 
>     4. Recheck the box for EVE-HR to go back to what we had in step 2. The conflict should be back, but the apply button remains enabled. Something's not going back to its previous state somehow.

Interesting, maybe connected to the above problem? EVE-HR doesn't get added to the changeset, though. So the relationship resolver is doing _something_ right, but somewhere along the chain the error seems to be lost.

DasSkelett:
Okay, turns out that by sending the more accurate list of installed mods the resolver is now "too good", which causes it to not find any candidates at all, which in turn causes a `DependencyNotSatisfiedKraken` kraken to be thrown:
https://github.com/KSP-CKAN/CKAN/blob/f053ee4e84c14e90cb3e4c4d79f985a960857d61/Core/Relationships/RelationshipResolver.cs#L456

... instead of going on with the resolution and adding the mod to a list of conflicts if `proceed_with_inconsistencies` is set:
https://github.com/KSP-CKAN/CKAN/blob/f053ee4e84c14e90cb3e4c4d79f985a960857d61/Core/Relationships/RelationshipResolver.cs#L500-L524

And in `UpdateChangeSetAndConflicts()` we're relying on

DasSkelett:
Changes of the last commit:
Remove the `stanzaSource` parameter of `ResolveStanza()` again, use `reason.Parent` instead.

`MightBeInstallable()` now takes a list of installed modules as argument as well, instead of reading it from `this.installed_modules`.

In the first search for candidates in `ResolveStanza()`, pass `installed_modules` as well, missed that before. Also pass `installed_modules` to `MightBeInstallable()`.

In the second search for candidates (if the previous didn't find any), pass an empty list for `installed` to `LatestAvailableWithProvides()` and `MightBeInstallable()`. Also send a `null` `stanzaSource`. This ensures that we find any potential satisfiers without considering conflicts to already installed mods, in case we want to `proceed_with_inconsistencies` or `soft_resolve`, and add it to the list of conflicts instead of throwing an exception (which some callers expect).

This also solved the conflict message not disappearing after deselecting the conflicting mod(s).

And `AnyOfRelationshipDescriptor.LatestAvailableWithProvides()` also passes along the `installed` argument now.

HebaruSan:
This search is really handy for testing btw:

image

HebaruSan:
The stuff that I was testing last time now works seemingly flawlessly, thanks for patching it up. :+1:

HebaruSan:
LGTM. Can we add a test or two so we have better guard rails the next time we want to change the resolver?

DasSkelett:
I can try to expand `UninstallingConflictingModule_InstallingRecursiveDependencies_ResolvesSuccessfully` to also cover the toggle-on toggle-off unexpected exception issue.

DasSkelett:
Okay, squashed the fixup commit, and expanded the new unit test to also cover what happened behind the scenes when doing the toggle switcharoo. It's basically scenario 2, which shouldn't have thrown any exceptions because we had `proceed_with_inconsistencies = true`, but it did. The test makes sure there is no exception thrown, and that the conflict list contains the two mods.

HebaruSan (approved):
Really nice, will be great to have this. :ram: 

Is it time to start preparing to release v1.30.2?

PR #3376, opened 2021-05-22
Create APT repo documentation

HebaruSan:
## Problem

Our release notes template needs to be revised, and one of the things that should go somewhere else is the APT repo instructions. However, the instructions involve several `sudo` commands, including one that adds CKAN's key to the system's trusted APT keys, so if we put this on a wiki page, there is a small risk that a random GitHub user would make an edit that would change the instructions to install a malicious key. We would revert that as soon as we noticed it, but there would be a risk to user systems as long as the edit was live.

## Considered and not done

We could lock down the wiki:

image

... but then it's not really a wiki anymore, and we would lose the benefits of user interaction and general editability.

As far as we have been able to determine, GitHub wikis cannot lock down individual pages, nor can they hold changes in a queue for review before they go live to the world.

## Changes

Now a new `doc/apt-repo.md` document holds the instructions behind the impenetrable barrier of push access and pull request code review. For the first time, the stable and nightly builds are documented in the same place, with hopefully enough text to help users determine which is right for them.

This will be linked from relevant wikis and possibly release notes (the plans there are not finalized, but this is part of it).

### DIFF
--- a/doc/apt-repo.md
+++ b/doc/apt-repo.md
@@ -1,0 +1,27 @@
+# CKAN's APT repository
+
+We have created an APT repository that you can add to your Debian-based OS to install CKAN. This will allow you to run CKAN from the system app menus or via `ckan` from your command line. Your system's package manager will pull in dependencies and update CKAN automatically. There's even a man page.
+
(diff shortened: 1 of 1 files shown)


DasSkelett (approved):
Yup, I think that's a good compromise with the options we have.

PR #3380, opened 2021-05-23
Invoke GUI update calls in SetupDefaultSearch() descendants

### LINKED ISSUE #3379: [Bug] CKAN "Not Responding" during/after loading.
Background
----------
* CKAN Version: 1.30.2
* KSP Version: 1.10.1.2939

I have

Problem
-------
When opening CKAN it loads, updates repositories, loads local goes to 'updating tray...' and stops.

**Steps to reproduce**
Open CKAN, 
Select my normal (modded) instance
* See error

**Expected behavior**
To see the mod list and

image

N/A

```
PLACE CODE HERE !
```

DasSkelett:
## Problem
With the restructuring to the search done in #3323 we added a call to `SetupDefaultSearch()`, which indirectly modifies GUI controls through other functions it calls.
As we all know, updating GUI components from another thread basically doesn't work.

In this case it manifests itself through freezes at the end of the repository update for some users.

## Changes
Now `ManageMods.Filter()` and `ManageMods.SetSearches()`, both called in `SetupDefaultSearch()`, wrap every GUI-related work and call in a `Util.Invoke()`.

The diff looks worse than the actual change :smile: 

I wonder if there's some static code analysis tool that can catch this? At least my IDE doesn't :/

I can't confirm the effectiveness of this patch, as my low core count tablet/laptop/convertible doesn't reproduce the hang/freeze (the runtime decides to run it all in a single thread there, I guess) and this only seems to affect Windows. I might dust off my Windows installation on my PC  tomorrow to see if it experiences the bug.

Fixes #3379 

@linuxgurugamer and @daantimmer, do you want to give this fix a try, after re-enabling the repo update on startup so it hits the critical code path?
If you click on the "Checks" tab in between the pull request header and body, there should be an "Artifacts" button on the right side, where you can download a ckan.exe built from this PR code.

### DIFF
--- a/GUI/Controls/ManageMods.cs
+++ b/GUI/Controls/ManageMods.cs
@@ -361,38 +361,45 @@ namespace CKAN
             var searches = search.Values.Select(s => ModSearch.Parse(s,
                 Main.Instance.ManageMods.mainModList.ModuleLabels.LabelsFor(Main.Instance.CurrentInstance.Name).ToList()
             )).ToList();
-            EditModSearches.SetSearches(searches);
 
-            // Ask the configuration which columns to show.
-            foreach (DataGridViewColumn col in ModGrid.Columns)
+            Util.Invoke(ModGrid, () =>
             {
-                // Some columns are always shown, and others are handled by UpdateModsList()
-                if (col.Name != "Installed" && col.Name != "UpdateCol" && col.Name != "ReplaceCol")
-                {
-                    col.Visible = !Main.Instance.configuration.HiddenColumnNames.Contains(col.Name);
-                }
-            }
+                EditModSearches.SetSearches(searches);
 
-            // If these columns aren't hidden by the user, show them if the search includes installed modules
-            setInstalledColumnsVisible(!SearchesExcludeInstalled(searches));
+                // Ask the configuration which columns to show.
+                foreach (DataGridViewColumn col in ModGrid.Columns)
+                {
+                    // Some columns are always shown, and others are handled by UpdateModsList()
+                    if (col.Name != "Installed" && col.Name != "UpdateCol" && col.Name != "ReplaceCol")
+                    {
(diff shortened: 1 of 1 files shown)


HebaruSan:
> I wonder if there's some static code analysis tool that can catch this? At least my IDE doesn't :/

Yeah, that would be a dream come true. That's why I mentioned `async`/`await` in chat, but I don't think that quite maps properly to the GUI threading problem. Thanks for working on this, I'll review it shortly.

HebaruSan (approved):
This looks like what's needed. Would be wise to have confirmation from somebody who has experienced the issue, so I'm happy to wait a day or three for that.

daantimmer:
I will have a look shortly and get back to you

daantimmer:
> > I wonder if there's some static code analysis tool that can catch this? At least my IDE doesn't :/
> 
> That's why I mentioned `async`/`await` in chat

That won't help. You must update a GUI element in the GUI thread. Async/await does not do this for you. To update a GUI from a different worker thread you have to use Invoke. I always do: GuiElement.Invoke(() => expression);

daantimmer:
@DasSkelett I can confirm this fixes my issue

DasSkelett:
A r/KSP Discord user just reported that they couldn't see the "clear search" cross:



The debug build from this PR fixed it for them as well, so it was also caused by the background thread GUI modifications.
Funny how different the effects can be.

HebaruSan:
> You must update a GUI element in the GUI thread.

Right, the problem is that the cognitive load for doing this by hand across an entire application is unsustainable/impracticable. If we tagged all of our code that updates a GUI property directly as `async`, then at least we could potentially get a compile error if we try to call it from a background thread that isn't `async`. And we could easily check whether all our instances of `DoWhatever().Result` were using `Invoke`.

PR #3382, opened 2021-05-25
Choose game instance if default is locked

### LINKED ISSUE #3375: [Feature] When the default instance is locked: open Manage Game Instances instead of quitting
The problem:
When there is a game instance marked as the default CKAN will refuse to start at all if it is locked. This means I can't open CKAN to work on another instance.

Solution: in this situation it should just open the "Manage Game Instances" dialog like when no default is selected.

Currently I have to switch the currently running instance to the non-default one and then start a new instance, or unset the default temporarily.

The reason I may want to have multiple instances is to (for example):
- compare between instances.
- leave an instance with pending changes running (because KSP is running) when working on another.

HebaruSan:
## Problem

If you have multiple game instances, you've selected one as the default instance, and you open that instance in CKAN, then opening another copy of CKAN will cause it to print an error about the instance being locked and quit.

However, if you have no default instance, then in the same flow you will be prompted to choose an instance. The user shouldn't lose options because they've selected a default instance.

## Changes

Now if the default instance is locked by another process but there are other unlocked instances, the Manage Game Instances popup appears. Instances that have a `CKAN/registry.locked` file will have "(LOCKED)" appended to their names, so you can tell which ones are open in other instances of CKAN:

image

Also the Linux man page's documentation for the `ckan ksp` subcommand is updated to the current `ckan instance` naming.

Fixes #3375.

@CodeLeopard, there should be [a test build under the Checks tab](https://github.com/KSP-CKAN/CKAN/suites/2828224736/artifacts/63025613) shortly after this is submitted, if you wouldn't mind giving it a try.

### DIFF
--- a/Core/GameInstance.cs
+++ b/Core/GameInstance.cs
@@ -72,14 +72,20 @@ namespace CKAN
             }
         }
 
-        public bool Valid
-        {
-            get
-            {
-                return game.GameInFolder(new DirectoryInfo(gameDir))
-                    && Version() != null;
-            }
-        }
+        /// <returns>
+        /// true if the game seems to be here and a version is found,
+        /// false otherwise
+        /// </returns>
+        public bool Valid => game.GameInFolder(new DirectoryInfo(gameDir)) && Version() != null;
+
+        /// <returns>
+        /// true if the instance may be locked, false otherwise.
+        /// Note that this is a tentative value; if it's true,
+        /// we still need to try to acquire the lock to confirm it isn't stale.
+        /// NOTE: Will throw NotKSPDirKraken if the instance isn't valid!
+        ///       Either be prepared to catch that exception, or check Valid first to avoid it.
+        /// </returns>
+        public bool IsMaybeLocked => RegistryManager.IsInstanceMaybeLocked(CkanDir());
 
         /// <summary>
         /// Create the CKAN directory and any supporting files.
--- a/Core/Registry/RegistryManager.cs
+++ b/Core/Registry/RegistryManager.cs
@@ -40,6 +40,16 @@ namespace CKAN
         /// </summary>
         public string previousCorruptedPath;
 
+        private static string InstanceRegistryLockPath(string path)
+        {
+            return Path.Combine(path, "registry.locked");
+        }
+
+        public static bool IsInstanceMaybeLocked(string path)
+        {
+            return File.Exists(InstanceRegistryLockPath(path));
+        }
+
         // We require our constructor to be private so we can
         // enforce this being an instance (via Instance() above)
         private RegistryManager(string path, GameInstance ksp)
@@ -47,7 +57,7 @@ namespace CKAN
             this.ksp = ksp;
 
             this.path    = Path.Combine(path, "registry.json");
-            lockfilePath = Path.Combine(path, "registry.locked");
+            lockfilePath = InstanceRegistryLockPath(path);
 
             // Create a lock for this registry, so we cannot touch it again.
             if (!GetLock())
@@ -146,7 +156,7 @@ namespace CKAN
         private void CheckStaleLock()
         {
             log.DebugFormat("Checking for stale lock file at {0}", lockfilePath);
-            if (File.Exists(lockfilePath))
+            if (IsInstanceMaybeLocked(path))
             {
                 log.DebugFormat("Lock file found at {0}", lockfilePath);
                 string contents;
--- a/GUI/Dialogs/ErrorDialog.cs
+++ b/GUI/Dialogs/ErrorDialog.cs
@@ -34,6 +34,9 @@ namespace CKAN
                 );
                 if (!Visible)
                 {
+                    StartPosition = Main.Instance.actuallyVisible
+                        ? FormStartPosition.CenterParent
+                        : FormStartPosition.CenterScreen;
                     ShowDialog(Main.Instance);
                 }
             });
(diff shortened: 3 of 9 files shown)


CodeLeopard:
I tested it and it works for the most part, but I did ran into a problem:

If I select a locked instance it just re-opens the selection dialog, in an ideal situation it would show the "Instance is locked" dialog at this point (but then still show the selection dialog again).

If I then select the locked instance *again* (after the selector re-appears) the full GUI starts up and immediately throws an exception: RegisteryInUseKraken. If I select continue the GUI has no window title and it does not populate the mods list.

HebaruSan:
OK, we need to de-spaghettify the instance selection/loading flow some more. I tried to start that gently by factoring out some of it into `InstancePromptAtStart`, but clearly it was not enough. I guess this should be a `while` or `for` loop that only terminates if the user picks a valid instance or gives up...

HebaruSan:
Thanks for testing, @CodeLeopard. The latest commit replaces the two sequential popup attempts with one big `do { ... } while (...);` loop that tries to handle all cases and guarantee an instance is picked before continuing. I had some difficulty getting the error popup to show up on the right screen before the main window is visible, so if it seems to freeze, look around for a stray window centered on one of your other monitors.

https://github.com/KSP-CKAN/CKAN/suites/2841888369/artifacts/63403475

Core/Registry/RegistryManager.cs
```diff
@@ -40,14 +40,24 @@ public class RegistryManager : IDisposable
         /// </summary>
         public string previousCorruptedPath;
 
+        private static string InstanceRegistryPath(string path)
```
DasSkelett, line 43: Shouldn't that be `InstanceRegistryLockFilePath` or so? Under `InstanceRegistryPath` I understand `CKAN/registry.json`.

Core/GameInstance.cs
```diff
@@ -72,14 +72,18 @@ public GameInstance(IGame game, string gameDir, string name, IUser user, bool sc
...
+        /// false otherwise
+        /// </returns>
+        public bool Valid => game.GameInFolder(new DirectoryInfo(gameDir)) && Version() != null;
+
+        /// <returns>
+        /// true if the instance may be locked, false otherwise.
+        /// Note that this is a tentative value; if it's true,
+        /// we still need to try to acquire the lock to confirm it isn't stale.
+        /// </returns>
+        public bool IsMaybeLocked => RegistryManager.IsInstanceMaybeLocked(CkanDir());
```
DasSkelett, line 86: If a KSP instance is invalid (or the NFS share it's on is not mounted, as in my case), `CkanDir()` throws a `NotKSPDirKraken`, which we should catch.
It's debatable whether it should return false or true in this case, but I'm leaning towards false, since I think I wouldn't expect the ManageGameInstancesDialog showing `(LOCKED)` here.
But maybe it should show `(INVALID)`? We could add another check to `UpdateInstancesList()` with a new string:
```csharp
 _manager.CurrentInstance != instance.Value && instance.Value.IsMaybeLocked
    ? string.Format(Properties.Resources.ManageGameInstancesLocked, instance.Key)
    : instance.Value.Valid
        ? instance.Key
        : string.Format(Properties.Resources.ManageGameInstancesInvalid, instance.Key),
```

HebaruSan: Now checking `Valid` first and labelling those instances `(INVALID)` if needed.

HebaruSan:
> However I did discover another issue that occurs when switching between instances (not opening a new one):
> In the full GUI, go to: file -> manage instances and select a locked instance, it throws RegisteryInUseKraken.

OK, now the GUI catches that exception, displays an error for it, and goes back to working with the original instance. There's a little delay because technically just the attempt to open the other instance invalidates certain things and we need to make sure everything is set up properly again.

https://github.com/KSP-CKAN/CKAN/suites/2848264095/artifacts/63574915

Core/GameInstance.cs
DasSkelett: I might still be a good idea to catch this exception, in case we ever use `IsMaybeLocked()` somewhere else, so we don't get surprised by it again. What do you think?

HebaruSan: I see your point, exceptions are annoying. But I'm not sure I like the idea of `IsMaybeLocked` returning true _or_ false for an invalid instance, since the lock won't even have been checked. It doesn't really lead to a safer coding pattern...

```csharp
    if (instance.IsMaybeLocked)
    {
        // Unsafe to use instance here if we return true for invalid
    }
    else
    {
        // Unsafe to use instance here if we return false for invalid
    }
```

Maybe longer-term we could replace `Valid` and `IsMaybeLocked` with a `Status` property returning what's wrong. Then the calling code would know the different possibilities and would be responsible for handling them.

```csharp
    enum InstanceStatus
    {
        Invalid,
        MaybeLocked,
        OK,
    };
```

DasSkelett: He, I also thought about a status enum. Something we can keep in mind.

What about a code comment at least, that instructs to check `Valid` first? If we make that the intended calling order, it would make sense to communicate that to the caller.

HebaruSan: Works for me. :+1:

DasSkelett:
Just pushed the German translation, and also renamed `ManageGameInstancesInvalid`/`ManageGameInstancesLocked` -> `ManageGameInstancesNameColumnInvalid`/`ManageGameInstancesNameColumnLocked` to differentiate them a bit from `ManageGameInstancesNotValid`.
The new name is admittedly not the best, but I couldn't think of a better one.

DasSkelett (approved):
Yup, should be good now. Nice enhancement, thanks for the suggestion @CodeLeopard!

PR #3383, opened 2021-05-25
ru-RU translation

nt0g:
Russian (ru-RU) translation.

I'd like this to be reviewed since there may be string length related problems. Some of the strings can be shortened, if needed.

### DIFF
--- a/Core/Utilities.cs
+++ b/Core/Utilities.cs
@@ -11,7 +11,7 @@ namespace CKAN
     {
         public static readonly string[] AvailableLanguages =
         {
-            "en-GB", "en-US", "de-DE", "zh-CN", "fr-FR", "pt-BR"
+            "en-GB", "en-US", "de-DE", "zh-CN", "fr-FR", "pt-BR", "ru-RU"
         };
 
         /// <summary>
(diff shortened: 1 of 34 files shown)


HebaruSan:
Hi @nt0g, thanks for the contribution!

> I'd like this to be reviewed since there may be string length related problems. Some of the strings can be shortened, if needed.

Your changes have been built into this test compile, so you can download it and try it out:

- https://github.com/KSP-CKAN/CKAN/suites/2831331986/artifacts/63111480

Just open the settings and select ru-RU in the language dropdown near the bottom.

nt0g:
Some layout errors persist and can't be fixed by changing the strings.

Capture d’écran 2021-05-26 141454

Capture d’écran 2021-05-26 141543

HebaruSan:
Thanks @nt0g, I took a look and made those buttons wider, plus some other usability tweaks while we're moving things in the UI. Newest build is here:

https://github.com/KSP-CKAN/CKAN/suites/2837535996/artifacts/63283466

HebaruSan (approved):
Thanks again for the new translation, @nt0g!
If there are any issuses with it, they'll be easy enough to fix. Merging.

PR #3384, opened 2021-05-26
Set bugtracker resource for SpaceDock mods with GitHub repos

HebaruSan:
## Motivation

The `resources.bugtracker` property links to a place where users can report problems with a mod. It is available in the GUI and the status page. It can be set manually in a netkan or automatically for mods hosted on GitHub (see #3061).

It would be nice to set this for mods on SpaceDock as well so we can jump to issues lists conveniently for more mods.

## Changes

Now if a mod on SpaceDock has a source code link of the form `https://github.com/owner/repo/whatever`, we will call out to the GitHub API to check whether that repo has issues enabled. If so, then `resources.bugtracker` will be set to the repo's issues link.

### DIFF
--- a/Netkan/Transformers/NetkanTransformer.cs
+++ b/Netkan/Transformers/NetkanTransformer.cs
@@ -35,7 +35,7 @@ namespace CKAN.NetKAN.Transformers
             {
                 new StagingTransformer(),
                 new MetaNetkanTransformer(http, ghApi),
-                new SpacedockTransformer(new SpacedockApi(http)),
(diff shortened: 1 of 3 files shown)


DasSkelett (approved):
Good idea. Let's see how big the wave will be.

HebaruSan:
> Let's see how big the wave will be.

I think the final count is somewhere around 350, plus or minus a few due to some GitHub rate limiting shenanigans. The next webhooks-only pass should be smaller but still quite a few, I would think.

PR #3385, opened 2021-05-26
Make current instance settings easier to find in ConsoleUI

HebaruSan:
## Motivation

We just saw this in Discord:

image

The referenced pinned message is:

image

It's there because this is the top FAQ for ConsoleUI by a large margin. People want/need to edit the compatible versions, and it's very non-obvious how to do that.

## Changes

Now a new "Game instance settings..." option is added as a shortcut to the screen to which the longer directions lead:

image

image

Just like in #3353, if you add or remove a repo, we will force a registry refresh after you save, and if you add or remove compatible versions, we will update the mod list.

### DIFF
--- a/ConsoleUI/ModListScreen.cs
+++ b/ConsoleUI/ModListScreen.cs
@@ -280,6 +280,9 @@ namespace CKAN.ConsoleUI {
                     "Save your mod list",
                     true, ExportInstalled),
                 null,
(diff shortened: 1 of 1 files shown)


DasSkelett (approved):
Nice, saves a few key strokes. Also really like those images :sunglasses:

PR #3394, opened 2021-06-13
Japanese Localization

utah239:
It's substantially my first contribution on github.
I have made some commit about Japanese Localization.
Plz check if it is correct, especially code matter.

### DIFF
--- a/Core/Utilities.cs
+++ b/Core/Utilities.cs
@@ -11,7 +11,7 @@ namespace CKAN
     {
         public static readonly string[] AvailableLanguages =
         {
-            "en-GB", "en-US", "de-DE", "zh-CN", "fr-FR", "pt-BR", "ru-RU"
+            "en-GB", "en-US", "de-DE", "zh-CN", "fr-FR", "pt-BR", "ru-RU", "ja-JP"
         };
 
         /// <summary>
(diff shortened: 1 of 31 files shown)


HebaruSan:
Yes, the code looks great to me! It compiles and I am able to switch into the ja-JP translation.
Have you had a chance to try it out yet? There is a test build here with your changes:
- https://github.com/KSP-CKAN/CKAN/suites/2981121803/artifacts/67269970
If you notice anything that does not look right, we can try to fix it, otherwise this looks ready to merge!

HebaruSan (approved):

HebaruSan:
(Note to others testing in Mono, I got missing character glyphs unless I set the locale in the environment as well:

```bash
LANG=ja_JP.utf-8 _build/ckan.exe
```

I don't think this can be helped, Mono needs to know which fonts to load somehow.)

utah239:
> Yes, the code looks great to me! It compiles and I am able to switch into the ja-JP translation.
> Have you had a chance to try it out yet? There is a test build here with your changes:
> 
> * https://github.com/KSP-CKAN/CKAN/suites/2981121803/artifacts/67269970
> 
> If you notice anything that does not look right, we can try to fix it, otherwise this looks ready to merge!

Yes, I have already tried making this build and checked if the layout is correct so far.

PR #3400, opened 2021-06-23
Clarify that downloading to cache does not install

HebaruSan:
## Problem

Mac users somewhat regularly ask why their mods are not installed after using this option:

image

## Cause

That hotkey is the equivalent of "Download Contents" in GUI; it's not supposed to install the mod, it just downloads the ZIP file into the cache, from which it can be installed later.

## Changes

Now the screen tip explains the option more verbosely, which hopefully will help to set more accurate expectations:

image

GUI's right click menu is updated the same way for the following locales:

- English
  image
- French
- German

Changes for other locales may follow...


DasSkelett (approved):
Thanks, should help directing people to the right buttons.

HebaruSan:
Also to benefit: :grin: 

image

PR #3414, opened 2021-07-14
Add versions table option to ckan show command

HebaruSan:
## Motivation

After KSP-CKAN/CKAN-meta#2389 updated the compatibility of 17 old modules and its parent PR used this illustration:

versions tab

... I said this and got a :+1: :

image

## Problems

There are also lots and lots of things I don't like about the current `ckan show` command.

- Only allows one mod to be searched at a time
- The format is almost but not quite YAML-like
- Extra blank line after description if it exists
- Always shows "Status" field even if it isn't set, even though it's optional, non-functional, and usually not set
- Always shows "Provides" section and includes the mod's own identifier in the list, introducing visual clutter
- Missing tags and localizations data
- Missing manual and version file links
- Resource URLs not aligned the same as the "module info" section above

## Changes

Now you can use `ckan show --with-versions <identifier>` to see the versions and compatible game versions, and all of the above annoyances are addressed. This will be particularly appreciated by all of our users in 1975 who use their PDP-11 computers by typing commands and inspecting the output from their dot matrix printers, or even those with brand new state of the art CRT displays:

image

## Next steps

After this is merged, we can add `--with-versions` here for NetKAN validation:

https://github.com/KSP-CKAN/xKAN-meta_testing/blob/00c0a0b4c3eb79c7dc740bb1daae9cd3f2f23074/ckan_meta_tester/ckan_install_template.txt#L4

This will produce duplicated output for CKAN-meta when a PR modified multiple historical versions of the same mod, but we can worry about that problem later if it bothers us.

### DIFF
--- a/Cmdline/Action/Search.cs
+++ b/Cmdline/Action/Search.cs
@@ -122,6 +122,8 @@ namespace CKAN.CmdLine
         /// <returns>List of matching modules.</returns>
         /// <param name="ksp">The KSP instance to perform the search for.</param>
         /// <param name="term">The search term. Case insensitive.</param>
+        /// <param name="author">Name of author to find</param>
+        /// <param name="searchIncompatible">True to look for incompatible modules, false (default) to look for compatible</param>
         public List<CkanModule> PerformSearch(CKAN.GameInstance ksp, string term, string author = null, bool searchIncompatible = false)
         {
             // Remove spaces and special characters from the search term.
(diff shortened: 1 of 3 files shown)


DasSkelett (approved):
Looks good.
I think we can do the change to xKAN-meta_testing first, since it won't get deployed until we rebuild from this repo, and that way we wouldn't have to deploy twice.

HebaruSan:
https://github.com/KSP-CKAN/CKAN-meta/runs/3192246022

```
  Version    Game Versions      
  ---------  -------------------
  1:1.6.5.0  KSP 1.8.1 and later
  1:1.6.4.2  KSP 1.8.1 and later
  1:1.6.4.1  KSP 1.8.1 and later
  1:1.6.4.0  KSP 1.8.1–1.9.1    
  1:1.6.3.1  KSP 1.3.1–1.4.5    
  1:1.6.3.0  KSP 1.3.1–1.4.4    
  1:1.6.2.2  KSP 1.3.1          
  1:1.6.2.1  KSP 1.3.1          
  1:1.6.1.2  KSP 1.3.1          
  1:1.6.1.1  KSP 1.3.1          
  1.6.1.0    KSP 1.3.1          
  1.6.0.1    KSP 1.3.1          
  1.6.0      KSP 1.3.1          
  1.5.99     KSP 1.3.1          
  1.5.88     KSP 1.3.1          
  1.5.3      KSP 1.3.1          
  1.5.2      KSP 1.3.0          
  1.5.1      KSP 1.3.0          
```

PR #3416, opened 2021-07-19
Label ordering buttons

### LINKED ISSUE #3395: [Feature] Allow ordering of mod labels
Using labels to classify the mods I need/want in my games, I found that there's a _de facto_ label ordering which apply when several one are selected for one mod. I.e. a label is always created at the end of the label list and when a mod is multi-labeled, the displayed color is the one of the first selected in label list.

Could we have a easier way to reorder this label list than manually editing the related json configuration file?

Nota: I'm only using labels on a global scope, but some care should be taken to have consistent behaviors with instances related ones.

HebaruSan:
## Motivation

The labels from #2936 have an implicit ordering that matters if a mod is included in multiple labels, but the user can't control this ordering except by editing a JSON file. Worse, this ordering is hidden in the Edit Labels window, which sorts each group of labels alphabetically by name.

## Changes

Now the Edit Labels window shows the labels in their actual ordering rather than alphabetical.

Now when you edit a label, up and down buttons appear to the right of the label tree:

image

Clicking the up button moves the label earlier in the list, and clicking the down button moves it later.

Fixes #3395.

### Side fixes

- The Japanese translation had a lot of duplicate references to resources, which are now removed
- #2671 added two image files that have never been used, `GUI/Resources/arrow_forward.png` and `GUI/Resources/arrow_upward.png`, which are now removed


HebaruSan:
@d-faure, there's a test build here if you'd like to try it out:

https://github.com/KSP-CKAN/CKAN/suites/3269592181/artifacts/76128410

d-faure:
Right-clicking to edit labels is still leading to the same dialog with the selection invite., but selecting a label doesn't lead to anything and no editing is allowed. The Create button is working and allows appending a new label to the list.

HebaruSan:
> Right-clicking to edit labels is still leading to the same dialog with the selection invite., but selecting a label doesn't lead to anything and no editing is allowed. The Create button is working and allows appending a new label to the list.

Thanks for testing and reporting that, I have a guess as to what weird thing Windows is doing to cause that (not setting `IsExpanded == True` for nodes without children when we call `ExpandAll`), so I'll make some more changes in a bit that should address it...

GUI/Dialogs/EditLabelsDialog.cs
```diff
@@ -70,14 +79,18 @@ private void LabelSelectionTree_BeforeSelect(Object sender, TreeViewCancelEventA
             {
                 e.Cancel = true;
             }
-            else if (!TryCloseEdit())
+            // This is a proxy for being in LoadTree (it calls ExpandAll)
+            else if (e.Node.IsExpanded)
```
HebaruSan, line 83: Apparently doesn't work on Windows based on testing by @d-faure

HebaruSan: Replaced with a strategy of removing the `BeforeSelect` event handler and re-adding it.

HebaruSan:
Thanks again @d-faure, the new test build is:

- https://github.com/KSP-CKAN/CKAN/suites/3388526713/artifacts/79920927

GUI/Properties/Resources.ja-JP.resx
```diff
@@ -119,29 +119,6 @@
...
-  <data name="ksp" type="System.Resources.ResXFileRef, System.Windows.Forms"><value>..\Resources\ksp.png;System.Drawing.Bitmap, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value></data>
-  <data name="search" type="System.Resources.ResXFileRef, System.Windows.Forms"><value>..\Resources\search.png;System.Drawing.Bitmap, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value></data>
-  <data name="settings" type="System.Resources.ResXFileRef, System.Windows.Forms"><value>..\Resources\settings.png;System.Drawing.Bitmap, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value></data>
-  <data name="refresh" type="System.Resources.ResXFileRef, System.Windows.Forms"><value>..\Resources\refresh.png;System.Drawing.Bitmap, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value></data>
-  <data name="CKANFileFilter" xml:space="preserve"><value>CKAN metadata (*.ckan)|*.ckan</value></data>
-  <data name="ExportInstalledModsDialogTitle" xml:space="preserve"><value>Modリストのエクスポート</value></data>
```
DasSkelett, line 129: Is this line removed intentionally? The property name does seem to be used in code.

HebaruSan: Oh, oops, no. I only meant to remove the `ResXFileRef` entries, will take a closer look.

DasSkelett:
Should we move instance groups above the "Global" group, so the more specifics overwrite less specifics? I think that's what you'd typically expect in similar situations.

HebaruSan:
Makes sense to me. Now the list model object sorts them that way on load (so the change will apply regardless of whether the user opens the editing form), and the editing UI maintains the ordering.

GUI/Dialogs/EditLabelsDialog.cs
```diff
@@ -275,6 +276,8 @@ private bool TrySave(out string errMsg)
                 {
                     labels.Labels = labels.Labels
                         .Concat(new ModuleLabel[] { currentlyEditing })
+                        .OrderBy(l => l.InstanceName == null)
+                        .ThenBy(l => l.InstanceName)
                         .ToArray();
```
DasSkelett, line 281: This doesn't work for newly created instance-specific labels (or those moved from global to instance or vice versa), they are still sorted below global labels because when this sorting logic is run, the label doesn't have `InstanceName` assigned yet. That happens 4 lines later.

If I'm not mistaken, we should be fine moving the whole ` if(!labels.Labels.Contains(currentlyEditing))` block right above the `EditDetailsPanel.Visible = false;` line, so that the new label is completely configured before we sort it into the array.

HebaruSan: Yes, I think you're right. Change made.

DasSkelett (approved):
Works very well now, nice feature!

PR #3419, opened 2021-07-22
Don't refresh search every keypress on Windows

### LINKED ISSUE #3402: [Feature] Delay search while typing
Problem
-------
- Open CKan,
- Clear the search bar,
- Type 'is:'

Suggestions
-----------

### LINKED ISSUE #3620: [Feature] Add a delay before searching
Problem
-------
When starting to type in the search box, it seems to immediately start searching, which causes a bit of a delay (on an i9-9900k) before further typing can be done.

Suggestions
-----------
I had the same problem with the QuickSearch mod, and was able to solve it by adding a 1/2 second delay (configurable in the settings) before actually starting to search.  This allows the user to get a few more characters typed which narrows the search.

kfsone:
Rebuilding the mod list dataview every time a key was pressed takes a non-trivial amount of time that makes the ui feel unresponsive and sluggish.

There already existed a workaround for this for linux and macos, but it was still horribly slow on windows.

This change simply normalizes this behavior across the OSes so that there is always a small delay from the last, non-enter, keypress to the start of the refresh.

Fixes #3402.
Fixes #3620.

### DIFF
--- a/GUI/Controls/EditModSearch.cs
+++ b/GUI/Controls/EditModSearch.cs
@@ -267,7 +267,6 @@ namespace CKAN
             filterTimer?.Stop();
             if (!string.IsNullOrEmpty(FilterCombinedTextBox.Text))
             {
-                // Delay updating to improve typing performance on OS X and Linux
                 RunFilterUpdateTimer();
             }
             else
@@ -277,9 +276,9 @@ namespace CKAN
         }
 
         /// <summary>
-        /// Start or restart a timer to update the filter after an interval since the last keypress.
-        /// On Mac OS X, this prevents the search field from locking up due to DataGridViews being
-        /// slow and key strokes being interpreted incorrectly when slowed down:
+        /// Start or reset a timer to allow multiple changes to the filter criteria before automatically
+		/// refreshing the display. Refresh may take longer than the time between key strokes,
+		/// which makes the UI seem unresponsive or buggy:
         /// http://mono.1490590.n4.nabble.com/Incorrect-missing-and-duplicate-keypress-events-td4658863.html
         /// </summary>
         private void RunFilterUpdateTimer()
@@ -289,13 +288,12 @@ namespace CKAN
                 filterTimer = new Timer();
                 filterTimer.Tick += OnFilterUpdateTimer;
                 filterTimer.Interval = 500;
-                filterTimer.Start();
             }
             else
             {
                 filterTimer.Stop();
-                filterTimer.Start();
             }
+            filterTimer.Start();
         }
 
         /// <summary>
--- a/GUI/Controls/EditModSearch.cs
+++ b/GUI/Controls/EditModSearch.cs
@@ -130,7 +130,7 @@ namespace CKAN
                     Main.Instance.ManageMods.mainModList.ModuleLabels.LabelsFor(Main.Instance.CurrentInstance.Name).ToList()
                 );
                 SearchToEditor();
-                TriggerSearchOrTimer();
+                QueueTriggerSearch();
             }
             catch (Kraken k)
             {
@@ -215,7 +215,7 @@ namespace CKAN
             }
             else
             {
-                TriggerSearchOrTimer();
+                QueueTriggerSearch();
             }
         }
 
@@ -246,8 +246,7 @@ namespace CKAN
                     // Bypass the timer for immediate update
                     e.Handled = true;
                     e.SuppressKeyPress = true;
-                    filterTimer?.Stop();
-                    TriggerSearch();
+                    QueueTriggerSearch();
                     break;
 
                 case Keys.Up:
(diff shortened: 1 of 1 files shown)


kfsone:
@hebarusan Is there additional work I need to do for this to advance?

GUI/Controls/EditModSearch.cs
```diff
@@ -263,9 +262,10 @@ private void FilterTextBox_KeyDown(object sender, KeyEventArgs e)
...
         }
 
-        private void TriggerSearchOrTimer()
+        private void QueueTriggerSearch()
         {
-            if (Platform.IsMono && !string.IsNullOrEmpty(FilterCombinedTextBox.Text))
+            filterTimer?.Stop();
+            if (!string.IsNullOrEmpty(FilterCombinedTextBox.Text))
             {
                 // Delay updating to improve typing performance on OS X and Linux
```
DasSkelett, line 270: We should probably remove the explicit mention of "OS X and Linux" now that it is the case for all OSes.

HebaruSan:
> @HebaruSan Is there additional work I need to do for this to advance?

Sorry for the delay, I needed to reboot into Windows to review this properly, which I have now finally done. Should have a chance to look at it in the next day or three.

HebaruSan:
Hmm, I'm still seeing essentially-instantaneous updates for most searches on Windows, without this change. Maybe there's something specific to one class of searches that's slower than it should be. If so, then fixing that would be preferable to hiding the problem with a workaround.

HebaruSan:
Specific example: as I type `dep:kopernicus` letter by letter, the list updates in real time and my cursor is not slowed down at all. Same for simple searches of names or descriptions.

However, the `is:` and `not:` prefixes slow the app down a lot. So there's probably something there that could be fixed.

kfsone:
I'd suggest the real problem is a non-trivial operation being done within the event pump loop. The application can't even receive subsequent key presses because the message pump is tied up doing the search from a previous keystroke, causing a cumulative delay effect where the app feels like its getting slower and slower. If the user is typing at 4cps and search is taking 600ms:

t+0 user types 'i', search begins
t+250 user types 's',
t+500 user types ':',
t+600 search finishes, message pump delivers 's', search begins,
t+750 user types 'c',
t+1000 user types 'o',
t+1200 search finishes, message pump delivers ':', search begins,
t+1250 user types 'm',
t+1500 user types 'p',
t+1750 user types 'a',
t+1800 search finishes, message pump delivers 'c', search begins,

On my i7-7700K 4.2GHz I can type "is:compatible kerb" before the UI has gotten "is:c" out: https://www.youtube.com/watch?v=bd7Mjd-QBJ8

The timer is just a pretty standard implementation of pre-emptive asynchronous dispatch of work from a ui thread.

Additional viable approaches would be:

- Always trigger an asynchronous search, blocking additional searches until the first completes, this is typically how microsoft did their UIs in the bad old days (try typing "\\1.1.1.1\" into a Windows 7 "Run" dialog and then try to use explorer before it times out)
- Always trigger an asynchronous search, restart if a new input is entered; responsiveness and behavior will vary based on user input speeds - typing fast will get you some cache benefits,
- set a second, larger threshold, so that a search will occur if Nms has elapsed since the last keypress or Mms since the first key/previous search, this will have different behaviors depending on the length of strings and may cause advanced users who regularly switch filters some discomfort,

kfsone:
> @kfsone, based on that analysis, why is `dep:kopernicus` not equally slow?

I have the same problem with 'dep:kopernicus', until the ':' is processed, so it takes about 1.2s from my finishing typing to when the UI finally shows the 'k'. Once it has the ':' tho, it can mostly keep up with my rate of typing. It feels a little like typing over a transatlantic ssh session, as opposed to a 1200/75 modem before the ':' was reached.

HebaruSan:
Quick note that I missed before:

> Additional viable approaches would be:
> 
> * Always trigger an asynchronous search, blocking additional searches until the first completes, this is typically how microsoft did their UIs in the bad old days (try typing "\1.1.1.1" into a Windows 7 "Run" dialog and then try to use explorer before it times out)
> * Always trigger an asynchronous search, restart if a new input is entered; responsiveness and behavior will vary based on user input speeds - typing fast will get you some cache benefits,

Since a short search term can be slow but a long one is fast, that means the slow part is adding a large number of matched rows to the grid, not the actual search logic itself (looping over the mods and checking whether they match). Unfortunately WinForms forces all UI updates to take place in the same foreground thread, so the slow part here can't be made asynchronous.

HebaruSan:
OK, after looking into this a bit more for #3620, I'm convinced that delaying on all platforms is a good idea. However, there is at least one new bug here (pressing Enter no longer triggers an immediate refresh), and I've never been particularly comfortable with this control owning its own timer and implementing all this logic for what in principle is a generic "[debounce](https://www.freecodecamp.org/news/javascript-debounce-example/)" operation that we could easily want to use in other places. I'm going to start a new PR with a generic `CKAN.GUI.Util.Debounce` for that...

Issue #3423, opened 2021-07-31
[Bug] CKAN process ignores termination signals while downloading/updating


mgalyean:
Background
----------
* Operating System: Debian Linux
* CKAN Version: 1.3.0
* KSP Version: 1.11.2

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
N/A

Problem
-------
Once CKAN has started updating its info, it is not possible to terminate it via r-click->close or ctrl-c.  So if you are on a really slow wifi and want to switch to a faster one you have to wait for CKAN to timeout on the network, then exit it, then connect to the faster wifi, then relaunch CKAN.  It doesn't appear to want to attempt to retry the update without exiting first even after the better connection is established.  There could be two issues here: no network retry after timeout under some vague condition and 2) all processes should be interruptible; if a critical atomic data operation is in process, it should be rolled back either prior to exiting, or rolled back by the restarted process
 
**Steps to reproduce**
Try to interrupt/close/ctrl-c/kill -TERM <ckanpid> while it is updating right after opening
* See error
It won't stop until network timeout

**Expected behavior**
Stops when told to stop.  Doesn't ignore user input or signals until a network timeout occurs

HebaruSan:
What does "updating its info" mean? What steps are you taking?

Trying this a few times with `ckan update`, it exits immediately when I press ctrl-C, every time. Need clearer specifics about what this complaint is.

mgalyean:
I'm referring to the GUI, I haven't tried from the command line.  I mean when the GUI client fires up, the first thing it does is refresh/update mod header info from the repos I assume.  It is during that first refresh/update upon startup that I am referring.  I do know that long, long ago Linux had some corners where it would block on network timeouts or something

This report does not apply to the command line client.  I don't know what happens there.  I only use the GUI and forget the command line client even exists, sorry about that

HebaruSan:
Thanks for the clarification. Still very weird; I'm able to terminate the CKAN GUI via ctrl-C even while updating the registry. Where is the window focus when you press ctrl-C? It's not going to try to stop the process if CKAN has focus, for a GUI app ctrl-C just means copy. You'd need to run ckan from a terminal and press ctrl-C while that terminal has focus.

mgalyean:
[...]

PR #3425, opened 2021-08-01
Set GNOME single window property

HebaruSan:
## Problem

In GNOME Shell currently, if you launch CKAN via the system menu, its icon has a "New Window" option:

image

Clicking it launches CKAN again, which invariably brings up the locked instance error:

image

It is a little embarrassing to have a highly visible option that doesn't work well by default.

Noticed while investigating #3423.

## Cause

The "New Window" option is added by GNOME Shell automatically (if the `.desktop` file's [`Actions=` list](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#extra-actions) doesn't include `new-window`), and probably works fine for 90% of apps where running the `Exec` again successfully opens a new window that works properly.

## Changes

Now our `.desktop` files set `X-GNOME-SingleWindow=true` to suppress the "New Window" option.

image

To find out this option existed, I had to read the GNOME source code:

- https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/src/shell-app.c#L662-665

I looked for a similar option for KDE or other desktops but could not find one, but I'm happy to add it here if it exists.

In addition, `StartupNotify` is removed because we do not "KNOW" that our application will send this message (because who knows what Mono's WinForms will do or not do in any situation):

- https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys

  > If true, it is KNOWN that the application will send a "remove" message when started with the DESKTOP_STARTUP_ID environment variable set. If false, it is KNOWN that the application does not work with startup notification at all (does not shown any window, breaks even when using StartupWMClass, etc.). If absent, a reasonable handling is up to implementations (assuming false, using StartupWMClass, etc.). (See the Startup Notification Protocol Specification for more details).


DasSkelett (approved):
Nope, can't find anything for KDE either. And ` X-GNOME-SingleWindow=true` or ` X-KDE-SingleWindow=true`don't have an effect.

Also not testing on Gnome, just gonna believe your screenshots there :laughing:

PR #3426, opened 2021-08-02
Override provides prompt with relationship property, check first recommendation in any_of group

HebaruSan:
## Motivation

The RSS/RO/RP-0/RP-1 team reached out to discuss ways to streamline their installation process, see KSP-CKAN/NetKAN#8615 and KSP-CKAN/NetKAN#8688. They hope to reduce the number of decisions the user has to make and provide more information to help the user make good choices.

We're still brainstorming options for improvements, but a few ideas seem to be clear wins so far. This is the first of possibly/hopefully several changes resulting from that discussion.

## Problems

- When a `depends` clause matches multiple modules (via either `provides` or `any_of`), only a standard prompt based on the identifier(s) is shown (and it varies between CmdLine/ConsoleUI/GUI). The RSS/RO/RP-0/RP-1 team would like to be able to override that prompt in order to give users additional help and clues about which module to pick.
- The RSS/RO/RP-0/RP-1 team has been taking advantage of `any_of` relationships for conflicting recommendations because they don't all show up as checked:
  image
  Rather, they're all _unchecked_, which is still not great because these are supposed to be _recommendations_, on-by-default:
  image

## Changes

- Now the standard prompt for choosing a dependency is the same across CmdLine/ConsoleUI/GUI (copied from the existing English text for GUI)
- Now you can add a `choice_help_text` property to a relationship to override the text the user sees when choosing among multiple modules:
  ```yaml
  depends:
    - any_of:
      - name: Mod1
      - name: Mod2
      choice_help_text: Pick Mod1 unless you like Mod2
    - name: RSSTextures
      choice_help_text: Pick the 16K version if your computer can handle it
  ```
  image
- Now if you recommend an `any_of` group or an identifier provided by multiple modules, the first one that shows up on screen will be checked, and the rest will be unchecked:
  image

After this, it should be possible to guide users into a good RSS/RO/RP-0/RP-1 install more often and more easily.

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -384,6 +384,10 @@
                             "max_version" : {
                                 "description" : "Optional maximum version",
                                 "$ref"        : "#/definitions/version"
+                            },
+                            "choice_help_text": {
+                                "description" : "Optional help text shown when user has to choose among multiple modules",
+                                "type"        : "string"
                             }
                         },
                         "required" : [ "name" ]
@@ -394,6 +398,10 @@
                             "any_of": {
                                 "description": "List of modules that can satisfy this relationship",
                                 "$ref": "#/definitions/relationship"
+                            },
+                            "choice_help_text": {
+                                "description" : "Optional help text shown when user has to choose among multiple modules",
+                                "type"        : "string"
                             }
                         },
                         "required": [ "any_of" ]
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -174,33 +174,25 @@ namespace CKAN.CmdLine
                 }
                 catch (TooManyModsProvideKraken ex)
                 {
-                    // Request the user selects one of the mods.
-                    string[] mods = new string[ex.modules.Count];
-
-                    for (int i = 0; i < ex.modules.Count; i++)
-                    {
-                        mods[i] = String.Format("{0} ({1})", ex.modules[i].identifier, ex.modules[i].name);
-                    }
-
-                    string message = String.Format("Too many mods provide {0}. Please pick from the following:\r\n", ex.requested);
-
+                    // Request the user selects one of the mods
                     int result;
-
                     try
                     {
-                        result = user.RaiseSelectionDialog(message, mods);
+                        result = user.RaiseSelectionDialog(
+                            ex.Message,
+                            ex.modules
+                                .Select(m => string.Format("{0} ({1})", m.identifier, m.name))
+                                .ToArray());
                     }
                     catch (Kraken e)
                     {
                         user.RaiseMessage(e.Message);
-
                         return Exit.ERROR;
                     }
 
                     if (result < 0)
                     {
-                        user.RaiseMessage(String.Empty); // Looks tidier.
-
+                        user.RaiseMessage("");
                         return Exit.ERROR;
                     }
 
--- a/Cmdline/Action/Upgrade.cs
+++ b/Cmdline/Action/Upgrade.cs
@@ -230,7 +230,7 @@ namespace CKAN.CmdLine
                     catch (TooManyModsProvideKraken k)
                     {
                         int choice = user.RaiseSelectionDialog(
-                            $"Choose a module to provide {k.requested}:",
+                            k.Message,
                             k.modules.Select(m => $"{m.identifier} ({m.name})").ToArray());
                         if (choice < 0)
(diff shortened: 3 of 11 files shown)


HebaruSan:
@NathanKell, @ppboyle, @StonesmileGit, @siimav, @jwvanderbeck:
There's a test build here if you want to try these changes:

- https://github.com/KSP-CKAN/CKAN/suites/3384640290/artifacts/79794073

To test the `comment` property thing, you can save a copy of the current .ckan files to your local drive, edit them to add the new property, then use File → Install from .ckan to install them:

- https://github.com/KSP-CKAN/CKAN-meta/blob/master/RP-0/RP-0-v1.10.5.0.ckan
- https://github.com/KSP-CKAN/CKAN-meta/blob/master/RealismOverhaul/RealismOverhaul-v14.3.0.0.ckan
- https://github.com/KSP-CKAN/CKAN-meta/blob/master/RealSolarSystem/RealSolarSystem-v18.1.5.ckan

NathanKell:
Awesome! I'll check this out after supper (am currently fighting with porting my Chrome profile over to a new PC, bleh). Thanks!!

HebaruSan:
One note, don't add `comment` properties to your metadata just yet, it's likely that the name will change during code review, since `comment` is pretty generic and not clearly related to this usage. But I did not think of a better name.

NathanKell:
Worked like a treat! ❤️

DasSkelett:
Yeah given that the spec defines (a top-level) comment as
> A comment field, if included, is ignored. It is not displayed to users, nor used by programs. Its primary use is to convey information to humans examining the CKAN file manually

https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#comment

I wouldn't reuse the key like that, even if it's not top-level. Maybe `help` or `help_text` instead? Or `hint`?

HebaruSan:
> Maybe `help` or `help_text` instead? Or `hint`?

Those are good. I'd like to add a short adjective prefix as well that explains what the hint is about.
`choice_hint`? `choice_help_text`?

NathanKell:
I think choice_help_text is clearest.

Sent by my thumbs, slowly.

HebaruSan:
`choice_help_text` it is! Force-pushed.

DasSkelett:
I was wondering why RCSBuildAid and RCSBuildAidCont both start checked, despite being in an  `any_of` in the RealismOverhaul netkan, turns out they are both listed as first-level recommendations in the RP-0 netkan.
```json
 		{ "name" : "RCSBuildAid" },
		{ "name" : "RCSBuildAidCont" },
```
@NathanKell I think you might want to adjust that to be in an `any_of` as well.

DasSkelett:
`choice_help_text` doesn't work yet for `any`of:
```
`any_of` should not be combined with other properties
```
https://github.com/KSP-CKAN/CKAN/blob/a8331592dda3bbfed3e42cdeb5e39f6858d23c5c/Core/Converters/JsonRelationshipConverter.cs#L29-L32

HebaruSan:
Fixed, I think.

image

DasSkelett:
Should we also deduplicate items from the list? Of course, the metadata should be changed as well, but I think it doesn't hurt to fix it in the client.

HebaruSan:
I think updating just `AnyOfRelationshipDescriptor.LatestAvailableWithProvides` is sufficient, we shouldn't be generating duplicates based on `provides`.

DasSkelett:
Yeah, that makes sense. Thanks!

DasSkelett (approved):
Okay, I think this PR works well as is. If we find more fixes that are needed for https://github.com/KSP-CKAN/NetKAN/pull/8701, it might be wise to do them as separate PR, especially if they _do_ dig into the RelationshipResolver.

HebaruSan:
> `choice_help_text` doesn't work yet for `any_of`:
> 
> ```
> `any_of` should not be combined with other properties
> ```
> 
> https://github.com/KSP-CKAN/CKAN/blob/a8331592dda3bbfed3e42cdeb5e39f6858d23c5c/Core/Converters/JsonRelationshipConverter.cs#L29-L32

@DasSkelett, should we bump the spec version for this? That would give us a way to protect old clients from newer metadata with extra properties alongside `any_of`.

DasSkelett:
Indeed, the same though came to my mind yesterday. I think we should do that.

PR #3429, opened 2021-08-04
Escape ampersands in mod info abstract label

### LINKED ISSUE #3428: [Bug] Ampersands not escaped in mod info, doubled in grid on Windows
Background
----------
Windows 10
CKAN 10.30.4 

Problem
-------
The text "&&" displays in the inline description, but not in the metadata panel to the right.

Specifically, for the mod Animate Generic Effects Module by linuxgurugamer the inline description shows "Added effects && part symmetry to ModuleAnimateGeneric", while the description in the metadata panel shows "Added effects _part symmetry to ModuleAnimateGeneric".

Expected behavior
-------
Assuming both strings come from the same source, the description in the metadata panel should match the inline description. 

CKAN metadata error

HebaruSan:
## Problem

If a mod has `&` in its abstract, it is hidden and the next character is underlind in the mod info abstract label:

windows

linux

A similar issue with the mod list was fixed in #3149.

## Cause

WinForms `Labels` (and some other controls) treat `&` as a hotkey indicator; the next character in the text is used as the hotkey such that if you press Alt and that key, the control (or the next control in the tab order) receives focus. This is indicated visually by hiding the `&` and underlining the next character.

Usually it's not a good idea to expose this functionality to external data sources such as abstracts entered into SpaceDock; they generally just want the `&` to display normally, which is accomplished by doubling it to `&&`.

## Changes

Now we double any ampersands in the abstract when populating them into the mod info abstract, which makes them display normally:

image

Fixes #3428.

### DIFF
--- a/GUI/Controls/ModInfo.cs
+++ b/GUI/Controls/ModInfo.cs
@@ -124,7 +124,7 @@ namespace CKAN
 
             Util.Invoke(MetadataModuleNameTextBox, () => MetadataModuleNameTextBox.Text = module.name);
             UpdateTagsAndLabels(module);
-            Util.Invoke(MetadataModuleAbstractLabel, () => MetadataModuleAbstractLabel.Text = module.@abstract);
+            Util.Invoke(MetadataModuleAbstractLabel, () => MetadataModuleAbstractLabel.Text = module.@abstract.Replace("&", "&&"));
             Util.Invoke(MetadataModuleDescriptionTextBox, () =>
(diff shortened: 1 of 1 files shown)


DasSkelett (approved):
A bit unfortunate that Mono and Framework decided to behave differently in the grid list case. But a double ampersand is probably more readable than no ampersand and a weird underscore.
At least this PR fixes it for the side panel.

PR #3432, opened 2021-08-10
Update old validate script to Python 3

busterbeam:
I haven't removed `jsonschema` but that can be added as a module in `python3`.  Sorry for the confusing backward logic of True and False being swapped.  your exit status was that way around so... yea your fault.  Just trying to make the code shorter simpler and neater

### DIFF
--- a/bin/ckan-validate.py
+++ b/bin/ckan-validate.py
@@ -33,7 +33,7 @@ def ckan_validates(ckan_paths):
                 except ValueError as error:
                     yield print(
                         "Failed! This error will be cryptic,\n",
-                        "but often a JSOqN or property error", error)
+                        "but often a JSON or property error", error)
         else:
             print(f"File '{ckan_path}' does not exist, skipping..")
 
@@ -45,6 +45,5 @@ def main(ckan_paths):
 if __name__ == "__main__":
     if len(argv) == 1: exit(print(f"Usage: {argv[0]} <.ckan files>"))
     else: exit(main(argv[1:]))
-
 
 
--- a/bin/ckan-validate.py
+++ b/bin/ckan-validate.py
@@ -1,58 +1,50 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
-import os, sys
+from os.path import exists
+from sys import argv, exit
 
-import json
+from json import load
 from jsonschema import validate, ValidationError
 
-def main():
-    if len(sys.argv) == 1:
-        print 'Usage:'
-        print sys.argv[0] + ' <.ckan files>'
-        sys.exit(0)
-        
-    SCHEMA_PATH = "CKAN.schema"
-    if not os.path.exists(SCHEMA_PATH):
-        print 'Cannot find JSON schema at %s' % SCHEMA_PATH
-        SCHEMA_PATH = "../" + SCHEMA_PATH
-        if not os.path.exists(SCHEMA_PATH):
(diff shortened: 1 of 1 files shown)


bin/ckan-validate.py
```diff
@@ -1,58 +1,50 @@
...
+                try:
+                    validate(load(ckan_file), schema)
+                    yield not print(f"Success!")
+                except ValidationError as error:
+                    yield print(
+                        "Failed! See below for error description.\n", error)
+                except ValueError as error:
+                    yield print(
+                        "Failed! This error will be cryptic,\n",
+                        "but often a JSOqN or property error", error)
```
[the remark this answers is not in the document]
busterbeam: Thanks must of fell asleep on my keyboard

HebaruSan:
This script is pretty old and no longer in use, but I guess if we're not going to delete it, we might as well update it in case someone finds a use for it.

HebaruSan:
Probably a good idea to test it first, though. :smirk: 

```
$ bin/ckan-validate.py ~/Downloads/KSP/CKAN-meta/Astrogator/Astrogator-v0.10.2.ckan 
Traceback (most recent call last):
  File "/home/user/Downloads/KSP/CKAN/bin/ckan-validate.py", line 47, in <module>
    else: exit(main(argv[1:]))
  File "/home/user/Downloads/KSP/CKAN/bin/ckan-validate.py", line 43, in main
    return all(ckan_validates(ckan_paths, schema))
TypeError: ckan_validates() takes 1 positional argument but 2 were given
```

bin/ckan-validate.py
```diff
@@ -1,58 +1,49 @@
...
+            return schema
+
+def ckan_validates(ckan_paths):
+    for ckan_path in ckan_paths:
+        if exists(ckan_path):
+            with open(ckan_path, 'r') as ckan_file:
+                print(f"Validating {ckan_path}..")
+                try:
+                    validate(load(ckan_file), schema)
+                    yield not print(f"Success!")
```
HebaruSan, line 29: My linter is telling me `print` doesn't return a value, so it probably doesn't make sense to put it in an expression like this.

bin/ckan-validate.py
```diff
@@ -1,58 +1,49 @@
...
-
-    sys.exit(error)
+    with open(schema_path, 'r') as schema_file:
+        schema = load(schema_file)
+        if schema == None:
+            return print("Could not parse JSON schema, exiting..")
+        else:
+            return schema
+
+def ckan_validates(ckan_paths):
```
HebaruSan, line 22: Looks like you meant to have `schema` as a second parameter here.

bin/ckan-validate.py
HebaruSan, line 29: My linter is telling me `print` doesn't return a value, so it probably doesn't make sense to put it in an expression like this.

HebaruSan:
> the confusing backward logic of True and False being swapped. your exit status was that way around

There's a module called `exitstatus` that's pretty good for this, you can return `ExitStatus.failure` or `ExitStatus.success` instead of hard coding numbers. Willing to try that?

HebaruSan:
Looks like this isn't . Better to

HebaruSan:
Looks like this isn't going to get updated. For now it's better to have an old, working script than an updated, broken one.

@busterbeam, if you have a chance to address the review comments here, please let us know and we can re-open this.

HebaruSan:
No updates, closing again.

PR #3436, opened 2021-08-13
More Netkan validation for any_of

HebaruSan:
## Motivation

In #3426 we added the ability to set `choice_help_text` for a relationship to change the prompt the user sees when choosing between multiple modules.

However, if you add `choice_help_text` to an `any_of` relationship, the current client will throw exceptions. So if any metanetkan does this, old clients could be disabled until we clean it up. This kind of incompatibility is usually managed with `spec_version` bumps.

## Changes

Now if Netkan finds `choice_help_text` in an `any_of` relationship, it requires the `spec_version` to be at least v1.31. This will ensure that the old clients will not attempt to load such metadata. This will require the next release to be v1.31.0.

### Side fixes

The spec is updated to explain `choice_help_text`, since it now has a version-specific component.

The "Set `x_netkan_license_ok` to suppress" sentence is removed from Netkan's output because this is bad advice. While that functionality does exist, almost all users should simply set the license to a valid license string.

### DIFF
--- a/Netkan/Validators/LicensesValidator.cs
+++ b/Netkan/Validators/LicensesValidator.cs
@@ -37,7 +37,7 @@ namespace CKAN.NetKAN.Validators
             {
                 if (licenses == null || licenses.Count < 1)
                 {
-                    throw new Kraken("License should match spec. Set `x_netkan_license_ok` to suppress");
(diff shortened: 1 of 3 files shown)


DasSkelett (approved):
👍

PR #3437, opened 2021-08-19
Add user guide and Discord to GUI help menu

HebaruSan:
## Motivation

CKAN has a pretty detailed user guide that answers most questions users might have. We frequently handle support questions by posting links to it:

- https://github.com/KSP-CKAN/CKAN/wiki/User-guide

However, GUI's help menu shows only:

image

This means the user guide isn't discoverable, and the help menu doesn't provide help!

## Changes

Now new menu items in the Help menu link to the user guide and CKAN Discord:

image

Hopefully this will mean that users seeking help will find the user guide and see their questions answered within it. If not, they can ask questions on Discord.

### Questions

Is "User guide" clear enough that a user would click on it? Checking a few older Windows applications, they use Help → View Help to open their main help documentation, which we could do as well.

image

### DIFF
--- a/GUI/Localization/de-DE/Main.de-DE.resx
+++ b/GUI/Localization/de-DE/Main.de-DE.resx
@@ -178,6 +178,7 @@
   <data name="GameCommandlineToolStripMenuItem.Text" xml:space="preserve"><value>Spielstart-Befehlszeile</value></data>
   <data name="compatibleGameVersionsToolStripMenuItem.Text" xml:space="preserve"><value>Kompatible Spielversionen</value></data>
   <data name="helpToolStripMenuItem.Text" xml:space="preserve"><value>Hilfe</value></data>
+  <data name="userGuideToolStripMenuItem.Text" xml:space="preserve"><value>Benutzerhandbuch</value></data>
   <data name="reportClientIssueToolStripMenuItem.Text" xml:space="preserve"><value>Fehler mit dem CKAN Client melden</value></data>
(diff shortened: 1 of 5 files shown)


vinix38:
Hi, I'm away from home currently, could you add the French translation for me? "User guide" --> "guide d'utilisation" (If you change your mind, I can adapt the translation, just @ me)

HebaruSan:
> Hi, I'm away from home currently, could you add the French translation for me? "User guide" --> "guide d'utilisation" (If you change your mind, I can adapt the translation, just @ me)

Sounds good, I'll add that. Thanks!

> EDIT : Just had a thought, do you think it would possible/useful to translate the documentation ? It might be pushing it, but I was wondering :)

If you are interested in producing a French version of the user guide, I think that would be fantastic, and we would find a way to fit it into the wiki somehow.

HebaruSan:
Maybe there should also be a Help menu option for the Discord...?

DasSkelett:
> Is "User guide" clear enough that a user would click on it? Checking a few older Windows applications, they use Help → View Help to open their main help documentation, which we could do as well.

I'm fine with "User guide", it gives a better hint on what to expect than a generic "View Help", which could be anything from a two sentence "click buttons to install" to a multi-volume essay.
Also fits better if we ever add a second link to a different "help" (like a multi-volume essay...)

DasSkelett (approved):
Excellent idea :+1:

PR #3438, opened 2021-08-19
YAMLize netkan spec

HebaruSan:
## Motivation

In #3367, we enabled YAML format for human-edited metadata, but most of the relevant documentation still uses JSON. This may be leading advanced users to submit JSON format netkans instead of YAML (see KSP-CKAN/NetKAN#8720).

## Changes

Now the Spec.md documentation uses YAML format when discussing netkans.

The wiki page for adding a mod was updated here, in case you would like to review the changes:

https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN/_compare/9e902784ae698b6f1f1fe8fca6e3b82bcead4e2e...2f5cacc582c207b8ffa3d1988f5e89ad17ef374f

Also the long-broken link to ksp.marce.at is replaced.

### DIFF
--- a/Spec.md
+++ b/Spec.md
@@ -843,7 +843,11 @@ x_netkan_jenkins:
 
 ###### `#/ckan/http/:url`
 
-Indicates data should be fetched from a HTTP server, using the `:url` provided. For example: `#/ckan/http/https://ksp.marce.at/Home/DownloadMod?modId=2`.
+Indicates data should be fetched from a HTTP server, using the `:url` provided. For example:
+
+```yaml
+$kref: '#/ckan/http/https://mysite.com/download_my_mod'
+```
 
 When used, the following fields will be auto-filled if not already present:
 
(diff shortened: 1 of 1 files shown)


DasSkelett:
I think this also should be adjusted, to `\.zip$` `

https://github.com/KSP-CKAN/CKAN/blob/c087675b60c05a2063e444bfc28f08f508850bba/Spec.md#L825

HebaruSan:
OK, changed.

DasSkelett:
Sorry, that was ambiguous. It needs to be either

```
(default: `\.zip$`)
or
(default: `'\.zip$'`)
or
(default: `"\\.zip$"`)
```

as in YAML in single quotes (or no quotes) the backslash doesn't need to be escaped, but in double quotes it behaves like JSON and we need `\\`.

So it would have been correct like it has been initially, but as we tend to use single or no quotes in YAMLkans, it might be misleading like that, so instead we should choose one of the first two options.

HebaruSan:
OK, quotes removed!

DasSkelett (approved):
Easier copy-paste!

PR #3441, opened 2021-08-28
Default install stanza multi-game support, catch missing install_to

HebaruSan:
## Problem

In KSP-CKAN/NetKAN#8735, this metadata:

```yaml
install:
  - find: GameData/InterstellarTechnologies
  - install_to: GameData
```

caused this error:

```
Error: 1459 [1] FATAL CKAN.NetKAN.Program (null) - Object reference not set to an instance of an object
```

## Cause

`install_to` is [required by the spec](https://github.com/KSP-CKAN/CKAN/blame/2a2b12e399818e39534070ccf888e3ac08fd47af/Spec.md#L292-L303), and `NetKAN.Validators.InstallValidator` assumes `install_to` will be set. In this instance, it wasn't, because that YAML translates to:

```json
  "install": [
    {
      "find": "GameData/InterstellarTechnologies"
    },
    {
      "install_to": "GameData"
    }
  ]
```

We should catch this explicitly rather than throwing a NRE.

## Motivation

That PR's changes were essentially trying to make the default install stanza's behavior more explicit, so they've been set aside. However, in the process of explaining this I realized that the default install stanza is KSP1-specific: it hard-codes `GameData`. This would make default stanzas unusable for any future supported game that uses a different folder for mod data.

## Changes

- Now if `install_to` isn't set, we throw an explicit error saying that
- Now the default install stanza uses the `PrimaryModDirectoryRelative` property of the module's game, which will be `GameData` for KSP1 and wherever mods are supposed to go instead for other games
- The spec is updated to hopefully be clearer about how the default install stanza works

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -514,7 +514,7 @@ namespace CKAN
                 else
                 {
                     files.AddRange(ModuleInstallDescriptor
-                        .DefaultInstallStanza(module.identifier)
+                        .DefaultInstallStanza(ksp.game, module.identifier)
                         .FindInstallableFiles(zipfile, ksp));
                 }
             }
--- a/Core/Types/CkanModule.cs
+++ b/Core/Types/CkanModule.cs
@@ -10,6 +10,7 @@ using Autofac;
 using log4net;
 using Newtonsoft.Json;
 using CKAN.Versioning;
+using CKAN.Games;
 
 namespace CKAN
 {
@@ -683,7 +684,7 @@ namespace CKAN
             return string.Format("{0} {1}", identifier, version);
         }
 
-        public string DescribeInstallStanzas()
+        public string DescribeInstallStanzas(IGame game)
         {
             List<string> descriptions = new List<string>();
             if (install != null)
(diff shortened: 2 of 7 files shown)


DasSkelett:
> `install_to` is [required by the spec](https://github.com/KSP-CKAN/CKAN/blame/2a2b12e399818e39534070ccf888e3ac08fd47af/Spec.md#L292-L303)

Interestingly, the client does work with a missing `install_to` in .ckan files, it fills in `GameData` when deserializing:
https://github.com/KSP-CKAN/CKAN/blob/c087675b60c05a2063e444bfc28f08f508850bba/Core/Types/ModuleInstallDescriptor.cs#L40-L42
(Just tested it, works indeed)

Also related to this comment, the `null` check is indeed always false:
https://github.com/KSP-CKAN/CKAN/blob/c087675b60c05a2063e444bfc28f08f508850bba/Core/Types/ModuleInstallDescriptor.cs#L75-L80

I guess we could keep it, as valid metadata should always have an `install_to`. Or we could also try to use `game.PrimaryModDirectoryRelative`, but this would require serialization context work etc.. Maybe we should just not populate it at all and throw the hard `BadMetadataKraken`.

HebaruSan:
> it fills in `GameData` when deserializing:
> (Just tested it, works indeed)

Oh, I think that was done to save space in `registry.json`, since there were thousands of occurrences of `install_to: GameData`. So everyone's existing registry files have missing `install_to` properties, and we can't remove the defaulting now.

DasSkelett:
Oh wow, you're right, they are indeed left out in registry.json. Didn't realize that before.

DasSkelett (approved):

PR #3443, opened 2021-09-03
Update Russian localization

nt0g:
A question I have: are empty strings allowed?

Asking specifically for the `"ModSearchYesPrefix"` string. Russian doesn't use copulas in the way English does, which means (besides the obvious) I can't find a translation that will also agree with mod state adjectives.


HebaruSan:
Thanks for the updates! No, an empty string won't work for that one. We need to be able to check whether a search string begins with it:

https://github.com/KSP-CKAN/CKAN/blob/2a2b12e399818e39534070ccf888e3ac08fd47af/GUI/Model/ModSearch.cs#L317-L343

nt0g:
I understand. The translations I used should work for most use cases.

HebaruSan (approved):
I have no way to assess the quality of Russian text, but since you contributed the originals, we have every reason to trust you.

Thanks for the update!

PR #3446, opened 2021-09-04
Show context menu on menu key press

### LINKED ISSUE #3445: [Bug] Menu key does not work on mods list (accessibility)
Background
----------
* CKAN Version: 1.30.4
* KSP Version: 1.12.2.3167

No

Problem
-------
**Describe the bug**
Pressing the "menu key" (between windows and right control) does not bring up "Labels/Reinstall/Download Contents/Purge Contents"

**Steps to reproduce**
Click on any random mod in list
Press "menu key" (https://en.wikipedia.org/wiki/Menu_key)

**Expected behavior**
The right-click menu pops up for highlighted mod after releasing menu key, navigable with arrows+enter (which it already is)

HebaruSan:
## Problem

Most keyboards have a key to open a context menu (mine requires pressing Fn+PrintScrn, but it's there). Currently this does nothing in CKAN, to the detriment of keyboard users.

## Cause

The mod list context menu is implemented in the grid's `MouseDown` event, and pressing the menu key triggers `KeyDown` and does nothing.

## Changes

Now the code to manage the menu is moved a new `ShowModContextMenu` function, which is called from `MouseDown` on a right click and `KeyDown` for `Key.Apps` (the code for the menu key).

Similar changes are also made for the URL link labels in mod info and the About dialog, as well as the list of instances in the manage game instances dialog. And this key also now opens and closes the F10 menu in ConsoleUI.

Fixes #3445.

### DIFF
--- a/ConsoleUI/Toolkit/ConsolePopupMenu.cs
+++ b/ConsoleUI/Toolkit/ConsolePopupMenu.cs
@@ -76,6 +76,7 @@ namespace CKAN.ConsoleUI.Toolkit {
                         break;
                     case ConsoleKey.F10:
                     case ConsoleKey.Escape:
+                    case ConsoleKey.Applications:
                         done = true;
                         break;
                 }
--- a/ConsoleUI/Toolkit/ConsoleScreen.cs
+++ b/ConsoleUI/Toolkit/ConsoleScreen.cs
@@ -18,7 +18,7 @@ namespace CKAN.ConsoleUI.Toolkit {
                 "F10", MenuTip(),
                 () => mainMenu != null
             );
-            AddBinding(Keys.F10, (object sender, ConsoleTheme theme) => {
+            AddBinding(new ConsoleKeyInfo[] {Keys.F10, Keys.Apps}, (object sender, ConsoleTheme theme) => {
                 bool val = true;
                 if (mainMenu != null) {
                     DrawSelectedHamburger(theme);
--- a/ConsoleUI/Toolkit/Keys.cs
+++ b/ConsoleUI/Toolkit/Keys.cs
@@ -75,6 +75,13 @@ namespace CKAN.ConsoleUI.Toolkit {
         );
 
         /// <summary>
+        /// Representation of Apps for key bindings
+        /// </summary>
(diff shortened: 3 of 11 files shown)


DasSkelett:
For some reason this isn't working for me, neither in the GUI nor the console UI, not when started through my IDE nor otherwise.
I've set a breakpoint in my IDE to check whether CKAN even receives the key press, and apparently it does, but with a `KeyCode` of `None` and basically everything else unset.

```
e = KeyEventArgs
 base = EventArgs
 Alt = {bool} false
 Control = {bool} false
 Handled = {bool} false
 KeyCode = {Keys} None
 KeyData = {Keys} None
 KeyValue = {int} 0
 Modifiers = {Keys} None
 Shift = {bool} false
 SuppressKeyPress = {bool} false
 event_handled = {bool} false
 key_data = {Keys} None
 supress_key_press = {bool} false
```

In other applications like Firefox or Dolphin or other KDE windows it works fine.
Given that there is no useful information to get from the `KeyEventArgs`, I don't see a way to fix this on our side.

I blame some misconfiguration of my system or KDE not handling it correctly.
@Xetaxheb already tested for us on Windows, so I'm fine moving on as is.

DasSkelett (approved):

HebaruSan:
> For some reason this isn't working for me, neither in the GUI nor the console UI, not when started through my IDE nor otherwise.

Hmm, I'm seeing the same thing now in GNOME Shell.

> I blame some misconfiguration of my system or KDE not handling it correctly. @Xetaxheb already tested for us on Windows, so I'm fine moving on as is.

I developed this in Windows, too. I'm guessing Mono has a problem handling this key.

HebaruSan:
Confirmed Mono bug (or something more complicated); this function receives keycode 135 and returns 0:

https://github.com/mono/mono/blob/5eb240a7fd5a1fee02a3851ccacca0925563258a/mcs/class/System.Windows.Forms/System.Windows.Forms/X11Keyboard.cs#L680-L697

The `LookupString` call seems to work based on my debugging output:

```
ks = XK_Menu, keysym = 65383
```

... but then the `keyc2vkey` lookup apparently doesn't have something usable for this key. I think that array is built by examining the X11 keyboard configuration, may take a while to get through it...

PR #3453, opened 2021-09-26
Suppress incompatibility warning at game launch

### LINKED ISSUE #2955: Add an option to disable the list of incompatible mods
At game startup.

HebaruSan:
## Background

In #2601 we added a pop-up that appears at game launch if the user has incompatible modules installed:

image

The goal was to alert users that some previously compatible mods may stop working after Steam auto-updates the game (e.g., this would have happened with the upgrade from KSP 1.7 to 1.8), similar to what KSP-AVC and MiniAVC do. It was also hoped that prompting users in this way would have positive side effects for metadata accuracy (i.e., if they know a mod is compatible but not yet marked as such, they might submit pull requests for that mod's version file to update it), but for the most part that seems not to have happened.

## Motivation

Some power users install incompatible mods on purpose and experience alert fatigue with this popup. They would like to be able to suppress it, especially after they know that their install is working.

This is currently one of our more frequently received feature requests.

## Changes

Now the warning popup has a new "Don't show this again for these mods on KSP X.XX.X" checkbox in the lower left:

image

If you check this checkbox and click Launch, then we write a file like this to `<KSP>/CKAN/suppressed_compat_warning_identifiers.json`:

```json
{"GameVersionWhenWritten":"1.12.2.3167","Identifiers":["HeapPadder","EvaFollower","AirParkContinued"]}
```

The next time you click Launch Game, we load this file and check whether `GameVersionWhenWritten` still matches your game version. If so, then the identifiers listed in the `Identifiers` property are excluded from the compatibility warning popup. If _all_ of your incompatible modules are already in the file, then the popup doesn't appear at all.

If your game version changes, or if you install new incompatible mods, then the warning will appear again, to ensure that we continue to warn in potentially crashy situations.

No means is provided to "revert" the suppression manually other than deleting that JSON file, since it's pretty unlikely a user would check the checkbox accidentally, and there are easier ways to check which of their mods are incompatible than launching the game.

Fixes #2955.

### DIFF
--- a/Core/SuppressedCompatWarningIdentifiers.cs
+++ b/Core/SuppressedCompatWarningIdentifiers.cs
@@ -1,0 +1,44 @@
+using System;
+using System.IO;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using log4net;
+
+using CKAN.Versioning;
+
+namespace CKAN
+{
+    class SuppressedCompatWarningIdentifiers
+    {
+        public GameVersion GameVersionWhenWritten;
+        public HashSet<string> Identifiers = new HashSet<string>();
+
+        public static SuppressedCompatWarningIdentifiers LoadFrom(GameVersion gameVer, string filename)
+        {
+            try
+            {
+                var saved = JsonConvert.DeserializeObject<SuppressedCompatWarningIdentifiers>(File.ReadAllText(filename));
+                // Reset warnings if e.g. Steam auto-updates the game
+                if (saved.GameVersionWhenWritten == gameVer)
+                {
+                    return saved;
+                }
+            }
+            catch (Exception exc)
+            {
+                log.Debug("Failed to load", exc);
+            }
+            return new SuppressedCompatWarningIdentifiers()
+            {
+                GameVersionWhenWritten = gameVer
+            };
+        }
+
+        public void SaveTo(string filename)
+        {
+            File.WriteAllText(filename, JsonConvert.SerializeObject(this));
+        }
+
+        private static readonly ILog log = LogManager.GetLogger(typeof(SuppressedCompatWarningIdentifiers));
+    }
+}
--- a/Core/GameInstance.cs
+++ b/Core/GameInstance.cs
@@ -132,6 +132,10 @@ namespace CKAN
             log.InfoFormat("Initialised {0}", CkanDir());
         }
 
+        #endregion
+
+        #region Settings
+
         public void SetCompatibleVersions(List<GameVersion> compatibleVersions)
         {
             this._compatibleVersions = compatibleVersions.Distinct().ToList();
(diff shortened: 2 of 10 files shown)


vinix38:
Hi, could you please add this line to the French translation? Thanks :)
<data name="MainLaunchDontShow" xml:space="preserve"><value>Ne plus demander pour ces mods sur {0} version {1}</value></data>

Core/SuppressedCompatWarningIdentifiers.cs
```diff
@@ -0,0 +1,34 @@
...
+        public GameVersion GameVersionWhenWritten;
+        public HashSet<string> Identifiers = new HashSet<string>();
+
+        public static SuppressedCompatWarningIdentifiers LoadFrom(GameVersion gameVer, string filename)
+        {
+            if (!File.Exists(filename))
+            {
+                return new SuppressedCompatWarningIdentifiers() { GameVersionWhenWritten = gameVer };
+            }
+            var saved = JsonConvert.DeserializeObject<SuppressedCompatWarningIdentifiers>(File.ReadAllText(filename));
```
DasSkelett, line 19: What about catching `JsonException` here in case the file got corrupted?
It's probably insignificant enough not to bother with showing any warning about this to the user (but maybe write a line to the log?)

DasSkelett:
(Did a force-push to add @vinix38 as co-author of the French translation commit, so it shows up in the stats and things, and to add the German translation)

Since we have the ability now, should we also emphasize the list of **new** incompatible mods?
E.g.
```
... Really launch?

Newly incompatible:
HeapAdder 0.0.1.5 (KSP 1.5.1-1.7.3

Existing incompatible:
EvaFollower 1.035 (KSP 1.1.2)
...
```

Core/SuppressedCompatWarningIdentifiers.cs
HebaruSan: Nice idea, consolidating into one exception handler lets us remove the `File.Exists` special case check and combine the "empty" return statements. Done.

HebaruSan:
> (Did a force-push to add @vinix38 as co-author of the French translation commit, so it shows up in the stats and things, and to add the German translation)

:+1: 

> Since we have the ability now, should we also emphasize the list of **new** incompatible mods?

Currently the new ones would be the _only_ mods shown, since the user chose previously not to show the message for the others. I think that's a good way to do it, less reading and parsing to get the basic idea.

HebaruSan:
> Also it looks like we lost my changes in the rebase, should I re-push them, or do you want to handle that so it doesn't happen again? Would be nice if we could also keep @vinix38 [as co-author](https://github.com/KSP-CKAN/CKAN/commit/964fe80ad1b724462baaecfcf9abcc722fffe85b) of the commit, now that GitHub also understands that keyword.

Go ahead and take care of both of those things, if you don't mind, I will make sure to pull your changes before doing anything else this time.

DasSkelett (approved):
Pushed another small change to add a short documentation comment to the new `SuppressableYesNoDialog()` function, to explain what the bool in the tuple means. Hope that's alright.

Looks good and works, I know a few people who will really like this feature ^^

PR #3454, opened 2021-09-26
Append resource links to staging PRs

HebaruSan:
## Motivation

This staging reason explicitly mentions "the forum thread" but fails to provide it:

KSP-CKAN/CKAN-meta#2460

image

That makes these PRs somewhat annoying to review, since you have to find that link yourself (probably via the status page or by checking the metadata).

## Changes

Now we append a table of the mod's `resources` links to the body of any staging pull request:

KSP-CKAN/CKAN-meta#2462 (manually inserted)

image

Behind the scenes:

- We add a new, later Netkan transformer because `StagingTransformer` has to run _before_ the `$kref` processors to be able to detect hard coded versions, but we usually only have the resources available _after_ those transformers run
- We use Markdown's `<link>` syntax for links because otherwise GitHub excludes a trailing `*` from the URL, which we use to abbreviate a lot of forum URLs
- Auto-epoch PRs are affected as well, which should be fine
- `--debug` prints the Markdown source of the table in case you want to see it

### DIFF
--- a/Netkan/CKAN-netkan.csproj
+++ b/Netkan/CKAN-netkan.csproj
@@ -121,6 +121,7 @@
     <Compile Include="Transformers\PropertySortTransformer.cs" />
     <Compile Include="Transformers\SpacedockTransformer.cs" />
     <Compile Include="Transformers\StagingTransformer.cs" />
+    <Compile Include="Transformers\StagingLinksTransformer.cs" />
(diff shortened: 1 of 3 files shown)


techman83 (approved):
I think that's super sensible. Lgtm!

DasSkelett:
Agree, very useful; also appears to be working fine!

PR #3457, opened 2021-09-30
Remove expired Let's Encrypt root certificate from Mono containers

DasSkelett:
## Problem
We are seeing hundreds of SSL errors from our indexer for SpaceDock mods, and metadata CI is broken for those as well.

errors

## Cause
An old root certificate used by Let's Encrypt (DST Root CA X3) has expired. It has largely been replaced by the ISRG Root X1, which is in the certificate trust stores of most modern systems.
However, to enhance compatibility with older clients (especially Android phones, which don't check whether root certs have expired), the X1 cert has a cross-signature from the old X3.

More information:
- https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/
- https://scotthelme.co.uk/lets-encrypt-old-root-expiration/
- https://www.openssl.org/blog/blog/2021/09/13/LetsEncryptRootCertExpire/

(The below is my best guess at where the exact issue with Mono is, I might be wrong)
Most clients don't care about this (e.g. OpenSSL as used by e.g. cURL), they go the certificate chain backwards and stop as soon as they see the X1 cert, which is valid and they trust. Mono however also looks at the cross-signature from X3, and throws an error, because it has expired.

Similar to older, broken OpenSSL versions:
> Unfortunately this does not apply to OpenSSL 1.0.2 which always prefers the untrusted chain and if that chain contains a path that leads to an expired trusted root certificate (DST Root CA X3), it will be selected for the certificate verification and the expiration will be reported.

This does not affect my normal Linux installation, as the X3 cert has already been removed from Ubuntu's trust store. Thus Mono "doesn't see" the expired cert and is happy with the valid X1.

## Workaround
If my above assumption is correct, it is a bug in Mono that needs fixing. And/or Debian should remove X3 from its trust store.

In the meantime, we can blocklist X3 in the Mono images, which forces Mono and other software to ignore it and only look at the X1 cert instead.
This works by prepending a `!` to the relevant line in `/etc/ca-certificates.conf` and running `update-ca-certificates` to apply the change.

We should monitor upstream changes and remove the line again when it is no longer needed. FWIW, `sed` doesn't fail when it didn't change something because the line is no longer present in the file.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,6 +52,17 @@ jobs:
         run: ./build --configuration=${{ matrix.configuration}}
       - name: Run tests
         run: xvfb-run ./build test+only --configuration=${{ matrix.configuration }} --where="Category!=FlakyNetwork"
+      - name: Run inflator container smoke test
+        run: |
+          cd _build
+          curl -fsSL https://get.docker.com -o get-docker.sh
+          sh get-docker.sh
+          docker build --tag inflator --file ../Dockerfile.netkan .
+          docker run --rm --name inflator --entrypoint /bin/bash inflator -c "
+            curl -O https://raw.githubusercontent.com/KSP-CKAN/NetKAN/master/NetKAN/ZeroMiniAVC.netkan && \
+            mono netkan.exe ZeroMiniAVC.netkan
+          "
+        if: matrix.configuration == 'release' && ( matrix.mono == '6.8' || matrix.mono == 'latest' )
 
       - name: Upload ckan.exe artifact
         uses: actions/upload-artifact@v2
--- a/Dockerfile.metadata
+++ b/Dockerfile.metadata
@@ -1,5 +1,7 @@
 FROM mono:latest
 
+RUN /bin/sed -i 's/^mozilla\/DST_Root_CA_X3.crt$/!mozilla\/DST_Root_CA_X3.crt/' /etc/ca-certificates.conf && \
+    /usr/sbin/update-ca-certificates
 RUN apt-get update && \
     apt-get install -y --no-install-recommends python3 python3-pip python3-setuptools git build-essential python3-dev && \
     apt-get clean
--- a/Dockerfile.netkan
+++ b/Dockerfile.netkan
@@ -1,4 +1,6 @@
 FROM mono:latest
+RUN /bin/sed -i 's/^mozilla\/DST_Root_CA_X3.crt$/!mozilla\/DST_Root_CA_X3.crt/' /etc/ca-certificates.conf && \
+    /usr/sbin/update-ca-certificates
 RUN useradd -ms /bin/bash netkan
 USER netkan
 WORKDIR /home/netkan


HebaruSan:
Can we add a test somewhere that will exercise this (succeed if the cert is removed, fail if it's present)? Is there one already?

DasSkelett:
> Can we add a test somewhere that will exercise this (succeed if the cert is removed, fail if it's present)? Is there one already?

`sed` itself doesn't really have a "fail if not present" option (looked that up just yesterday for something else), but we could grep for it, which exits with 1 if it doesn't find anything.

HebaruSan:
No, I meant a basic "smoke test" to see whether ckan.exe Mono is able to access a letsencrypt-signed site, such as SpaceDock.

DasSkelett:
Ah I see. No, I don't think we run any tests in the containers yet. We only run some network tests in the .NET Core CI builds.
We could try to set something up, but it needs some work.

.github/workflows/build.yml
```diff
@@ -51,7 +51,12 @@ jobs:
       - name: Build ckan.exe and netkan.exe
         run: ./build --configuration=${{ matrix.configuration}}
       - name: Run tests
-        run: xvfb-run ./build test+only --configuration=${{ matrix.configuration }} --where="Category!=FlakyNetwork"
+        run: |
+          xvfb-run ./build test+only --configuration=${{ matrix.configuration }} --where="Category!=FlakyNetwork"
+          # Smoke test for netkan.exe, test for SpaceDock reachability
+          cd _build && \
+            curl -O https://raw.githubusercontent.com/KSP-CKAN/NetKAN/master/NetKAN/ZeroMiniAVC.netkan && \
+            mono netkan.exe ZeroMiniAVC.netkan
```
HebaruSan, line 59: Will this test the container? I thought it would have to be added to the Dockerfiles.

DasSkelett: The workflow runs in mono containers, so it is mostly the same.
The problem is that we don't build the images in `build.yml`, only `deploy.yml`.

We could also build the images in `build.yml` (manually, not using `./build docker-*`), but we would also need to split the Dockerfile into a multi-stage file so we don't bloat the production images with unrelated files.
This is more work than I'd like to include in this PR, I'd prefer to get our infrastructure working again first.

Btw., I just found out that if we remove the `--where="Category!=FlakyNetwork"` we also reproduce today's problem:
```
1) Failed : Tests.Core.Net.Net.SpaceDockSSL
  But was:  <System.Net.WebException: Error: TrustFailure (Authentication failed, see inner exception.) ---> System.Security.Authentication.AuthenticationException: Authentication failed, see inner exception. ---> Mono.Btls.MonoBtlsException: Ssl error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED
  
2) Error : Tests.NetKAN.Sources.Jenkins.JenkinsApiTests.GetLatestBuild_ModuleManager_Works
System.Net.WebException : Error: TrustFailure (Authentication failed, see inner exception.)
  ----> System.Security.Authentication.AuthenticationException : Authentication failed, see inner exception.
  ----> Mono.Btls.MonoBtlsException : Ssl error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED

3) Error : Tests.NetKAN.Sources.Spacedock.SpacedockApiTests.GetsModCorrectly
System.Net.WebException : Error: TrustFailure (Authentication failed, see inner exception.)
  ----> System.Security.Authentication.AuthenticationException : Authentication failed, see inner exception.
  ----> Mono.Btls.MonoBtlsException : Ssl error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED

4) Failed : Tests.NetKAN.Sources.Spacedock.SpacedockApiTests.ThrowsWhenModMissing
  Expected: instance of <CKAN.Kraken>
  But was:  <System.Net.WebException: Error: TrustFailure (Authentication failed, see inner exception.) ---> System.Security.Authentication.AuthenticationException: Authentication failed, see inner exception. ---> Mono.Btls.MonoBtlsException: Ssl error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED
```

However I'm not sure if we want to run so many network tests in our CI, for one everything breaks if SpaceDock is down, and it generates load for it as well.

HebaruSan: > The workflow runs in mono containers, so it is mostly the same.

Except now _all_ pull requests will fail the tests, right? We need something that will succeed after this PR's changes.

DasSkelett: Latest changes should do it the way you have in mind. It's ugly but it works...
...on those Docker images that run on a Debian release that is still supported by Docker.

I can move it into a separate step and let it only run for certain mono versions. I could choose all supported ones, or maybe just `latest`?

DasSkelett: (Btw., if you have a better test candidate than ZeroMiniAVC, suggestions welcome, that was just the one I spotted at the end of the message flood in #netkan-bot earlier)

HebaruSan: This makes sense, thanks!

HebaruSan (approved):
Let's try it! :adhesive_bandage:

PR #3458, opened 2021-10-05
Global install filters

### LINKED ISSUE #2490: [Feature] Alternate MiniAVC handling idea
## Background

MiniAVC is a mod for modders that automatically checks whether a mod is compatible with the current game version and displays a warning if it isn't. It can also check a remote URL for new versions of a mod. It's related to KSP-AVC, and it parses .version files.

Many mods bundle a copy of MiniAVC. Mod authors see this as a way to reduce problems caused by incompatible game versions. Currently CKAN's policy is to install those bundled MiniAVC DLLs.

## Problem

Some users may consider MiniAVC redundant with CKAN and not helpful. CKAN can already tell you whether there are updates and whether your mods are compatible, as well as installing the updates for you. If you're aware of the risk and want to try your incompatible mods anyway, the extra warnings when launching the game are kind of annoying to click through.

Evidence of this is the existence of [ZeroMiniAVC](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ZeroMiniAVC.netkan).

## Previous developments

KSP-CKAN/NetKAN#1455 outlines the history of CKAN's attempts to deal with MiniAVC. There was an effort to support it as a standalone mod, which can't work because it uses the DLL's location to determine the scope of scanning:

- If you only install it to GameData/MiniAVC/MiniAVC.dll, it will only scan the MiniAVC folder for .version files, so you won't be warned about incompatible mods or available updates
- If you install it to GameData/MiniAVC.dll, it will scan the entire GameData folder, including mods that don't want MiniAVC support, and including manually installed mods with bundled copies of MiniAVC that then cause messages to be displayed multiple times

To work properly, MiniAVC.dll has to be exactly where a bundling mod puts it. This means we can't share it between modules.

## Suggestions

- In the settings window, add a checkbox titled, "Suppress MiniAVC"
- When installing a module, check the setting. If it's checked, then don't install any files named MiniAVC.dll, MiniAVC.xml, or LICENSE-MiniAVC.txt
- ~~To serve MiniAVC's purpose, we should add a check to the Launch KSP button: scan the installed modules and warn the user if any aren't compatible with the current game version.~~ Done in #2601.

The filter would be easy to add here:

https://github.com/KSP-CKAN/CKAN/blob/ab714c224b9b446ac4a6d9c9b209b73f14bfa752/Core/ModuleInstaller.cs#L514-L541

Since concerns have been raised about whether de-bundling MiniAVC violates modders intentions, the checkbox should be unchecked by default. CKAN would continue to install bundled MiniAVC by default for new and upgrading users, and only if a user was aware of the new setting and willing to take on the risk of enabling it would mods be installed differently.

### LINKED ISSUE #3129: [Feature] Allow blacklisting of particular files under installed mods.
Problem
-------
When installing mods, I sometimes have things I don't want to be installed with them. Whether it's a particular patch for compatibility that's more like combatability, or a mod that bundles other mods that I may not want, there doesn't seem to be a way in CKAN to say I don't want something.

Suggestions
-----------
I'd like to see an option somewhere (whether it's in a separate menu, a config file, a right-click menu, or integrated into the bottom-right area with the various tabs, possibly even enabled by an "Advanced Mode" switch) to blacklist a file or folder from being installed when I install a mod.

Sure, I could write a powershell script to remove the ones I find offensive, but that requires running it manually every time I update or pointing Steam to a PS script that starts KSP after it cleans up or something like that. Similarly, I could just remove the files by hand every time I update a mod. It just seems to go against how CKAN is automated for everything else.

Example:
KSPI-E bundles InterstellarFuelSwitch, and has MM entries to patch a lot of things when Near Future Electrical is installed. I have custom fuel tanks that I use for KSPI-E fuels and I dislike what the NFE patches do.

Similarly, a bunch of mods bundle ModuleManager, but I just install it separately. Multiple MM .dlls can break stuff from what I've read, so maybe I want to explicitly prevent any other mods from installing the .dll.

I should be able to specify something like the following:

```
Blacklist
{
  WarpPlugin
  {
    \InterstellarFuelSwitch
    WarpPlugin\Patches\USI_NF_Mode.cfg
  }
  !ModuleManager
  {
    *ModuleManager*.dll
  }
}
```

This would, in theory:

- Prevent the InterstellarFuelSwitch folder from being installed ONLY from the WarpPlugin mod (I imagine this would be some internal identifier or a GUID instead of "WarpPlugin" - that's just the mod folder's name.)
- Prevent ModuleManager*.dll from being installed UNLESS it's being installed explicitly as the ModuleManager mod. The first * would match ANY folder under a mod - this would be wise if a mod author stuffed MM in some random subdirectory for some reason (something that actually happened in a previous version of a mod I was looking at...)
-- This is probably a pretty specific use case, the first functionality would be _far_ more useful if you had to pick one.

Obviously I'm writing the above in a .cfg format with MM flavoring, but I don't particularly care _how_ it's done - just that it's possible. :)

I considered submitting this as a request to MM and mentioned it on the thread asking if the capability was there, but I feel like there's just too much possibility for bad behavior from mods that think they have a hard incompatibility with something and screwing up a game. It's a kettle of fish that probably shouldn't be opened.

(KSPI-E no longer bundles IFS, but the general idea stands - it used to install GameData\InterstellarFuelSwitch. Lots of mods bundle stuff that might be essential for gameplay the way the author intended, but not necessarily required for basic functionality.)

HebaruSan:
## Motivation

As of some recent API changes, users might install somewhat older mods that _would_ work except that they bundle `MiniAVC.dll`, which crashes the game on startup. Even before the crashes, MiniAVC was annoying and unnecessary thanks to CKAN's own compatibility checking. It would be nice to prevent MiniAVC from being installed.

> NOTE: The crash was specific to KSP 1.12.2, and was fixed in 1.12.3.

Similarly, some power users may wish to customize the details of how certain mods are installed, often a cfg file that patches how two or more mods interact. Currently they can only do this by deleting them manually after installation, which means they will re-appear after upgrades.

## Changes

A new menu option under Settings opens a window with text boxes for installation filters that apply to all mods:

image

image

If you add a string to either of these boxes, then CKAN will not install any file where that string is contained in the full relative path. The top box applies to all game instances, and the bottom box applies only to the current instance. The target audience for this window is power users who are comfortable typing paths, but an Add MiniAVC button is available so typical users can more easily avoid installing MiniAVC.

Cmdline now has a `ckan filter` subcommand to manage these lists, and ConsoleUI has a new screen that works like the one in GUI.

Fixes #2490.
Fixes #3129.

### DIFF
--- a/Cmdline/Action/Filter.cs
+++ b/Cmdline/Action/Filter.cs
@@ -1,0 +1,289 @@
+using System.Linq;
+using System.Collections.Generic;
+
+using Autofac;
+using CommandLine;
+using CommandLine.Text;
+using log4net;
+
+namespace CKAN.CmdLine
+{
+    /// <summary>
+    /// Subcommand for managing installation filters
+    /// </summary>
+    public class Filter : ISubCommand
+    {
+        /// <summary>
+        /// Initialize the subcommand
+        /// </summary>
+        public Filter() { }
+
+        /// <summary>
+        /// Run the subcommand
+        /// </summary>
+        /// <param name="mgr">Manager to provide game instances</param>
+        /// <param name="opts">Command line parameters paritally handled by parser</param>
+        /// <param name="unparsed">Command line parameters not yet handled by parser</param>
+        /// <returns>
+        /// Exit code
+        /// </returns>
+        public int RunSubCommand(GameInstanceManager mgr, CommonOptions opts, SubCommandOptions unparsed)
+        {
+            string[] args = unparsed.options.ToArray();
+            int exitCode = Exit.OK;
+            // Parse and process our sub-verbs
+            Parser.Default.ParseArgumentsStrict(args, new FilterSubOptions(), (string option, object suboptions) =>
+            {
+                // ParseArgumentsStrict calls us unconditionally, even with bad arguments
+                if (!string.IsNullOrEmpty(option) && suboptions != null)
+                {
+                    CommonOptions options = (CommonOptions)suboptions;
+                    options.Merge(opts);
+                    user     = new ConsoleUser(options.Headless);
+                    manager  = mgr ?? new GameInstanceManager(user);
+                    exitCode = options.Handle(manager, user);
+                    if (exitCode != Exit.OK)
+                        return;
+
+                    switch (option)
+                    {
+                        case "list":
+                            exitCode = ListFilters((FilterListOptions)suboptions, option);
+                            break;
+
+                        case "add":
+                            exitCode = AddFilters((FilterAddOptions)suboptions, option);
+                            break;
+
+                        case "remove":
+                            exitCode = RemoveFilters((FilterRemoveOptions)suboptions, option);
+                            break;
+
+                        default:
+                            user.RaiseMessage("Unknown command: filter {0}", option);
+                            exitCode = Exit.BADOPT;
+                            break;
+                    }
+                }
+            }, () => { exitCode = MainClass.AfterHelp(); });
+            return exitCode;
+        }
+
+        private int ListFilters(FilterListOptions opts, string verb)
+        {
+            int exitCode = opts.Handle(manager, user);
+            if (exitCode != Exit.OK)
+            {
+                return exitCode;
+            }
+
+            var cfg = ServiceLocator.Container.Resolve<Configuration.IConfiguration>();
+            user.RaiseMessage("Global filters:");
+            foreach (string filter in cfg.GlobalInstallFilters)
+            {
+                user.RaiseMessage("\t- {0}", filter);
+            }
+            user.RaiseMessage("");
+
+            var instance = MainClass.GetGameInstance(manager);
+            user.RaiseMessage("Instance filters:");
+            foreach (string filter in instance.InstallFilters)
+            {
+                user.RaiseMessage("\t- {0}", filter);
+            }
+            return Exit.OK;
+        }
+
+        private int AddFilters(FilterAddOptions opts, string verb)
+        {
+            if (opts.filters.Count < 1)
+            {
+                user.RaiseMessage("Usage: ckan filter {0} filter1 [filter2 ...]", verb);
+                return Exit.BADOPT;
+            }
+
+            int exitCode = opts.Handle(manager, user);
+            if (exitCode != Exit.OK)
+            {
+                return exitCode;
+            }
+
+            if (opts.global)
+            {
+                var cfg = ServiceLocator.Container.Resolve<Configuration.IConfiguration>();
+                var duplicates = cfg.GlobalInstallFilters
+                    .Intersect(opts.filters)
+                    .ToArray();
+                if (duplicates.Length > 0)
+                {
+                    user.RaiseError(
+                        "Global filters already set: {0}",
+                        string.Join(", ", duplicates)
+                    );
+                    return Exit.BADOPT;
+                }
+                else
+                {
+                    cfg.GlobalInstallFilters = cfg.GlobalInstallFilters
+                        .Concat(opts.filters)
+                        .Distinct()
+                        .ToArray();
+                }
+            }
+            else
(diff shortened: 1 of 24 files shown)


vinix38:
Hi, here is the usual French translation :)

installfiltersdialog
```
<data name="GlobalFiltersGroupBox.Text" xml:space="preserve"><value>Filtres Globaux</value></data>
<data name="InstanceFiltersGroupBox.Text" xml:space="preserve"><value>Filtres de l'Instance</value></data>
<data name="AddMiniAVCButton.Text" xml:space="preserve"><value>Ajouter MiniAVC</value></data>
<data name="$this.Text" xml:space="preserve"><value>Installer des Filtres</value></data>
```
main
```
<data name="installFiltersToolStripMenuItem.Text" xml:space="preserve"><value>Installation de filtres</value></data>
```

DasSkelett:
I was thinking whether a simple `.Contains()` check might be too broad and filter out files that people didn't want to, but since I can't think of a case where this could happen I guess it's okay. And every attempt to change this would make it a lot more complex and less accessible (e.g. if we made it behave like shell globs).

HebaruSan:
Yeah, if a filter is too broad, the user can add more characters to match only the right file(s), as in changing `MiniAVC` to `MiniAVC.dll`. If it's not broad enough, they can do the reverse. I think `.Contains()` gives us a pretty good level of flexibility here.

HebaruSan:
Should we re-scan the already installed files after the user changes the filters? I can imagine that some might expect `MiniAVC.dll` to disappear after they add it to a filter, but it seems like doing that could get messy (i.e., what about installing files that were filtered out previously when a filter is removed?).

DasSkelett:
> Should we re-scan the already installed files after the user changes the filters? I can imagine that some might expect `MiniAVC.dll` to disappear after they add it to a filter, but it seems like doing that could get messy (i.e., what about installing files that were filtered out previously when a filter is removed?).

I would keep it as is, people can still reinstall the mod they want. We could put a sentence in the "Install filters" window like
> Changes to the install filters only take effect on future mod installations, already installed files are unaffected

HebaruSan:
Coming up in a few moments:

image

HebaruSan:
OK, should be all set now. :heavy_check_mark:

vinix38:
Just added a few bits that I had missed during translation

DasSkelett (approved):
All three UI are fine, and it works very well. Thanks!

Clayell:
@HebaruSan Am I using the instance filters incorrectly? I can get the installation filters to work when I put them as global, but not as instance.

Global:
> non_wip_RP0.cfg
> non_WIP_RO_FE_config.cfg

image

(RealismOverhaul)
image

Instance:
> RP-1\Filter_Extensions_Configs\non_wip_RP0.cfg
> RealismOverhaul\Filter_Extensions_Configs\non_WIP_RO_FE_config.cfg

image

(RealismOverhaul)
image

Asking here since the "?" on the installation filter UI leads to this PR.

HebaruSan:
> Instance:
> 
> > RP-1\Filter_Extensions_Configs\non_wip_RP0.cfg
> > RealismOverhaul\Filter_Extensions_Configs\non_WIP_RO_FE_config.cfg
> 
> image

Use forward slashes (`/`), not backslashes. Or just leave out the directory part of the path like you did for the global filter.

Clayell:
Ah, thanks. I forgot windows gives backslashes when you hit "Copy path".

> Or just leave out the directory part of the path like you did for the global filter.

I'm realizing now that I misinterpreted what Global and Instance mean here. I thought that Global meant across all mods in that save file, while Instance required you to specify a mod, like how you have it for the example at the top comment. I suppose I should have read further to see:
> The top box applies to all game instances, and the bottom box applies only to the current instance.

As always, you're a great help Hebaru, thanks!

PR #3460, opened 2021-10-11
Negated search for name, desc, author, lang, relationships, tags

### LINKED ISSUE #3397: [Feature] Negative statements in search
Problem
-------
As you noted [here](https://github.com/KSP-CKAN/CKAN/issues/3071#issuecomment-791656558) - any boolean expression can in fact be written in [DNF](https://en.wikipedia.org/wiki/Disjunctive_normal_form) (or [CNF](https://en.wikipedia.org/wiki/Conjunctive_normal_form) to that matter, but no one likes it).

Except afaik, rn you only added local `and`'s and global `or`'s, and to complete DNF you also need `not`'s.

Well, as for the more practical side of this issue:
1. There are a few kinds of mods I would not want in my mod pack, like flag mods.
2. I've installed a few planet packs, a tech pack, etc. It's safe to assume, I wouldn't need any more of these in the current mod pack.
So it would be great to exclude them, to more comfortably check out the rest of misc mods.

Suggestions
-----------
Many advanced search engines (at least those with `:` syntax, like google and Github issues) use `-` to negate the term.
So for instance: `is:compatible -flags -dep:Kopernicus`

That said, it's definitely not a universal syntax, so would be great to consider more possibilities.
Some search engines use `!`. And IMO `~` would be far more readable with the font you've chosen.

On the other hand - you seem to be leaning towards using more words and fewer symbols.
And I can see `not:compatible` being more readable to non-programmers than `~is:compatible`.
(not 100% sure tho, been a programmer for too long)

HebaruSan:
## Motivation

The GUI search currently supports a limited disjunctive normal form in which only some terms can be negated (the ones using the tri-state toggle control). It would be nice to be able to negate the remaining ones.

## Changes

Now if you add `-` to the start of any of the following search terms (_before_ the prefix), that term will be flipped so it matches only the mods that it didn't match previously:

- Name (no prefix)
- Description (`desc:` prefix in English)
- Authors (`@` prefix)
- Localizations (`lang:` prefix in English)
- Depends (`dep:` prefix in English)
- Recommends (`rec:` prefix in English)
- Suggests (`sug:` prefix in English)
- Conflicts (`conf:` prefix in English)
- Tag (`tag:` prefix in English)

For example, while `@HebaruSan` would include my mods, `-@HebaruSan` would exclude them. The prefixes are omitted from the expanded per-term search boxes, so if you expand the latter search, the authors field would contain `-HebaruSan`.

The label search is not negatable because you can already configure a label to hide the mods it contains (and because it would be annoying technically to do it since unlike the other search terms we track it as a list of `ModuleLabel` objects with no place to store whether it's negated).

Fixes #3397.

### DIFF
--- a/GUI/Model/ModSearch.cs
+++ b/GUI/Model/ModSearch.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Linq;
 using System.Collections.Generic;
-using System.Text.RegularExpressions;
 
 namespace CKAN
 {
@@ -16,9 +15,9 @@ namespace CKAN
         /// Null or empty parameters are treated as matching everything.
         /// </summary>
         /// <param name="byName">String to search for in mod names, identifiers, and abbreviations</param>
-        /// <param name="byAuthor">String to search for in author names</param>
+        /// <param name="byAuthors">String to search for in author names</param>
         /// <param name="byDescription">String to search for in mod descriptions</param>
-        /// <param name="localization">Language to search for in mod localizations</param>
+        /// <param name="localizations">Language to search for in mod localizations</param>
         /// <param name="depends">Identifier prefix to find in mod depends relationships</param>
         /// <param name="recommends">Identifier prefix to find in mod recommends relationships</param>
         /// <param name="suggests">Identifier prefix to find in mod suggests relationships</param>
--- a/GUI/Model/ModSearch.cs
+++ b/GUI/Model/ModSearch.cs
@@ -32,9 +32,11 @@ namespace CKAN
             bool? upgradeable, bool? replaceable,
             string combined = null)
         {
-            Name         = CkanModule.nonAlphaNums.Replace(byName, "");
+            Name = (ShouldNegateTerm(byName, out string subName) ? "-" : "")
+                + CkanModule.nonAlphaNums.Replace(subName, "");
             initStringList(Authors, byAuthors);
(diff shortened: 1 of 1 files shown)


GUI/Model/ModSearch.cs
```diff
@@ -32,9 +32,11 @@ public sealed class ModSearch
             bool? upgradeable, bool? replaceable,
             string combined = null)
         {
-            Name         = CkanModule.nonAlphaNums.Replace(byName, "");
+            Name = (ShouldNegateTerm(byName, out string subName) ? "-" : "")
+                + CkanModule.nonAlphaNums.Replace(subName, "");
             initStringList(Authors, byAuthors);
-            Description  = CkanModule.nonAlphaNums.Replace(byDescription, "");
+            Description = (ShouldNegateTerm(byDescription, out string subDesc) ? "-" : "")
+                + CkanModule.nonAlphaNums.Replace(byDescription, "");
```
DasSkelett, line 39: While the regex replacement saves us here, I think this should technically be `CkanModule.nonAlphaNums.Replace(subDesc, "");

HebaruSan: Good catch! That was indeed the point of using `out string` there.

DasSkelett (approved):
Noticed one unused import and two code doc parameters not matching their respective actual parameter names.
I like it!

PR #3471, opened 2021-10-31
Use CRLF for resx files

### LINKED ISSUE #2847: [Bug] user.RaiseMessage does not properly handle newlines on GUI
Background
----------
* CKAN Version: v1.26.4
* KSP Version: N/A

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**: N/A

Problem
-------
Example: user.RaiseMessage in https://github.com/KSP-CKAN/CKAN/blob/56eded6390f3f63787c06ca9e9e01c2fabb43729/Cmdline/Action/Install.cs#L236-L247 renders as

`Oh no!It looks like you're trying to install a mod which is already installed,or which conflicts with another mod which is already installed.As a safety feature, the CKAN will *never* overwrite or alter a filethat it did not install itself.If you wish to install ExtraPlanetaryLaunchpads 6.6.1.0 via the CKAN,then please manually uninstall the mod which owns:Ships/SPH/Ore Scanner.craftand try again.`

**Steps to reproduce**
* Install a mod with a file that is already installed

**Expected behavior**
* user.RaiseMessage handles newlines properly

image

**CKAN error code (if applicable):**<details><summary></summary>
</details>

### LINKED ISSUE #3470: [Bug] Localized line breaks missing on Windows
Background
----------
* Operating System:  Windows 10
* CKAN Version:  v1.30.4
* KSP Version: 1.12.2

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** No. 

Problem
-------
Grammatical Error

**Steps to reproduce**
1. Start a removal of a mod
2. Have the removal of the mod fail

Lack of a space

image

```
PLACE CODE HERE !
```

HebaruSan:
## Problem

image

image

We do already have a line break there:

https://github.com/KSP-CKAN/CKAN/blob/41a0143b5cfb48b5767db306f045ae1640143e0a/GUI/Properties/Resources.resx#L272-L274

## Cause

[It seems that `.resx` files are supposed to use CRLF line endings](https://github.com/dotnet/msbuild/issues/2206), but we have them as normalized to LF in the repo and OS-specific in the working dir (CRLF when checked out on Windows, LF on Linux and Mac), so it would be LF when we build in an Ubuntu container:

https://github.com/KSP-CKAN/CKAN/blob/41a0143b5cfb48b5767db306f045ae1640143e0a/.gitattributes#L31

Refresher for `.gitattributes` usage:

- `text` means to treat the matching files as text format. LF will be stored to the server, LF will be used on Linux and Mac, CRLF will be used on Windows.
  - `text=auto` means to run git's built-in algorithm to check whether each matching file is in text format, then treat it as `text` if so
- `eol=lf` means to use LF everywhere
- `eol=crlf` means to use CRLF at checkout on all platforms, and who cares what's on the server because you can't see it

So the resource DLLs are built with LF in them, which Windows sometimes doesn't like when those strings get passed to GUI controls. Especially helpfully, if you check out CKAN on Windows and build it there, git will give you `.resx` files with CRLF and therefore you will get functionally different resource DLLs and possibly not be able to reproduce the problem!

## Changes

Now our `.gitattributes` file specifies `eol=crlf` for `.resx` files. This should make git convert these files' line endings to CRLF at checkout on all platforms, so our resource files will contain CRLF as intended in .NET.

As far as I can tell this does not break anything on Linux/Mono. I ran this:

```
find . -name \*.resx -print0 | xargs -0 unix2dos
git add --renormalize
```

... as every tutorial says you should do, and it converted all my local `.resx` files to CRLF. Nothing was staged, however, and I think that's because these files will still have LF on the server, so auto-normalizing the local files from CRLF to LF in the background results in no change to be committed.

Fixes #2847.
Fixes #3470.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -28,7 +28,7 @@
 *.config        text
 *.crt           text
 *.DotSettings   text
-*.resx          text
+*.resx          text eol=crlf
 *.settings      text
 
 # CKAN files


DasSkelett (approved):
I like the solution of setting the endings on checkout, in my experience trying to commit CRLF files and handle them in Git gets pretty messy

PR #3478, opened 2021-11-17
Properly determine the game when cloning instances

DasSkelett:
## Problem
A user reported a problem when trying to clone a KSP instance, they're getting a `Clone failed: Object reference not set to an instance of an object` error.
We couldn't quite figure out what the user was doing exactly, the messages were kinda vague:
> I remember I was messing around with the specific path, and I changed it, but the changes didn't show up, so I figured it hadn't saved my changes or something

But after looking at the clone UI logic, these lines stood out as a possible cause for a NRE:
https://github.com/KSP-CKAN/CKAN/blob/2564dfb3a0784cf82cedaf9cac1c0064a2b6b58a/GUI/Dialogs/CloneFakeGameDialog.cs#L149-L156

They were refactored in #3223 and didn't account for the possibility of a path being entered that didn't already belong to a known instance, which would result in a `null` `sourceInstance`.

## Changes
The underlying issue that forced this instance lookup was that we didn't have a good way to determine the game of the instance to clone.
In `GameInstanceManager.AddInstance()` we already had a basic layout for this, but it was missing the option to prompt the user to choose between multiple games if it detected multiple potential ones.

This logic is now moved into a new `DetermineGame()` method, which calls User.`RaiseDelectionDialog` in the aforementioned case.

Both the CloneFake dialog and the AddInstance button use the new `DetermineGame()` method.

For testing you can simply copy `KerbalSpaceProgram.cs`, rename the file and class, and add it to `GameInstanceManager.knownGames` to force multiple games to be detected.

### DIFF
--- a/Core/GameInstanceManager.cs
+++ b/Core/GameInstanceManager.cs
@@ -623,7 +623,7 @@ namespace CKAN
                 default:
                     // Prompt user to choose
                     int selection = user.RaiseSelectionDialog(
-                        $"Please select the game that is installed at {path}",
+                        $"Please select the game that is installed at {path.FullName.Replace('/', Path.DirectorySeparatorChar)}",
                         matchingGames.Select(g => g.ShortName).ToArray());
                     return selection >= 0 ? matchingGames[selection] : null;
             }
--- a/Core/GameInstanceManager.cs
+++ b/Core/GameInstanceManager.cs
@@ -206,7 +206,7 @@ namespace CKAN
         {
             var game = DetermineGame(new DirectoryInfo(path), user);
             if (game == null)
-                throw new NotKSPDirKraken(path);
+                return null;
 
             return AddInstance(new GameInstance(game, path, name, user));
         }
@@ -608,14 +608,15 @@ namespace CKAN
         /// </summary>
         /// <param name="path">A DirectoryInfo of the path to check</param>
         /// <param name="user">IUser object for interaction</param>
-        /// <returns>An instance of the matching game, or null if none could be found</returns>
+        /// <returns>An instance of the matching game or null if the user cancelled</returns>
+        /// <exception cref="NotKSPDirKraken">Thrown when no games found</exception>
         public IGame DetermineGame(DirectoryInfo path, IUser user)
         {
             var matchingGames = knownGames.Where(g => g.GameInFolder(path)).ToList();
             switch (matchingGames.Count)
             {
                 case 0:
-                    return null;
+                    throw new NotKSPDirKraken(path.FullName);
 
                 case 1:
                     return matchingGames.First();
--- a/Core/GameInstanceManager.cs
+++ b/Core/GameInstanceManager.cs
@@ -22,6 +22,11 @@ namespace CKAN
             new KerbalSpaceProgram()
         };
 
+        /// <summary>
+        /// An IUser object for user interaction.
+        /// It is initialized during the startup with a ConsoleUser,
+        /// do not use in functions that could be called by the GUI.
+        /// </summary>
         public IUser User { get; set; }
         public IConfiguration Configuration { get; set; }
         public GameInstance CurrentInstance { get; set; }
@@ -69,7 +74,7 @@ namespace CKAN
         }
 
         /// <summary>
-        /// Returns the preferred KSP instance, or null if none can be found.
+        /// Returns the preferred game instance, or null if none can be found.
         ///
         /// This works by checking to see if we're in a KSP dir first, then the
         /// config for an autostart instance, then will try to auto-populate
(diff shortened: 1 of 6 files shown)


HebaruSan:
I'm seeing GUI use the CmdLine prompt:

image

This seems vaguely familiar. I think the `GameInstanceManager` is created before GUI is launched to handle the `--instance` and `--gamedir` command line options.

DasSkelett:
Ah, that's why it's passed to `AddInstance()`. So going to pass it to `DetermineGame()` as well, and try to leave a comment for the `User` field, maybe it helps the next time.

DasSkelett:
I just got the idea that we could reassign the `User` property of the `GameInstanceManager` at start-up of the GUI instead, and eliminate the problem once and for all.
Do you see any pitfall that we could hit when we do this?

HebaruSan:
Maybe this would no longer work?

```
$ ckan prompt
Welcome to CKAN!

To get help, type help and press enter.

CKAN v1.30.5> gui
(close the GUI to get back to the prompt)
CKAN v1.30.5: KSP 1.12.2.3167 (Steam)> instance clone Steam newname c:\test
```

I think `prompt` uses the same instance manager all the way through, so changing its `IUser` object would affect future calls after GUI closes.

HebaruSan:
Maybe the instance manager's `IUser` reference should be removed and replaced with parameters for each function that needs it. Then the calling code would be responsible or supplying the right one.

HebaruSan:
Oh, there's another quirk: that class passes its `User` to `GameInstance`, so they would all have old references from before GUI's override, for whatever it is they do with `IUser`.

HebaruSan:
Sorry for the delay, I didn't notice the latest update until now and I don't think GitHub sent me a notification. The prompt now appears in GUI as intended:

image

Can we format the path with the OS's path separator?

GUI/Dialogs/CloneFakeGameDialog.cs
```diff
@@ -144,13 +145,17 @@ private async void buttonOK_Click(object sender, EventArgs e)
...
                 {
+                    IGame guessedGame = manager.DetermineGame(new DirectoryInfo(existingPath), user);
                     await Task.Run(() =>
                     {
-                        GameInstance sourceInstance = manager.Instances.Values
-                            .FirstOrDefault(i => i.GameDir() == textBoxClonePath.Text);
+                        if (guessedGame == null)
+                        {
+                            throw new NotKSPDirKraken(existingPath);
+                        }
```
HebaruSan, line 154: This can be null not just because there's a problem with the path, but also if the user clicks Cancel in the game selection popup. In that case, the instance is valid but we say it's not.

HebaruSan: Reworked this to throw `NotKSPDirKraken` from inside `DetermineGame` when we know it's true, then treat null only as cancellation and skip the error message.

DasSkelett: Yes, I kept it like this because the user will know that they cancelled, so they will understand why they got the message, even if it isn't 100% accurate.

I explicitly avoided throwing an exception inside `DetermineGame`, because we already return `null` in the bad case. Mixing the two isn't good code style IMHO, you'd expect _either_ null for errors _or_ exceptions. Being aware that we commonly overlook the possibility of exceptions being thrown and don't catch them, leading to uncaught ones crashing the whole client, I did decide for `null` here.
If you want to switch to exceptions, please also throw one for the "User cancelled" case, and never return `null`.

Core/GameInstanceManager.cs
```diff
@@ -608,14 +608,14 @@ public static bool IsGameInstanceDir(DirectoryInfo path)
         /// </summary>
         /// <param name="path">A DirectoryInfo of the path to check</param>
         /// <param name="user">IUser object for interaction</param>
-        /// <returns>An instance of the matching game, or null if none could be found</returns>
+        /// <returns>An instance of the matching game, or null if the user cancelled, or throws NotKSPDirKraken if no games found</returns>
```
DasSkelett, line 611: C# uses `exception` tags to document possible exceptions.
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#recommended-tags
Like here:
https://github.com/KSP-CKAN/CKAN/blob/2f55a1e67801f05a5fd612de2db5a4db29e23862/Core/GameInstanceManager.cs#L220-L224

HebaruSan: OK, changed to use that syntax. :+1:

GUI/Dialogs/CloneFakeGameDialog.cs
HebaruSan: The user cancelling isn't a bad case or an error, though, it's a normal flow that is available for the user to choose. Whereas the folder not having a game in it is an error.

HebaruSan:
One more update, it's weird that the instance list shows that we know each instance's game, but then when we go to clone them we may prompt the user. Pushed a commit to make the prompting a fallback if we can't get the game from the instance (i.e., if the paths don't match).

HebaruSan (approved):
This looks good to me now, but in case you want to make some more changes, I'll wait a couple of days before merging it.

PR #3479, opened 2021-11-17
Case insensitive installed file lookup on Windows

HebaruSan:
## Problems

Recently, Windows users have been encountering weird install/uninstall errors relating to Scatterer:

screenshot1
screenshot2
screenshot3

## Causes

In its [latest release](https://spacedock.info/mod/141/scatterer/download/0.0825b), Scatterer renamed its install folder from `scatterer` to `Scatterer` (capitalizing the S), but the metadata was unchanged until KSP-CKAN/NetKAN#8862. There was no inflation error because install stanza matching is case insensitive, so `find: scatterer` matched a path in the ZIP of `Scatterer`:

https://github.com/KSP-CKAN/CKAN/blob/35927bcc70dd3febb62f34cbb0e2bbb5ab24bbee/Core/Types/ModuleInstallDescriptor.cs#L230-L245

One user who encountered these errors shared a screenshot confirming that the mod was installed into `GameData/scatterer`:

GameData

But their `registry.json` has it as `GameData/Scatterer` in the `installed_files` hashes:

https://1drv.ms/u/s!AlCbO5GR6RGwj7RlQn2vSUdicVFNQA?e=pJQfMM

On Windows, filenames are treated as case insensitive, so if we take a filename with mismatched capitalization from the registry and look for it on disk, it will be found. But `Dictionary<string, whatever>` lookups are case sensitive by default, so if we take a filename with mismatched capitalization from disk and look for it in the registry, it won't be found unless we take extra steps to make that happen.

One place this happens is when we look for manually installed DLLs. This is done by iterating over all the DLLs in GameData and checking whether each one is registered to an installed module. If the capitalization is mismatched, then that code does not find the file and adds it to the list of manually installed DLLs. Later the `RelationshipResolver` will trip on Scatterer's self-conflict when it sees it in the manually installed DLL list and emit the above strange errors.

## Changes

Now on Windows we initialize the `installed_file` dictionaries with `StringComparer.OrdinalIgnoreCase`, which makes their key lookup case insensitive. This should allow us to find files properly when going from disk to the registry, so CKAN will no longer think that `Scatterer.dll` is manually installed.

### DIFF
--- a/Core/Registry/Registry.cs
+++ b/Core/Registry/Registry.cs
@@ -38,7 +38,7 @@ namespace CKAN
         // name => path
         [JsonProperty] private  Dictionary<string, string>          installed_dlls;
         [JsonProperty] private  Dictionary<string, InstalledModule> installed_modules;
-        // filename => module
+        // filename (case insensitive on Windows) => module
         [JsonProperty] private  Dictionary<string, string>          installed_files;
 
         [JsonProperty] public readonly SortedDictionary<string, int> download_counts = new SortedDictionary<string, int>();
@@ -157,7 +157,10 @@ namespace CKAN
             {
                 log.Warn("Older registry format detected, normalising paths...");
 
-                var normalised_installed_files = new Dictionary<string, string>();
+                // We need case insensitive path matching on Windows
+                var normalised_installed_files = Platform.IsWindows
+                    ? new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+                    : new Dictionary<string, string>();
 
                 foreach (KeyValuePair<string,string> tuple in installed_files)
                 {
@@ -189,6 +192,12 @@ namespace CKAN
                 // because that needs a registry, and we chicken-egg.)
 
                 log.Warn("Registry upgrade complete");
+            }
+            else if (Platform.IsWindows)
+            {
+                // We need case insensitive path matching on Windows
+                // (already done when replacing this object in the above block, hence the 'else')
+                installed_files = new Dictionary<string, string>(installed_files, StringComparer.OrdinalIgnoreCase);
             }
 
             // Fix control lock, which previously was indexed with an invalid identifier.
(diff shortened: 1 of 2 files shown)


Core/Registry/Registry.cs
```diff
@@ -231,6 +234,12 @@ private void DeSerialisationFixes(StreamingContext context)
                     ReindexInstalled();
                 }
             }
+            else if (Platform.IsWindows)
+            {
+                // We need case insensitive path matching on Windows
+                // (already done when replacing this object in the above blocks, hence the 'else')
+                installed_files = new Dictionary<string, string>(installed_files, StringComparer.OrdinalIgnoreCase);
+            }
```
DasSkelett, line 242: Shouldn't this be after the `if (registry_version == 0)` block? Like this we're skipping the fix with `registry_version` `1`, no?

HebaruSan: Ahh yes, I misremembered which block the `StringComparer.OrdinalIgnoreCase` change was in, will move it...

HebaruSan: No, wait, I had it right originally. The `registry_version < 2` block calls `ReindexInstalled`, which takes care of setting up the case insensitivity. Reverting...

DasSkelett (approved):
Did some testing on Windows now, both the metadata changes and this PR solve it on their own, i.e. either updating the metadata or using the updated client are enough to mitigate it, both together obviously works as well.
And I realized that there were more Scatterer releases since the folder name changed, but they have been deleted from SpaceDock. So I'll do some backporting of https://github.com/KSP-CKAN/NetKAN/pull/8862 just to be safe

HebaruSan:
> both the metadata changes and this PR solve it on their own, i.e. either updating the metadata or using the updated client are enough to mitigate it

Excellent news, that means we don't need to rush the next release. :+1:

PR #3481, opened 2021-11-19
Add Release Date column to GUI modinfo versions list

### LINKED ISSUE #3480: [Feature] Add release date to version tab
Problem
-------
Often release dates by version would make it easier to mentally correlate when a symptom began showing up in the game and when a version was installed, even if more than one update has occurred since the symptom began appearing

Suggestions
-----------
Put the release date as a column in the list of versions in the versions tab in the lower right when a mod is selected in the main window.  Put on far right so one only occasionally has to horiz scroll to see it as one would probably only need to see it occasionally

HebaruSan:
## Motivation

image

## Changes

image

The existing translations of the Release date column header in the mod grid are copied to the new column. Also changed the English capitalization of these column headers to match the mod grid (sentence capitalization instead of title capitalization) and removed "Compatible" to save space.

Fixes #3480.

### DIFF
--- a/GUI/Controls/AllModVersions.resx
+++ b/GUI/Controls/AllModVersions.resx
@@ -118,8 +118,9 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="label1.Text" xml:space="preserve"><value>All available versions of selected mod</value></data>
-  <data name="ModVersion.Text" xml:space="preserve"><value>Mod Version</value></data>
-  <data name="CompatibleGameVersion.Text" xml:space="preserve"><value>Compatible Game Versions</value></data>
+  <data name="ModVersion.Text" xml:space="preserve"><value>Mod version</value></data>
+  <data name="CompatibleGameVersion.Text" xml:space="preserve"><value>Game versions</value></data>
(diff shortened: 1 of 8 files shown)


GUI/Controls/AllModVersions.resx
```diff
@@ -118,8 +118,9 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="label1.Text" xml:space="preserve"><value>All available versions of selected mod</value></data>
-  <data name="ModVersion.Text" xml:space="preserve"><value>Mod Version</value></data>
-  <data name="CompatibleGameVersion.Text" xml:space="preserve"><value>Compatible Game Versions</value></data>
+  <data name="ModVersion.Text" xml:space="preserve"><value>Mod version</value></data>
+  <data name="CompatibleGameVersion.Text" xml:space="preserve"><value>Compatible game versions</value></data>
```
DasSkelett, line 122: Should we get rid of the word `Compatible` here, to make the string a bit shorter? It's kind of implied by the fact that it's a mod version->game version mapping, and should be rather obvious from looking at the table.

That would relax the horizontal real estate situation a bit.

HebaruSan: Done!

DasSkelett (approved):
Also shortened the column header for other languages where I could figure it out. (and apparently my IDE removed a bunch of trailing spaces)
Neat addition!

PR #3489, opened 2021-11-28
Theme all buttons, checkboxes, groupboxes, and listboxes

HebaruSan:
## Background

When Windows users ask for a dark theme, our answer has been to configure the WinForms system colors to be dark because the CKAN GUI obeys those colors. However, Microsoft decided to make these colors uneditable in recent versions of Windows.

One of my long-term TODO items has been to create a `DarkTheme.reg` file so Windows users could configure a dark theme even when their operating system does not allow them to access these settings. (I think I had previously tried to create a `.theme` file, but it didn't go very well.)

## Problems

I started working on that reg file, and I made some progress, but some of the buttons, checkboxes, groupboxes, and listviews could not be fixed by tweaking the colors:

<details><summary>Expand for "before" screenshots</summary>

image

image

image

image

image

image

image

image

</details>

## Causes

- [`GroupBox` seems to set its text color to black regardless of the system colors](https://developercommunity.visualstudio.com/t/groupbox-forecolor-text-is-black-when-forecolor-pr/451637)
- See #2937, `ListView` requires some tweaking to look good, and these were left out of the previous pass of that
- If `Button.FlatStyle` and `CheckBox.FlatStyle` aren't set, then buttons and checkboxes are rendered with a garish "3D" style that obeys the system color for the text but ignores it for the background

## Changes

- Now we set `GroupBox.ForeColor` to `SystemColors.ControlText`
- Now our `ThemedListView` control replaces the few remaining plain `ListView`s
- Now we set `FlatStyle` for all of our buttons and checkboxes
- Now the `GUI/DarkTheme.reg` file in the CKAN repo contains dark theming registry keys for Windows. It is a text file with 2 bytes per character that `regedit.exe` uses for exports, I think UTF-16 format? Once this is merged, we can add a link to this file on the wiki/forum/Discord, and users clicking it would be prompted like this:
  image
  After they click Yes, restarting Windows will load the dark theme. It would be nice to host the file itself on the wiki as well, but the wiki can only host plain text content with no file extension, so adding it to the repo is the next best option.
- The visible file path in the compatible game versions window now uses the OS's path separator instead of using `/` on Windows

<details><summary>Expand for "after" screenshots</summary>

image

image

image

image

image

image

image

image

</details>


HebaruSan:
Just noticed the checkboxes on Edit Labels aren't flat. One moment, please.

DasSkelett:
Nice work!
Is there an easy way to revert the color changes in the registry again? I think that would be neat to have, some users might not like the colors (different tastes etc. etc.) and wish to go back to the default colors.

HebaruSan:
> Is there an easy way to revert the color changes in the registry again? I think that would be neat to have, some users might not like the colors (different tastes etc. etc.) and wish to go back to the default colors.

Good point. I reverted my color settings to a saved version from before I started this and captured the same keys in `GUI/DefaultTheme.reg`, so we can provide a link to that as the "undo" button. In the process I noticed that none of the keys in the second group (`HKCU\Control Panel\Desktop\Colors`) are changed back even though my colors are back to normal, so I removed those from `DarkTheme.reg` as well.

DasSkelett (approved):
Code changes and Linux testing look good

PR #3492, opened 2021-12-02
Sort Netkan warning lists

HebaruSan:
## Problem

It was a surprise to see this for an established mod that probably isn't changing much from one release to the next:

image

It seems like the warning should remain the same and therefore not be sent to Discord.

## Cause

Same files, different order. The order of the list in the message depends on the order in the ZIP file, so if a succession of releases has the same warnings but the ZIP is created slightly differently, we get extra warning messages sent to Discord:

image

This could happen for any warning that contains a list of multiple things.

## Changes

Now the following messages are sorted alphabetically. This will probably cause a one-time cascade of new warnings sent to Discord as they all switch over to the proper sorted order:

- Multiple version files
- Craft files in wrong path
- Found copies of the Harmony DLL
- Paths with GameData contained in GameData
- Attempts to install same paths multiple times
- Config files using ModuleManager syntax without ModuleManager depends
- DLLs not matching the identifier for detecting manual installs

Also the `localizations` property will now be in alphabetical order. This might cause a large number of one-time re-indexings.

### DIFF
--- a/Netkan/Services/ModuleService.cs
+++ b/Netkan/Services/ModuleService.cs
@@ -271,7 +271,7 @@ namespace CKAN.NetKAN.Services
             {
                 throw new Kraken(
                     string.Format("Too many .version files located: {0}",
-                              string.Join(", ", files.Select(x => x.Name))));
+                              string.Join(", ", files.Select(x => x.Name).OrderBy(f => f))));
             }
             else
             {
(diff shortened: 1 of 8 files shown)


DasSkelett (approved):
I would've never thought that files in zips could be unsorted :smile: 
But it makes sense, it just packs whatever you pass it.

Sorted localizations make a lot of sense as well (a bit unrelated: we should consider sorting the list in our GUI language selection dropdown as well)

PR #3505, opened 2022-01-04
Enforce a few more spec version requirements

HebaruSan:
#3494 and KSP-CKAN/NetKAN#8943 were fun, so I double checked the spec and found a few more requirements we aren't checking yet:

- Installing to Scenarios requires spec version v1.14
- Installing to Missions requires spec version v1.25
- `licenses` as an array requires spec version v1.8
- `supports` requires spec version v1.2


DasSkelett (approved):
:+1:

Issue #1875, opened 2016-08-25
CKAN Error in Regards to Sorting


DonKalypso:
Thanks for wanting to bring an issue to our attention!

Before you open a ticket, please do a quick search of the existing issues to see if it's already been brought up.

**CKAN Version:** 
1.18.1-0-g0d2c4f0 (beta)
**Operating System:** 
Windows 10
**The issue you are experiencing:** 
On sorting by Max KSP Version, CKAN will spit the following error: http://pastebin.com/QPuCpckw
**How to recreate this issue:** 
Sort by Max KSP Version, not replicating in --debug.
**CKAN error codes (if applicable):** 
http://pastebin.com/QPuCpckw

DonKalypso:
I've no idea why, but now it's just not at all replicating.

If no one else can replicate this problem, then I'll chalk it up to my comp being haunted by the angry ghosts of a thousand long dead, 1950s era programmers getting their revenge on me for looking at porn or something.

ayan4m1:
You were likely experiencing this issue: https://github.com/KSP-CKAN/CKAN/pull/1866

Were you sorting by the "Max KSP Version" column before this exception occurred?

DonKalypso:
Before the error began happening, I would keep it on sorted by "Max KSP Version," yes.

Issue #1803, opened 2016-06-27
Consistent crash when installing up updating any mod


maekern:
CKAN version 1.18.0-0, Windows 10

When attempting to upgrade or install any mods, the status window shows at least a couple of them downloading, and then CKAN throws an exception. Here is the data from that:

````
See the end of this message for details on invoking 
just-in-time (JIT) debugging instead of this dialog box.

************** Exception Text **************
System.NullReferenceException: Object reference not set to an instance of an object.
   at CKAN.Main.DefaultSorter(DataGridViewRow row)
   at System.Linq.EnumerableSorter`2.ComputeKeys(TElement[] elements, Int32 count)
   at System.Linq.EnumerableSorter`1.Sort(TElement[] elements, Int32 count)
   at System.Linq.OrderedEnumerable`1.<GetEnumerator>d__1.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at CKAN.Main._UpdateFilters()
   at CKAN.Util.Invoke[T](T obj, Action action)
   at CKAN.Main._UpdateModsList(Boolean repo_updated)
   at CKAN.Util.Invoke[T](T obj, Action action)
   at CKAN.Main.PostInstallMods(Object sender, RunWorkerCompletedEventArgs e)
   at System.ComponentModel.BackgroundWorker.OnRunWorkerCompleted(RunWorkerCompletedEventArgs e)


************** Loaded Assemblies **************
mscorlib
    Assembly Version: 4.0.0.0
    Win32 Version: 4.6.1080.0 built by: NETFXREL3STAGE
    CodeBase: file:///C:/Windows/Microsoft.NET/Framework64/v4.0.30319/mscorlib.dll
----------------------------------------
ckan
    Assembly Version: 0.0.0.0
    Win32 Version: 0.0.0.0
    CodeBase: file:///D:/Program%20Files%20(x86)/Steam/SteamApps/common/Kerbal%20Space%20Program/CKAN.exe
----------------------------------------
System.Configuration
    Assembly Version: 4.0.0.0
    Win32 Version: 4.6.1038.0 built by: NETFXREL2
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Configuration/v4.0_4.0.0.0__b03f5f7f11d50a3a/System.Configuration.dll
----------------------------------------
System
[...]

PR #3509, opened 2022-01-07
Rename GH1866 test, fix invalid char test, fix equality assertion order

HebaruSan:
## Problems

- The `ZipValid_ContainsFilenameWithBadChars_NoException` test is failing on Windows:

  ```
  1) Failed : Tests.Core.Cache.ZipValid_ContainsFilenameWithBadChars_NoException
    Expected string length 128 but was 27. Strings differ at index 0.
    Expected: "Error in step EntryHeader for GameData/FlagPack/Flags/Weyland..."
    But was:  "Illegal characters in path."
    -----------^
     at Tests.Core.Cache.ZipValid_ContainsFilenameWithBadChars_NoException() in C:\Users\User\github\CKAN\Tests\Core\Cache.cs:line 202
  ```
- The above error message has swapped the expected and actual values of the string
- The `GH1866` test does not have a meaningful name. Every time I see it, I wonder what it does. It's hard to maintain it correctly when we are guessing about its purpose.

## Causes

- I think `ZipValid`'s error format has changed over time, and the test has an old value
- This test and a few others have the parameters to `Assert.AreEqual` in the wrong order, with the actual value first and the expected value second

## Changes

- The invalid chars test is updated to use the current error messgae, so it will succeed on Windows
- The parameters of all `Assert.AreEqual` calls are expected first, actual second
- The `GH1866` test is moved to `Tests.GUI.ModList.InstallAndSortByCompat_WithAnyCompat_NoCrash`, based on descriptions in #1803, #1875, #1866, and #1882
  - It is now self-contained in one function rather than distributed around an entire class (private variables become locals, all steps in sequence)
  - I have attempted to organize the code into Arrange / Act / Assert sections

I might self-review this, since it's just tests and shouldn't affect users.


DasSkelett (approved):
Yup, the name of the test was weird. On the upside, you still knew exactly where to look for more information ^^
But the links do the job even better.

techman83:
As the one who likely named the test :100:

PR #3512, opened 2022-01-11
Disable tx timeouts, add tx debug logging, static DLL pattern, fix docs

HebaruSan:
## Problems

I've been investigating several transaction-related issues lately.

- Something about trying to enlist in multiple transactions (see #3502)
- The manually installed DLL scan, which happens in a tx (see #3506)
  - `Registry.DllPattern` is compiled for every call to `RegisterDll`, which is not great because this regex will be evaluated on every DLL in GameData
  - A comment says the regex `works great for things like GameData/Foo/Foo-1.2.dll`, but that's wrong, it's for `GameData/Foo/Foo.1.2.dll`, the former would detect the identifier as `Foo-1`
  - `EnlistTransaction` is called in `RegisterDll` even if one of its validation checks decides we don't need to make any changes
- Cloning an instance can throw an exception if it takes more than 10 minutes
- Installing mods can throw the same exception if it takes more than 10 minutes

```
Unhandled exception:
System.Transactions.TransactionException: Failed to roll back. ---> System.IO.IOException: The directory is not empty.

   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.Directory.DeleteHelper(String fullPath, String userPath, Boolean recursive, Boolean throwOnTopLevelDirectoryNotFound, WIN32_FIND_DATA& data)
   at System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
   at ChinhDo.Transactions.FileManager.Operations.CreateDirectory.Rollback()
   at ChinhDo.Transactions.FileManager.TxEnlistment.Rollback(Enlistment enlistment)
   --- End of inner exception stack trace ---
   at ChinhDo.Transactions.FileManager.TxEnlistment.Rollback(Enlistment enlistment)
   at System.Transactions.VolatileEnlistmentAborting.EnterState(InternalEnlistment enlistment)
   at System.Transactions.TransactionStateAborted.EnterState(InternalTransaction tx)
   at System.Transactions.EnlistableStates.Timeout(InternalTransaction tx)
   at System.Transactions.Bucket.TimeoutTransactions()
   at System.Transactions.BucketSet.TimeoutTransactions()
   at System.Transactions.TransactionTable.ThreadTimer(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[...]

PR #3514, opened 2022-01-12
Rewrap ConsoleUI textbox for scrollbar and resize

HebaruSan:
## Problems

- ConsoleUI's textbox (used on the progress screen and mod info) wraps lines longer than its width at word boundaries, but unbroken words longer than its width (e.g. URLs) underlap the scrollbar if there is one, hiding one character from view
- If you resize ConsoleUI and press <kbd>Ctrl+L</kbd>, the text within the textbox is still formatted according to its previous layout, which may leave big gaps or allow the text to overflow the box

Noticed while testing RP-1 express install in ConsoleUI for #3512, because [a Mac user asked about RP-1 on the forum](https://forum.kerbalspaceprogram.com/index.php?/topic/197082-ckan-the-comprehensive-kerbal-archive-network-v1304-hubble/&do=findComment&comment=4075886).

## Causes

`ConsoleTextBox` performs word wrap on each line as it is added, using the full width without the scrollbar. Words longer than the textbox width are split at the width of the textbox at the time they are added. If that width later changes, by either the addition of a scrollbar or the resizing of the window, the old wrapping is retained as-is.

## Changes

Now we store the original full strings in `lines` and the word-wrapped strings in `displayLines`. `AddLine` appends its raw input to the former and word-wraps it into the latter. A scrollbar-aware `rewrapLines()` function re-does this word wrapping as needed, and we call this function when we first add the scrollbar as well as anytime the textbox size changes. This will ensure that the scrollbar never overwrites characters, and the word wrap always fits the box properly.

### DIFF
--- a/ConsoleUI/Toolkit/ConsoleTextBox.cs
+++ b/ConsoleUI/Toolkit/ConsoleTextBox.cs
@@ -31,6 +31,7 @@ namespace CKAN.ConsoleUI.Toolkit {
             align          = ta;
             getFgColor     = fgFunc;
             getBgColor     = bgFunc;
+            prevTextW      = r - l + 1;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


DasSkelett (approved):
Funny _edge_ cases we have there :P
Tested with CameraToolsContinued, which has a URL in the description, and while I had to downscale the terminal window to size where it would be unusable anyway, it did display the wrapped but complete.

PR #3515, opened 2022-01-16
Tab completion for `ckan prompt`

### LINKED ISSUE #888: Feature Request: Tab Completion on the CLI for mods in database
_Saturday Dec 27, 2014 at 12:22 GMT # Sample: Friday Sep 13, 2013 at 22:58 GMT_
_Originally opened as https://github.com/KSP-CKAN/CKAN-support/issues/25_

---

_From @Felger on November 30, 2014 19:40_

Typing out the full mod name every time can get tiring, and I'm lazy.  CKAN should finish typing the name for me when I get most of the way there so I can stop typing and let my tired fingers rest.

Examples:

```
Re->"tab"->Real->"i""tab"->RealismOverhaul  
ckan install RealismOverhaul RSS -> "tab" -> ckan install RealismOverhaul RSSTextures
```

And so on.  This would be a nice addition to an already amazing tool!

_Copied from original issue: KSP-CKAN/CKAN#491_

HebaruSan:
## Motivation

Currently with CmdLine you have to type everything, including all the commands and subcommands and the full identifier of every mod. This is inconvenient.

## Changes

Now if you run `ckan prompt`, you can use tab completion for commands/subcommands, `--` options, module identifiers, instance names:

image

Press tab:

image

Type part of an identifier, case insensitive:

image

Press tab:

image

To make this work:

- A project reference for `ReadLine` 1.2.0 is added to CmdLine (more recent versions require .NET Core)
- All of the subcommands' option classes are now publicly available and used in `Options.cs` to describe the options of the corresponding subcommands
- New attributes `[AvailableIdentifiers]`, `[InstalledIdentifiers]`, and `[GameInstances]` are defined and used to tag options that use each kind of input
- `Prompt` uses `ReadLine.Read` instead of `Console.ReadLine`, and we set `ReadLine.AutoCompletionHandler` to a function that inspects the command line option classes with reflection to determine what strings are available

As a side effect, `ckan prompt` has also gained many standard terminal editing hotkeys like ctrl-W to delete a word, ctrl-U to delete to start of the line, ctrl-E to move cursor to end of line, etc.

Fixes #888.

### DIFF
--- a/Cmdline/Action/Prompt.cs
+++ b/Cmdline/Action/Prompt.cs
@@ -1,4 +1,8 @@
 using System;
+using System.Reflection;
+using System.Linq;
+using System.Collections.Generic;
+
 using CommandLine;
 using CommandLine.Text;
 using log4net;
@@ -8,9 +12,12 @@ namespace CKAN.CmdLine
 
     public class Prompt
     {
-        public Prompt() { }
+        public Prompt(GameInstanceManager mgr)
+        {
+            manager = mgr;
+        }
 
-        public int RunCommand(GameInstanceManager manager, object raw_options)
+        public int RunCommand(object raw_options)
         {
             CommonOptions opts = raw_options as CommonOptions;
             // Print an intro if not in headless mode
@@ -21,6 +28,7 @@ namespace CKAN.CmdLine
                 Console.WriteLine("To get help, type help and press enter.");
                 Console.WriteLine("");
             }
+            ReadLine.AutoCompletionHandler = GetSuggestions;
             bool done = false;
             while (!done)
             {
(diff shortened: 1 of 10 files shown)


Cmdline/Action/Prompt.cs
```diff
@@ -55,6 +63,121 @@ public int RunCommand(GameInstanceManager manager, object raw_options)
...
+
+        private bool WantsAvailIdentifiers(TypeInfo ti)
+        {
+            return ti.DeclaredProperties
+                .Any(p => p.GetCustomAttribute<AvailableIdentifiersAttribute>() != null);
+        }
+
+        private string[] GetAvailIdentifiers(string prefix)
+        {
+            CKAN.GameInstance inst = MainClass.GetGameInstance(manager);
```
DasSkelett, line 140: This throws a `NoGameInstanceKraken` if there is no default instance and also non selected with `--instance` / `--gamedir`, and makes the prompt close and exit with code 1.
```
CKAN v1.30.5> install AsI don't know where a game instance is installed.
Use 'ckan instance help' for assistance in setting this.

Process finished with exit code 1.
```

We should catch it and just write the message

HebaruSan: Good catch, `upgrade` has the same problem through a different code path. I'm going to look for a solution that handles both `GetAvailIdentifiers` and `GetInstIdentifiers` consistently...

HebaruSan: A couple of findings...

- If we just `Console.WriteLine` in the completion handler and return an empty array, it messes up the display (at least on Windows); the `ReadLine` code doesn't know that the cursor has moved, so the user is still editing the string but can't see it
- There doesn't seem to be any way to tell `ReadLine` that this has happened
- Given the above, the best option seems to be to catch the exception after it interrupts the `ReadLine` call, so we can start a new `ReadLine` call after that with fresh state
- If we do that, we can't restore the previously entered comment, even if we capture it via the exceptions; there's no way to pass it back to `ReadLine` as the initial command to edit. However, maybe that's desirable, as there is no way to finish an `install` command successfully if you have no game instance

HebaruSan: Pushed a commit that should improve this:

image

techman83 (approved):
I don't see any standout issues, and looks to be well scoped to the ckan prompt. Assuming @DasSkelett comments are resolved, it all seems relatively sensible to me.

HebaruSan:
Rebased and made completion work for the `filter` command.

Issue #3410, opened 2021-07-08
CKAN not working at start


ThatUnreal:
**Background**
----------
* Operating System: Windows
* CKAN Version: latest version
* KSP Version: 1.12, steam

Have you made any manual changes to your GameData folder (i.e., not via CKAN)?
Not yet

Problem
-------
I installed CKAN from the Github page and opened it up, the first time it opened it showed me the update page (no screenshot for that), but than I noticed that something was wrong and since I couldn't understand why I deleted it and reinstalled it, but than it completely stop working, giving me two error pages, one with code and one with the list of errors

**Expected behavior**
CKAN opening

**Screenshots (if applicable)**
![ckan error 1](https://user-images.githubusercontent.com/87145454/124969067-8fbf7380-e026-11eb-835a-cce6b0b4d601.png)
![ckan error 2](https://user-images.githubusercontent.com/87145454/124969087-95b55480-e026-11eb-9b67-cfa44db080a5.png)
![ckan error 3](https://user-images.githubusercontent.com/87145454/124969112-9bab3580-e026-11eb-83e2-90887e972949.png)

DasSkelett:
Looks like you grabbed an older release when reinstalling, this error happens after downgrading to a pre-v1.30.0 release.

Make sure to remove each and every ckan.exe you have on your PC, then download the [latest](https://github.com/KSP-CKAN/CKAN/releases/latest) release again.

DasSkelett:
Btw. this is why putting `* CKAN Version: latest version` into the issue body is pretty useless. We don't know what you think the latest version is. Instead, check what it saysin the window title bar, and put this exact version into the issue. This is guaratneed to be the one you are running.

ThatUnreal:
> Btw. this is why putting `* CKAN Version: latest version` into the issue body is pretty useless. We don't know what you think the latest version is. Instead, check what it saysin the window title bar, and put this exact version into the issue. This is guaratneed to be the one you are running.

Ok sorry about that

HebaruSan:
Looks like the cause of this was figured out. @ThatUnreal, if you still have this problem on the actual latest version of CKAN, let us know and we can re-open for investigation.

HebaruSan:
Duplicate of #3331

Issue #3331, opened 2021-03-27
CKAN.BadMetadataKraken: The license MPL-2.0 is invalid


typicxl11:
Background
----------
* Operating System: Unknown
* CKAN Version: v1.30.0 previously, then some unknown older version
* KSP Version: Unknown

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
Yes

Problem
-------

Can't reinstall CKAN, Refuses to open and gives me an error page saying

```
Unhandled exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
---> CKAN.BadMetadataKraken: The license MPL-2.0 is invalid
    at CKAN.License..ctor(String license)
    --- End of inner exception stack trace ---
```

. Im operating on a old gaming laptop that can still fluently run the game. CKAN And KSP Are up to date. I Have messed with the Gamedata folder before unistalling due to weird things happening within CKAN. The day before both KSP And CKAN We're operating fine. need help!

typicxl11:
CKAN Refuses to open and gives me an error page saying ; Unhandled exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> CKAN.BadMetadataKraken: The license MPL-2.0 is invalid    at CKAN.License..ctor(String license)    --- End of inner exception stack trace ---. Im operating on a old gaming laptop that can still fluently run the game. CKAN And KSP Are up to date. I Have messed with the Gamedata folder before unistalling due to weird things happening within CKAN. The day before both KSP And CKAN We're operating fine. After redownloading CKAN Itself wont open. needing a little help.

HebaruSan:
This means you updated the registry with CKAN v1.30.0 and then downgraded to an older version.
Use v1.30.0:

https://github.com/KSP-CKAN/CKAN/releases/tag/v1.30.0

HebaruSan:
Full stack trace in case we can find a way to make this more foolproof:

```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> CKAN.BadMetadataKraken: The license SomethingWeird is invalid
   at CKAN.License..ctor(String license)
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
[...]

PR #3526, opened 2022-01-26
Get licenses from embedded schema, skip bad modules in deserialize

HebaruSan:
## Problems

- Every time we add a new license, the previous version of the CKAN client is unable to read modules that use it. This is supposed to be handled by `spec_version`, but the user can bypass this protection by updating the registry with a new client and then running an old client, and they have actually done this (see #3331 and #3410 and a reply on the forum thread).
- We currently maintain the list of valid licenses in three places, which is annoying:
  - `CKAN/Core/Types/License.cs`
  - `CKAN/CKAN.schema`
  - `NetKAN-Infra/netkan/netkan/mirrorer.py`

## Changes

- Now if any `CkanModule` in the available modules list throws an exception during deserialization, including but not limited to an unsupported license, we log a warning and omit that module from its `AvailableModule` collection. This allows the user to keep using the client, albeit without access to such modules.
  - This is done with a new `JsonConverter` that adds each version/module pair to a `SortedDictionary` one at a time and returns it containing whatever values didn't throw.
- Now `CKAN.schema` is an embedded resource in Core, and `Licenses.cs` retrieves its license data from there.
  - A `CKANSchema` class holds a `schema` object generated by extracting the embedded resource and then parsing it with `NJsonSchema`
  - Netkan's schema validator now uses `CKANSchema.schema` instead of its own embedded resource
  - A `.definitions.redistributable_license` enum is added to the schema to document which licenses allow us to redistribute.

### DIFF
--- a/Core/Converters/JsonLeakySortedDictionaryConverter.cs
+++ b/Core/Converters/JsonLeakySortedDictionaryConverter.cs
@@ -1,0 +1,64 @@
+using System;
+using System.Collections.Generic;
+
+using log4net;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace CKAN
+{
+    /// <summary>
+    /// [De]serializes a dictionary that might have some questionably
+    /// valid data in it.
+    /// If exceptions are thrown for any key/value pair, leave it out.
+    /// Removes CkanModule objects from AvailableModule.module_version
+    /// if License throws BadMetadataKraken.
+    /// </summary>
+    public class JsonLeakySortedDictionaryConverter<K, V> : JsonConverter
+    {
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            var dict = new SortedDictionary<K, V>();
+            foreach (var kvp in JObject.Load(reader))
+            {
+                try
+                {
+                    dict.Add(
+                        (K)Activator.CreateInstance(typeof(K), kvp.Key),
+                        kvp.Value.ToObject<V>());
+                }
+                catch (Exception exc)
+                {
+                    log.Warn($"Failed to deserialize {kvp.Key}: {kvp.Value}", exc);
+                }
+            }
+            return dict;
+        }
+
+        /// <summary>
+        /// Use default serializer for writing
+        /// </summary>
+        public override bool CanWrite => false;
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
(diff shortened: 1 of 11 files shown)


HebaruSan:
Oops, I accidentally pressed Ctrl+Enter while composing the description. Editing now...

Core/Converters/JsonLeakySortedDictionaryConverter.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                try
+                {
+                    dict.Add(
+                        (K)Activator.CreateInstance(typeof(K), kvp.Key),
+                        kvp.Value.ToObject<V>());
+                }
+                catch (Exception exc)
+                {
+                    log.WarnFormat($"Failed to deserialize {kvp.Key}", exc);
```
DasSkelett, line 32: `kvp.Key` is the module version, so we get warning messages like this one, which isn't terribly useful for debugging:
> 12560 [1] WARN CKAN.JsonLeakySortedDictionaryConverter`2[[CKAN.Versioning.ModuleVersion, ckan, Version=1.30.0.0, Culture=neutral, PublicKeyToken=null],[CKAN.CkanModule, ckan, Version=1.30.0.0, Culture=neutral, PublicKeyToken=null]] (null) - Failed to deserialize 1

Ideally we could print the identifier as well. If I've read the [docs](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_JsonConverterAttribute.htm) correctly, it should be possible to pass custom parameters for a custom constructor in the attribute: 
```csharp
[JsonConverter(typeof(JsonLeakySortedDictionaryConverter<ModuleVersion, CkanModule>),this.identifier)]
internal SortedDictionary<ModuleVersion, CkanModule> module_version = ...
```
I haven't tried this out yet, though.

DasSkelett: I feared it wouldn't be that easy 😢
Then we'll just have to live with it.

HebaruSan: Hmm, if we logged some or all of `kvp.Value.ToString()`, I think that should give us the JSON of the module that failed to deserialize. Is that better than just the key?

DasSkelett: That's actually a good idea, this way we also have the "faulty" module right away without having to ask for a upload of the registry. (Assuming we get the logs, of course)

HebaruSan: OK, just force-pushed that. Also I should have been using `Warn` instead of `WarnFormat` all along because there were no format parameters and I wanted the special exception handling.

PR #3543, opened 2022-03-07
Play Time

PizzaRules668:
Made it so CKAN can track how long you play the game each time you launch the game. Then it saves it to a file. You can view the total amount of playtime by going under File->Play Time

### DIFF
--- a/GUI/CKAN-GUI.csproj
+++ b/GUI/CKAN-GUI.csproj
@@ -690,6 +690,9 @@
     <EmbeddedResource Include="Controls\PlayTime.resx">
       <DependentUpon>PlayTime.cs</DependentUpon>
     </EmbeddedResource>
+    <EmbeddedResource Include="Localization\fr-FR\PlayTime.fr-FR.resx">
+      <DependentUpon>..\..\Controls\PlayTime.cs</DependentUpon>
+    </EmbeddedResource>
     <EmbeddedResource Include="Controls\ManageMods.resx">
       <DependentUpon>ManageMods.cs</DependentUpon>
     </EmbeddedResource>
--- a/GUI/CKAN-GUI.csproj
+++ b/GUI/CKAN-GUI.csproj
@@ -44,6 +44,7 @@
   <ItemGroup>
     <Reference Include="System" />
     <Reference Include="System.Core" />
+    <Reference Include="System.Data" />
     <Reference Include="System.Drawing" />
     <Reference Include="System.Transactions" />
     <Reference Include="System.Windows.Forms" />
@@ -229,6 +230,12 @@
     <Compile Include="Controls\ManageMods.Designer.cs">
       <DependentUpon>ManageMods.cs</DependentUpon>
     </Compile>
+    <Compile Include="Controls\PlayTime.cs">
+      <SubType>UserControl</SubType>
+    </Compile>
+    <Compile Include="Controls\PlayTime.Designer.cs">
+      <DependentUpon>PlayTime.cs</DependentUpon>
+    </Compile>
     <Compile Include="Controls\ThemedTabControl.cs">
       <SubType>Component</SubType>
     </Compile>
@@ -239,7 +246,7 @@
       <SubType>Component</SubType>
     </Compile>
     <Compile Include="Controls\TriStateToggle.cs">
-      <SubType>Component</SubType>
+      <SubType>UserControl</SubType>
     </Compile>
   </ItemGroup>
   <ItemGroup>
@@ -292,6 +299,9 @@
     </Compile>
     <Compile Include="Main\MainTabControl.cs">
       <SubType>Component</SubType>
+    </Compile>
+    <Compile Include="Main\MainTime.cs">
+      <SubType>Form</SubType>
     </Compile>
     <Compile Include="Main\MainTrayIcon.cs">
       <SubType>Form</SubType>
@@ -676,6 +686,9 @@
     </EmbeddedResource>
     <EmbeddedResource Include="Controls\ModInfo.resx">
       <DependentUpon>ModInfo.cs</DependentUpon>
+    </EmbeddedResource>
+    <EmbeddedResource Include="Controls\PlayTime.resx">
+      <DependentUpon>PlayTime.cs</DependentUpon>
     </EmbeddedResource>
     <EmbeddedResource Include="Controls\ManageMods.resx">
       <DependentUpon>ManageMods.cs</DependentUpon>
--- a/GUI/Controls/PlayTime.Designer.cs
+++ b/GUI/Controls/PlayTime.Designer.cs
@@ -1,0 +1,156 @@
+﻿namespace CKAN
+{
+    partial class PlayTime
+    {
+        /// <summary> 
+        /// Required designer variable.
+        /// </summary>
+        private System.ComponentModel.IContainer components = null;
+
+        /// <summary> 
+        /// Clean up any resources being used.
+        /// </summary>
+        /// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing && (components != null))
+            {
+                components.Dispose();
+            }
+            base.Dispose(disposing);
+        }
+
+        #region Component Designer generated code
+
+        /// <summary> 
+        /// Required method for Designer support - do not modify 
+        /// the contents of this method with the code editor.
+        /// </summary>
+        private void InitializeComponent()
+        {
+            this.components = new System.ComponentModel.Container();
+            System.ComponentModel.ComponentResourceManager resources = new SingleAssemblyComponentResourceManager(typeof(PlayTime));
+            this.EditHelpLabel = new System.Windows.Forms.Label();
+            this.PlayTimeGrid = new System.Windows.Forms.DataGridView();
+            this.NameColumn = new System.Windows.Forms.DataGridViewTextBoxColumn();
+            this.TimeColumn = new System.Windows.Forms.DataGridViewTextBoxColumn();
+            this.BottomButtonPanel = new System.Windows.Forms.Panel();
+            this.TotalLabel = new System.Windows.Forms.Label();
+            this.OKButton = new System.Windows.Forms.Button();
+            this.SuspendLayout();
+            //
+            // EditHelpLabel
+            //
+            this.EditHelpLabel.AutoSize = true;
+            this.EditHelpLabel.Dock = System.Windows.Forms.DockStyle.Top;
+            this.EditHelpLabel.BackColor = System.Drawing.Color.Transparent;
+            this.EditHelpLabel.Location = new System.Drawing.Point(4, 5);
+            this.EditHelpLabel.Margin = new System.Windows.Forms.Padding(4, 5, 4, 5);
+            this.EditHelpLabel.Name = "EditHelpLabel";
+            this.EditHelpLabel.Size = new System.Drawing.Size(147, 20);
+            this.EditHelpLabel.TabIndex = 0;
+            this.EditHelpLabel.Visible = true;
+            resources.ApplyResources(this.EditHelpLabel, "EditHelpLabel");
+            // 
+            // PlayTimeGrid
+            // 
+            this.PlayTimeGrid.AutoGenerateColumns = false;
+            this.PlayTimeGrid.Dock = System.Windows.Forms.DockStyle.Fill;
+            this.PlayTimeGrid.AllowUserToAddRows = false;
+            this.PlayTimeGrid.AllowUserToDeleteRows = false;
+            this.PlayTimeGrid.AllowUserToResizeRows = false;
+            this.PlayTimeGrid.BackgroundColor = System.Drawing.SystemColors.Window;
+            this.PlayTimeGrid.EnableHeadersVisualStyles = false;
+            this.PlayTimeGrid.ColumnHeadersDefaultCellStyle.BackColor = System.Drawing.SystemColors.Control;
+            this.PlayTimeGrid.ColumnHeadersDefaultCellStyle.ForeColor = System.Drawing.SystemColors.ControlText;
+            this.PlayTimeGrid.DefaultCellStyle.ForeColor = System.Drawing.SystemColors.WindowText;
+            this.PlayTimeGrid.CellBorderStyle = System.Windows.Forms.DataGridViewCellBorderStyle.None;
+            this.PlayTimeGrid.ColumnHeadersBorderStyle = System.Windows.Forms.DataGridViewHeaderBorderStyle.None;
+            this.PlayTimeGrid.ColumnHeadersHeightSizeMode = System.Windows.Forms.DataGridViewColumnHeadersHeightSizeMode.AutoSize;
+            this.PlayTimeGrid.Columns.AddRange(new System.Windows.Forms.DataGridViewColumn[] {
+            this.NameColumn,
+            this.TimeColumn});
+            this.PlayTimeGrid.Location = new System.Drawing.Point(0, 111);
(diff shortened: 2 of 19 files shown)


GUI/Main/Main.cs
```diff
@@ -765,17 +767,45 @@ public void LaunchGame()
...
+                
                 Directory.SetCurrentDirectory(CurrentInstance.GameDir());
-                Process.Start(binary, args);
+
+                Process kspProcess = new Process();
+                kspProcess.StartInfo.FileName = binary;
+                kspProcess.Exited += new EventHandler(KspExit);
+                kspProcess.EnableRaisingEvents = true;
+
+                kspProcess.Start();
```
HebaruSan, line 783: The `args` parameter is no longer being passed to the process, which will break many common user configs.

GUI/Main/Main.cs
```diff
@@ -765,17 +767,45 @@ public void LaunchGame()
...
+            TimeLog t = new TimeLog(CurrentInstance.CkanDir());
+            t.InstanceName = CurrentInstance.Name;
+            t.addTime((int)ts.TotalSeconds);
+            
+            t.Save();
+
+            Console.WriteLine(t.Time);
+            Console.WriteLine(t.InstanceName);
+
+            MessageBox.Show(String.Format("You played for {0:00}:{1:00}.{2:00}(H:M.s)", ts.Hours, ts.Minutes, ts.Seconds));
```
HebaruSan, line 806: Needs internationalization (see #2749).

GUI/Main/MainTime.cs
```diff
@@ -0,0 +1,25 @@
...
+using CKAN.Versioning;
+
+namespace CKAN
+{
+    public partial class Main
+    {
+        private void playTimeMenuItem_Click(object sender, EventArgs e)
+        {
+            // Run in a background task so GUI thread can react to user
+            Task.Factory.StartNew(() => playTime());
```
HebaruSan, line 16: This should probably run on the GUI thread because it manipulates the GUI (adds items to a list view).

GUI/TimeLog.cs
```diff
@@ -0,0 +1,59 @@
...
+                this.Time = 0;
+            }
+        }
+
+        [JsonProperty("instance_name", NullValueHandling = NullValueHandling.Ignore)]
+        public string InstanceName;
+
+        [JsonProperty("time", NullValueHandling = NullValueHandling.Ignore)]
+        [DefaultValue(0)]
+        public int    Time;
```
HebaruSan, line 43: This should probably be a `TimeSpan` object.

GUI/Main/Main.cs
```diff
@@ -765,17 +767,45 @@ public void LaunchGame()
...
+            TimeSpan ts = gameTime.Elapsed;
+
+            TimeLog t = new TimeLog(CurrentInstance.CkanDir());
+            t.InstanceName = CurrentInstance.Name;
+            t.addTime((int)ts.TotalSeconds);
+            
+            t.Save();
+
+            Console.WriteLine(t.Time);
+            Console.WriteLine(t.InstanceName);
```
HebaruSan, line 804: Don't `Console.WriteLine`. If you need to print something, use one of the `log4net` objects.

GUI/Main/Main.cs
```diff
@@ -765,17 +767,45 @@ public void LaunchGame()
             var binary = split[0];
             var args = string.Join(" ", split.Skip(1));
 
+            gameTime = new Stopwatch();
+
+            TimeLog t = new TimeLog(CurrentInstance.CkanDir());
```
HebaruSan, line 772: This object isn't being used (since we don't need to touch the file till after the game exits). Can it be removed?

GUI/CKAN-GUI.csproj
```diff
@@ -233,7 +239,7 @@
       <SubType>Component</SubType>
     </Compile>
     <Compile Include="Controls\TriStateToggle.cs">
-      <SubType>Component</SubType>
+      <SubType>UserControl</SubType>
```
HebaruSan, line 242: Was this change intentional? This file doesn't seem to be involved in the other code changes. If Visual Studio did this automatically, maybe I need to investigate whether something is set up wrong in this control.

GUI/Controls/PlayTime.cs
```diff
@@ -0,0 +1,53 @@
...
+                    ts = TimeSpan.FromSeconds(t.Time);
+
+                    this.PlayTimeListView.Items.Add(new ListViewItem(new string[] { 
+                        instance.Key, String.Format("{0:00}:{1:00}.{2:00}", ts.Hours, ts.Minutes, ts.Seconds) }));
+                }
+                
+                ts = TimeSpan.FromSeconds(time);
+
+                this.PlayTimeListView.Items.Add(new ListViewItem(new string[] { 
+                    "Your Total Playtime", String.Format("{0:00}:{1:00}.{2:00}", ts.Hours, ts.Minutes, ts.Seconds) }));
```
HebaruSan, line 44: Needs internationalization (see #2749).

GUI/Controls/PlayTime.Designer.cs
```diff
@@ -0,0 +1,106 @@
+﻿namespace CKAN
+{
+    partial class PlayTime
```
HebaruSan, line 3: I'm not sure that users will want to open a separate screen to see just this information. What do you think of adding it as a column in the Manage Game Instances dialog instead? And possibly the title bar or status bar for the active game instance?

image

GUI/TimeLog.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public string dir;
```
HebaruSan, line 45: Looks like the `dir` member variable isn't used, the `dir` parameter is only used to build `path`. Let's remove this.

GUI/CKAN-GUI.csproj
PizzaRules668: I think I did by mistake

GUI/Main/Main.cs
PizzaRules668: I only had that there for testing for got to remove it

GUI/Main/Main.cs
PizzaRules668: That was another thing that I just needed for testing, will remove

PizzaRules668:
Ok, I think I have fixed all of the comments you made

vinix38:
I found 3 strings that could be translated, so here they are in French:

`<data name="GamePlayTime.Text" xml:space="preserve"><value>Temps de Jeu (H:M.s)</value></data>`
`<data name="PlayTimeTabPage.Text" xml:space="preserve"><value>Temps de Jeu</value></data>`
`<data name="TotalPlayTime" xml:space="preserve"><value>Votre Temps de Jeu Total</value></data>`

Tell me if there is anything else to translate 🙂

PizzaRules668:
@vinix38 Thank your for the translation there is also [GUI/Controls/PlayTime.resx](https://github.com/PizzaRules668/CKAN/blob/time_count/GUI/Controls/PlayTime.resx)

vinix38:
Indeed, there you go:
`<data name="nameHeader.Text" xml:space="preserve"><value>Nom de l'Instance</value></data>`
`<data name="timeHeader.Text" xml:space="preserve"><value>Votre temps de jeu (H:M.s)</value></data>`
`<data name="PlayTimeOkButtom.Text" xml:space="preserve"><value>Ok</value></data>`

GUI/CKAN-GUI.csproj
```diff
@@ -233,7 +239,8 @@
       <SubType>Component</SubType>
     </Compile>
     <Compile Include="Controls\TriStateToggle.cs">
-      <SubType>Component</SubType>
+      <SubType>            TimeLog t = new TimeLog(CurrentInstance.CkanDir());
```
HebaruSan, line 242: Looks like an accidental paste of code where this should just say `Component`.

PizzaRules668: Ok i've fixed it

HebaruSan:
Some of my game instances aren't entirely valid, which causes exceptions to be thrown when opening the Play Time screen and the Manage Game Instances dialog. We need to be fault tolerant and not crash for this, like the existing code does currently. Probably makes sense to return the empty string for such instances.

```
CKAN.NotKSPDirKraken: Could not find KSP version in buildID.txt or readme.txt
   at CKAN.GameInstance.CkanDir()
   at CKAN.PlayTime.loadAllPlayTime()
   at CKAN.Main.playTimeMenuItem_Click(Object sender, EventArgs e)
   at System.Windows.Forms.ToolStripItem.RaiseEvent(Object key, EventArgs e)
   at System.Windows.Forms.ToolStripMenuItem.OnClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleMouseUp(MouseEventArgs e)
   at System.Windows.Forms.ToolStrip.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.ToolStripDropDown.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.ToolStrip.WndProc(Message& m)
   at System.Windows.Forms.ToolStripDropDown.WndProc(Message& m)
```

```
CKAN.NotKSPDirKraken: Could not find KSP version in buildID.txt or readme.txt
   at CKAN.GameInstance.CkanDir()
   at CKAN.ManageGameInstancesDialog.<UpdateInstancesList>b__8_1(KeyValuePair`2 instance)
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at CKAN.ManageGameInstancesDialog.UpdateInstancesList()
   at CKAN.ManageGameInstancesDialog..ctor(Boolean centerScreen, IUser user)
   at CKAN.Main.manageGameInstancesMenuItem_Click(Object sender, EventArgs e)
   at System.Windows.Forms.ToolStripItem.RaiseEvent(Object key, EventArgs e)
   at System.Windows.Forms.ToolStripMenuItem.OnClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleMouseUp(MouseEventArgs e)
   at System.Windows.Forms.ToolStrip.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.ToolStripDropDown.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.ToolStrip.WndProc(Message& m)
   at System.Windows.Forms.ToolStripDropDown.WndProc(Message& m)
```

PizzaRules668:
@HebaruSan I wasn't able to replica the error but I added some stuff to stop it from crashing

HebaruSan:
> @HebaruSan I wasn't able to replica the error but I added some stuff to stop it from crashing

That's a bit better, but still not usable because it displays a popup and then the form doesn't work. I will push a commit in a moment to fix this plus several other small things that I noticed that are easier to just change than describe and check...

HebaruSan:
Remaining things I want to address before merge:

- [ ] Does anything bad happen if I close CKAN while the game is still running?
  - [ ] Should we try to capture partial time data at close to account for that use case?
- [ ] Is it possible to launch the game, switch to another instance with the game still running, and launch again?
  - [ ] If so, does it work?
- [ ] What happens if the launch fails (e.g., for a fake instance)? Does the timer stop?
- [ ] Consider adding the instance play time to the status bar next to the instance name
- [ ] Consider showing total play time somewhere on the manage game instances popup
- [ ] Should there be a setting to turn this functionality on and off? I.e., will any users be annoyed by it enough to want that?
- [ ] Squash commits to the minimum reasonable number
- [ ] Is there any way we could capture play time when launching the game from Steam or the command line?
  - [ ] If not, should we make it easier for users who want to sync their data to edit the numbers manually (other than by editing the JSON file in a text editor)?
- [ ] Should the play time be cleared when cloning an instance, to avoid double counting?
- [ ] Should the play time be a public property of `GameInstance`?
  - [ ] It could be loaded at instantiation, then saved as needed without more loads

PizzaRules668:
> Remaining things I want to address before merge

I can start working on it. I should be able to get stuff done next week b/c it's spring break for me.

PizzaRules668:
> Does anything bad happen if I close CKAN while the game is still running?

When CKAN closes it now save the current time on the stopwatch

> Is it possible to launch the game, switch to another instance with the game still running, and launch again? 

You can now Launch an instance then switch to a new instance and it will save the play time to the correct instance

> What happens if the launch fails (e.g., for a fake instance)? Does the timer stop?

If the launch fails it should be caught by the `try catch` and not start the stopwatch

> Consider adding the instance play time to the status bar next to the instance name

Added

> Consider showing total play time somewhere on the manage game instances popup

I believe this was already added, unless I'm thinking of something else

> Should there be a setting to turn this functionality on and off? I.e., will any users be annoyed by it enough to want that?

I don't think there is a reason to turn it off, I think most users would like this.

>Is there any way we could capture play time when launching the game from Steam or the command line?

This can be done, I believe we would have to use the [Steam WebAPI](https://steamcommunity.com/dev), Users would also have to have a public profile

> Should the play time be cleared when cloning an instance, to avoid double counting?

Added

> Should the play time be a public property of `GameInstance`?

Added

GUI/Main/Main.cs
```diff
@@ -243,7 +243,7 @@ private void CurrentInstanceUpdated(bool allowRepoUpdate)
             CurrentInstance.Scan();
             Util.Invoke(this, () =>
             {
-                Text = $"CKAN {Meta.GetVersion()} - {CurrentInstance.game.ShortName} {CurrentInstance.Version()}    --    {CurrentInstance.GameDir().Replace('/', Path.DirectorySeparatorChar)}";
+                Text = $"CKAN {Meta.GetVersion()} - {CurrentInstance.game.ShortName} {CurrentInstance.Version()}    --    {CurrentInstance.GameDir().Replace('/', Path.DirectorySeparatorChar)}  -  {CurrentInstance.playTime.ToString()}h";
```
HebaruSan, line 246: "h" is English-specific, we'll need to internationalize that.

HebaruSan:
> I don't think there is a reason to turn it off, I think most users would like this.

The counterexample that comes to mind is a user who always launches the game through Steam rather than CKAN. For that user, the time counts would always be 0, so they might get tired of seeing them.

PizzaRules668:
Ok, I think that is a fair argument to be made, will think of away to turn it on an off

HebaruSan:
Just pushed a somewhat big commit, the general idea of which is to let the user edit the instances' play times in the grid instead of using an "external time" concept/popup:

image

This way we don't have to explain what "external time" means, and the user can split up their previous Steam play time across different instances if they want to. Also moved the total time out to a separate label so the user doesn't try to edit that.

> > Consider showing total play time somewhere on the manage game instances popup
> 
> I believe this was already added, unless I'm thinking of something else

I meant on this form:

image

But on second thought, this doesn't seem important. A user who's really curious can check the dedicated Play Time screen.

> > Should there be a setting to turn this functionality on and off? I.e., will any users be annoyed by it enough to want that?
> 
> I don't think there is a reason to turn it off, I think most users would like this.
> Ok, I think that is a fair argument to be made, will think of away to turn it on an off

The column in Manage Game Instances is now auto-hidden if all the values are 0, which means it'll become visible automatically if such a user starts launching from CKAN but otherwise stay out of the way. No need for an explicit setting now.

___

I'm pretty happy with how this works now. If you want to make any other changes, go ahead and do that, otherwise I'll squash this down to a small number of commits and merge.

HebaruSan (approved):
Let's try it! :fireworks: :crossed_fingers:

PR #3554, opened 2022-03-29
Combine staging reasons

HebaruSan:
## Background

Since about 2022-02-23 (earliest example I found was KSP-CKAN/CKAN-meta#2585), we have been getting auto-epoch pull requests from the bot indicating that it inflated an older release from GitHub instead of the latest release. The reason for this is unknown, but we presume the GitHub API is acting weird. We currently have no ideas for how to address this (but it would be nice!).

## Problems

That phenomenon generated these pull requests:

- KSP-CKAN/CKAN-meta#2620
- KSP-CKAN/CKAN-meta#2621
- KSP-CKAN/CKAN-meta#2622
- KSP-CKAN/CKAN-meta#2623

I thought these were for a new release, but in fact they were auto-epochs for an old release.

Also, they don't have the link table, which would be nice for investigating what happened.

## Cause

The pull request description was populated from the netkan's staging reason only. The auto-epoch staging reason was set into `TransformOptions.StagingReason`, but overridden by the one from the netkan when the module was sent to the queue.

## Changes

- Now `TransformOptions.StagingReason` is replaced by a list `TransformOptions.StagingReasons`
- Now the values for staged and staging reason are copied from the netkan to `TransformOptions`
- Now transformers that set staging (including the one that appends the link table) will append their staging reason to the list rather than overwriting it (or being ignored)
- When we finally send the inflated module to the queue, we join all the staging reasons together separated by blank lines

The next time we get an auto-epoch pull request for a mod with a staging reason set in the netkan, the auto-epoch reason will also be visible, and we will see the link table.

### DIFF
--- a/Netkan/Processors/QueueHandler.cs
+++ b/Netkan/Processors/QueueHandler.cs
@@ -148,7 +148,7 @@ namespace CKAN.NetKAN.Processors
             IEnumerable<Metadata> ckans = null;
             bool   caught        = false;
             string caughtMessage = null;
-            var    opts          = new TransformOptions(releases, null, highVer);
+            var    opts          = new TransformOptions(releases, null, highVer, netkan.Staged, netkan.StagingReason);
             try
             {
                 ckans = inflator.Inflate($"{netkan.Identifier}.netkan", netkan, opts);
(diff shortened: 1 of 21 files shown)


HebaruSan:
Self-reviewing so we can see this in action the next time it happens.

PR #3560, opened 2022-04-06
Show dependencies of upgrading mods in change set

### LINKED ISSUE #3559: [Bug] New dependencies of upgraded mods missing from change set
CKAN 1.30.4; Debian GNU/Linux 11; Mono 6.12.0.122

Problem
-------
I went to update a mod recently (X Science! Continued) and it now contains new dependencies after having been taken over by another developer.

Apart from confirming the list of mods to update, the new dependencies were not listed or needing to be confirmed during this step. It was only when the download started that I noticed the additional mods.  I had to kill the CKAN process in order to abort the update (the "Cancel" button was, for some reason, unresponsive)

Suggestions
-----------
I'd like CKAN to pop up a similar confirmation as happens when first installing a mod; that is, listing all new dependencies during the confirmation step.

Alternatively/relatedly, there is no way to check whether a new version of a mod has new dependencies. The dependencies list in the UI only shows the dependencies of the currently installed version. It would be nice to be able to display the information about another version of the mod.

Semi-Relatedly, it would be nice to "pin" a mod to its current version, to prevent it from being updated when using the "Add available updates" button.

Semi-Relatedly, it would be nice to be able to blacklist mods, which would prevent any updates to existing mods which depend on blacklisted mods. (EDIT: This is actually raised in #3129)

HebaruSan:
## Problem

1. Install xScience 5.26
2. Upgrade to the latest version
3. The change set shows just the upgrade:
   image
4. However, the install screen shows that new dependencies will be installed:
   image
   The user should be told about this in the change set screen.

## Cause

The change set screen has always only looked for dependencies for modules being installed, not upgraded.

I think the expectation was that all of the complicated stuff would happen for installs, while upgrades would be simple replacements of one version of a module with a newer version, even though the underlying relationship logic in Core can handle more complicated scenarios.

## Changes

Now upgrades will be treated more like installs, in that any new dependencies they introduce will be displayed:

image

To make this work, we also need to fix `ComputeChangeSetFromModList` to use the upgraded version in calculating the change set. It was using the old version; luckily this only affects the change set screen and not the actual install, which is handled in Core.

Fixes #3559.

### DIFF
--- a/GUI/Controls/Changeset.cs
+++ b/GUI/Controls/Changeset.cs
@@ -20,19 +20,17 @@ namespace CKAN
             ChangesListView.Items.Clear();
             if (changes != null)
             {
-                // We're going to split our change-set into two parts: updated/removed mods,
-                // and everything else (which right now is replacing and installing mods, but we may have
+                // We're going to split our change-set into two parts: removed mods,
+                // and everything else (right now, replacing, upgrading, and installing mods, but we may have
                 // other types in the future).
 
                 sortedChangeSet.Clear();
                 sortedChangeSet.AddRange(changes.Where(change => change.ChangeType == GUIModChangeType.Remove));
-                sortedChangeSet.AddRange(changes.Where(change => change.ChangeType == GUIModChangeType.Update));
 
                 // Now make our list more human-friendly (dependencies for a mod are listed directly
                 // after it.)
                 CreateSortedModList(changes
-                    .Where(change => change.ChangeType != GUIModChangeType.Remove
-                                  && change.ChangeType != GUIModChangeType.Update)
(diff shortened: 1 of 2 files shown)


DasSkelett:
This comment needs updating,
https://github.com/KSP-CKAN/CKAN/blob/d09ac3a34eada4899c343d4be1b38d78fdfd0f5d/GUI/Controls/Changeset.cs#L23-L25
"updated" is no longer part of the first group

Otherwise looks good to me

HebaruSan:
Updated! Sometimes writing extra documentation can be a liability of sorts...

DasSkelett (approved):
Doesn't break more complex changeset scenarios, so that's a plus :smile:

PR #3563, opened 2022-04-13
Context sensitive help

HebaruSan:
## Motivation

We had somebody stop by the Discord asking how to use the Manage Game Instances popup, which appears by itself, blank, if you have no game instances and CKAN can't find one in Steam.

We added some help menu options in #3437, but these not accessible from the Manage Game Instances popup. (And that pull request's changes have not been released yet.)

## Changes

- Now [`.HelpButton`](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.form.helpbutton?view=netframework-4.7.2) is `true` for most of our popup dialogs
- Now `.MinimizeBox` and `.MaximizeBox` are `false` for Manage Game Instances and Install Filters, which they probably should have been anyway because maximizing and minimizing these popups is not useful
- Now we override [`OnHelpButtonClicked`](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.form.onhelpbuttonclicked?view=netframework-4.7.2#system-windows-forms-form-onhelpbuttonclicked(system-componentmodel-canceleventargs)) and [`OnHelpRequested`](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.onhelprequested?view=netframework-4.7.2#system-windows-forms-control-onhelprequested(system-windows-forms-helpeventargs)) to open the most relevant help page for each context (the user guide by default, more specific links such as PRs when available)

Now to access help, users will be able to press F1 anywhere or click this question mark button for most of our popups:

image

(Note that this button is _not_ added to the main form because it is mutually exclusive with the minimize and maximize buttons, which we should keep for the main form.)

This should allow some users to find help on their own, and it will simplify many support requests down to telling users, "You can press F1 for help on that".

### DIFF
--- a/Core/HelpURLs.cs
+++ b/Core/HelpURLs.cs
@@ -1,0 +1,23 @@
+namespace CKAN
+{
+    /// <summary>
+    /// Holds URLs that we launch for F1 help
+    /// </summary>
+    public static class HelpURLs
+    {
+        public const string UserGuide = "https://github.com/KSP-CKAN/CKAN/wiki/User-guide";
+        public const string ManageInstances = "https://github.com/KSP-CKAN/CKAN/wiki/User-guide#adding-game-folders";
+        public const string CompatibleGameVersions = "https://github.com/KSP-CKAN/CKAN/wiki/User-guide#choosing-compatible-game-versions";
+        public const string ModPacks = "https://github.com/KSP-CKAN/CKAN/wiki/Sharing-a-modlist-%28metapackages%29";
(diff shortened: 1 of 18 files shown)


techman83 (approved):
My only comment would be here, that we might want to centralise these URLs as constants somewhere. Digging through the code base if we had a to change a bunch of URLs would make for a maintenance burden (though I appreciate we probably already have this burden.)

HebaruSan:
Good idea, moved them to a new `HelpURLs` file/class. Also updated some existing links to the user guide and Discord.

PR #3568, opened 2022-04-17
Add install size to metadata and display in clients

HebaruSan:
## Motivation

This looks a little silly, since each of these modules installs a different, smaller subset of the download:

image

And this isn't accurate because it sums the `download_size` properties of the installed modules, not their installed sizes:

image

## Changes

- Now an `install_size` property is added to the schema, spec, and `CkanModule`, and `InstallSize` is added to `GUIModule`
- Now Netkan populates `install_size` based on the sum of the `Size` properties of the list of the `ZipEntry` objects that the module installs
- Now the `ConsoleUI` uses this property instead of and falling back to `download_size`
- Now a column for install size is added to GUI

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -82,6 +82,10 @@
             "description" : "The content type of the download",
             "type"        : "string"
         },
+        "install_size": {
+            "description" : "The size of the installed files",
+            "type"        : "integer"
+        },
         "license" : {
             "description" : "Machine readable license, or array of licenses",
             "$ref"        : "#/definitions/licenses"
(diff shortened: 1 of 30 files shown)


HebaruSan:
Rebased to resolve conflict.

techman83 (approved):
Looks like a useful addition. I note that the bot will do a bunch of work once merged. Do we have any tests validating this transformer or will it fail gracefully?

HebaruSan:
Good point, I will have a look at writing some tests. Especially since it's a new transformer...

HebaruSan:
Update of the original screenshot, now we can see how big each package is when decompressed and installed:

image

PR #3570, opened 2022-04-22
Fix play time column if no playtime, hide game column if single game

DasSkelett:
## Problem
#3543 introduced play time tracking and tries to conditionally display it in the "Manage game instances" dialog only if any instance has a playtime > 0.

This is broken, because the `0,0` string for the playtime is added to the `ListViewItem` contents array all the time, and if the column is removed it shows under `Path`:
broken-game-time

## Changes
Now the string array for the column contents is factored out into a separate function. It includes the play time string only if the column (header) will actually be present in the end.
The same is applied to the "Game" column now as well, for those that only have instances of a single type of game registered with CKAN (currently the case all the time for everyone as we don't support another game yet).
The auto-resizing needs to be done after removing columns as it applies it to the incorrect columns otherwise.

new-shiny-and-fixed
(Ignore the Factorio instance being present, it is detected as KSP as this current branch doesn't have Factorio support)

I've also added the German localization for the play time strings.

### DIFF
--- a/GUI/Dialogs/ManageGameInstancesDialog.cs
+++ b/GUI/Dialogs/ManageGameInstancesDialog.cs
@@ -74,20 +74,11 @@ namespace CKAN
             GameInstancesListView.Items.Clear();
             UpdateButtonState();
 
-
-            if (!GameInstancesListView.Columns.Contains(Game))
-            {
-                // Always show the game column so our rows load correctly
-                GameInstancesListView.Columns.Insert(Game.Index, Game);
-            }
-            if (!GameInstancesListView.Columns.Contains(GamePlayTime))
-            {
-                // Always show the play time column so our rows load correctly
-                GameInstancesListView.Columns.Insert(GamePlayTime.Index, GamePlayTime);
-            }
-
             var allSameGame = _manager.Instances.Select(i => i.Value.game).Distinct().Count() <= 1;
             var hasPlayTime = _manager.Instances.Any(instance => (instance.Value.playTime?.Time ?? TimeSpan.Zero) > TimeSpan.Zero);
+
+            AddOrRemoveColumn(GameInstancesListView, Game, !allSameGame);
+            AddOrRemoveColumn(GameInstancesListView, GamePlayTime, hasPlayTime);
 
             GameInstancesListView.Items.AddRange(_manager.Instances
                 .OrderByDescending(instance => instance.Value.Version())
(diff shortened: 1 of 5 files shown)


GUI/Dialogs/ManageGameInstancesDialog.cs
```diff
@@ -73,42 +74,66 @@ public void UpdateInstancesList()
...
             UpdateButtonState();
 
+
+            if (!GameInstancesListView.Columns.Contains(Game))
+            {
+                // Always show the game column so our rows load correctly
+                GameInstancesListView.Columns.Insert(Game.Index, Game);
+            }
             if (!GameInstancesListView.Columns.Contains(GamePlayTime))
             {
```
HebaruSan, line 84: Hmm, the add-then-remove logic of this was based on the belief that the _contents_ of the columns were hidden when the columns were hidden (and I thought I confirmed that to be the case in testing, but I may be mis-remembering). If that's not the case, then we can probably simplify this to add/remove the columns all together and then just add the column data we need when we add the rows, as you have it now in `rowItems`. In other words, we don't need to call `Insert` for either of these columns if we aren't planning to populate them.

GUI/Dialogs/ManageGameInstancesDialog.cs
```diff
@@ -73,42 +74,66 @@ public void UpdateInstancesList()
...
                     Tag = instance.Key
                 })
                 .ToArray()
             );
 
-            GameInstancesListView.AutoResizeColumns(ColumnHeaderAutoResizeStyle.ColumnContent);
-            GameInstancesListView.AutoResizeColumns(ColumnHeaderAutoResizeStyle.HeaderSize);
-
-            var hasPlayTime = _manager.Instances.Any(instance => (instance.Value.playTime?.Time ?? TimeSpan.Zero) > TimeSpan.Zero);
+            if (allSameGame && GameInstancesListView.Columns.Contains(Game))
```
HebaruSan, line 101: Cool idea! :sunglasses: :+1:

GUI/Dialogs/ManageGameInstancesDialog.cs
HebaruSan: Added a commit based on this idea, does it look OK to you?

DasSkelett: > based on the belief that the _contents_ of the columns were hidden when the columns were hidden (and I thought I confirmed that to be the case in testing, but I may be mis-remembering)

I can imagine that this is true on Windows, but broken on Mono (assuming you also were on Windows at that time).

> Added a commit based on this idea, does it look OK to you?

Very nice, makes it cleaner.

HebaruSan (approved):
Looks good, let's do this.

PR #3571, opened 2022-04-23
Sort GitHub releases

HebaruSan:
## Problems

For several months now the bot has been submitting bogus auto-epoch pull requests, about 0.7 times per day:

<https://github.com/KSP-CKAN/CKAN-meta/pulls?q=is%3Apr+author%3Anetkan-bot+is%3Aunmerged+is%3Aclosed>

This seems to happen completely randomly, only for GitHub krefs, as if the most recent release for that mod has disappeared. However, when we check the repo, it is still there. So somehow the wrong release is being inflated.

## Cause (guesses)

We don't know for sure why this happens (it's very hard to investigate something that only happens once out of 297 unique unfrozen GitHub krefs * 2 inflations per hour * 24 hours per day / 0.7 per day = 20366 attempts), but I have two main guesses at this point:

1. The GitHub API somehow glitches out and skips the most recent release completely, so the most recent release appears to be the one before that
2. The GitHub API returns all the releases, but it somehow glitches out and returns the releases out of order, so the most recent release is included, but it's listed further back in the list than it should be

To further the investigation of this problem, I will be attempting to rule out cause 2 here.

## Changes

Now when we receive a group of releases from the API, before we choose which ones to inflate, we sort them by the `published_at` property, in descending order. If the API has returned the releases out of order, this will ensure that we will identify the most recent one correctly.

If the problem stops after this, then that confirms cause 2 as the culprit. If it continues, then cause 1 looks likely.

In addition, the interpretation of the `JToken` objects returned from the API for releases is moved from `GithubApi` to a new `GithubRelease` constructor, which allows `GithubApi` to deal only in properties of a release object rather than having to cast JSON tokens, etc.

### DIFF
--- a/Netkan/Sources/Github/GithubApi.cs
+++ b/Netkan/Sources/Github/GithubApi.cs
@@ -3,9 +3,11 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Text.RegularExpressions;
+
 using log4net;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+
 using CKAN.NetKAN.Services;
 using CKAN.Versioning;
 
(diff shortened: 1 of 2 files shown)


DasSkelett:
We recently had this error in #netkan-bot, for a mod with only one release. The "missing required field download" also happens when there is no (full) release for a mod. This makes me think that guess 1 is the correct one, i.e. the release is "hidden" completely.
Screenshot_20220423-180649

That said, your changes are still worth a try and sorting the releases by publish date makes sense even without this problem.

HebaruSan:
> We recently had this error in #netkan-bot, for a mod with only one release. The "missing required field download" also happens when there is no (full) release for a mod. This makes me think that guess 1 is the correct one, i.e. the release is "hidden" completely.

Oof, good point, I didn't make that connection when that error came up. My only idea for cause 1 so far, since we can't prevent it or detect it, is to switch to the GraphQL API in the hopes that the bug is specific to the old API. Hopefully it doesn't come to that. :crossed_fingers:

HebaruSan:
> 1322 unfrozen modules * 2 inflations per hour * 24 hours per day / 0.7 per day = 90651 attempts

This number is inflated in a few ways:

- Mods that share krefs also share API data within 5 minutes, so the number of API accesses for one pass is less than 1322 (I knew this when I typed the above)
- Not all mods are on GitHub! The ones on SpaceDock are included in the module count but won't call out to GitHub (I just realized this now)

So the true frequency is greater than 1 in 90651, possibly by a factor of 2 or more. A more accurate estimate would replace 1322 with the number of unfrozen unique GitHub krefs.

HebaruSan:
The correct number of API calls per pass appears to be:

```
>>> len({'/'.join(nk.kref_id.split('/')[0:2]) for nk in nkr.netkans() if nk.on_github})
297
```

So the correct numerator then is 297 * 2 * 24 / 0.7 = 20366. Still way higher than I'm going to be trying manually, but not nearly as close to 100k.

DasSkelett (approved):
Let's give it a try :crossed_fingers:

HebaruSan:
> If it continues, then cause 1 looks likely.

Cause 1 confirmed:  KSP-CKAN/CKAN-meta#2686

HebaruSan:
> My only idea for cause 1 so far, since we can't prevent it or detect it, is to switch to the GraphQL API in the hopes that the bug is specific to the old API.

If we tried this, would we be able to test it before going live? It seems like we would have to run it at least 20k times to detect it if the same bug was still present.

PR #3576, opened 2022-05-02
Auto-size buttons in bottom panels

HebaruSan:
## Problems

- The bottom right button in Delete Directories is supposed to say "Delete Checked," since only the folders with checked checkboxes will be deleted, but on Windows it only says "Delete":
  https://github.com/KSP-CKAN/CKAN/blob/4f3da572f14a3a46e9ca7ddf3190f341574f7dc0/GUI/Controls/DeleteDirectories.resx#L126
  screenshot
  - This potentially can happen anywhere, especially if a translator picks a long string for a button
- The text of the progress bar screen's log text box is unreadable black-on-dark when I install things in Windows using the dark theme from #3489, but readable light-on-dark for refreshing the mod list and repo updates.

## Causes

- Most of our buttons use `Anchor` to position themselves and have hard coded sizes, and if those sizes aren't big enough to fit the text, it is truncated. Due to differences in fonts, we have enough space for "Delete Checked" on Linux bot not on Windows.
- The `ForeColor` we set is ignored because we aren't setting `BackColor`. The text color change happens because `MainTabControl` is disabled, which cascades into disabling the log text box.

## Changes

- A new `LeftRightRowPanel` control hosts two groups of controls, one in `.LeftControls` and the other in `.RightControls`. These are passed through to a pair of `FlowLayoutPanel`s, which themselves are hosted in the cells of a one-row `TableLayoutPanel`. Controls hosted in such a way will be positioned nicely while also being able to autosize themselves.
  - Most of the controls where we have a `BottomButtonPanel` now use `LeftRightRowPanel` for it, and the buttons contained therein have autosizing enabled in `GrowOnly` mode, so they will expand if needed to fit their contents but otherwise keep a nice friendly width for easy clicking:
    image
    As a bonus, the buttons even wrap nicely when space is very limited:
    image
  - The exception is the add repo popup, which has several non-Button controls on its `BottomButtonPanel` which cannot be easily split into left and right flow layout groups. This panel is now renamed `BottomPanel` to distinguish it from the ones that only contain Buttons.
  - The Depends/Recommends/etc. buttons in Export Modpack are radio buttons, which have an `AutoSize` property but not `AutoSizeMode`, so we have to choose between static sizes and making them smaller in some cases. For now I have left them with static sizes.
  - Since this layout logic is more robust than what we had before, the "Retry" and "Cancel" buttons on the progress bar tab are now hidden rather than disabled when they're not usable. And the update repo action is now properly treated as non-cancellable.
- Now we set the log text box's `BackColor`, which makes the text color render correctly in all cases. The disabling of `MainTabControl` is replaced with calls to `tabController.SetTabLock`, which is how we are supposed to disable the tab control.
- The splitter in mod info is now set to treat its top panel as fixed size, which should make it less likely to end up weird sizes.

### DIFF
--- a/GUI/CKAN-GUI.csproj
+++ b/GUI/CKAN-GUI.csproj
@@ -224,6 +224,9 @@
     <Compile Include="Controls\EditModSearchDetails.Designer.cs">
       <DependentUpon>EditModSearchDetails.cs</DependentUpon>
     </Compile>
+    <Compile Include="Controls\LeftRightRowPanel.cs">
+      <SubType>UserControl</SubType>
+    </Compile>
     <Compile Include="Controls\ManageMods.cs">
       <SubType>UserControl</SubType>
     </Compile>
--- a/GUI/Controls/Changeset.Designer.cs
+++ b/GUI/Controls/Changeset.Designer.cs
@@ -34,10 +34,11 @@
             this.Mod = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
             this.ChangeType = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
             this.Reason = ((System.Windows.Forms.ColumnHeader)(new System.Windows.Forms.ColumnHeader()));
-            this.BottomButtonPanel = new System.Windows.Forms.Panel();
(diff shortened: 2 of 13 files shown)


techman83 (approved):
Looks like a good enhancement :tada:

PR #3586, opened 2022-06-02
Fix tracking of paths with trailing spaces on Windows

### LINKED ISSUE #3585: [Bug] Can't uninstall some mods, CKAN tries and fails to install something instead.
Background
----------
* Operating System: Windows 10 Version 21H1 (OS Build 19043. 1706)
* CKAN Version: v1.30.4
* KSP Version: 1.12(On Final Approach)

Problem
-------
I'm deleting some mods because they seem to have somehow removed all manned pods, and it's giving me the following error:

About to remove:

 * DaMichel's AeroRadial (DAR) 1:1.1.0.1
 * DaMichel's SphericalTanks (DST) 1:1.1.0.0
 * DaMichel's Fuselage (DMF) 1.1.0.2
 * BD Armory Continued 1:v1.3.4
 * MrScienceMeister's Nuclear Warfare 0.5.1
 * Nehemiah Engineering Orbital Science 0.9.1
 * BD Modular Missile Parts v3.0
 * Procedural Parts v2.3.0
 * Physics Range Extender  1.21.0
 * B9 Aerospace Procedural Wings - Fork 3:0.43.0.11
Could not find a part of the path 'D:\SteamLibrary\SteamApps\common\Kerbal Space Program\GameData\NuclearWarfare\Parts\freeFallNukeBrake'.
Error during installation!

The really annoying thing is that I can actually locate the named file by just typing in the exact path it says it can't find, and I have no idea why it's trying to install anything, given that I'm actually wanting to uninstall the listed mods.

```
PLACE CODE HERE !
```

HebaruSan:
## Problem

If you install `MrScienceMeistersNuclearWarfare` (latest version is for KSP 1.0.5) and then try to uninstall it, you get this error:

```
About to remove:

 * MrScienceMeister's Nuclear Warfare 0.5.1
Could not find a part of the path 'D:\SteamLibrary\SteamApps\common\Kerbal Space Program\GameData\NuclearWarfare\Parts\freeFallNukeBrake'.
Error during installation!
```

## Cause

The path can't be found because the `installed_files` entry in `registry.json` doesn't match the path on disk.

`ModuleInstaller` assumes that `Directory.CreateDirectory` and `File.Create` will either create the paths they're given or fail. Astoudingly, this assumption is **false** on Windows, a purportedly real professional operating system that is sold to users across the world! Instead, both of these functions silently trim trailing spaces from paths. This code:

```csharp
Directory.CreateDirectory("test path      ");
```

... creates a directory called `test path` with no trailing spaces. This happens silently; there's no exception or other explicit indication that something unexpected has happened. If the application code needs to know the path, it has to know about this potential problem (despite it not being particularly clearly documented) and implement a workaround for it.

Two directories in `MrScienceMeistersNuclearWarfare` end with trailing spaces. So when we uninstall this module, we try to remove the path with the trailing space, but Windows doesn't find it because it's not there, having been replaced by a different path during installation. :facepalm: 

## Changes

- Now `ModuleInstaller.CopyZipEntry` returns the path it created
  - For directories, the `TxFileManager.CreateDirectory` function we use has no return value, but `Path.GetDirectoryName` apparently performs the trimming logic, so we use `Path.GetDirectoryName(Path.Combine(fullPath, "DUMMY"))`
  -  For files, `File.Create` returns a `FileStream` object with a `Name` property that has it
- Now `ModuleInstaller.InstallModule` collects the created paths from `CopyZipEntry` and returns them instead of simply assuming that every `InstallableFile.destination` will be created as requested

This will ensure that the entries in `registry.json` always match the real paths on disk, so uninstallation will no longer be tripped up by this Windows quirk.

Fixes #3585.

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -327,6 +327,7 @@ namespace CKAN
         private IEnumerable<string> InstallModule(CkanModule module, string zip_filename, Registry registry, ref HashSet<string> possibleConfigOnlyDirs)
         {
             CheckKindInstallationKraken(module);
+            var createdPaths = new List<string>();
 
             using (ZipFile zipfile = new ZipFile(zip_filename))
             {
@@ -387,7 +388,7 @@ namespace CKAN
                     foreach (InstallableFile file in files)
                     {
                         log.DebugFormat("Copying {0}", file.source.Name);
-                        CopyZipEntry(zipfile, file.source, file.destination, file.makedir);
+                        createdPaths.Add(CopyZipEntry(zipfile, file.source, file.destination, file.makedir));
                         if (file.source.IsDirectory && possibleConfigOnlyDirs != null)
                         {
                             possibleConfigOnlyDirs.Remove(file.destination);
@@ -403,9 +404,8 @@ namespace CKAN
                     kraken.owningModule = registry.FileOwner(kraken.filename);
                     throw;
                 }
-
-                return files.Select(x => x.destination);
             }
+            return createdPaths.Where(p => p != null);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


HebaruSan:
This makes me wonder if the validation scripts should _uninstall_ mods after installing them. This could have been caught 6+ years ago if that was a thing back then.

DasSkelett (approved):
Oh Windows, always good for surprises.
Didn't test on Windows, but it doesn't break anything on Linux and the code looks sensible.

PR #3605, opened 2022-07-17
Generate RPM repo for releases and under dev builds

HebaruSan:
## Motivation

Currently we provide `.rpm` packages, but somewhat ironically (given what CKAN is for), the user must download a new `.rpm` file and install it manually when a new version is released.

See #3197 for when we did this for APT/deb, and #2757 for when we started making RPMs.

## Changes

Now a new `./build rpm-repo` command creates a `_build/rpm/repo` folder containing the `.rpm` file (signed with our GPG key) and the additional metadata files needed to host an RPM repo. Just like we did with APT, the test builds (odd build number in version) go into a `nightly` repo, and the releases go into a `stable` repo.

I followed these instructions:

- https://earthly.dev/blog/creating-and-hosting-your-own-rpm-packages-and-yum-repo/

User instructions will be in [doc/rpm-repo.md](https://github.com/HebaruSan/CKAN/blob/feature/rpm-repo/doc/rpm-repo.md).

Using this (with some hacks to work around the servers and GPG keys being different), I was able to use the GNOME Software app to upgrade my 3-year-old Fedora VM from CKAN v1.26.3 to v1.31.0.

PR #3609, opened 2022-07-21
Switch build from createrepo-c to createrepo

HebaruSan:
## Problem

After #3605, the build failed:

screenshot

## Cause

Ubuntu and Debian have `createrepo-c` for generating RPM repos. I hoped that `mono:latest` would be the same :crossed_fingers:, but it was not to be.

```
$ docker run -it --entrypoint sh mono:latest

# apt-cache search createrepo
createrepo - tool to generate the metadata for a yum repository
```

## Changes

- The `deploy.yml` file is updated to install `createrepo` instead of `createrepo-c`
- The RPM `Makefile` now uses `createrepo` if you have it, or `createrepo_c` if you don't, so local dev will still work

I'll merge this shortly.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -30,7 +30,7 @@ jobs:
           fi
 
       - name: Installing build dependencies
-        run: apt-get update && apt-get install -y git make sed gzip fakeroot lintian dpkg-dev gpg createrepo-c
+        run: apt-get update && apt-get install -y git make sed gzip fakeroot lintian dpkg-dev gpg createrepo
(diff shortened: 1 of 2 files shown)


HebaruSan:
That worked, but the nightly repo only has `ckan-1.31.1-1.noarch`, which lacks a daily incrementing version number part, so all the builds until the next release will share the same version number and RPM won't update them...

HebaruSan:
Updated `release.yml` in 5917e37cf8650d9f1748885c565d9e872fd209e1.

PR #3610, opened 2022-07-21
Add year and day of year to RPM versions

HebaruSan:
## Problem

#3609 worked, but the nightly repo only has `ckan-1.31.1-1.noarch`, which lacks a daily incrementing version number part, so all the builds until the next release will share the same version number and RPM won't update them.

## Changes

Now the version has a new piece added to it made up of the year concatenated with the day of the year, just like we did for the `.deb`s in #3215.

I'll merge this as soon as the validation scripts finish.


HebaruSan:
Updated `release.yml` in bf92016da270988d71c3a166d516e39a38383eca.

PR #3612, opened 2022-07-24
Implement a new tab in mod info to show changelogs where available

jkavalik:
I wanted to see the changelog of updated modules.
This change shows changelogs (well, a list of release notes) from github based repositories.
It checks the module repository and if it is github.com then it loads releases info from github api.
Current limitations:
- it can only load release notes from github
- it is subject to rate limiting by github api (60req/h/IP) without a token, I can use my personal token for development, getting 5000req/hour but I am not sure if such limit would be reasonable for release usage

Todo:
- [x] Add some kind of caching - cache response json per repo in memory could be enough, with some etag check or something similar to be able to invalidate without restart
- [x] Do not load the changelog automatically but instead add a button (only for supported repos) or info (for unsupported repos)
- [x] After ^ maybe load changelogs automatically for mods which have an update available
- [ ] Format the changelog text in some way - at least show current version in bold and highlight newer available versions
- [ ] try other sources - it might be possible to scrap data from https://spacedock.info/mod/918/Deflatable%20Heat%20Shield#changelog or https://forum.kerbalspaceprogram.com/index.php?/topic/64987-18x-110x-smokescreen-2814-extended-fx-plugin-18-april-2020/ etc

### DIFF
--- a/GUI/CKAN-GUI.csproj
+++ b/GUI/CKAN-GUI.csproj
@@ -51,6 +51,12 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="Controls\Changelogs.cs">
+      <SubType>UserControl</SubType>
+    </Compile>
+    <Compile Include="Controls\Changelogs.Designer.cs">
+      <DependentUpon>Changelogs.cs</DependentUpon>
+    </Compile>
     <Compile Include="Dialogs\AboutDialog.cs">
       <SubType>Form</SubType>
     </Compile>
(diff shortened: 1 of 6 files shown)


HebaruSan:
(The SSL thing is a false positive, it should clear up on a re-run. Please ignore.)

HebaruSan:
Here's SpaceDock's API documentation, in case you want to try to get the changelogs out of there as well:

- https://github.com/KSP-SpaceDock/SpaceDock/blob/master/api.md

That (plus GitHub) would cover a pretty high proportion of mods.

jkavalik:
> Here's SpaceDock's API documentation, in case you want to try to get the changelogs out of there as well:
> 
> * https://github.com/KSP-SpaceDock/SpaceDock/blob/master/api.md
> 
> That (plus GitHub) would cover a pretty high proportion of mods.

Thank you! Will look into it :)

HebaruSan:
Hi @jkavalik, are you still planning to work on this? It's marked as a draft, which we've taken as a signal not to start reviewing it yet, so please let us know if you were expecting some eyes on it.

HebaruSan:
Hi @jkavalik, if you do find time to return to this and finish it, please let us know and we can re-open it. Closing as incomplete and inactive.

PR #3622, opened 2022-08-08
Create a system menu entry for command prompt

HebaruSan:
## Motivation

Now that we have integrated tab completion in #3515, it might be nice to make the `ckan prompt` UI more easily discoverable, similar to what we did for ConsoleUI in #3052.

## Changes

Now a new "CKAN Command Line" option is added to the system menus when installing the .deb or .rpm packages:

image

Clicking it runs `ckan prompt` in a terminal.


techman83 (approved):
That's pretty neat!

Issue #2848, opened 2019-08-20
[Discussion] GUI Redesign Mock Up


jbrot:
As I mentioned in #2799, I am planning on creating a new GUI for Mac OS X that better integrates with the system. Right now, my plan is to write a new GUI in Xamarin, which will provide a native UI instead of using mono's implementation of Winforms.

Since I will be writing an entire UI either way, this seems like a good opportunity to redesign how the CKAN UI looks. Since Xamarin also supports Windows and Linux (via GTK), I could bring this UI to Windows and Linux if it is received well. This is my first mock up for what a redesigned UI could look like:

![mock](https://user-images.githubusercontent.com/26423613/63377389-08978280-c35e-11e9-94e9-52dcc9169a8f.png)

I would love to hear any feedback. Is this a reasonable direction to take CKAN? If this seems like a reasonable approach, I will then mock up the rest of the UI to make sure the new UI has feature parity with the old. Then, after further discussion, I will actually implement the UI per the mock-ups.

HebaruSan:
For comparison, this was what I came up with for an ElectronUI rewrite (@DasSkelett, an old build on Windows was still working, so finally a screenshot!):

![image](https://user-images.githubusercontent.com/1559108/63379217-62f70a00-c383-11e9-99b3-2d98f9319ad4.png)

The left drawer slides in and out to allow game instance selection.

![image](https://user-images.githubusercontent.com/1559108/63379233-68eceb00-c383-11e9-8b66-840201703eae.png)

Can't show the modules list because fails to load due to file format changes, but it's similar to what you have there.

![image](https://user-images.githubusercontent.com/1559108/63379239-6be7db80-c383-11e9-99f1-99144dc8e052.png)

HebaruSan:
Ahh there we go, mod list:

![image](https://user-images.githubusercontent.com/1559108/63382245-d56ae880-c389-11e9-8014-40a59e25d973.png)

Searching:

![image](https://user-images.githubusercontent.com/1559108/63382282-e7e52200-c389-11e9-85c7-ce6424a5c493.png)

Mod info:

![image](https://user-images.githubusercontent.com/1559108/63382293-ee739980-c389-11e9-91c0-29ddc5e565c2.png)

jbrot:
Very cool! I didn't know there was an Electron UI. Are you still developing it? An Electron UI would solve all the issues I'm setting out to solve with the Xamarin UI.

HebaruSan:
[...]

Issue #2108, opened 2017-09-05
CKAN not friendly with REAL satellites


ndotb:
**CKAN Version:** 
1.22.3

**Operating System:** 
Windows 10 Creators

**The issue you are experiencing:** 
Upon opening, CKAN immediately begins making internet requests without direct authorization.  It does not check if the current connection on Windows 10 is metered.  Its download system does not:

- Prime DNS
- Work with high latency connections
- Resume on failure

In short, CKAN works with toy satellites but has no idea how to manage a real one.

In addition, there is no cancellation token associated with download requests.  The only way to stop it is to go into task manager and kill the program.

**How to recreate this issue:** 
Open it, then try to download on a satellite internet connection.

**CKAN error codes (if applicable):** 
Request timed out, server not found.

I would offer to help, but the build system has become an exercise in obsessive compulsive disorder for a relatively simple application.

politas:
There are settings to turn off automatic updates and automatic refresh, and the CKAN client should be asking you about those two things on first load. The inability to resume on failure is an issue, but we are pretty much limited to what our libraries support in that regard, I believe. You can pre-download mod contents individually from the mod details "Content" tab if that helps.

HebaruSan:
The API for metering _might_ be...

```cs
var cost = NetworkInformation.GetInternetConnectionProfile().GetConnectionCost().NetworkCostType;
bool metered = (cost == NetworkCostType.Fixed || cost == NetworkCostType.Restricted);
```

HebaruSan:
@ndotb, what is meant by "does not work with high latency connections"? What problem specifically happens on high latency connections?

GargamelKSP:
@HebaruSan, per your request, couple comments:
Satellite internet has extremely high ping times.  It's a very rare day if I can get a ping of less than 2000ms to my ISP host site.   Online gaming is not even possible with that type of lag.  I have not encountered any latency issues, yet, but like I said in the 1.4 thread, first time running it.

A scheduler would be a very nice feature to have, to limit updates and downloads to specific times of day, when our connections are not throttled.

HebaruSan:
No response from OP in over 2 years.

PR #3624, opened 2022-08-11
Cache remote build map, fetch in registry refresh

HebaruSan:
## Problems

My network has been a bit unreliable lately, and I've found that during periods of downtime, not only can I not download mods, but even commands like `ckan instance list` simply hang. (First noticed while working on the Electron UI project previewed in the comments of #2848, which just wraps around `CKAN.dll` without additional C# code being run.)

## Cause

When CKAN starts, it loads all the game instances pretty early. To do this, it first has to load the known game versions so it can check each instance's version. Currently we always use the remote build map for this, via an HTTPS request to GitHub. If the network isn't working, it just sits there waiting for timeout.

This isn't a great model. As pointed out in #2108, CKAN should refrain from creating network connections without the user's awareness, especially if it isn't necessary. Furthermore, we already have an embedded build map compiled into CKAN that we essentially never use, because the remote build map is always retrieved first.

An additional quirk is that the command line parameters are handled _after_ all this, so even if you pass `--debug`, the `log.Debug` output of these steps is never printed.

## Changes

- Now instead of retrieving the remote build map at every startup, we check for a locally cached copy (in the same directory as `config.json`). If found, we use it, otherwise we fall back to the embedded build map. This eliminates the network call at startup and would allow commands like `ckan instance list` to complete while offline.
- The updating code used by `ckan update` and GUI's Refresh button is updated to fetch the remote build map and save a cached copy for future startups, since this represents a time when the user has authorized network activity.
- Now before full parsing of the command line parameters, we check for `--debug` or `--verbose` and set the log level if found. This way these arguments can be used to see activity before the full parsing happens, including the build map's initialization.

### DIFF
--- a/Core/Games/KerbalSpaceProgram/GameVersionProviders/KspBuildMap.cs
+++ b/Core/Games/KerbalSpaceProgram/GameVersionProviders/KspBuildMap.cs
@@ -1,9 +1,12 @@
 ﻿using System;
+using System.Linq;
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
+
 using log4net;
 using Newtonsoft.Json;
+
 using CKAN.Versioning;
 using CKAN.Configuration;
 
@@ -30,6 +33,9 @@ namespace CKAN.GameVersionProviders
         // TODO: Need a way for the client to configure this
         private static readonly Uri BuildMapUri =
             new Uri("https://raw.githubusercontent.com/KSP-CKAN/CKAN-meta/master/builds.json");
+        private static readonly string cachedBuildMapPath = Path.Combine(
+            Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData),
+            "CKAN", "builds.json");
 
         private static readonly ILog Log = LogManager.GetLogger(typeof(KspBuildMap));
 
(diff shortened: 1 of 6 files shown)


Core/Games/KerbalSpaceProgram/GameVersionProviders/KspBuildMap.cs
```diff
@@ -106,20 +112,28 @@ private bool TrySetBuildMap(string buildMapJson)
...
         }
+
+        private bool TrySetCachedBuildMap()
+        {
+            try
+            {
+                Log.Debug("Getting cached build map");
+                return TrySetBuildMap(File.ReadAllText(cachedBuildMapPath));
+            }
+            catch (Exception e)
```
techman83, line 123: Are we doing anything with this exception we're catching? We're getting a few notes about it.

techman83 (approved):
That seems like a super sensible improvement! Do we have any tests for this?

From my cursory read, nothing stood out apart from the unused variable. So lgtm!

Core/Games/KerbalSpaceProgram/GameVersionProviders/KspBuildMap.cs
HebaruSan: No; unlike the other `TrySet*BuildMap` functions, we don't want to print a warning for this one because we _expect_ it to be missing the first time. I've removed this variable.

HebaruSan:
As for tests, I don't know how this could be tested, since it depends on writing to and reading from `~/.local/share/CKAN/builds.json`. And as far as I can tell, there were no existing tests for any of the old build map stuff.

HebaruSan:
For fun I took the latest build for a spin in my Debian and Fedora VMs with their network devices disabled, and the commands that don't need network access all worked fine. :tada:

PR #3631, opened 2022-08-16
Check free space before downloading

### LINKED ISSUE #3581: Unhandled Exception: Not Enough Space on Disk
Unhandled exception when trying to download to a full hard drive causes CKAN to crash. Should check for enough space before downloading and notify the user if there is not enough space on the disk from which CKAN was launched (NOT the disk that KSP is installed on.)

As is, the download crawls and stalls as it reaches the limit of free space, then the crash report looks more dev-oriented than user oriented so it's not immediately obvious what is the cause of the issues.

HebaruSan:
## Motivation

Users periodically report issues with running out of disk space. (Given how cheap storage is these days, we can only credit the extremely creative developers of planet packs for taking up so many gigabytes with their beautiful textures.) This can happen at three points:

- The downloader downloads ZIP files to the `%TEMP%` folder, which can fill up
- Then it copies/moves downloaded files into the cache folder, which can fill up
- Then the installer extracts mod content from the ZIPs into the game folder, which can fill up

Currently CKAN does not attempt to handle or prevent this. The exception is simply thrown and the user gets to read a stack trace.

## Previous investigations

Last time I looked at this I hit a speedbump due to the lack of a built-in mapping from directories to drives in .NET, which is probably because of cross-platform considerations. Whereas on Windows, all `DriveInfo.RootDirectory.FullName` properties return something like `C:\`, `D:\`, etc., on Unix these can be just about any path since Unix is much more flexible regarding where you can mount devices. The logic needs to work for both.

## Considered and not done

It would be nice to catch and handle these exceptions. Unfortunately, they're generic `IOException`s rather than something specific to disk space, and there's no programmatic way to check whether the exception relates to disk space. So we would be catching a lot of other unrelated exceptions in the same spot, and we would not be able to generate a useful error message from them.

## Changes

- Now a new extension method allows `DirectoryInfo.GetDrive()` to return the `DriveInfo` object associated with a folder. It works by comparing (case-insensitive on Windows) the pieces of the `FullName` properties of the given dir and `DriveInfo.RootDirectory`. Empty pieces are pruned to make sure `"C:\\".Split("\\")` doesn't return `["C:", ""]` which would fail to match a longer path without a blank piece in that position.
- Now `CKANPathUtils` is `static` (it always should have been, just an oversight)
- Now `CKANPathUtils.CheckFreeSpace` uses the new `.GetDrive()` to check the free space of a given folder and throws an instance of new class `NotEnoughSpaceKraken` if it isn't enough, with a custom error message, which GUI catches and pretty-prints in the install log
- Now we `CheckFreeSpace` in the game dir before installing, using the `install_size` metadata from #3568
- Now we `CheckFreeSpace` in the temp dir and the cache dir before downloading
- Now we `CheckFreeSpace` in the game dir again after the downloads complete, just in case the game dir and cache dir are on the same device
- Now `CkanModule.FmtSize` goes up to TiB rather than just GiB, since we're going to be reporting disk sizes
- Now `ckan cache list` prints the free space of your cache's device
- Now GUI's settings window also shows the free space of the cache dir

This will alert users to disk space problems earlier in the install flow so they can free up space in response to a friendlier message.

Fixes #3581.

image

image

image

### DIFF
--- a/Cmdline/Action/Cache.cs
+++ b/Cmdline/Action/Cache.cs
@@ -1,8 +1,9 @@
 using CommandLine;
 using CommandLine.Text;
 using log4net;
-using CKAN.Configuration;
 using Autofac;
+
+using CKAN.Configuration;
 
 namespace CKAN.CmdLine
 {
@@ -229,10 +230,11 @@ namespace CKAN.CmdLine
 
         private void printCacheInfo()
         {
-            int fileCount;
-            long bytes;
-            manager.Cache.GetSizeInfo(out fileCount, out bytes);
-            user.RaiseMessage(Properties.Resources.CacheInfo, fileCount, CkanModule.FmtSize(bytes));
+            manager.Cache.GetSizeInfo(out int fileCount, out long bytes, out long bytesFree);
+            user.RaiseMessage(Properties.Resources.CacheInfo,
+                              fileCount,
+                              CkanModule.FmtSize(bytes),
+                              CkanModule.FmtSize(bytesFree));
         }
 
         private IUser               user;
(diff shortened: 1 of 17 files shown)


techman83 (approved):
Neat, that looks good. Disk handling can be a chore, so nice work!

PR #3557, opened 2022-04-06
One concurrent download per host for all hosts

HebaruSan:
## Background

In #3054 we changed the async downloader to sequentialize downloads from GitHub to avoid triggering GitHub's download throttling. Only one GitHub download will ever be active concurrently, and this seems to have solved the problems we were having with throttling.

## Problems

- I have been installing big groups of mods to test #3525, and I've found that even SpaceDock does not seem to like having a lot of parallel downloads; any beyond 2 or 3 seemed to fail and fallback to archive.org or require retrying the installation. #3314 may be a report of this same problem, if the hosts of the affected downloads were mainly SpaceDock.
- In addition, `Mono.Unix.UnixIOException` can be thrown if there are many concurrent downloads in progress (see #3343 and #3555).

## Causes

- It's hard to know for sure, but I suspect that the first few downloads from SpaceDock quickly fill the pipe, and then the remaining ones either fail to start due to congestion or are deliberately throttled by the server.
- Too many UI updates at once can overwhelm Mono. A fix for this was submitted in mono/mono#21136, but it shows no signs of having been looked at by the Mono team in the last 9 months, if such a team still exists. CKAN can be affected by this Mono bug if there are many concurrent downloads, because each of them has its own progress bar that we update every time another chunk of the download arrives.

## Changes

Now the single concurrent download behavior is applied to all download hosts, per host. When you start a large group of downloads, we will start one from each host in the group (GitHub, SpaceDock, etc.) in parallel, and as each download finishes the next one from that host will be started, until they're all done. This should fix problems arising from too many downloads from one host.

Fixes #3314 (probably, depending on which downloads were failing for that user).
Fixes #3555 (hopefully, by reducing the number of GUI updates during downloads).

PR #3635, opened 2022-08-17
Many improvements for failed downloads

### LINKED ISSUE #873: Generate (and install?) a reduced changeset after failing to download a mod
when installing a mod and for somereason the download fails this kills the install.  Now on a single mod this is not a big deal you just restart it and carry on.  But whn you have a whole list of mods checked this stops them all and you must go back and redo all of them.

So the idea is if a mod fails to DL rather than fail the install just retry it or ignore that one and do the rest of the list and then just fail that single mod.

### LINKED ISSUE #1636: Using download button on contents tab causes System.Net.WebException on 404 error
**CKAN Version:** 
 v1.16.1
**Operating System:** 
Windows 8.1
When downloads fail after pressing button in contents tab. There is a bug in the 404 error handling leading to a crash. System.Net.WebException: The remote server returned an error: (404) Not Found. There is nothing we can do if host breaks down but an external remote server error should not be crashing CKAN.
Select a mod with a known broken download link that produces the standard 404 dialogue box. Like one we know is still hosed on Kerbal Stuff for example. In the mod description section we have three tabs. Metadata, Relationships and Contents.

On the contents tab there is a download button to for mods not cached. When this is pressed the download naturally fails to happen. We get the standard 404 error message and dismiss button which is ok. On closing this we then get a crash  

*************\* Exception Text **************
Uh oh, the following things went wrong when downloading...

System.Net.WebException: The remote server returned an error: (404) Not Found.
   at System.Net.HttpWebRequest.EndGetResponse(IAsyncResult asyncResult)
   at System.Net.WebClient.GetWebResponse(WebRequest request, IAsyncResult result)
   at System.Net.WebClient.DownloadBitsResponseCallback(IAsyncResult result)

### LINKED ISSUE #3588: [Feature] Keep downloading mods on error
Problem
-------
when one mod fails to download and forces you to redownload all the mods and reselect mods

Suggestions
-----------
Don't stop on error, finish and then retry errored mods

### LINKED ISSUE #3604: [Bug] Uninstallation should be reverted if subsequent installation in same changeset fails
Background
----------
* CKAN Version: 1.31.0
* KSP Version: 1.12.3.3173

Yes, however not to any of the mods involved.

Problem
-------
**Describe the bug**
**Steps to reproduce**
(Combining these two as it frankly makes more sense in this case.)
I had thought to replace ScienceAlert Relarted with [x]Science!, and while at it, try out Parallax.
CKAN correctly uninstalled Science Alert Realerted, then proceeded to the download step of the installation half of my changes.
Due to dodgy internet, the sizable download of Parallax got interrupted, resulting in a download error.
Upon choosing "Retry", CKAN alerted me that ScienceAlert Realerted isn't installed, no doubt since it was still trying to use the same changelist as when I started out, when I asked it to uninstall the mod.

**Expected behavior**
Either CKAN undoing the entire process and restoring the pre-uninstallation state, or not worrying about the uninstallation issue, either because it's unimportant, or due to the program handling uninstallations and installations as two entirely separate processes with their independent checklists. (If such a thing happens to someone who's assembling a large list of changes, it could be quite annoying, as backing out of the error appears to have made CKAN clear the change list.)

Sadly, while switching back to the main tab to double check which folders ScienceAlert touches so I can be sure it was indeed uninstalled fully, I lost the tab with the error so I can't share it. However, I expect "[X Mod] isn't installed" to identify it well enough.
```
PLACE CODE HERE !
```

Overall, just a minor inconvenience for me, but it might be worth taking a look at it to avoid others' headaches. Thank you for your great work!

HebaruSan:
## Background

We have had a lot of recent user frustration with how CKAN reacts when downloads fail. Such errors are hard to reproduce with the full app running because they happen randomly according to users' network conditions. However, I recently hit upon a strategy: edit `CKAN/registry.json` and change the `download` properties of the latest versions of some mods so they're invalid (I added `/BROKEN_LINK/` at the start of each URL's path). Note that this will only work for mods with non-redistributable licenses (generally `restricted`), because otherwise the downloader will fall back to `archive.org` and succeed. The ones in the screenshot below work well for this, if it saves reviewers the trouble of finding them; you can right click them to purge them from the cache if they're already downloaded.

## Problems

With my mildly corrupted registry in place, I was finally able to conduct a proper investigation with failures-on-demand, and I discovered several more issues, some of which may simply be root causes of mysterious unreproducible issues that were reported previously:

- `ckan gui --show-console` gets confused and thinks `--show-console` is an identifier that you want it to highlight, resulting in a superfluous "Not found" message in the log :facepalm: 
- The async downloader has a potential race condition in its manipulation of `queuedDownloads` with `FirstOrDefault`: one thread can change the list while the other is searching it, which throws an exception
- The async downloader considers failed downloads to still be active, so they block downloads from the same host from starting (see #3557)
- If you try to cancel a download, the async downloader halts the currently in progress downloads but then proceeds to start up any remaining queued downloads
- The async modules downloader can't be re-used to retry a batch of downloads because it already has all the modules in `this.modules` and therefore thinks they're in progress
- If you choose the `not:cached` filter and then download a mod to cache, it doesn't disappear even though it no longer matches the filter
- If your changeset contains both removals and installs, and one of the installs fails, the removal step is not rolled back. This is bad if you were intending to switch from the removed mods to the installed mods, because now you need to fall back to your old mods and have to go find them and reinstall, if you can even remember what they were.
- The code related to the `Wait` tab is messy and hard to change without breaking something else

## Changes

Now if any downloads in your changeset fail, this popup appears over the install progress screen:

image

Clicking either checkbox in a row will toggle both of that row's checkboxes; rather than representing this with a single checkbox, I wanted to make it clear that both choices are actions, and you are sorting the mods into one group or the other.

If you click "Abort whole changeset", then any uninstallations are rolled back and you are returned to the main mod list with your changeset intact. If you click "Retry", then any mods with a check mark in the "Skip" column will be removed from your changeset, along with any mods depending on them, and CKAN will try again with the mods with check marks in the Retry column plus the ones that didn't fail.

- Now `--show-console` isn't passed to GUI since it's handled before that
- Now a `lock` mutex protects the queued downloads collection, so only one thread can change it at a time
- Now if a download fails, the async downloader allows a queued download from the same host to replace it (since we will be reporting and handling download errors later in the popup)
- Now cancelling downloads purges the queued downloads
- Now the async module downloader clears `this.modules` after a batch finishes, successfully or otherwise, so it can be used to retry failed downloads
- Now all of GUI's many `BackgroundWorker`s used for interacting with `Wait` are replaced by one that lives inside `Wait`, for better encapsulation. The functions formerly assigned to `DoWork` and `RunWorkerCompleted` are now parameters to `StartWaiting`. In addition, some of `Wait`'s functions are changed from `public` to `private` and only manipulated from well defined, well known interfaces. This will make it easier to keep track of where these things are being used and ensure they don't end up in a weird state.
- `Wait.OnCancel` is now used by other code to react to cancellation instead of `Main.cancelCallback`, and the Cancel button is only available if the `cancelable` parameter to `StartWaiting` is `true` (which it is for installation and downloading to cache)
- The Retry button starts out hidden and is shown with `Wait.RetryEnabled = true` when the failure handler supports it
- Now `ModuleDownloadErrorsKraken`'s list of modules and their corresponding download errors is public, so we can use it in the popup
- `InstallMods` now catches `ModuleDownloadErrorsKraken` and passes its module/exception info to the new popup, then either throws a cancellation kraken or removes the skipped mods and their depending mods from `toInstall`
- Now we always show the selected mod's info on returning to `ManageMods`
- Now if you cancel an install or a download to cache, we return you to `ManageMods` immediately since you don't benefit from reading text that says you cancelled
- Now the changeset is **not** cleared after a failed install, so users will no longer have the experience of selecting dozens of mods and then trying to re-select them after a failure
- Now we refresh the mod list after downloading to cache
- Now one big transaction covers the entire remove / install / upgrade flow, so if the install fails, the whole changeset is reverted
- `Registry.FindReverseDependencies` now has a `satisfiedFilter` parameter that accepts a function from `RelationshipDescriptor` to `bool`, which we use to avoid removing a depending mod from the changeset if the broken dependency is virtual; this way, the user can Skip a virtual dependency and then make a different choice

This should provide a much more stable and pleasant experience when downloads fail.

Fixes #873.
Fixes #1636.
Fixes #3588.
Fixes #3604.

### DIFF
--- a/Cmdline/Main.cs
+++ b/Cmdline/Main.cs
@@ -288,7 +288,10 @@ namespace CKAN.CmdLine
             // TODO: Sometimes when the GUI exits, we get a System.ArgumentException,
             // but trying to catch it here doesn't seem to help. Dunno why.
 
-            GUI.GUI.Main_(args, manager, options.ShowConsole);
+            // GUI expects its first param to be an identifier, don't confuse it
+            GUI.GUI.Main_(args.Except(new string[] {"--verbose", "--debug", "--show-console"})
+                              .ToArray(),
+                          manager, options.ShowConsole);
 
             return Exit.OK;
         }
--- a/ConsoleUI/InstallScreen.cs
+++ b/ConsoleUI/InstallScreen.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Transactions;
 using System.Collections.Generic;
+
 using CKAN.ConsoleUI.Toolkit;
 
 namespace CKAN.ConsoleUI {
--- a/Core/HelpURLs.cs
+++ b/Core/HelpURLs.cs
@@ -17,6 +17,7 @@ namespace CKAN
         public const string Filters = "https://github.com/KSP-CKAN/CKAN/pull/3458";
         public const string Labels = "https://github.com/KSP-CKAN/CKAN/pull/2936";
         public const string PlayTime = "https://github.com/KSP-CKAN/CKAN/pull/3543";
+        public const string DownloadsFailed = "https://github.com/KSP-CKAN/CKAN/pull/3635";
 
         public const string Discord = "https://discord.gg/Mb4nXQD";
     }
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -4,11 +4,13 @@ using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Transactions;
+
 using ICSharpCode.SharpZipLib.Core;
 using ICSharpCode.SharpZipLib.Zip;
 using log4net;
 using ChinhDo.Transactions.FileManager;
 using Autofac;
+
 using CKAN.Extensions;
 using CKAN.Versioning;
 using CKAN.Configuration;
--- a/Core/Net/NetAsyncDownloader.cs
+++ b/Core/Net/NetAsyncDownloader.cs
@@ -105,6 +105,7 @@ namespace CKAN
         /// </summary>
         public event Action<Net.DownloadTarget, long, long> Progress;
 
+        private readonly object dlMutex = new object();
         private List<NetAsyncDownloaderDownloadPart> downloads = new List<NetAsyncDownloaderDownloadPart>();
         private List<Net.DownloadTarget> queuedDownloads = new List<Net.DownloadTarget>();
         private int completed_downloads;
@@ -143,11 +144,13 @@ namespace CKAN
         {
             if (shouldQueue(target))
             {
+                log.DebugFormat("Enqueuing download of {0}", target.url);
                 // Throttled host already downloading, we will get back to this later
                 queuedDownloads.Add(target);
             }
             else
             {
+                log.DebugFormat("Beginning download of {0}", target.url);
                 // We need a new variable for our closure/lambda, hence index = 1+prev max
                 int index = downloads.Count;
 
@@ -167,7 +170,7 @@ namespace CKAN
 
                 // And schedule a notification if we're done (or if something goes wrong)
                 dl.Done += (sender, args) =>
-                    FileDownloadComplete(index, args.Error);
+                    FileDownloadComplete(index, args.Error, args.Cancelled);
 
                 // Start the download!
                 dl.Download(dl.target.url, dl.path);
@@ -187,7 +190,9 @@ namespace CKAN
         {
             return downloads.Any(dl =>
                     dl.target.url.Host == target.url.Host
-                    && dl.bytesLeft > 0);
+                    && dl.bytesLeft > 0
+                    // Consider done if already tried and failed
+                    && dl.error == null);
         }
 
         /// <summary>
@@ -196,7 +201,7 @@ namespace CKAN
         /// <param name="urls">The downloads to begin</param>
         public void DownloadAndWait(ICollection<Net.DownloadTarget> urls)
         {
-            if (downloads.Count > completed_downloads)
+            if (downloads.Count + queuedDownloads.Count > completed_downloads)
             {
                 // Some downloads are still in progress, add to the current batch
                 foreach (Net.DownloadTarget target in urls)
@@ -230,6 +235,8 @@ namespace CKAN
             // If the user cancelled our progress, then signal that.
             if (old_download_canceled)
             {
+                // Ditch anything we haven't started
+                queuedDownloads.Clear();
                 // Abort all our traditional downloads, if there are any.
                 foreach (var download in downloads.ToList())
                 {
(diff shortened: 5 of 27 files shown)


HebaruSan:
Hi @NathanPeake, @epower53, and @Ictiv, there is a test build for these changes under the Artifacts dropdown here:

https://github.com/KSP-CKAN/CKAN/pull/3635/checks

If you have a chance, please try it out and let us know if you still have problems with downloads.

HebaruSan:
Noting some scenarios that came up in Discord...


1. Start installing a mod with a virtual dependency
2. CKAN prompts you to pick a mod to satisfy the depends
3. You pick a mod
4. The dependency fails
5. You Skip the dependency in the popup

At this point, the current code would remove the depending mod from the changeset. But it would be nice to have special handling for virtual dependencies so the user could choose a different one instead.

If the popup contains a mod and its dependency, you are able to choose Skip for the dependency and Retry for the depending, but the depends handling logic will remove both from the changeset despite your choice of Retry. As a "stretch goal" it would be cool to enforce relationships within the UI.

If you choose to install a mod with dependencies, and you Skip that mod, do the dependencies vanish from the changeset as well since they're no longer needed? I _think_ I saw that happening but it would be good to be sure.

techman83 (approved):
Nothing stands out from my read over, should make downloads that barf halfway far less annoying :tada:

HebaruSan:
Investigating the virtual dependency idea, I discovered that failed **non**-virtual dependencies can cause the popup to appear but aren't listed in it; the changeset only includes the user's selections, not the dependencies that automatically get pulled in. I'll have to add some logic to handle that...

HebaruSan:
> 1. Start installing a mod with a virtual dependency
> 2. CKAN prompts you to pick a mod to satisfy the depends
> 3. You pick a mod
> 4. The dependency fails
> 5. You Skip the dependency in the popup
>    At this point, the current code would remove the depending mod from the changeset. But it would be nice to have special handling for virtual dependencies so the user could choose a different one instead. This would actually not be that hard; if we simply don't remove the depending mod, the prompt would appear when you Retry. So all we really need is a way to detect virtual dependencies.

Addressed this in latest commit; we now no longer remove mods from the changeset if their broken dependencies are virtual, so the user can go back and make another choice.

> If you choose to install a mod with dependencies, and you Skip that mod, do the dependencies vanish from the changeset as well since they're no longer needed? I _think_ I saw that happening but it would be good to be sure.
> Investigating the virtual dependency idea, I discovered that failed **non**-virtual dependencies can cause the popup to appear but aren't listed in it; the changeset only includes the user's selections, not the dependencies that automatically get pulled in. I'll have to add some logic to handle that...

The dependencies aren't in the changeset, but the latest commit retrieves the full changeset where needed for the changeset pruning logic.

Haven't looked at enforcing relationships in the UI. This feels like something we can do without and return to if there's demand.

HebaruSan:
> Note that this will only work for mods with non-redistributable licenses (generally `restricted`), because otherwise the downloader will fall back to `archive.org` and succeed. The ones in the screenshot below work well for this, if it saves reviewers the trouble of finding them; you can right click them to purge them from the cache if they're already downloaded.

Another approach to this caveat is to change the `license` property of whatever mod you want to use to 'restricted' while you're editing `CKAN/registry.json`. That's all it takes to make CKAN ignore the fallback URL.

NathanPeake:
> Hi @NathanPeake, @epower53, and @Ictiv, there is a test build for these changes under the Artifacts dropdown here:
> 
> https://github.com/KSP-CKAN/CKAN/pull/3635/checks
> 
> If you have a chance, please try it out and let us know if you still have problems with downloads.

This is what I have been wanting for a long long time. thanks for adding this feature.

PR #3638, opened 2022-08-19
Show reverse relationships

### LINKED ISSUE #3616: [Feature] Show reverse relationships
Problem
-------
One of my 267 installed mods is causing a problem.  To narrow it down, I use a quasi binary search - delete half the mods, test.  If the problem went away, the offending mod is in the half deleted, if not, it's in the half still installed.  Rinse and repeat.  In this way, even with very large modlists, you can find the offending mod in only 7 or 8 tries.

However, dependencies complicate this.  In a large modlist, it's often the case that deleting half the mods will also remove a great many mods (that are dependent on some of the ones you deleted) from the group you wish to leave installed.

It would be great if information about which mods depend on a particular other mod were easily viewable.  For example, ASET Props and ASET Avionics depend on ASET Agency.  ASET Props and ASET Avionics both show this dependency in their relationship tab.  ASET Agency, however, does not mention that there are two other mods that depend on *it*.

This information would be *really* useful when going through the modlist and deciding which mods to uninstall and which to leave as part of the quasi binary search.

I know you can search for e.g. "dep:ASETAgency" (which shows that CKAN has this information) but frankly, having to interact with a search bar and type in a search query for every single mod in the group I want to examine (133 mods in the first round of testing) is just way too much work and friction.  Especially when there is an info tab labeled "Relationships" that seems tailor made for displaying just this sort of info.

Suggestions
-----------
I'd like the Relationships tab to display information on *what other mods depend on this mod*.  I don't know what you'd call it; I think you already use "supports" to mean something different.

**Describe alternatives you've considered**
The search bar, but it's just too damn tedious for 100+ mods at a time.

**Additional context**
I asked about this on discord, and HebaruSan was kind enough to respond.  In addition to educating me a bit (thank you!), he also gave some pushback.  His argument against had to do with 'which mod logically "owns" a relationship', and metadata, and 'some modders complaining that the recommendations/suggestions screen shows things they didn't approve'.

Respectfully, that is a mod-author-centric, data-structure-centric argument.  I'm a software engineer so I totally understand that viewpoint.  But as a *user*, I have a very different point of view.  My *user-centric* argument is this:

CKAN has this information, and as a user I really don't care where it came from or how CKAN knows, just that it does.  But that information is currently only accessible through a tedious bit of UI, when there is an already-allocated bit of screen real-estate dedicated to displaying just this sort of information.  I think it should go there.

HebaruSan:
## Motivations

- The relationships tab only shows the current mod's dependencies (etc.) on other mods, but some users may wish to browse other mods' dependencies (etc.) on the current mod, for example if they are trying to decide which mods would be the least disruptive to uninstall when trying to narrow down which mod is causing a bug in-game. This information is available via a `dep:identifier` search but not easily accessible.
- The mod info control loads everything in the foreground, which can slow down response times when the user clicks a mod:
  - The relationships tab can take a long time to load fully if you use the new reverse relationships option described below because we have to search all mods and check them (see `ModuleManager`)
  - The contents tab can load slowly if a cached mod has a big ZIP with a lot of files (see `BluedogDB`)
  - The versions tab can load slowly if a mod has many versions and a lot of deep, complex dependencies because we have to check them to determine whether each version of the mod is installable (see `RP-0`)
- The contents tab is a user-unfriendly flat list of long pathnames, despite the underlying data being tree-shaped
- The relationships tab's thumb-up icon for recommendations is very heavy-looking in contrast to the other relationship icons, which makes it stand out more than it needs to:
  image
- The relationships tab shows installed mods in a different format for no good reason (version and sub-relationships missing), and shows mods that conflict with installed mods as non-indexed:
  image

## Changes

- Now only the active tab of mod info loads initially to reduce CPU load, and the others load if you activate them
- Now the relationships tab has a "Reverse relationships" checkbox:
  image
  When unchecked, the usual relationships are shown as before. When checked, the reverse relationships are shown instead and the checkbox will revert to unchecked when you switch mods. If you shift-click or ctrl-click, it becomes checked-and-gray (the indeterminate state), the reverse relationships are shown, and the state of the checkbox is "sticky" (persists when switching mods, for when the user wants to look at reverse relationships for many mods in a row)
- Now the relationships tab is loaded in a background thread to keep the GUI responsive, and the mouse cursor will be an hourglass if hovered over it while loading. The grandchildren dependencies are loaded separately in yet more background threads, because they are the slowest part (especially for reverse relationships). We start one thread for each group of ten dependencies (because it flickers a lot if we update for each mod), and if the last group is smaller than ten, we combine it with the one before to reduce the chance of it finishing before the other groups. In practice this means you can scroll through the list and watch the `+` icons appear for nodes with children.
- Now the relationships tab uses the outlined version of the thumb-up icon for recommendations, which will blend in better with the other icons:
  image
- Now the relationships tab shows the version and sub-relationships for installed modules, and mods that conflict with installed mods are displayed normally
- `ModuleInstaller.GetModuleContentsList` now returns an empty sequence instead of `null` if it can't find any files, to make it easier for `System.Linq`-style code to use it
- `ModuleInstaller.GetModuleContentsList` now skips directories and only returns files, because otherwise we would need to know which entries are directories and which are files, and it's really hard to do that just from strings. We don't need the directories anyway.
- Now the contents tab is loaded in the background to keep the GUI responsive, and the mouse cursor will be an hourglass if hovered over it while loading.
- Now the contents tab displays a collapsible tree of the contents of the ZIP, with file, folder, and zipped-folder icons care of [Google's Material Icons](https://material.io/icons/) (as we did for the relationship icons in #2226, Apache 2.0 license):
  image
- Now the top node of the contents tab shows the file name in cache instead of the mod name for cached mods, since this information wasn't otherwise available in the GUI
- Now the top node of the contents tab shows both the name and version for uncached mods, since the version may be useful
- Now the fast part of the mod versions list, the text, is loaded in the foreground, and the slow part, the colors indicating compatibility, is loaded in a background thread to keep the GUI responsive, and the mouse cursor will be an hourglass if hovered over it while loading. In practice this means you can watch the rows turn green one at a time for mods that load slowly.
- Very minor unrelated usability improvement, the columns in the changeset tab now resize themselves to fit the contents, so you don't need to resize them manually to read long strings

After these changes, there is no noticeable delay when you click a row in the main mod list.

Fixes #3616.


HebaruSan:
@tjanke42, FYI, we got some feedback that the tri-state UI was a bit confusing as-is, and I agreed, so I've changed it so the checkbox works normally unless you ctrl-click or shift-click it, in which case it goes to the special "sticky" state.

techman83 (approved):
This is a really neat, constrained PR. LGTM :tada:

tjanke42:
Damn, you guys are awesome.  Based on the comments, that's a lot of work you did, HebaruSan.  I appreciate it very, very much.  You da man!

PR #3639, opened 2022-08-20
Hide tray icon at exit via ctrl-C

HebaruSan:
## Problems

- On Windows, the tray icon doesn't disappear at close if you kill CKAN with ctrl-C in the console. (You need to use the `--show-console` parameter to keep it visible after the GUI appears.) You can even accumulate several of them if you do it over and over:
  image
- If you use the tray icon to open the settings while CKAN doesn't have focus, it appears at the upper left corner of your monitor
- Tray icons have the ability to display tooltips, but we don't have one

## Changes

- Now the `Console.CancelKeyPress` ctrl-C event handler hides the tray icon
- Now we always pass a reference to the main form when we show the settings dialog, which will ensure it is properly centered on the main window
- Now the tray icon has a tooltip that says "CKAN"

### DIFF
--- a/GUI/Main/Main.cs
+++ b/GUI/Main/Main.cs
@@ -48,9 +48,6 @@ namespace CKAN.GUI
 
         public GUIUser currentUser;
 
-        private bool enableTrayIcon;
-        private bool minimizeToTray;
-
         public static Main Instance { get; private set; }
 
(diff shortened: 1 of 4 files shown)


techman83 (approved):
LGTM, this is something I've had to do in other applications

PR #3641, opened 2022-08-20
Debounce search events on all platforms

### LINKED ISSUE #3402: [Feature] Delay search while typing

### LINKED ISSUE #3620: [Feature] Add a delay before searching

HebaruSan:
## Problem

Mod searches in GUI that return many results can be noticeably slow on Windows. Usually this means one- or two-character searches.

## Cause

The search logic that we perform is quick, but adding a lot of rows to `DataGridView` is unavoidably slow. Since WinForms requires that all UI updates happen in the foreground GUI thread, we can't improve this by simply moving logic to a background thread.

## Background

Mono platforms already have a delay between keystrokes and search, so they don't suffer as much from this today. However, the implementation of the delay is embedded within and very specific to `EditModSearch`, even though this is a generic operation that could be useful elsewhere.

## Changes

- Now there is no longer anything platform-specific in the mod search logic
- Now a new `Util.Debounce` function implements a generic "[debounce](https://www.freecodecamp.org/news/javascript-debounce-example/)" operation, which is a term for handling a very busy event queue by postponing slow operations until there's a given delay between the events
- Now `EditModSearch` uses `Util.Debounce` instead of its own `Timer` object to debounce the search editing events. The logic to control it is now consolidated into new functions `ImmediateHandler`, `SkipDelayIf`, `AbortIf`, and `DelayedHandler`.
  - Pressing Enter still triggers an immediate search
  - Typing more than 4 characters without a prefix (e.g `dep:`) also triggers an immediate search, because such a search will match a small number of rows and therefore be fast
- Now `EditModSearchDetails` forwards events to `EditModSearch` rather than determining on its own when an immediate search should happen, so it can share in the generic debounce logic

Fixes #3402.
Fixes #3620.
Closes #3419 (breaks immediate searching on Enter keypress).

### DIFF
--- a/GUI/Controls/EditModSearch.Designer.cs
+++ b/GUI/Controls/EditModSearch.Designer.cs
@@ -80,8 +80,6 @@
             this.FilterCombinedTextBox.Size = new System.Drawing.Size(247, 26);
             this.FilterCombinedTextBox.TabIndex = 17;
             this.FilterCombinedTextBox.Enter += new System.EventHandler(this.FilterTextBox_Enter);
-            this.FilterCombinedTextBox.TextChanged += new System.EventHandler(this.FilterCombinedTextBox_TextChanged);
-            this.FilterCombinedTextBox.KeyDown += new System.Windows.Forms.KeyEventHandler(this.FilterTextBox_KeyDown);
             resources.ApplyResources(this.FilterCombinedTextBox, "FilterCombinedTextBox");
             //
             // ExpandButton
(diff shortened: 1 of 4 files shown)


techman83 (approved):
Looks sensible! :tada:

PR #3642, opened 2022-08-21
Clean up window disabling

HebaruSan:
## Problem

In current dev builds, if you turn on automatic registry refreshing at launch in the settings, you can sometimes end up with the main menu disabled and the close button not working at startup.

## Causes

- In #3635 and #3637 we tried to improve the `Wait` tab's encapsulation and how it's used, and some of those updates must have created a state where we disable the window and don't re-enable it, but I have not been able to pin down exactly how it happened.
- `Main.SwitchEnabledState` is a toggle that keeps state in a private `_enabled` variable, but every place that calls it wants specifically to either enable or disable the window; there is no point where a toggle actually makes sense like it does with a checkbox. If that state gets out of sync unexpectedly, then those calls all flip, we disable when we mean to enable and vice versa, and we end up with a disabled window that can't be re-enabled.

## Changes

- `Main.SwitchEnabledState` from `MainRepo.cs` is split into `Main.EnableMainWindow` and `Main.DisableMainWindow` in `Main.cs`, and all calling code is updated to use the right one based on what is being done at that point
- Every other place that I could find that enables or disables the tabs and menus is also updated to use `Main.EnableMainWindow` and `Main.DisableMainWindow`
- The unused parameter of `Wait.Finish` and `HideWaitDialog` is removed

This fixes the bug and should help keep things more consistent in the future.

### DIFF
--- a/GUI/Controls/Wait.cs
+++ b/GUI/Controls/Wait.cs
@@ -241,7 +241,7 @@ namespace CKAN.GUI
             });
         }
 
-        public void Finish(bool success)
+        public void Finish()
         {
             OnCancel = null;
             Util.Invoke(this, () =>
--- a/GUI/Main/Main.cs
+++ b/GUI/Main/Main.cs
@@ -590,8 +590,6 @@ namespace CKAN.GUI
                         continue;
                     }
 
-                    menuStrip1.Enabled = false;
-
                     InstallModuleDriver(registry_manager.registry, module);
                 }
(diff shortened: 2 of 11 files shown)


HebaruSan:
I'm going to self-review and merge this so I can rebase other PRs and hand out test builds that don't have this problem, since I'm more confident in this branch than I am in master at this point...

PR #3648, opened 2022-08-29
Improve translation process

Olympic1:
## Motivation

The current translation process is difficult for users who that don't know or can't work with Git. The users that can, can sometimes forget some translations.

This will improve that process for those users as they just have to create an account on Crowdin and can start translating with very easy steps. Every translatable string is shown for them so none is forgotten.

## Changes

Added a configuration file to upload the translatable files to Crowdin.

### DIFF
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -1,0 +1,9 @@
+files:
+  - source: /GUI/**/*.resx
+    ignore:
+      - Properties/
+      - Localization/
+      - HintTextBox.resx
+    translation: /GUI/Localization/%locale%/%file_name%.%locale%.%file_extension%
+  - source: /**/Resources.resx
+    translation: /**/%file_name%.%locale%.%file_extension%


crowdin.yml
```diff
@@ -0,0 +1,9 @@
+files:
+  - source: /GUI/**/*.resx
+    ignore:
+      - Properties/
+      - Localization/
+      - HintTextBox.resx
```
HebaruSan, line 6: Is `HintTextBox.resx` excluded because it doesn't have any text resources? Or is there something else special about it?

HebaruSan (approved):
Well whatever the reason for ignoring `HintTextBox.resx`, it's working fine and we can make more updates later if appropriate. Let's merge.

(I'm thinking along the lines of deleting `HintTextBox.resx` completely, or at least removing its `$this.TrayLargeIcon` property, which looks unnecessary and weird.)

crowdin.yml
Olympic1: Sorry for the late answer, was a busy week(end).

I exluded it because Crowdin just copied the file for each translation because everything was supposedly translated when there is just nothing to translate.

PR #3649, opened 2022-08-29
Fix missing mod info tab alerts

HebaruSan:
## Background

The `<!>` alerts from #3271 are supposed to call user attention to the reason that an incompatible mod is incompatible; if the mod itself is incompatible, the Versions tab shows an alert, and if it's a problem with a relationship, the Relationships tab shows an alert:

image

## Problem

In the current build (not released yet), these alerts don't appear unless you click the Metadata tab.

## Causes

`UpdateModInfo` was a hybrid of general and tab-specific updates that #3638 tried to split into separate pieces, and these alerts anded up associated with the metadata tab instead of the general update where they belong.

## Changes

Now the tab alerts are updated along with the header info, so they are always visible.

### DIFF
--- a/GUI/Controls/ModInfo.cs
+++ b/GUI/Controls/ModInfo.cs
@@ -54,7 +54,7 @@ namespace CKAN.GUI
                         {
                             ReverseRelationshipsCheckbox.CheckState = CheckState.Unchecked;
                         }
-                        UpdateHeaderInfo(module);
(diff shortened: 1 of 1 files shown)


HebaruSan:
Really just submitted this for documentation; I'm going to self-review and merge, since it's small and should have been included in #3638...

PR #3650, opened 2022-08-29
Self review option for merge script

HebaruSan:
## Motivation

For a very small, uncontroversial fix like #3649, it may make sense to submit a pull request with documentation of the bug and the fix (so team members know what happened), but then merge it without review, to keep the `master` branch working.

Currently the merge script from #3263 can't do this because it absolutely requires approval by at least one reviewer. I've been making a live temporary edit to add `False and` to bypass that check as needed, which is not a great solution.

## Changes

Now the merge script has a `--self-review` option that skips the approval check if present. The default is `False` (obviously). This should only be used for small fixes in which we have high confidence. This has already been tested in the merge of #3649.

### DIFF
--- a/bin/ckan-merge-pr.py
+++ b/bin/ckan-merge-pr.py
@@ -82,8 +82,8 @@ class CkanPullRequest:
     def merge_commit_message(self) -> str:
         return f'Merge #{self.pull_request.number} {self.pull_request.title}'
 
-    def merge_into(self, repo: CkanRepo) -> bool:
-        if not self.approvers():
+    def merge_into(self, repo: CkanRepo, self_review: bool) -> bool:
+        if not self_review and not self.approvers():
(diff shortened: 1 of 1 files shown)


DasSkelett (approved):

HebaruSan:
> a very small, uncontroversial fix

> small fixes in which we have high confidence

A further suggestion for the scope of acceptable usage of this option, it should probably also be limited to fixes of unreleased bugs, when we've broken something that was working in the latest release (and therefore the correct behavior is trivial to define and confirm). I think my usages have all been along those lines.

Maybe eventually we can indicte that with GitHub labels, so the `--self-review` option might print an error unless the PR has the `Unreleased Bug` label.

techman83:
That sounds useful and appropriate.

PR #2963, opened 2020-01-13
Check all dependencies for compatibility checking

HebaruSan:
## Problems

The Compatible and Incompatible filters in GUI check dependencies, but only one level. For example, OuterPlanetsMod depends on Kopernicus depends on ModularFlightIntegrator; if the first two are compatible and the third isn't, then OuterPlanetsMod appears to be installable but will fail.

We do extra work when loading the GUI filters, since `Registry.Available` and `Registry.Incompatible` both loop over all modules and check their dependencies, and the results are not re-used. Similarly, various operations that refresh the mod list are slower than they need to be because we re-check all those same relationships over again even if they haven't changed.

Sometimes "Available" includes incompatible modules:

- `Registry.available_modules`
- `Registry.AllAvailable`
- `Registry.SetAllAvailable`
- `Registry.HasAnyAvailable`
- `Registry.AddAvailable`
- `Registry.RemoveAvailable`
- `Registry.GetAvailableMetadata`

And sometimes it excludes them:

- `Registry.Available`
- `ckan available`

This is confusing.

## Changes

Now `Registry` uses a new `CompatibilitySorter` class to track module compatibility, which has public properties `Compatible` and `Incompatible` and a `CompatibleVersions` property to represent which versions they're compatible with. When we need the compatible or incompatible mods, we check whether the compatible versions are the same as last time, and if so, we return the same lists as previously (which solves the problem of `Registry.Available` and `Registry.Incompatible` duplicating work), otherwise we create a new `CompatibilitySorter` to compute the new compatibilities.

`CompatibilitySorter` uses a fully recursive dependency checking algorithm, so modules will only be counted as compatible if all of their dependencies are compatible. It also uses the lists of compatible and incompatible mods to speed up the search as it goes, rather than re-checking over and over whether dependencies are compatible (now that it's recursive, we don't want to re-check all of Kopernicus's dependencies over and over for all of its many depending mods).

Now "Available" means all modules, and anything that excludes incompatible mods is renamed to "Compatible". `AllAvailable` is renamed to `AvailableByIdentifier` since it doesn't return "all" of anything. `ckan available` retains its old name so users don't have to switch to a new command.

Fixes #1105.
Fixes #1646.
Fixes #2231.
Fixes #2849.

PR #3651, opened 2022-08-30
Highlight incompatible mods recursively

HebaruSan:
## Problem

If you try to install RP-1 on KSP 1.12 without following [the instructions (which call for marking KSP 1.11 compatible)](https://github.com/KSP-RO/RP-0/wiki/RO-&-RP-1-Express-Installation-for-1.12.3), you'll find you can't install it, and the reason is not apparent on the Relationships tab, which has increasingly become the place to investigate compatibility:

screenshot1

If you expand dependencies (or try to install the depending mod from the Versions tab), you can discover that RP-0 depends on SXTContinued, which depends on Firespitter, which is only compatible up to KSP 1.11:

screenshot2

Noted in KSP-CKAN/NetKAN#9306.

## Cause

The relationships tab's red highlighting only indicates whether the mod itself is compatible, not including its dependencies. It doesn't load all relationships recursively, for performance reasons; instead, it only loads two generations, so it can show which nodes are expandable, and then more generations are loaded if the user expands a node. This structure means that it may not know whether a mod is indirectly incompatible until multiple layers have been expanded and the user finds the issue.

But another component does know all mods' recursive compatibility: the `CompatibilitySorter` from #2963.

## Changes

- Now the registry provides `IdentifierCompatible` to allow quick recursive compatibility checks via the `CompatibilitySorter` object
- Now the relationships tab uses `Registry.IdentifierCompatible` to check compatibility instead of simple game version checks

This causes SXTContinued to be highlighted red even though it's individually compatible, and you can still expand it to see Firespitter:

image

image

### DIFF
--- a/Core/Registry/IRegistryQuerier.cs
+++ b/Core/Registry/IRegistryQuerier.cs
@@ -125,6 +125,14 @@ namespace CKAN
         /// <param name="with_provides">If set to false will not check for provided versions.</param>
         /// <returns>The version of the mod or null if not found</returns>
         ModuleVersion InstalledVersion(string identifier, bool with_provides = true);
+
+        /// <summary>
+        /// Check whether any versions of this mod are installable (including dependencies) on the given game versions
+        /// </summary>
+        /// <param name="identifier">Identifier of mod</param>
+        /// <param name="crit">Game versions</param>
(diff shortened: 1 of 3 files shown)


techman83 (approved):
That's a good QoL improvement!

PR #3654, opened 2022-09-01
update repo instructions for new Debian policy

isantop:
Current Debian policy guidelines state that 3rd party repositories should use the "signed-by" option and install a keyring file to `/usr/share/keyrings` named `reponame-archive-keyring.gpg` rather than using a key file in `trusted.gpg.d`, as the latter is deprecated and will no longer work in future versions of Debian and derivatives. This change updates the repo installation instructions to comply with current Debian Third-Party repository guidelines. The guidelines as currently outlined are available here: https://wiki.debian.org/DebianRepository/UseThirdParty

Additionally, there is a small change to add the `arch=amd64` option, as this prevents an error otherwise presented by APT due to the repository lacking support for i386.

### DIFF
--- a/doc/apt-repo.md
+++ b/doc/apt-repo.md
@@ -4,13 +4,16 @@ We have created an APT repository that you can add to your Debian-based OS to in
 
 (These instructions are here instead of a wiki because the `curl` command adds our trusted key to your APT configuration, and we do not want that line to be editable by the entire world, in case it might be altered maliciously.)
 
+See https://wiki.debian.org/DebianRepository/UseThirdParty for the Debian policy that we used to generate these steps.
+
 ## Stable builds
 
 These are [the main releases](https://github.com/KSP-CKAN/CKAN/releases), recommended for most users. You will have the same features at the same time as everyone else, but you will have the added conveniences of APT managing the updates for you.
 
 ```
-sudo curl -sS -o /usr/share/keyrings/ckan-archive-keyring.gpg https://raw.githubusercontent.com/KSP-CKAN/CKAN/master/debian/ksp-ckan.gpg
-sudo apt-add-repository -u -y 'deb [arch=amd64 signed-by=/usr/share/keyrings/ckan-archive-keyring.gpg] https://ksp-ckan.s3-us-west-2.amazonaws.com/deb stable main'
+sudo curl -sS -o /usr/share/keyrings/ksp-ckan-archive-keyring.gpg https://raw.githubusercontent.com/KSP-CKAN/CKAN/master/debian/ksp-ckan.gpg
+echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/ksp-ckan-archive-keyring.gpg] https://ksp-ckan.s3-us-west-2.amazonaws.com/deb stable main' | sudo tee /etc/apt/sources.list.d/ksp-ckan.list > /dev/null
+sudo apt update
 sudo apt install ckan
 ```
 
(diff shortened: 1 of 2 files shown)


doc/apt-repo.md
```diff
@@ -21,7 +21,7 @@ If you like to live dangerously, these are the bleeding edge builds that are gen
 Things may break! But if they do and you report it to us, you'll be a hero to CKAN users everywhere, whether they know it or not.
 
 ```
-sudo curl -sS -o /etc/apt/trusted.gpg.d/ksp-ckan.gpg https://raw.githubusercontent.com/KSP-CKAN/CKAN/master/debian/ksp-ckan.gpg
-sudo apt-add-repository -u -y 'deb https://ksp-ckan.s3-us-west-2.amazonaws.com/deb nightly main'
+sudo curl -sS -o /usr/share/keyrings/ckan-archive-keyring.gpg https://raw.githubusercontent.com/KSP-CKAN/CKAN/master/debian/ksp-ckan.gpg
+sudo apt-add-repository -u -y 'deb [arch=amd64 signed-by=/usr/share/keyrings/ckan-archive-keyring.gpg] https://ksp-ckan.s3-us-west-2.amazonaws.com/deb nightly main'
```
HebaruSan, line 25: Thanks so much for the update!

This command doesn't work on my Ubuntu 20.04 VM:

```
$ sudo apt-add-repository -u -y 'deb [arch=amd64 signed-by=/usr/share/keyrings/ckan-archive-keyring.gpg] https://ksp-ckan.s3-us-west-2.amazonaws.com/deb nightly main'
[sudo] password for user: 
Error: 'deb [arch=amd64 signed-by=/usr/share/keyrings/ckan-archive-keyring.gpg] https://ksp-ckan.s3-us-west-2.amazonaws.com/deb nightly main' invalid
```

Any idea what's wrong?

isantop: Hmm, it looks like there's a bug in `apt-add-repository` which is preventing it from parsing that deb line correctly. I personally wrote/maintain the parser used in Pop_OS, and adding it with my tool it does work correctly (Apt tells me it's verified and updates/installs work correctly) with the repo configured as in the PR.

A possible alternative is adding the repo using cat and tee, like so:
```
cat "deb [arch=amd64 signed-by=/usr/share/keyrings/ckan-archive-keyring.gpg] https://ksp-ckan.s3-us-west-2.amazonaws.com/deb nightly main" | sudo tee /etc/apt/sources.list.d/ckan.list
```
This would, however, require running `sudo apt update` afterwards to fetch the updated lists. 

Another option would be to put together a `.list` (or a DEB822-formatted `.sources`) file in the repo along side the key, which the user can install using curl similar to installing the key. This would also require `sudo apt update` following to fetch the lists, but it would be a bit cleaner looking I think. Downside would be that it's a bit less transparent to the user what exactly is being added, but it's not major. 

In either case, the key should change and the "signed-by" option added to point at the key. trusted.gpg.d is deprecated and will not be supported for much longer in the future.

HebaruSan: OK, that makes sense. Could you please pick an option and update this branch? We can't merge a PR that would have most of the affected users reporting errors in a process that worked previously.

HebaruSan:
Note to self: After we get this resolved (to avoid git conflicts), "report it to us" should be a link to the bug tracker in both repo documents, so users choosing the nightly builds have an obvious place to go to report problems.

doc/apt-repo.md
HebaruSan: The `apt-add-repository` issue seems to be tracked here:

https://bugs.launchpad.net/ubuntu/+source/software-properties/+bug/1862764

HebaruSan: Pushed some changes that resolve this (`echo ... | sudo tee`). I tried them in my test VM and they worked.

HebaruSan (approved):
Thanks again! We appreciate you saving us from having to figure this out ourselves when distros start enforcing the new rules and users come to us with error messages.

PR #3655, opened 2022-09-03
Support mouse back/forward buttons

HebaruSan:
## Background

CKAN's GUI has back/forward buttons that navigate through the mods the user recently clicked.

Some mice have extra buttons for back and forward (typically used in a web browser), and currently these do nothing in CKAN.

## Changes

Now we listen for messages that contain XButton events and map them to the existing back/forward navigation. This is done via `IMessageFilter.PreFilterMessage` in order to react to events regardless of the focus or position of the mouse cursor.

Learned from this:  https://stackoverflow.com/a/67698917

### DIFF
--- a/GUI/Controls/ManageMods.cs
+++ b/GUI/Controls/ManageMods.cs
@@ -1489,13 +1489,13 @@ namespace CKAN.GUI
             navHistory.AddToHistory(module);
         }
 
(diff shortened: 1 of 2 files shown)


techman83 (approved):
That's a neat little QoL improvement!

PR #3656, opened 2022-09-06
Fix second search box disappearing

HebaruSan:
## Problem

1. Type a search
2. Click the `+` button to add a second search box
3. Type into the second search box
4. Second search box disappears

## Cause

In #3641 we refactored the search box timer to use the same event handler for keystrokes and text change events. It uses a timer to delay the search by default, but this is bypassed for an immediate search if the user presses Enter or if the text box is empty.

The keystroke handler fires before the runtime updates the text box's contents, so it was always empty for the first keystroke, and we were firing the update immediately with an empty search string.

## Changes

Now we don't bypass the delay for a keystroke event for any keys other than Enter.


DasSkelett (approved):
Works!

PR #3645, opened 2022-08-24
Improvements for failed repo updates

HebaruSan:
## Problems

- `Repo.Update` defines an inherently incoherent operation: Refreshing metadata from just one repository to the exclusion of any others that may be configured. This is not something you would ever want to do, because you would end up with an incomplete set of modules in your registry. Registry updates should be from all repos or none.
- Repo updates can't be cancelled, and if one fails, you can't recover

## Changes

- Now the ability to update only one repo in isolation is eliminated
  - `Repo.Update` is removed
  - The `ckan update` command no longer takes a parameter and always update all repos
  - `ckan update`'s `--all` parameter is removed (note that its documentation incorrectly described upgrading modules, which it never did)
- Now `DownloadTarget.filename` is used as the default for `NetAsyncDownloaderDownloadPart.path`, so users of `NetAsyncDownloader` can find the files they requested
- Now `NetAsyncDownloaderDownloadPart.Done` and `NetAsyncDownloader.onOneCompleted` pass the download's ETag (see #2682) as a third parameter if available, so users of `NetAsyncDownloader` can save this information after downloading if desired
- Now `Repo.UpdateAllRepositories` accepts a `NetAsyncDownloader` parameter and uses it to perform its downloads asynchronously, in parallel, with progress bar updates, and with the possibility of cancellation
- Now `Repo.UpdateRegistryFromTarGz` is renamed to `Repo.ModulesFromTarGz`
- Now `Repo.UpdateRegistryFromZip` is renamed to `Repo.ModulesFromZip`
- Now `Repo.ModulesFromTarGz` and `Repo.ModulesFromZip` provide progress bar updates as they process their input files
- Now the parts of `Wait` that were specific to `CkanModule` are made generic and able to handle `string`s instead
- Now the `DownloadsFailedDialog` from #3635 is made generic and accepts `object`s instead of `CkanModule`s, and the module-specific strings it contained are parameters to be supplied by the calling code
  - `MainInstall.cs` is updated to pass module-specific strings to the `DownloadsFailedDialog`
- `GUI.Main.UpdateRepo` now:
  - Loops inside a transaction to allow retries and rollback
  - Enables the Cancel button in the `Wait` dialog, and clicking it cancels in progress downloads
  - Calls the `DownloadsFailedDialog` from #3635 to allow the user to retry or abort the refresh or remove a bad repository:
    image
[...]

Issue #3343, opened 2021-04-10
[Bug][Solved] Mono.Unix.UnixIOException: Resource temporarily unavailable on Arch Linux after Util.Invoke and long chain of runtime calls 


BlazingPL:
Background
----------
* Operating System: Arch Linux
* CKAN Version: 1.30.0-1 (AUR)
* KSP Version: 1.11.2

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
* No

## Problem

Crash when downloading multiple mods at once.

```
exception inside UnhandledException handler: (null) assembly:/usr/lib/mono/gac/Mono.Posix/4.0.0.0__0738eb9f132ed756/Mono.Posix.dll type:UnixIOException member:(null)

exception inside UnhandledException handler: (null) assembly:/usr/lib/mono/gac/Mono.Posix/4.0.0.0__0738eb9f132ed756/Mono.Posix.dll type:UnixIOException member:(null)

exception inside UnhandledException handler: (null) assembly:/usr/lib/mono/gac/Mono.Posix/4.0.0.0__0738eb9f132ed756/Mono.Posix.dll type:UnixIOException member:(null)

exception inside UnhandledException handler: (null) assembly:/usr/lib/mono/gac/Mono.Posix/4.0.0.0__0738eb9f132ed756/Mono.Posix.dll type:UnixIOException member:(null)

exception inside UnhandledException handler: (null) assembly:/usr/lib/mono/gac/Mono.Posix/4.0.0.0__0738eb9f132ed756/Mono.Posix.dll type:UnixIOException member:(null)

[ERROR] FATAL UNHANDLED EXCEPTION: Mono.Unix.UnixIOException: Resource temporarily unavailable [EWOULDBLOCK].
  at Mono.Unix.UnixMarshal.ThrowExceptionForLastError () [0x00005] in <7bcd6d815ebc477ab0521e0361c44f6c>:0 
  at Mono.Unix.UnixStream.Write (System.Byte[] buffer, System.Int32 offset, System.Int32 count) [0x00055] in <7bcd6d815ebc477ab0521e0361c44f6c>:0 
  at System.Windows.Forms.XplatUIX11.WakeupMain () [0x00000] in <4b6c441381804088ab0fff508a3fbabf>:0 
[...]

PR #3659, opened 2022-09-07
Report progress in validating downloads

HebaruSan:
## Motivation

- After a mod finishes downloading and before we install it, we check that it's a valid ZIP file and that it matches the hashes in that module's metadata. This takes a noticeable amount of time for mods larger than about ~30MB (several seconds), which frighteningly probably scales linearly to the GB+ mods that we index, and there is no indication of what's happening; it looks like CKAN has frozen. As part of the general improvements for downloading in the next release, I want to increase the transparency of this process.
- The per-module progress bars in the install flow can be overlapped by their labels when a module name is long
  - The code from #3054 that handles these progress bars does a lot of low-level logic to position them, which isn't easy for maintenance
- Per-module progress bars aren't shown when downloading a mod to cache without installing, even though you can download multiple in parallel by jumping back to the mod list and starting more

## Changes

- Now a new `CKAN.Extensions.CryptoExtensions.ComputeHash` extension function supports progress updates from 0 to 100 via an `IProgress` parameter and cancellation via a cancellation token (cancellation not used yet but could be taken advantage of in the future)
- Now `NetFileCache.ZipValid`, `NetFileCache.GetFileHashSha1`, `NetFileCache.GetFileHashSha256`, and `NetModuleCache.Store` also have an `IProgress` parameter for reporting progress from 0 to 100 in each of their respective tasks. `NetModuleCache.Store` scales and combines the updates from its sub-tasks in a 60:20:20 ratio for `ZipValid`:`GetFileHashSha1`:`GetFileHashSha256`, since that seems to be about how long they take in practice.
  - The nearly identical `NetFileCache.GetFileHashSha1` and `NetFileCache.GetFileHashSha256` are mostly factored out into a new generic function `NetFileCache.GetFileHash<T>` that they call
- Now the `Wait` tab uses a `TableLayoutPanel` to manage the per-module progress bars
  - The labels and progress bars always fit neatly into 2 columns without overlapping
  - The code is simpler and no longer micro-manages locations or sizes
- Now `IDownloader` has a new `StoreProgress` event for reporting progress as it validates a ZIP file
- `NetAsyncModulesDownloader` fires its `StoreProgress` event in response to progress updates from `NetModuleCache.Store`
- Now the install and download to cache flows use the new `StoreProgress` event to show a per-module progress bar while downloads are being validated so the user can see what's happening
  - Download to cache also shows regular per-module download progress bars
  - We also print one additional message to the log when we have finished downloading and are starting validation of a download, so users can see what's happening there as well (and for CmdLine)

### Additional fixes

With the major features of my ongoing network handling revamp completed, but mainly developed on Windows, I decided to take the master branch for a spin in an Ubuntu VM. I found myself on a very lively bug hunt, but understanding what was happening in various confusing cases required this PR's progress bars for ZIP validation, so I rebased and added the fixes here:

- The failed downloads dialog from #3635 was opened on a background thread instead of the UI thread, which completely crashed it on Mono (but worked on Windows). Now it opens on the UI thread as it should, and a `TaskCompletionSource` tells us when it's done, as in several existing spots.
  - The failed downloads dialog's auto-sizing wasn't tall enough on Mono, now it includes the grid headers and padding
  - #3645 renamed the `DownloadRow.Module` property to `Data`, but the grid was still looking for `Module`. Now it's updated to `Data`.
- You can cancel downloads, and they really cancel, leaving behind the partial file as per #3666
  - The async downloader no longer tries to cancel downloads that already finished or failed
  - The async downloader no longer starts queued downloads after cancellation
- If `ResumingWebClient` finds that it already has the entire file in the in progress folder, it closes the stream, but Mono's WebClient seems to re-open it (whereas Windows leaves it closed), so now we also treat a download as unnecessary if `contentLength` is 0
  - Download progress updates now only fire once per 200 milliseconds, which should help with the `EWOULDBLOCK` errors from #3343
  - The activity timeout for downloads is now set to 30 seconds
- The checkbox above the mod list that can uninstall all mods or clear the changeset now operates much quicker by updating the changeset once at the end instead of for every row
- #3667 broke mod upgrades because the `ModUpgrade` class's constructor no longer handled the parameters `GUIMod` gave it correctly, now this is fixed
- `Wait.SetProgress` will no longer divide by 0 if you pass 0 in its `total` param
- `ManageMods.ModGrid` was renamed from `ModList` in #3041, but its event handlers still had a `ModList_` prefix; now they finally start with `ModGrid_`
- Unreachable code is removed from `Repo.ModulesFromFile`
- Kraken classes that used `\r\n` now use `Environment.NewLine` instead to match the host platform
  - Various public fields in the `Kraken` classes are now marked `readonly` where possible
- Several debug messages are shortened from printing whole exceptions to just `Exception.Message`, to make them easier to browse

### DIFF
--- a/Core/Extensions/CryptoExtensions.cs
+++ b/Core/Extensions/CryptoExtensions.cs
@@ -1,0 +1,51 @@
+using System;
+using System.IO;
+using System.Threading;
+using System.Security.Cryptography;
+
+namespace CKAN.Extensions
+{
+    /// <summary>
+    /// Extensions for cryptographic operations
+    /// </summary>
+    public static class CryptoExtensions
+    {
+
+        /// <summary>
+        /// A version of ComputeHash with progress updates.
+        /// Based on https://stackoverflow.com/a/53966139 with lots of cleaning up.
+        /// </summary>
+        /// <param name="hashAlgo">The crypto object to use for the operation, like SHA1, SHA256, etc.</param>
+        /// <param name="stream">Input stream to hash</param>
+        /// <param name="progress">Callback to notify as we traverse the input, called with percentages from 0 to 100</param>
+        /// <param name="cancelToken">A cancellation token that can be used to abort the hash</param>
+        /// <returns>The requested hash of the input stream</returns>
+        public static byte[] ComputeHash(this HashAlgorithm hashAlgo, Stream stream,
+            IProgress<long> progress, CancellationToken cancelToken = default(CancellationToken))
+        {
+            const int bufSize = 1024 * 1024;
+            var buffer = new byte[bufSize];
+            long totalBytesRead = 0;
+            while (true)
+            {
+                var bytesRead = stream.Read(buffer, 0, bufSize);
+                cancelToken.ThrowIfCancellationRequested();
+
+                if (bytesRead < bufSize)
+                {
+                    // Done!
+                    hashAlgo.TransformFinalBlock(buffer, 0, bytesRead);
+                    break;
+                }
+                else
+                {
+                    hashAlgo.TransformBlock(buffer, 0, bytesRead, buffer, 0);
+                }
+
+                totalBytesRead += bytesRead;
+                progress.Report(100 * totalBytesRead / stream.Length);
+            }
+            return hashAlgo.Hash;
+        }
+    }
+}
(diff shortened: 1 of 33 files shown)


HebaruSan:
I'm planning to add some more fixes here, partly from this PR, partly from previous ones. Please hold off on reviewing or approving.

HebaruSan:
OK, this should now be ready! And it fixes some pretty serious problems.

techman83 (approved):
That looks like a significant Quality of Life improvement! :tada:

PR #3660, opened 2022-09-08
Fix already contains module error at upgrade

HebaruSan:
## Problem

After #3643, upgrades are broken in the dev builds:

1. Install an old version of `KSP-Recall`
2. Click the upgrade checkbox on that mod's row

image

```
System.ArgumentException: Already contains module: KSP-Recall
   at CKAN.RelationshipResolver.Add(CkanModule module, SelectionReason reason)
   at CKAN.RelationshipResolver..ctor(IEnumerable`1 modulesToInstall, IEnumerable`1 modulesToRemove, RelationshipResolverOptions options, IRegistryQuerier registry, GameVersionCriteria GameVersion)
   at CKAN.IRegistryQuerierHelpers.FindRemovableAutoInstalled(IRegistryQuerier querier, List`1 installedModules, HashSet`1 dlls, IDictionary`2 dlc, GameVersionCriteria crit)
   at CKAN.IRegistryQuerierHelpers.FindRemovableAutoInstalled(IRegistryQuerier querier, List`1 installedModules, GameVersionCriteria crit)
   at CKAN.GUI.ModList.ComputeChangeSetFromModList(IRegistryQuerier registry, HashSet`1 changeSet, ModuleInstaller installer, GameVersionCriteria version)
   at CKAN.GUI.ManageMods.<UpdateChangeSetAndConflicts>d__117.MoveNext()
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at CKAN.GUI.ManageMods.<ModList_CellValueChanged>d__74.MoveNext()
```

Reported on Discord by @TwiceDaIce, who was not able to submit an issue at the time.

## Cause

The exception indicates that the list of modules passed to the resolver contained a duplicate. This was supposed to be prevented by the way `FindRemovableAutoInstalled`'s parameter is constructed, but `modules_to_remove` doesn't contain the installed version (as I assumed it would by its name since that version would be removed). This duplicate would always have been there for upgrades, but prior to #3643 it didn't matter because we weren't passing the modules to a resolver.

## Tried and failed

I tried adding the installed module to `modules_to_remove`, and this broke switching from one module version to another. There are too many uses of these variables and too many implicit assumptions about how they behave for me to sort it out that way.

## Changes

Now a new `InstalledAfterChanges` function generates exactly the list of modules that I want, without duplicates, so upgrading and switching versions both work.

### DIFF
--- a/GUI/Model/ModList.cs
+++ b/GUI/Model/ModList.cs
@@ -179,14 +179,7 @@ namespace CKAN.GUI
             }
 
             foreach (var im in registry.FindRemovableAutoInstalled(
-                registry.InstalledModules
-                    .Where(im =>
-                        modules_to_remove.All(m => m.identifier != im.identifier)
-                        || modules_to_install.Any(m => m.identifier == im.identifier))
-                    .Concat(modules_to_install.Select(m =>
-                        new InstalledModule(null, m, new string[0], false)))
-                    .ToList(),
-                version))
+                InstalledAfterChanges(registry, changeSet, version).ToList(), version))
             {
                 changeSet.Add(new ModChange(im.Module, GUIModChangeType.Remove, new SelectionReason.NoLongerUsed()));
                 modules_to_remove.Add(im.Module);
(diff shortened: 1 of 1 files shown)


techman83 (approved):
That looks sensible, thanks @HebaruSan !

PR #3662, opened 2022-09-09
Trivial code style clean-up

HebaruSan:
## Motivation

I've been accumulating trivial changes while working on unrelated pull requests.
Now they've been dumped into a branch so I don't have to keep stashing and popping them.

## Changes

- Redundant namespace specifications are removed
- BOMs are removed from lines where they don't belong
- Missing or misaligned `{` and `}` characters are fixed
- Some `using` statements are moved to the top of the file and split into system, third party, and CKAN sections
- The `X` icon in the `HintTextBox` is made to fit the text box rather than just being centered on top of it

I'll probably self-review this since it's so minor.

### DIFF
--- a/Cmdline/Action/Upgrade.cs
+++ b/Cmdline/Action/Upgrade.cs
@@ -1,8 +1,10 @@
 using System;
 using System.Linq;
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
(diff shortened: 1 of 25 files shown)


techman83 (commented):
LGTM! :tada:

PR #3661, opened 2022-09-08
GitLab kref

HebaruSan:
## Motivation

See KSP-CKAN/NetKAN#9325, the author of a mod hosted on GitLab has requested to have it added to CKAN. Currently we don't have a way to do that without hard coding the download URL, which is bad for maintenance.

## Changes

- A `netkan.exe --gitlab-token` command line option is now added and works just like `--github-token` but for GitLab
- If `PurgeDownloads` fails after a failed inflation, we ignore it so the original error can be printed
- A new kref format `#/ckan/gitlab/<account>/<project>` is added for indexing GitLab projects
- A new `GitlabTransfomer` uses a `GitlabApi` and accompanying data classes from a `Netkan/Sources/Gitlab` folder to inflate modules with GitLab krefs
- The spec is updated to describe the new kref

After this, we can inflate a netkan for the mod from KSP-CKAN/NetKAN#9325.

Fixes #2554.

### Known limitations

Whereas GitHub allows the user to custom-build a special ZIP and attach it as a release asset, GitLab appears to only support source ZIPs (which we can support for GitHub as well using `x_netkan_github.use_source_archive`). This means that mods with build steps such as compiling a plugin may not be indexable from GitLab, but some mods without such steps can be correctly installed from a source ZIP if the repo is structured the same as the installed folder.

While it would be great if GitLab added support for non-source assets, we don't have to wait for that to get some value out of a GitLab kref. However, we do need to make sure that any future changes don't break the mods we index in the meantime. For this reason, `x_netkan_gitlab.use_source_archive` **must** be set to `true` for now! This will ensure that we only index GitLab mods that can be properly installed from a source ZIP, while allowing for future enhancements to the API for non-source assets.

PR #3665, opened 2022-09-11
Pass GitLab token from env to inflator param

HebaruSan:
## Motivation

In KSP-CKAN/CKAN#3661 we added the ability to index mods hosted on GitLab. This included the ability to provide a GitLab authentication token from the command line, but the Inflator isn't using this capability yet. We should add that eventually for completeness, see https://github.com/KSP-CKAN/CKAN/pull/3661#discussion_r966544529.

## Changes

Now the `$GL_Token` environment variable is passed to the inflator's `--gitlab-token` parameter.
KSP-CKAN/NetKAN-Infra#276 will handle providing this env var based on the configured secrets.

I tested `netkan.exe --gitlab-token ''` and it still worked, so I think there is no harm in setting this parameter before we create a GitLab token.


techman83 (approved):
LGTM :tada:

PR #3670, opened 2022-09-21
ModInfo usability improvements

HebaruSan:
## Motivation

Mod info has some limitations and quirks currently:

- There's a splitter between the abstract/description and the tabstrip, and the description has a vertical scrollbar to compensate for this. Making the user choose where to put the splitter is unnecessary because there's an obvious right answer: show the whole description.
- The metadata tab has a horizontal scrollbar so the whole URLs can be viewed, but scrolling to the right hides the labels
- When an incompatible mod is selected, `<!>` shows up in one of the tabs, which was supposed to draw the user's attention to the reason the mod is incompatible (see #3271), but going by our user support questions, it has not been effective
- Sometimes you want to know what a mod `provides`, but it's not shown
- Relationships can include `version`, `min_version`, and `max_version` properties, but they aren't shown, which can be confusing. For example, when a mod only works with older versions of Scatterer, but it's not required, we add Scatterer to the `recommends` and `conflicts`, with a `min_version` property in the conflict (see KSP-CKAN/NetKAN#8880 and KSP-CKAN/NetKAN#8881). But to the user this just looks like it's both recommended and conflicted.
- The relationship tree shows a tooltip of "Depends", "Recommends", etc., but these strings are not localized/translatable

## Changes

Mod info generally:

- The tabs of mod info are separated into their own source code files
- The mod info splitter and the description scrollbar are eliminated, and instead the abstract and description are automatically sized to fit:
  image
- The fake stop sign is replaced by a real red stop sign (using the "report" icon from Material Icons, Apache-2.0 licensed):
  image

Metadata tab:

- The horizontal scrollbar is eliminated and the URLs wrap instead:
  image

Relationships tab:

- The `provides` list is now shown in the relationships tab so users can see it (using the "ballot" icon from Material Icons, Apache-2.0 licensed):
  image
- The relationships tab tooltip shows the `version`, `min_version`, and `max_version` properties of relationships so users can see them:
  image
- Relationship tree tooltips are internationalized

Other:

- `HintTextBox` had a resource file with no resources, which would have made Crowdin make copies of it for every language, so #3648 filtered it out. Now it and the filter are removed.


techman83 (approved):
That looks like a really good usability improvement! :tada:

PR #3672, opened 2022-10-06
Authors as clickable filter links, combine search links w/ ctrl or shift

HebaruSan:
## Motivation

Continuing recent efforts to improve mod info (see #3670), and remembering my quest to find FruitGoose's mods (see KSP-CKAN/NetKAN#9351), influenced by the "sticky" checkbox on the Relationships tab (see #3638), I thought it would be nice to be able to click author names in mod info to set the filter, like the tab/label links. And if we're going to have both tag/label links and author links, that raises the possibility of wanting to use them together somehow.

## Changes

- The authors field in mod info is now an autosized wrapping flow layout collection of links that can be clicked to search by that author, similar to the tag/label links
  - Since authors can contain characters that aren't allowed or don't work in searches, non-alphanumeric characters are removed before searching
  - Can be Ctrl- or Shift-clicked to merge with existing search instead of replacing
- Mod tag/label links are no longer oddly offset to the right by 2px
  - Can be Ctrl- or Shift-clicked to merge with existing search instead of replacing
- Filter dropdown menu options can be Ctrl- or Shift-clicked to merge with existing search instead of replacing
- The `[+]` button to add a filter moves up and down to align with the bottommost search, since you'd want to use it after populating that field

Merging two searches produces their **intersection** (AND logic). I think this makes sense because the user's typical goal is to narrow down the list of mods more and more until the desired mod is found, and using OR logic would expand the list instead. Similarly, if you already have multiple OR searches and choose to merge a new filter into it, the new filter is merged into **all** of those existing ones, to make sure mods not matching the new filter drop off.

Since Ctrl- and Shift-clicking are not discoverable, a tooltip calls it out in each place where it's available.

image

image

image

### DIFF
--- a/GUI/Controls/EditModSearches.cs
+++ b/GUI/Controls/EditModSearches.cs
@@ -55,6 +55,22 @@ namespace CKAN.GUI
             Apply();
         }
 
+        /// <summary>
+        /// Merge the given searches with the currently active ones
+        /// </summary>
+        /// <param name="searches">New searches to add</param>
+        public void MergeSearches(List<ModSearch> searches)
+        {
+            // Merge inputs once for all editors
+            var merged = searches.Aggregate((search, newSearch) => search.MergedWith(newSearch));
+            foreach (var editor in editors)
+            {
+                // Combine all new with each existing (old AND new)
+                editor.Search = editor.Search?.MergedWith(merged) ?? merged;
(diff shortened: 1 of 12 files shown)


techman83 (approved):
That looks like a neat enhancement!

PR #3675, opened 2022-10-14
Fix modpack installation, install incompatible with confirmation

HebaruSan:
## Problems

- After #3667, trying to install modpacks doesn't work in dev builds anymore, because the relationship resolver's mod list property now tries to get the reason for each mod's _parent_, and mods from a modpack have a metapackage parent, which wouldn't have its reasons stored. An exception is thrown and not caught, so it fails messily. This definitely needs to be fixed before the next release.
- The default identifier for a modpack wasn't internationalized, but instead always had an English `installed-` prefix
- The "KSP all versions" string used in many places wasn't internationalized but instead always English

## Additional motivation

I felt like assembling my own modpack that I'm calling "Stock Plus," focused on enhancing the gameplay of the stock solar system and parts; if anyone is interested, I may be publishing it once this PR's changes are released. Working on this with the current dev build was how I discovered that modpack installation is broken.

A few mods that I really wanted are not marked compatible with KSP 1.10–1.12 (but they do work), which means I can still install them via the Versions tab by overriding a proper compatibility warning, but the options for adding them to a metapackage .ckan file are not satisfactory:

- Leave them as dependencies, which will prevent the modpack from being installed unless you add some really old game versions as compatible
- Change these mods to recommendations, which will simply hide them at install time

Either way, the incompatible mods that I want in my pack can't be installed via a modpack. I think we've had other modpack creators ask about this before, so it's potentially of general interest.

## Changes

- Now installing a modpack works again because the relationship resolver now saves the reason for metapackages (requested by user) even though they're excluded from the mod list
- Now the default identifier for a modpack is internationalized
- Now the "KSP all versions" string is internationalized using the already translated strings for `CkanModule.HighestCompatibleKSP`, and hence also changed capitalization to "KSP All versions" in English
- Now if some of a modpack's dependencies are incompatible, instead of erroring out, we alert the user and ask for confirmation, just like installing from the Versions tab (side note, the mods in this screenshot are **not** the affected mods from "Stock Plus"):
  image
  This will make modpack installation more closely replicate the user experience of installing the same set of mods without a modpack.
- Now the Version tab's compatibility warning includes your currently configured compatible game versions, similar to the above warning. To make this work without updating the current translation strings, the new string is added with a later `string.Format` index, despite occurring earlier in the main string. For consistency, this same ordering is also used for the above warning.
- Now when you export a modpack, more properties are set, and some are set better:
  - `ksp_version_min` and `ksp_version_max` are set based on your current compatibility settings
  - The default unknown license is retrieved from `License.UnknownLicense` instead of hard coded in the registry manager
  - The `download_content_type` is set based on the `DefaultValue` attribute instead of hard coded in the registry manager
  - The `depends`, `recommends`, and `suggests` properties will now be omitted instead of `[]` if they are empty
- Now if you choose multiple .ckan files to install via File→Install from ckan..., all of the files are grouped together into one big changeset rather than being installed one at a time
- Now if you install a modpack via File→Install from ckan..., all of its dependencies are added to the changeset individually so their compatibility checks can be bypassed
- The labels for the per-mod progress bars are now better aligned with the progress bars (8px down)

### DIFF
--- a/Core/Properties/Resources.Designer.cs
+++ b/Core/Properties/Resources.Designer.cs
@@ -591,5 +591,9 @@ namespace CKAN.Properties {
         internal static string SanityCheckerConflictsWith {
             get { return (string)(ResourceManager.GetObject("SanityCheckerConflictsWith", resourceCulture)); }
         }
+
+        internal static string ModpackName {
+            get { return (string)(ResourceManager.GetObject("ModpackName", resourceCulture)); }
+        }
     }
 }
--- a/Core/Properties/Resources.resx
+++ b/Core/Properties/Resources.resx
@@ -308,4 +308,5 @@ Free up space on that device or change your settings to use another location.
   <data name="NetModuleCacheModuleResuming" xml:space="preserve"><value>{0} {1} ({2}, {3} remaining)</value></data>
   <data name="ModuleInstallerUpgradeInstallingUncached" xml:space="preserve"><value> * Install: {0} {1} ({2}, {3} remaining)</value></data>
   <data name="ModuleInstallerUpgradeUpgradingResuming" xml:space="preserve"><value> * Upgrade: {0} {1} to {2} ({3}, {4} remaining)</value></data>
+  <data name="ModpackName" xml:space="preserve"><value>installed-{0}</value></data>
 </root>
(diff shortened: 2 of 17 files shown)


HebaruSan:
Rebased to resolve conflict with #3687.

techman83 (approved):
Looks all sensible to me :tada:

PR #3684, opened 2022-10-25
Replace apt with dnf in the docs for the RPM repo

htmlcsjs:
99% sure that apt isnt the package mangier for rpm distros

### DIFF
--- a/doc/rpm-repo.md
+++ b/doc/rpm-repo.md
@@ -4,7 +4,7 @@ We have created an RPM repository that you can add to your RPM-based OS to insta
 
 ## Stable builds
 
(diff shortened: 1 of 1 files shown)


HebaruSan (approved):
I did a search for "What package manager does Fedora use?", and it confirmed that DNF is the right name.

htmlcsjs:
> Well I'm **100%** sure you're right. ;)
> 
> This file was [created](https://github.com/KSP-CKAN/CKAN/pull/3605) by copying [the APT repo doc](https://github.com/KSP-CKAN/CKAN/blob/master/doc/apt-repo.md) and changing the APT-specific bits. Thanks for catching this one we missed.

np :+1:

PR #3698, opened 2022-11-07
Red highlight for dependencies on missing DLC

HebaruSan:
## Problem

If a mod depends on a DLC that you don't have installed, the mod will be treated as incompatible with a red stop sign on the relationships tab due to the missing dependency (since CKAN can't install DLC even if there's a version of it that's compatible with your game version). But if you look at the relationships tab of mod info to try to figure out why it's incompatible, there's no indication that this is the reason:

image

## Changes

Now if a mod depends on a DLC that isn't installed, the relationship will be highlighted red, just like for an incompatible dependency:

image

### DIFF
--- a/Core/Types/CkanModule.cs
+++ b/Core/Types/CkanModule.cs
@@ -607,21 +607,9 @@ namespace CKAN
             return this.identifier == identifier || provides.Contains(identifier);
         }
 
-        public bool IsMetapackage
-        {
-            get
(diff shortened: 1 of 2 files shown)


techman83 (approved):
That makes sense. The only thing I'll note, there isn't an indicator beyond the colour. Which could be a challenge for those that are colour blind.

PR #3702, opened 2022-11-11
Mark new deps of upgrades as auto-installed

HebaruSan:
## Problem

I upgraded Sandcastle after KSP-CKAN/NetKAN#9428, and WildBlueCore was installed as intended, but it was not marked as auto-installed. It should be, because it's only installed due to Sandcastle's dependency.

## Cause

`ModuleInstaller.AddRemove`, which is called by Upgrade and Replace, sets the auto-installed flag to false for any module that isn't being removed in the same call. This was the most cautious thing to do when the auto-installed flag was being added (setting this flag to true unnecessarily could remove modules the user wants to keep), but on further investigation, it's not correct.

## Changes

- Now `ModuleInstaller.AddRemove` has a `newModulesAreAutoInstalled` parameter that calling code can use to set the value of the auto-installed flag for modules that aren't being removed. Upgrade sets it to true because newly installed modules are dependencies, and replace sets it to false because newly installed modules are the replacement modules.
- Since `ModuleInstaller.AddRemove` was only used internally to `ModuleInstaller` with all of its parameters set, it is now `private` and its parameters no longer have defaults, to make maintenance easier.
- The mod list that Upgrade generates when it resolves relationships is now memoized to ensure we do not re-run the potentially expensive `RelationshipResolver.ModList()` logic (see #2953 and #3667)

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -940,9 +940,10 @@ namespace CKAN
         /// No relationships will be processed.
         /// This *will* save the registry.
         /// </summary>
-        /// <param name="add">Add.</param>
-        /// <param name="remove">Remove.</param>
-        public void AddRemove(ref HashSet<string> possibleConfigOnlyDirs, RegistryManager registry_manager, IEnumerable<CkanModule> add = null, IEnumerable<InstalledModule> remove = null, bool enforceConsistency = true)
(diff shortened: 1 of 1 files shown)


techman83 (approved):
Looks sensible to me, good explanation!

PR #3706, opened 2022-11-14
Remove duplicate Install changes for upgrades

HebaruSan:
## Problem

When you upgrade a module, the GUI changeset shows two entries for the same module, an Update step and an Install step.

image

## Cause

I think these extra changes existed along, but were hidden in GUI by `CreateSortedModList` until #3667 simplified the changeset UI to show what's actually being done. The `!sortedChangeSet.Any(c => c.Mod.identifier == change.Mod.identifier` check would have prevented an Install change in the changeset from being added to the list in the UI if an Update was already present:

https://github.com/KSP-CKAN/CKAN/blob/32517ba0fdc6f3116c0817572cd5ae76dfcf307d/GUI/Controls/Changeset.cs#L133-L149

## Changes

Now the resolver's Install step is only added to the changeset if there isn't already an Update step for the same mod.

### DIFF
--- a/GUI/Model/ModList.cs
+++ b/GUI/Model/ModList.cs
@@ -126,6 +126,7 @@ namespace CKAN.GUI
         {
             var modules_to_install = new List<CkanModule>();
             var modules_to_remove = new HashSet<CkanModule>();
+            var upgrading = new HashSet<CkanModule>();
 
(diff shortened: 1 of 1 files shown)


techman83 (approved):
Aha, that makes sense!

PR #3713, opened 2022-11-19
Fix Netkan check for Ships/Script spec_version

HebaruSan:
## Background

kOS scripts are installed to `Ships/Script`. We added support for this in #3180, including a Netkan validation check that the `spec_version` was `v1.29` or later, since earlier versions of CKAN wouldn't support it.

## Problem

The check only triggers for one specific case:

```yaml
install:
  - install_to: Ships/Script
```

There are other ways to install into that path:

```yaml
install:
  - file: some/path/Script
    install_to: Ships
```

```yaml
install:
  - find: some/path/Script
    install_to: Ships
```

```yaml
install:
  - file / find: whatever
    install_to: Ships
    as: Script
```

Remembered while reviewing KSP-CKAN/NetKAN#9443, which almost slipped by with a bad `spec_version` because of it.

## Changes

Now the Netkan validation checks for these other cases. It may still miss some things, but it's better than no check at all.

### DIFF
--- a/Netkan/Validators/InstallValidator.cs
+++ b/Netkan/Validators/InstallValidator.cs
@@ -1,4 +1,7 @@
+using System.Linq;
+
 using Newtonsoft.Json.Linq;
+
 using CKAN.Versioning;
 using CKAN.NetKAN.Model;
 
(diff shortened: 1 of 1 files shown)


techman83 (approved):
That makes sense!

PR #3726, opened 2022-11-24
Use changeset tab for reinstall

### LINKED ISSUE #3725: [Bug] Trying to reinstall a mod which has a dependency doesn't work due to the dependency
Background
----------
* Operating System:  Windows 10, latest build
* CKAN Version:  1.31.2
* KSP Version: 1.12.4

Problem
-------
Clean game, installed X-Science only.  Dependencies are ClickThroughBlocker, SpaceTuxLibrary and ToolbarController
No installation filters 

1. Select one of the mods
2. Right-click and select "Reinstall"
3. Get the next page which shows Recommendations (on a reinstall, why?)
4. Don't change anything
5. Click Continue
6. Get the message the it already contains module (see attached image)

Doing a Retry then appears to work

However, after some testing, it appears that the mod did NOT get reinstalled
This was tested by removing the dll and seeing if it was reinstalled while investigating an issue with the mod.  This raises the issue from an annoyance to a significant problem

**Expected behavior**

Do the reinstall without this error page

reinstall

Message at bottom of screen:  :Error during installation!"
```
PLACE CODE HERE !
```

HebaruSan:
## Problem

1. Install `xScienceContinued` (which will also install three dependencies)
2. Right-click it and select "Reinstall"
3. Click Yes in the confirmation popup
4. Click Continue on the recommendations screen
5. Install flow fails:
   reinstall

## Cause

The changeset passed to the install flow for reinstallation includes the dependencies, so they're added to the resolver once as dependencies and then again later as user-selected, which triggers an exception.

## Changes

- Now the reinstall option jumps to the changeset tab instead of popping up a confirmation dialog, so you can see what it's going to do
- Now the changeset tab saves the changeset it displays and passes it to the install flow upon confirmation, rather than re-calculating the changeset from the mod list
- Now the reinstall option generates a changeset that includes removing the depending/dependency mods but excludes installation of dependencies that will be pulled in automatically

After this, reinstallation succeeds for these mods. I also tried installing RP-1 express and re-installing some of it, and that worked as well. :crossed_fingers: 

Fixes #3725.

### DIFF
--- a/GUI/Controls/Changeset.cs
+++ b/GUI/Controls/Changeset.cs
@@ -17,6 +17,7 @@ namespace CKAN.GUI
 
         public void LoadChangeset(List<ModChange> changes, List<ModuleLabel> AlertLabels)
         {
+            changeset = changes;
             alertLabels = AlertLabels;
             ChangesListView.Items.Clear();
             if (changes != null)
@@ -46,7 +47,7 @@ namespace CKAN.GUI
 
         public event Action<ListView.SelectedListViewItemCollection> OnSelectedItemsChanged;
 
-        public event Action OnConfirmChanges;
+        public event Action<List<ModChange>> OnConfirmChanges;
         public event Action<bool> OnCancelChanges;
 
         private void ChangesListView_SelectedIndexChanged(object sender, EventArgs e)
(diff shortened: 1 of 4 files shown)


HebaruSan:
@linuxgurugamer, when the build finishes, could you please take it for a test drive? It'll be under the Artifacts dropdown here:

https://github.com/KSP-CKAN/CKAN/pull/3726/checks

techman83 (approved):
That all seems to make sense :tada:

HebaruSan:
I need to review the pull requests that added the dependencies and dependent mods here before merging, to make sure we haven't broken some obscure known use cases...

PR #3727, opened 2022-11-27
Show conflicts in changeset

HebaruSan:
## Background

When a user tries to install mods that conflict with one another or with an installed mod, the apply changes button is disabled and a description of the conflict appears in the leftward part of the status bar.

## Motivation

It has been fairly common recently for users to ask why the apply changes button is disabled; many or most users don't notice the text in the lower left until it's pointed out.

## Changes

Now when the changeset has conflicts, you can still click the apply changes button, but the conflicting mods are highlighted in red in the changeset, the conflict reasons take over the description column, and the apply button is disabled:

image

Hopefully this will make it more discoverable.

### DIFF
--- a/GUI/Controls/Changeset.cs
+++ b/GUI/Controls/Changeset.cs
@@ -15,7 +15,10 @@ namespace CKAN.GUI
             InitializeComponent();
         }
 
-        public void LoadChangeset(List<ModChange> changes, List<ModuleLabel> AlertLabels)
+        public void LoadChangeset(
+            List<ModChange> changes,
+            List<ModuleLabel> AlertLabels,
+            Dictionary<CkanModule, string> conflicts)
         {
             changeset = changes;
             alertLabels = AlertLabels;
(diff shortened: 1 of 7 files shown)


vinix38:
I have a suggestion: if there are conflicts, clicking the apply button will bring a popup describing the conflicts that are present and why changes cannot be applied.
I don't know if that would be useful or disruptive, what do you think?

HebaruSan:
> I have a suggestion: if there are conflicts, clicking the apply button will bring a popup describing the conflicts that are present and why changes cannot be applied. I don't know if that would be useful or disruptive, what do you think?

That's a good point, it would be impossible to miss the reason if it popped up like that. It would be a much bigger change, but might end up with overall simpler code.

Now I'm wondering whether we could take the user to the changeset tab with conflicts, but highlight the conflicts in red there...

HebaruSan:
OK, finally revisited this and got red highlighting working in the changeset, see revised description and screenshot.

@techman83, if you have time to review this, I think it's finally ready and takes us a step closer to the next release.

techman83 (approved):
Looks great! Making things clearer is always a great thing!

PR #3739, opened 2022-12-07
Fix auto-removal

### LINKED ISSUE #3738: [Bug] CKAN 1.31.3 Zero MiniAVC doesn't uninstall
Background
----------
* Operating System: Windows 11
* CKAN Version: 1.31.2 and 1.31.3 (dev build)
* KSP Version: 1.12.3

No

Problem
-------
**Describe the bug**
I had the same Zero MiniAVC crashing bug mentioned in #3711. However when I tried the new development build (1.31.3) mentioned under that issue that fixes the crash I noticed:
1. Selecting install boxes have longer than before delay, like the program freezes UI updates for 300ms or similar
2. Program tries to automatically remove Zero MiniAVC due to its dependencies being uninstalled. However when I click "apply", it runs the remove operation with no errors but it also fails to remove it. Which causes it to have the "apply" button enabled constantly as it tries to remove the mod every time

After this I ran the program in debug mode to troubleshoot it myself.
There was a mention of a broken registry in the logs after the failed uninstall despite there being no errors on the GUI.
It seems like in the 1.31.3 version, if there is a broken registry (in my case it was "retrofuture spaceplane parts modernized") there is no user side error given and removal of Zero MiniAVC is prevented. After reinstalling the aforementioned mod the issue seemed to resolve itself.

**Steps to reproduce**
1. Same as #3711
2. Swap to 1.31.3 from 1.31.2
4. Start ckan
5. You might need to remove the mods depending on zero miniavc, but I personally didn't have to do it myself.

**Expected behavior**
Removal of Zero MiniAVC mod.

image

```
PLACE CODE HERE !
```

HebaruSan:
## Problem

1. Install MiniAVC (or any other mod)
2. Check the checkbox for auto-installed
3. Click Apply changes
4. The changeset tab appears and shows the mod being removed
5. Click Continue
6. The install flow does nothing

## Cause

#3726 made it so that when you click to confirm a changeset, instead of recalculating the changeset from the mod grid as we did previously, it used the data from the same changeset object that the user just viewed and approved in the changeset tab. This allows us to use the changeset tab for changesets other than the one selected on the main mod list (such as reinstallation via the right click menu, which cannot otherwise be represented by the main mod list's checkboxes).

#3728 changed it to filter out non-user-requested changes. But this means auto-removals are filtered out of the changeset because their reason is not `SelectionReason.UserRequested` but rather `SelectionReason.NoLongerUsed`. They should be kept.

## Changes

Now all non-install changes are passed from the changeset tab to the install flow. Only non-user-requested installs are filtered out.

Fixes #3738.

### DIFF
--- a/GUI/Main/MainChangeset.cs
+++ b/GUI/Main/MainChangeset.cs
@@ -40,7 +40,9 @@ namespace CKAN.GUI
                     new KeyValuePair<List<ModChange>, RelationshipResolverOptions>(
                         changeset
                             // Only pass along user requested mods, so auto-installed can be determined
-                            .Where(ch => ch.Reasons.Any(r => r is SelectionReason.UserRequested))
+                            .Where(ch => ch.Reasons.Any(r => r is SelectionReason.UserRequested)
+                                // Include all removes and upgrades
+                                || ch.ChangeType != GUIModChangeType.Install)
                             .ToList(),
                         RelationshipResolver.DependsOnlyOpts()));
             }


techman83 (approved):
Makes sense! :tada:

PR #3749, opened 2022-12-23
Basic .NET 4.5 -> 4.8 update

unquietwiki:
Hey there. I wanted to update the program to .NET 6/7, to see if it resolved recent issues pulling metadata. I managed to clean up the build process to wherein the .NET 4.5 stuff is now .NET 4.8; anything .NET 5 should be unimpacted. The [.NET project upgrade tool](https://codemag.com/Article/2211082/Upgrade-Tooling-for-.NET-7) still struggles on this, but the program does work for me (albeit not as a single EXE, unless I picked the wrong folder to test).

No major code changes happened; I think it just adjusted some whitespace trying to re-implement references to different assemblies.


unquietwiki:
Now if this is accepted, I'd hit it with [some of this tooling](https://unquietwiki.com/en/TechFindings/op1) I've used in the past for re-factoring.

HebaruSan:
> recent issues pulling metadata

Are you referring to the recent issue of #3707, fixed in #3708?

I don't think we want to break things for users who have .NET 4.5 but not .NET 4.8 installed unless there's a good reason.

unquietwiki:
Thanks for the issue links; I'll put in any observations from my end on
that.

Regarding this, I suggest it be used for a branch to support .NET 7; which
the exes can be made for any OS via including required runtime parts. I
already noticed some references to .NET 5 as well. The tool from MS was
breaking on some duplicate references I couldn't find.

Michael Adams, unquietwiki.com

HebaruSan:
> I don't think we want to break things for users who have .NET 4.5 but not .NET 4.8 installed unless there's a good reason.

I just wanted to reiterate this to make sure it's clear: If we merged this, then some number of users would find that CKAN no longer works from them, and we would have to handle the support load of triaging their bug reports and telling them how to upgrade to .NET 4.8. I realize that you probably spent a lot of time and effort on this, and I don't want there to be hurt feelings if and when we close this without merging it.

unquietwiki:
@HebaruSan That's fine. (ponders) If there's something else to inform your decision-making, 4.5 went EOL years ago, and 4.5.2 went EOL earlier this year. Any Win 7 deployment I've worked on before that went EOL was supporting 4.7 (or even 4.8) via a Windows Update. I know Kerbal uses a bastardized Mono runtime that's roughly 4.7; so if you want to accept that as a baseline; otherwise, my admin experience suggests you should be fine.

https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-framework

unquietwiki:
@HebaruSan Hey, the .NET 4.8 build successfully pulled metadata today; it's re-downloading many mods as I write this. I also did set up a Hurricane Electric IPv6 tunnel on my router last night; so that could be a part of it too (maybe v4 isn't working on an endpoint, but v6 is).

HebaruSan:
That doesn't mean much to me, since I don't know what network problem you were having or whether you've done an apples-to-apples comparison with the master branch.

unquietwiki:
@HebaruSan Well, I already did that before the tunneling; the new build didn't pull the metadata same day as the old one. I made that tunnel change last night, and then it worked. I agree this is an edge case, but one that probably comes up for folks that may not realize its an issue (most websites you use will be maybe 100ms latency; this is almost 300).

unquietwiki:
@HebaruSan I also realize this may not be something you can immediately solve; but rather can suggest a possible workaround, until folks come up with a better solution. You can't do too much about IPv4 MTU issues with the app; that's VPN/network adapter turf.

HebaruSan:
The thing is that I don't know what we'd be working around. What problem are you trying to solve with this pull request?

unquietwiki:
@HebaruSan this PR was just for code modernization so you could eventually migrate to .NET 7, do cross-OS support, and resolve some of the other bugs via updated libraries. The network issue in #3751 is something that would have to be fixed via additional mirrors, CDNs, and offering workarounds.

HebaruSan:
OK, let's leave this as it is for now, then. We already have cross-OS support, and the other suggested benefits are entirely hypothetical as far as I can tell, as opposed to the real support load of breaking things for users who haven't upgraded to .NET 4.8.

unquietwiki:
@HebaruSan Acknowledged. At some point in the near future, I plan to do a deep dive on the .NET situation; you're not the only one that's had to worry about making such a change. I'll offer that up for your review when I do have that ready, so you can plan accordingly.

PR #3755, opened 2023-01-01
Attempt to fix .NET 4.5 compilation issue with VS 2022

unquietwiki:
@HebaruSan Happy New Year! This is a follow-up to #3749; it attempts to correct compilation issues on Visual Studio 2022 by adding references to .NET 4.5 assemblies in the project files, minor dependency updates, and instructions on how to further "hack" the dev environment to accept these projects without triggering the dead Dev Pack download.


unquietwiki:
Aborted attempt: ends up with SSL/TLS errors on connection.

PR #3762, opened 2023-01-11
Indy the dog patch 1

IndyTheDog:
Test security issue of public repo.
Letting other contributors submit pull request on a public repo with GH actions might allow them to disclose secrets

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,119 +23,13 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Adding HTTPS support to APT for old Mono images
-        run: |
-          apt-get update || true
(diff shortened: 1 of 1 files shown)


IndyTheDog:
Ok, so someone with intentions to abuse would need at least an approved PR before being able to use this technique.
Also I apologize the maintainer is offended. My intentions are not to hack but to raise awareness (I am learning defensive security). Feel free to discard my PR.
Note in the case my PR had built automatically, nothing would have been exposed as I only printed the MD5 of the secrets.

HebaruSan:
Right, GitHub already protects against these sorts of shenanigans by default. I hope you feel your awareness has been sufficiently raised. :raised_eyebrow:

PR #3766, opened 2023-01-18
Set git safe.directory in meta tester

HebaruSan:
## Problem

See KSP-CKAN/NetKAN#9527; the metadata tester is broken, but that attempt has not fixed the problem.

## Causes

Probing the docker image, I have slightly more confirmation now that this relates somehow to the permissions of the git repo (see the later comments in KSP-CKAN/NetKAN#9527).

However, there's a big gap in this explanation: **NOTHING CHANGED**. It's using the same old version of `git`, GitPython is probably the same, and our code is the same. For this to start happening when it did would imply that GitHub made some change to how Actions handle permissions right at the moment when we were rebuilding this docker image. So :shrug:.

## Changes

Now the Dockerfile for the meta tester sets git's `safe.directory` setting to `*`, which should allow our code to access the repo regardless of the mess that the GitHub Actions framework makes of the ownership or permissions.

I'll self-review this because it's low risk (the thing is already broken) and I want to test it right away.


HebaruSan:
I didn't work.

Issue #2636, opened 2019-01-07
Offer an AppImage for Linux?


zilti:
It would be nice to have an AppImage for the Linux users. It's a self-contained executable packed a bit similar to a dmg on macOS and would faciliate a lot.

HebaruSan:
How would this differ from downloading and running ckan.exe?

politas:
> How would this differ from downloading and running ckan.exe?

You wouldn't need to have a compatible Mono installation.

HebaruSan:
So we would bundle *Mono* into our package? Wouldn't that be hundreds of MB?

zilti:
Hundreds? It's interesting how today's devs have completely lost all sense of space. `mono-core` takes up 22.5 MB. But 1. you don't necessarily have to include mono - see for example OpenRA. And 2. having an AppImage gets you as the user quality-of-life features like update checks, updates, and a menu entry in whatever launcher you use.

HebaruSan:
I'm happy to admit when I'm wrong, but why be insulting about it? Remember that you're trying to convince someone to do something to help you here. :)

Regarding "see for example OpenRA", can you be more specific? Apparently it's a Red Alert clone, but what about it is relevant here? They don't seem to explain their Mono deployment strategy on their home page.

HebaruSan:
As for the actual size, it's not clear that mono-core is enough to run CKAN. The size of the actual downloads from mono-project.com are obscured by the debian repo folder structure (where are the actual .debs??), but for comparison the Windows download is 129 MB.

zilti:
What I meant to say is: OpenRA does not include Mono into their AppImage, so the user has to install it separately, is what I meant: "OpenRA is distributed as portable AppImages that run on most modern 64-bit Linux distributions. Mono 4.2 or later is required." Mono seems to be installed by default on most distributions anyway.

DasSkelett:
But the problem with CKAN is, that we need a version of mono that is _not_ installed by default on most distributions.
And if th user has to install mono and download the AppImage, what's the difference to the install process right now, where he has to install mono and download the ckan.exe?
Seems like the user just downloads a different file, to get the same, with the same effort.

zilti:
Ok, so what mono components does ckan use exactly? I'm guessing more than just mono-core and mono-wpf

zilti:
[...]

PR #3767, opened 2023-01-20
Fix mkbundle executable crash

memchr:
In `CKAN.GameInstance.PortableDir (CKAN.Games.IGame game)`:

```cs
            // Find the directory our executable is stored in.
            string exeDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
````

An empty string would be assigned to `exeDir` when ckan is packaged
using mkbundle. This caused following method call
`game.GameInFolder(new DirectoryInfo(exeDir))` to crash with
`System.ArgumentException: The specified path is not of a legal form (empty).`

```
mkbundle --deps -o ckan --simple --machine-config /etc/mono/4.5/machine.config --config /etc/mono/config --simple ckan.exe
```

### DIFF
--- a/Core/GameInstance.cs
+++ b/Core/GameInstance.cs
@@ -5,6 +5,8 @@ using System.Linq;
 using System.Reflection;
 using System.Text.RegularExpressions;
 using System.Transactions;
+using System.Diagnostics;
+
 using ChinhDo.Transactions.FileManager;
 using log4net;
 using Newtonsoft.Json;
@@ -241,9 +243,15 @@ namespace CKAN
 
             // Find the directory our executable is stored in.
             string exeDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
-            if (exeDir.Length == 0)
+            if (string.IsNullOrEmpty(exeDir))
             {
-                return null;
+                exeDir = Path.GetDirectoryName(Process.GetCurrentProcess().MainModule.FileName);
+                if (string.IsNullOrEmpty(exeDir))
+                {
+                    log.InfoFormat("Executing assembly path and main module path not found");
+                    return null;
+                }
+                log.InfoFormat("Executing assembly path not found, main module path is {0}", exeDir);
             }
 
             log.DebugFormat("Checking if {0} is in my exe dir: {1}",
--- a/Core/GameInstance.cs
+++ b/Core/GameInstance.cs
@@ -241,6 +241,10 @@ namespace CKAN
 
             // Find the directory our executable is stored in.
             string exeDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+			if (exeDir.Length == 0)
+			{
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Core/GameInstance.cs [hunk 1]
```diff
@@ -241,6 +241,10 @@ public static string PortableDir(IGame game)
 
             // Find the directory our executable is stored in.
             string exeDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+			if (exeDir.Length == 0)
+			{
+				return null;
+			}
```
HebaruSan, line 247:
```suggestion
            if (exeDir.Length == 0)
            {
                return null;
            }
```

Please keep the indentation consistent with the existing code.

HebaruSan:
For anyone else who wasn't familiar with `mkbundle`, it generates "self-contained executables that do not rely on Mono being installed on the system to simplify deployment of .NET Applications":

- https://www.mono-project.com/docs/tools+libraries/tools/mkbundle/

That sounds potentially useful as an alternate download format. Though for this specific fix, it would be better if we could fall back to the path of the mkbundle-generated executable when the executing assembly path can't be found, so users could still put their native-compiled `ckan` executable in the game root.

@memchr, would you mind checking the size of the executables generated in this way? I'm not currently in a Mono environment.

HebaruSan:
This might be the way to get the bundle path:

- https://stackoverflow.com/a/57649113/2422988

```
Path.GetDirectoryName(Process.GetCurrentProcess().MainModule.FileName)
```

memchr:
> would you mind checking the size of the executables generated in this way? I'm not currently in a Mono environment.

Size of executable generated with mono `6.12.0.177` on Linux is **36.1 MB**

HebaruSan:
Thanks! That's surprisingly reasonable compared to other modern day downloads (even if it's much larger than `ckan.exe` by itself). Bigger than Python, smaller than VLC or Git. This might make a nice middle ground between the current downloads and the full-on bloat packaging requested in #2636.

HebaruSan:
What command line are you using to build? My Ubuntu VM is giving me an error related to `I18N.dll`, but if you've already got it building enough to crash, maybe you already know the solution to this?

```
$ mkbundle --deps ckan.exe
Failure to load i18n assemblies, the following directories were searched for the assemblies:
   Path: .
In Custom mode, you need to provide the directory to lookup assemblies from using -L
ERROR: Couldn't load one or more of the i18n assemblies: Failed to load I18N.dll
```

memchr:
The command i am using is:
```bash
mkbundle --deps -o ckan --simple --machine-config /etc/mono/4.5/machine.config --config /etc/mono/config --simple ckan.exe
```

HebaruSan:
Thanks, that also worked on Ubuntu. 🎉

HebaruSan (approved):
Thanks for improving compatibility with `mkbundle`!

PR #3771, opened 2023-02-06
Clarify mod list saving options, add menu hotkeys

HebaruSan:
## Problem

If a user wants to install their current mod list on another computer or share it with friends, they might click File &rarr; Save installed mod list... to generate a text file with their list of mods. However, there is no way to use this file to install that same mod list somewhere else. (It's mainly useful for posting your mod list to the forum.)

Thanks to Discord user SpaceTrash67 for bringing this up.

## Cause

It's not easy to tell the difference between the menu options:

image

In the given use case, the user would have to try to guess and maybe experiment with these options to learn what they do.

## Changes

Now the English versions of these menu options have a parenthetical clarification explaining whether the files they create can be re-imported, similar to what we did in #3400:

image

Other locales would have to be updated subsequently by their individual translators.

I also added hotkeys to the menus and made the existing hotkeys more consistent; <kbd>i</kbd> should really install rather than import, since the latter is much less commonly used.

### DIFF
--- a/GUI/Main/Main.resx
+++ b/GUI/Main/Main.resx
@@ -123,27 +123,27 @@
   <data name="$this.Localizable" type="System.Boolean, mscorlib">
     <value>True</value>
   </data>
-  <data name="fileToolStripMenuItem.Text" xml:space="preserve"><value>File</value></data>
-  <data name="manageGameInstancesMenuItem.Text" xml:space="preserve"><value>Manage game instances</value></data>
-  <data name="openGameDirectoryToolStripMenuItem.Text" xml:space="preserve"><value>Open game directory</value></data>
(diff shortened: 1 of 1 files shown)


techman83 (approved):
I'm unsure what the purpose of the addition of all the `&amp;` is. It would make it easier to review if a refactor is in a separate commit for clarity. But overall LGTM!

HebaruSan:
The `&amp;` in a resource XML file translates to `&` in the string, which acts as a prefix to set the hotkey for a menu, button, or label in WinForms.

techman83:
I thought so. Would be good to split the commit in future, but that makes sense :slightly_smiling_face:

PR #3804, opened 2023-03-06
Cache path setting fixes

### LINKED ISSUE #3801: [Bug] Cache path setting can be just a drive letter but it doesn't work
Background
----------
* Win11
* CKAN Version: V1.31.2 
* KSP Version: V1.12.5.3190

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** No

Problem
-------
Installation fails every time for what ever mods used.

**Steps to reproduce**
It seems what causes the error during installation is changing the cache directory in the settings as shown in the picture:(Blue is before change, red after)
image
then i try to change the directory by just entering through the text box 
image
at first it seems like it works, but do not be fooled, it's sneaky.
then after leaving and re-entering the settings page, this is what you get
image
you can no longer edit the download cache location with the change button or any button.

then when attempting to install a mod an error will be out put as so:
Access to the path 'E:\System Volume Information' is denied.
Error during installation!

**Be advised - if you first install a mod or two, then try to change the directory. it will all work as it should with no problems!**
**Expected behavior**
Installation succesful

image
[CKAN DOWNLOAD CACHE EDIT FAILURE.txt](https://github.com/KSP-CKAN/CKAN/files/10889497/CKAN.DOWNLOAD.CACHE.EDIT.FAILURE.txt)

HebaruSan:
## Problem

If you go to the settings and try to type a valid full path with drive letters into the cache path field, it silently gets "stuck" after the drive letter behind the scenes. Upon re-opening the settings the field is blank and can't be edited:

image

The setting saved to disk is like:

```json
  "DownloadCacheDir": "E:",
```

Installing mods then fails with some strange text about a `System Volume Information` folder:

image

## Cause

`TrySetupCache` is supposed to determine whether a cache path is valid, but an exception was being thrown in `NetModuleCache.GetSizeInfo` when the path was `E:`, _after_ `TrySetupCache` succeeded, which corrupted the display and caused `updatingCache` to be `true` forever, which suppressed all further calls to update the setting while you typed the rest of the path.

Later, when we tried to use the cache during an install, `NetModuleCache.GetSizeInfo` again threw its exception and broke everything.

## Changes

- Now before we check whether the path segments match up, `GetDrive()` directly compares the given path to the drive's `RootDirectory` in the hopes of avoiding NREs for paths like `E:`
- Now `TrySetupCache` tries to check the free space associated with a cache path before changing the setting to it, to ensure it only ends up with valid values, which should make it fail for `E:`
- Now if we fail to initialize the cache at startup, we revert the cache path setting to its default and try again, to make sure we don't end up with an invalid cache object, to provide a recovery path for affected users
- Now instead of trying to change the cache setting with every keystroke, the settings window creates its own temporary cache object to validate the path, then tries to change the setting when the window is closed. If that fails, it'll display an error (again) and make you go back and fix it.
- Now we don't create the `downloading` folder until you actually _use_ a cache folder (previously if you typed a long path, we'd try creating it at each level because it was in `NetModuleCache`'s constructor)
- Now the English strings for the settings window say "MiB" instead of "MB" since that's technically more correct

Fixes #3801.

@navuek, there should be a test build that you can try here in a few minutes on the Checks tab under the Artifacts dropdown.

### DIFF
--- a/Core/Extensions/IOExtensions.cs
+++ b/Core/Extensions/IOExtensions.cs
@@ -28,6 +28,8 @@ namespace CKAN.Extensions
             return true;
         }
 
+        private static readonly char[] pathDelims = new char[] {Path.DirectorySeparatorChar};
+
         /// <summary>
         /// Check whether a given path is an ancestor of another
         /// </summary>
@@ -36,10 +38,8 @@ namespace CKAN.Extensions
         /// <returns>true if child is a descendant of parent, false otherwise</returns>
         public static bool IsAncestorOf(this DirectoryInfo parent, DirectoryInfo child)
             => StringArrayStartsWith(
-                child.FullName.Split(new char[] {Path.DirectorySeparatorChar},
-                                     StringSplitOptions.RemoveEmptyEntries),
-                parent.FullName.Split(new char[] {Path.DirectorySeparatorChar},
-                                      StringSplitOptions.RemoveEmptyEntries));
+                child.FullName.Split(pathDelims, StringSplitOptions.RemoveEmptyEntries),
+                parent.FullName.Split(pathDelims, StringSplitOptions.RemoveEmptyEntries));
 
         /// <summary>
         /// Extension method to fill in the gap of getting from a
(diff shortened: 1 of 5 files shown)


techman83 (approved):
That makes a lot of sense, LGTM.

navuek:
Thank you very much for the quick assist!!

PR #3807, opened 2023-03-11
Make grid ampersand workaround platform specific

### LINKED ISSUE #3806: [Bug] Ampersands doubled in mod grid on Windows
Background
----------
* Operating System: Windows 10 Home
* CKAN Version: 1.31.2
* KSP Version: N/A; Metadata bug

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** Yes

Problem
-------
**Describe the bug:** Crew R and R mod (CrewQueueTwo) has 2 &s in its name on CKAN 

**Steps to reproduce:** Open CKAN
* Search "Crew R"
* Read name field for Crew R&R mod

**Expected behavior**
Text matches mod name

image

**CKAN error code (if applicable):** N/A

HebaruSan:
## Background

Originally, mod names, abstracts, and authors containing `&` characters would have them removed and the next character underlined on Mono:

image

#3149 fixed that by escaping the ampersands, which is done by adding another one.

## Problem

Windows clients currently see `&&` in the grid:

image

## Cause

.NET is cross-platform in principle, but not quite "write once run anywhere" in practice. Whenever Mono deviates from how .NET does something, pain ensues for application developers. In this case, Mono treats `&` as a hotkey prefix character in the data grid view, but .NET treats it as just a normal character.

## Changes

Now we only double the ampersand on Mono. Windows gets the single ampersand and likes it:

image

Fixes #3806.

### DIFF
--- a/GUI/Model/ModList.cs
+++ b/GUI/Model/ModList.cs
@@ -373,8 +373,8 @@ namespace CKAN.GUI
                     Value = "-"
                 };
 
-            var name   = new DataGridViewTextBoxCell { Value = mod.Name.Replace("&", "&&") };
-            var author = new DataGridViewTextBoxCell { Value = string.Join(", ", mod.Authors).Replace("&", "&&") };
+            var name   = new DataGridViewTextBoxCell { Value = ToGridText(mod.Name)                       };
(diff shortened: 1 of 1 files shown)


techman83 (approved):
Make it cross platform they said, it'll be fun they said! :laughing:  - Looks like a clean implementation :slightly_smiling_face:

PR #3808, opened 2023-03-11
Default GAME env var to KSP in Inflator Dockerfile, fix redeploy

HebaruSan:
## Problem

If the `$GAME` environment variable isn't set, the Inflator won't start, as of #3797.

Also the `docker-inflator` Cake build target tries to redeploy the `Inflator` container, but that was renamed to `InflatorKsp` in KSP-CKAN/NetKAN-Infra#284. And the `InflatorKsp2` container isn't touched at all.

## Changes

- Now if it's not set, we fall back to a default value of `KSP` using `bash`'s `${ENVVAR:-defaultval}` syntax.
  - <https://stackoverflow.com/questions/2013547/assigning-default-values-to-shell-variables-with-a-single-command-in-bash>
- Now we redeploy `InflatorKsp` followed by `InflatorKsp2`

### DIFF
--- a/Dockerfile.netkan
+++ b/Dockerfile.netkan
@@ -5,6 +5,6 @@ RUN useradd -ms /bin/bash netkan
 USER netkan
 WORKDIR /home/netkan
 ADD netkan.exe .
-ENTRYPOINT /usr/bin/mono netkan.exe --game $GAME --queues $QUEUES \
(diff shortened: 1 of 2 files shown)


techman83 (approved):
LGTM! Thanks!

PR #3810, opened 2023-03-11
Fix NRE on purging cache in GUI

### LINKED ISSUE #3809: [Bug] NRE thrown on purge all cache files
Background
----------
* Operating System: Windows 11 Home 64-bit (10.0, Build 22621) (22621.ni_release.220506-1250)
* CKAN Version: 1.31.1
* KSP Version: 1.12.5 (Although I don't think it matters here)

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** Yes, but not to the CKAN cache - nor do I have the CKAN cache open elsewhere.

Problem
-------
**Describe the bug**
With a 12.8gb CKAN cache, purge all throws an NRE.

**Steps to reproduce**
* Open CKAN
* Go to CKAN settings
* Press purge
* Select purge all

**Expected behavior**
Cache files are purged

image

**CKAN error code (if applicable):** N/A; Exception text can be found below:

```

   at CKAN.GUI.Contents._UpdateModContentsTree(CkanModule module, Boolean force)
   at CKAN.GUI.SettingsDialog.PurgeAllMenuItem_Click(Object sender, EventArgs e)
   at System.Windows.Forms.ToolStripItem.RaiseEvent(Object key, EventArgs e)
   at System.Windows.Forms.ToolStripMenuItem.OnClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleClick(EventArgs e)
   at System.Windows.Forms.ToolStripItem.HandleMouseUp(MouseEventArgs e)
   at System.Windows.Forms.ToolStrip.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.ToolStripDropDown.OnMouseUp(MouseEventArgs mea)
   at System.Windows.Forms.ToolStrip.WndProc(Message& m)
   at System.Windows.Forms.ToolStripDropDown.WndProc(Message& m)
```

HebaruSan:
## Problem

If you use the purge all option in the GUI settings, an exception is thrown:

image

## Causes

- The `Contents` tab has a poor public interface, with two ways to change the active mod, the `SelectedModule` property and `UpdateModContentsTree`
- `UpdateModContentsTree` has a `module` parameter, which is malsuited to this function's usage, which is simply to refresh the currently displayed mod after caching or purging, which requires calling code to do unnecessary work to retrieve and pass the currently shown mod
- `UpdateModContentsTree` assumes that its `CkanModule` parameter will never be `null`
- `SettingsDialog` passes that parameter as `null` after purging, to try to refresh the display in case a previously cached mod is shown, because it seemed like a natural way to request an operation on the same mod

## Changes

- Now the public interface of the `Contents` tab has _only_ `SelectedModule` for setting the mod, and `UpdateModContentsTree` is replaced by a new `Refresh` function for refreshing, which importantly does not have a module parameter and only re-uses the currently displayed one
- Similarly `Main.UpdateModContentsTree` and `ModInfo.UpdateModContentsTree` are replaced by zero-parameter functions `Main.RefreshModContentsTree` and `ModInfo.RefreshModContentsTree` calling `Contents.Refresh`
- Now all usages of `UpdateModContentsTree` are replaced by calls to `Refresh`, which eliminates the possibility of a null reference exception

Fixes #3809.

@GreenKeldeo, a few minutes after this is submitted, assuming there are no compile errors or failed tests, there should be a test build here under the Artifacts dropdown of the Contents tab, if you'd like try out the fix.

### DIFF
--- a/GUI/Controls/ManageMods.cs
+++ b/GUI/Controls/ManageMods.cs
@@ -1042,7 +1042,7 @@ namespace CKAN.GUI
                     Main.Instance.Manager.Cache.Purge(mod);
                 }
                 selected.UpdateIsCached();
-                Main.Instance.UpdateModContentsTree(selected.ToCkanModule(), true);
+                Main.Instance.RefreshModContentsTree();
             }
         }
 
--- a/GUI/Controls/ModInfo.cs
+++ b/GUI/Controls/ModInfo.cs
@@ -42,9 +42,9 @@ namespace CKAN.GUI
             get => selectedModule;
         }
 
-        public void UpdateModContentsTree(CkanModule module, bool force = false)
+        public void RefreshModContentsTree()
         {
-            Contents.UpdateModContentsTree(module, force);
+            Contents.Refresh();
         }
 
         public event Action<GUIMod>            OnDownloadClick;
--- a/GUI/Controls/ModInfoTabs/Contents.cs
+++ b/GUI/Controls/ModInfoTabs/Contents.cs
@@ -23,15 +23,18 @@ namespace CKAN.GUI
                 if (value != selectedModule)
                 {
                     selectedModule = value;
-                    UpdateModContentsTree(selectedModule.ToModule());
+                    Util.Invoke(ContentsPreviewTree, () => _UpdateModContentsTree(selectedModule.ToModule()));
                 }
             }
             get => selectedModule;
         }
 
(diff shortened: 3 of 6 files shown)


techman83 (approved):
Sensible improvement, thanks @HebaruSan !

PR #3811, opened 2023-03-11
Multi-game fakery

HebaruSan:
## Motivation

As we advance toward the finish line of KSP2 support, we are getting closer to the point where we will need to be able to validate, review, and merge `KSP2-NetKAN` pull requests. This will require updating `xKAN-meta_testing` to support the new game. I started that and discovered that `ckan.exe` needs some more updates first.

## Problem

`ckan instance fake` is used by the validator to generate dummy game instances, but it currently can only do this for KSP1.

## Changes

- Now the `ckan instance fake` command has an optional `--game` argument that can be `KSP` (default) or `KSP2`, which determines the files and folders to be created to fake the instance according to the game. The metadata validator will eventually pass the value of a `$GAME` environment variable here to adapt itself to the right context.
- To accomplish this, remaining game-specific logic in `GameInstance.FakeInstance` is updated to use the `IGame game` parameter it already has
- `IGame.RebuildSubdirectories` now takes a path instead of a game instance so we can use it to create the game instance's files and folders _before_ instantiating the object
- `KerbalSpaceProgram2.RebuildSubdirectories` now creates the `KSP2_x64_Data` folder

This should allow us to generate fake KSP2 instances for the pull request validator.

### Known limitations

We can't generate a valid `KSP2_x64.exe` file on the fly, and this is how we get the game version!

To work around this, we just create it as if it's a `buildID.txt` file, and then if our attempt to get the version from that file fails, we fall back to using the most recent known version. This allows the fake instances to be added as valid, but for now they will not be able to represent exclusively older versions. Hopefully there will be a better way to get the game version eventually.

### DIFF
--- a/Cmdline/Action/GameInstance.cs
+++ b/Cmdline/Action/GameInstance.cs
@@ -51,7 +51,7 @@ namespace CKAN.CmdLine
                 {
                     // First the commands with three string arguments
                     case "fake":
-                        ht.AddPreOptionsLine($"{Properties.Resources.Usage}: ckan instance {verb} [{Properties.Resources.Options}] name path version [--MakingHistory <version>] [--BreakingGround <version>]");
+                        ht.AddPreOptionsLine($"{Properties.Resources.Usage}: ckan instance {verb} [{Properties.Resources.Options}] name path version [--game KSP|KSP2] [--MakingHistory <version>] [--BreakingGround <version>]");
                         break;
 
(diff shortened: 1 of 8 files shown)


HebaruSan:
Self-reviewing to get the ball rolling on this...

techman83 (commented):
Retro review! Looks sensible to me

PR #3822, opened 2023-04-12
Only update Versions tab when the mod changes

HebaruSan:
## Problem

If you switch to the Versions tab, then to another tab and back, the Versions tab fully refreshes itself, which it shouldn't do because it can take a long time (see #3638).

## Cause

The call to `Refresh` is unconditional and is only skipped if the mod is null.

## Changes

Now we don't refresh the versions list if it's already loaded for the same mod. There's already a code block checking if the mod is the same, so the call to `Refresh` is moved in there.


techman83 (approved):
That makes sense, clean fix!

PR #3827, opened 2023-04-18
Warnings for missing swinfo.json deps

HebaruSan:
## Problem

See KSP-CKAN/KSP2-NetKAN#33, a dependency was missing from a KSP2 module. This dependency was documented (only) in the `swinfo.json` file, and the netkan didn't have it.

This could easily happen again with other mods, either when initially adding them or as new versions are released.

## Cause

The inflator parses the `swinfo.json` file but ignores its `dependencies` property. I thought about automatically importing it into `CkanModule.depends`, but that would have trapped us if:

- The identifier was different in CKAN (there's no guarantee of matches)
- The `swinfo.json` was wrong

## Changes

Now when we inflate a module with a `swinfo.json` file, if it has `dependencies` listed, we'll check them and log a warning for any identifier that isn't in the `CkanModule.depends` list. It's a warning instead of an error so a human can decide what to do.

```
$ netkan.exe --game KSP2 NetKAN/CheatsMenu.netkan
1788 [1] WARN CKAN.NetKAN.Transformers.SpaceWarpInfoTransformer (null) - Dependencies from swinfo.json missing from module: ShadowUtilityLIB
```

For now, min/max versions are ignored. From what I've seen, these tend to be filled in unnecessarily with non-useful data, so it would be a lot of noise / alert-fatigue to see these all if we logged them.

This way we will get a notification in Discord if a KSP2 module adds a new dependency and documents it in the `swinfo.json` file, or in the GitHub Action output in a pull request.

### Considered and not done

In principle, this code belongs in a `Validator` file/class.

In practice, I started that and it got really messy. I had to duplicate essentially _all_ of the logic of the `Transformer` class, including retrieving the file from the ZIP and checking the remote copy. The last straw was when I had to add an `IGithubApi` object to `CkanValidator`, which really does not belong there.

Since I was ending up with an extended copy of the `Transformer` class, I thought it would be cleaner to extend the original. We can look into adding some sort of shared retrieval and caching for `swinfo.json` data in the future and revisit this then.

This means these warnings will only fire for `.netkan` files, not for validating `.ckan` files. I think that's fine for now.

### DIFF
--- a/Netkan/Transformers/SpaceWarpInfoTransformer.cs
+++ b/Netkan/Transformers/SpaceWarpInfoTransformer.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 using ICSharpCode.SharpZipLib.Zip;
 using log4net;
@@ -10,6 +11,7 @@ using CKAN.NetKAN.Sources.SpaceWarp;
 using CKAN.NetKAN.Services;
 using CKAN.NetKAN.Extensions;
 using CKAN.NetKAN.Sources.Github;
+using CKAN.Extensions;
 using CKAN.Versioning;
 using CKAN.Games;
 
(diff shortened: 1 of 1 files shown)


HebaruSan:
Self-reviewing so I can get a sense of what's out there in the already indexed mods...

HebaruSan:
This worked, but we didn't catch the message (eventually addressed in KSP-CKAN/KSP2-NetKAN#44):

image

PR #3828, opened 2023-04-19
Treat reinstalling a changed module as an update

### LINKED ISSUE #3825: [Bug]: Repeated "metadata has changed" messages

### Operating System

Windows 11

### CKAN Version

1.32.0

### KSP Version

KSP2 v0.1.2.0

### Did you make manual changes?

No

### Describe the bug

image

I periodically check for mod updates in CKAN throughout the day.  This message showed up a few days ago, and I clicked "yes" and reinstalled the mod.  Later that day, and a few times every day since, on subsequent mod list refreshes, the same message reappears for the same mod.

Initially reported in Discord; HebaruSan asked me to open an issue for it.

### Steps to reproduce

* Have CheatsMenu installed via CKAN on KSP2
* Periodically refresh mod list in CKAN

### Relevant log output

_No response_

HebaruSan:
## Background

If you install a module, and then we later change it in the metadata repo, you'll get this popup when you refresh your registry:

popup

If you click Yes, the module is reinstalled. If you click No, then it's not. The idea is to keep users up to date with the latest fixes and improvements as we maintain the metadata, such as missing dependencies or broken install stanzas.

## Problems

- This popup is confusing in general. What is it saying? Should I say Yes or No?
- If you click No, there's no way to go back if you change your mind, other than refreshing again, which will only work if _other_ modules in the repo have been changed in the meantime
- Currently reinstalling does not reliably update the installed module's metadata in the registry (which is the whole reason we're doing it), so the popup keeps reappearing for the same mods.

## Causes

I think this line from #3344 re-installs from the _current/old_ metadata, not the new:

https://github.com/KSP-CKAN/CKAN/blob/4aa4db6aabb6af1e97b930784187f410b74bfa2b/GUI/Main/MainRepo.cs#L164

`ModuleInstaller.Upgrade` makes promises about reinstalling, into which I read too much.

## Motivations

@JonnyOThan had some questions about refactoring how CKAN installs RPM. While investigating, I had a thought:

image
(Don't worry, it wasn't actually _that_ hard.)

The logic here is that we sometimes need to make related changes to the metadata, for example adding a new module and updating other modules to depend on it. It makes sense to apply such changes together in the same filesystem transaction, but currently we can't do that; reinstalling changed modules happens first, during repo update, and updating to new versions of modules happens separately. This allows users to end up with inconsistent metadata, some from before we merged a PR and some from after.

## Changes

- Now the repo update code no longer deals with module changes at all. The kraken class and the associated code, popup, and i18n resources for triggering a reinstall are deleted.
- Now `IRegistryQuerier.HasUpdate` returns `true` if the metadata for an installed module differs from the same mod version from the repo, using the same comparison logic the repo update code used previously (which has been migrated to `CkanModule.MetadataEquals`)
  - This causes the upgrade checkbox to show up for such mods
    - If checked, the changeset will include an upgrade entry containing the module from the repo (if available) instead of the installed module, to ensure it is updated properly
    - If checked, the changelog will say "Re-install (metadata changed)" (with translations cobbled together from other strings where possible)

You can now freely refresh your registry and not worry about what "metadata changed" means. Any modules with changed metadata will be included in your next pass of updates, which will happen all together in one filesystem transaction. This will be easier for users (fewer steps and screens to learn) and better for maintainers (related changes will be applied together).

Fixes #3825 (I think, needs confirmation). @Silvia-Dragoness, if you have time to take a look at it, there should be a test build here under the Checks tab, Artifacts dropdown a few minutes after this is submitted.


techman83 (approved):
Looks good to me! Some tests on the comparison logic would be good, but I am not across the value vs implementation time here.

PR #3829, opened 2023-04-21
Allow GUI users to delete registry lockfiles

### LINKED ISSUE #3558: [Feature] Option to delete registry.locked and continue in GUI
**Extremely low priority; me just being lazy after a recent issue which caused CKAN to crash repeatedly.**

Problem
-------
It's a tiny bit annoying to have to manually delete the "registry.locked" file after CKAN has crashed or was killed/incorrectly shut down.

Suggestions
-----------
It would be nice if CKAN could warn the user (as at present), but then also allow the user to continue without having to manually delete the registry.locked file.

Granted this introduces a higher level of risk with users accidentally launching a second instance of CKAN, continuing, and then presumably causing filesystem/registry corruption. Perhaps gate this functionality behind an "experienced user" option?

HebaruSan:
## Motivation

ConsoleUI can do one thing that GUI can't:

image

image

It would be helpful if GUI users also had the option to delete the lockfile after an explanatory warning.

## Problems

- ConsoleUI's lockfile message confusingly says "instance" twice with different meanings
- The GUI window take a long time to appear at startup (for KSP1); as a rule, the app should be visible as early as possible so users can see it's running, and they should be able to tell what it's doing if it's busy.
- `Main`'s constructor was calling `Application.Run`, which is not good coding style; a constructor should just create the object (even under RAII), not trigger side effects such as _running the entire application_

## Cause

- GUI calls `Registry.Instance` before the window appears, which makes it load `registry.json` which is around 30 MiB nowadays and causes a substantial delay

## Changes

- ConsoleUI's lockfile message moved to `Core.RegistryInUseKraken.ToString` to be shared, and the English text is edited to be clearer (one "instance" is now "CKAN process" and the other is "game folder")
- Now GUI's lockfile error popup is replaced with a yes/no dialog with custom button captions (translations copied from ConsoleUI where available):
  image
  If you click Cancel, the manage game instances popup will appear so you can pick a different game instance (if you have any others). If you click Force, it'll try to delete the lockfile and proceed with the chosen instance.
- Now the GUI window appears before `registry.json` is loaded, and a progress bar is shown while it loads
- The sometimes-slow `GameInstance.Scan` call is moved from startup to `UpdateModsList` with its own log message to keep the GUI responsive while it happens (since this was previously _between_ the two progress bar sections)
- We now pass the parent form to all the calls of `ShowDialog` that I was able to find, which is good practice in case we want to center the child dialogs on the parents
- `Main.manager` is now private and external users are switched to using the public `Main.Manager` instead, and various other members of `Main` are also made private or readonly
- All potentially slow actions are removed from `Main`'s constructor and from `OnLoad` and moved to `OnShown`, where they now occur within a `Wait.StartWaiting` block that shows a progress bar. The roles of these functions are now:
  - `Program.Main_`: Instantiate the main form, hide the console window, and run the application
  - Constructor: Just set up the basic core properties so `Main` can be loaded, such as the GUI events
  - `OnLoad`: Get the GUI configuration (without locking the instance!) so we can get the saved window geometry/layout before the window appears
  - `OnShown`: Load the registry once the window appears and prompt the user for various things
- GUI configuration is now only saved at exit and when we change instances rather than every time we change any setting
- `Main.tabController` and `Main.Instance` are now marked with the `[Obsolete]` attribute to generate code style warnings, and the `Main*.cs` file now all have `#pragma warning disable 0618` to suppress the warning when used by the same class

This will generally make GUI better behaved, since it will appear very soon after launch and tell the user what it's doing. And if the lockfile exists, users will have the option to delete it.

Fixes #3558.

### Deferred to later

We should also show a progress bar when the user changes instances via the manage game instances popup. I looked into that and it was more complicated than I could manage at this point due to conflicting usages of the progress bar tab.

### DIFF
--- a/ConsoleUI/GameInstanceListScreen.cs
+++ b/ConsoleUI/GameInstanceListScreen.cs
@@ -185,7 +185,7 @@ namespace CKAN.ConsoleUI {
                 } catch (RegistryInUseKraken k) {
 
                     ConsoleMessageDialog md = new ConsoleMessageDialog(
-                        string.Format(Properties.Resources.InstanceListLocked, k.lockfilePath),
+                        k.ToString(),
                         new List<string>() {
                             Properties.Resources.Cancel,
                             Properties.Resources.Force
--- a/ConsoleUI/Properties/Resources.Designer.cs
+++ b/ConsoleUI/Properties/Resources.Designer.cs
@@ -362,9 +362,6 @@ namespace CKAN.ConsoleUI.Properties {
         internal static string InstanceListLoadingError {
             get { return (string)(ResourceManager.GetObject("InstanceListLoadingError", resourceCulture)); }
         }
-        internal static string InstanceListLocked {
-            get { return (string)(ResourceManager.GetObject("InstanceListLocked", resourceCulture)); }
-        }
         internal static string InstanceListNoVersion {
             get { return (string)(ResourceManager.GetObject("InstanceListNoVersion", resourceCulture)); }
         }
--- a/ConsoleUI/Properties/Resources.fr-FR.resx
+++ b/ConsoleUI/Properties/Resources.fr-FR.resx
@@ -421,13 +421,6 @@
     <value>Erreur lors du chargement de {0} :
 {1}</value>
   </data>
-  <data name="InstanceListLocked" xml:space="preserve">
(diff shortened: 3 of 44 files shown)


techman83 (approved):
Yep, that will solve a few discord/issue raises. LGTM!

PR #3831, opened 2023-04-24
Add download link to staging PRs

HebaruSan:
## Motivation

Dealing with KSP-CKAN/CKAN-meta#3183, KSP-CKAN/CKAN-meta#3184, KSP-CKAN/CKAN-meta#3185, KSP-CKAN/CKAN-meta#3186, and KSP-CKAN/CKAN-meta#3187 requires downloading the mod, but how to do this must be figured out by the PR reviewer (generally the answer is to click the spacedock or repository link and then click the download link on the host). This is inconvenient.

## Changes

- The `StagingLinksTransformer` now executes after `DownloadAttributeTransformer` and `InstallSizeTransformer`, which from reviewing #3454 I think should be fine
- Before it loops over the resources, `StagingLinksTransformer` adds a row containing `metadata.Download`

### DIFF
--- a/Netkan/Transformers/NetkanTransformer.cs
+++ b/Netkan/Transformers/NetkanTransformer.cs
@@ -55,12 +55,12 @@ namespace CKAN.NetKAN.Transformers
                 new VersionEditTransformer(),
                 new ForcedVTransformer(),
                 new EpochTransformer(),
-                new StagingLinksTransformer(),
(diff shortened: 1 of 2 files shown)


HebaruSan:
Self-reviewing because this is small and so we can start benefitting from it in the next pull request from SpaceDock.

PR #3834, opened 2023-04-30
Installation history tab

### LINKED ISSUE #3741: [Feature] Browse `CKAN/history/*.ckan` files in GUI
Problem
-------
It's really annoying when ckan crashes or some other problem that restarts ckan and the list of mods to install/remove is wiped.

Suggestions
-----------
A history/changelist so you can see the changes you have made recently through restarts.
Or the ability to save a list of mods to a file to install a bunch of mods at once.
I would try coding it myself but i don't know C#

HebaruSan:
## Motivation

Users often ask for a way to see what they have installed previously, and when we tell them this already exists in the `<GameRoot>/CKAN/history` folder, they then request an easier way to access those files.

## Changes

Now a new menu option opens an Installation History tab containing a list of all the timestamps for which there is a history entry in the `<GameRoot>/CKAN/history` folder:

image

image

The OK button closes the tab and returns the user to Manage Mods.

If the user clicks an entry at the left, the mods that were installed at that point in time are listed on the right, split into a "Not currently installed" group at the top and a "Currently installed" group below that to make it easier to see the differences:

image

The user can click a mod to open ModInfo:

image

If the user clicks the Install toolbar button, the currently displayed mods that aren't installed are put into the changeset, which the user can then install:

image

If there are conflicts with already installed mods, these will be highlighted in the changeset and the user will not be able to continue:

image

Fixes #3741.

### DIFF
--- a/Core/GameInstance.cs
+++ b/Core/GameInstance.cs
@@ -122,20 +122,6 @@ namespace CKAN
                 User.RaiseMessage(Properties.Resources.GameInstanceCreatingDir, InstallHistoryDir());
                 txFileMgr.CreateDirectory(InstallHistoryDir());
             }
-
-            // Clear any temporary files we find. If the directory
-            // doesn't exist, then no sweat; FilesystemTransaction
-            // will auto-create it as needed.
-            // Create our temporary directories, or clear them if they
-            // already exist.
-            if (Directory.Exists(TempDir()))
-            {
-                var directory = new DirectoryInfo(TempDir());
-                foreach (FileInfo file in directory.GetFiles())
-                    txFileMgr.Delete(file.FullName);
-                foreach (DirectoryInfo subDirectory in directory.GetDirectories())
-                    txFileMgr.DeleteDirectory(subDirectory.FullName);
-            }
             log.InfoFormat("Initialised {0}", CkanDir());
         }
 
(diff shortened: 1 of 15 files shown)


HebaruSan:
Not a real test failure, will re-run after a few minutes.

image

Core/GameInstance.cs
```diff
@@ -122,20 +122,6 @@ private void SetupCkanDirectories(bool scan = true)
...
-            // Create our temporary directories, or clear them if they
-            // already exist.
-            if (Directory.Exists(TempDir()))
-            {
-                var directory = new DirectoryInfo(TempDir());
-                foreach (FileInfo file in directory.GetFiles())
-                    txFileMgr.Delete(file.FullName);
-                foreach (DirectoryInfo subDirectory in directory.GetDirectories())
-                    txFileMgr.DeleteDirectory(subDirectory.FullName);
-            }
```
techman83, line 138: This is the only bit I'm not clear on

techman83 (approved):
Looks like a really cool addition! Only had one minor comment, I'm sure there is a reason, I just can't see it :slightly_smiling_face:

Core/GameInstance.cs
HebaruSan: `GameInstance.TempDir` corresponds to `<GameRoot>/CKAN/temp`, but it has never been used for anything since I've been on the team. Since no code ever adds files or directories, there's no need to clear them out.

techman83: Excellent, love code cleanup!

PR #3837, opened 2023-05-05
Scan for DLLs with or without primary mod dir

HebaruSan:
## Problem

If you manually install SpaceWarp for KSP2, it will be detected. But if you then delete the `BepInEx` folder, all your mods are gone, but any that were manually installed will still show up as such.

## Cause

`GameInstance.Scan` aborts its work if the primary mod dir (`BepInEx/plugins` in this case) doesn't exist. This includes clearing the list of manually installed mods.

## Changes

Now the DLL scan partly happens if the primary mod dir is missing; we clear the DLL list unconditionally, and then if the primary mod dir exists, we scan it and register the DLLs. Otherwise it just stays blank, as it should.

### DIFF
--- a/Core/GameInstance.cs
+++ b/Core/GameInstance.cs
@@ -352,25 +352,22 @@ namespace CKAN
         /// </returns>
         public bool Scan()
         {
-            if (Directory.Exists(game.PrimaryModDirectory(this)))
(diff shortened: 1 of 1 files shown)


techman83 (approved):
Makes sense to me :slightly_smiling_face:

Issue #3658, opened 2022-09-06
game instances


lordshitass:
Background
----------
* Operating System: 22000.795
* CKAN Version: 1.31.0
* KSP Version: 1.12.2

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
No

Problem
-------
**Describe the bug**
I'm stupid and messed up the game instances I've uninstalled both KSP and CKAN but it doesn't help and it makes things worse
(sorry if this isn't enough I'm not exactly sure how I messed it up also I'm sorry if I've upset y'all)

**Steps to reproduce**
* Go to...
* Click on...
* See error

**Expected behavior**
im just trying to open my game with mods 

**Screenshots (if applicable)**

**CKAN error code (if applicable):**

```
PLACE CODE HERE !
```

HebaruSan:
Please describe the problem you're having with CKAN.

lordshitass:
> Please describe the problem you're having with CKAN.

I've tried to get CKAN to recognize that I have uninstalled KSP once but its not doing that so I tried my best to make a new instance but I did something wrong and it wont open KSP

HebaruSan:
> I tried my best to make a new instance

Did you use the "Fake new KSP instance" option to do that?

![image](https://user-images.githubusercontent.com/1559108/188696738-ecc55800-f06a-4f9a-95f9-fff82c9dd9ed.png)

If so, "it wont open KSP" is expected, that option creates a folder without the game for testing. To play KSP, you need to purchase and install it outside of CKAN.

lordshitass:
> > I tried my best to make a new instance
> 
> Did you use the "Fake new KSP instance" option to do that?
> 
> ![image](https://user-images.githubusercontent.com/1559108/188696738-ecc55800-f06a-4f9a-95f9-fff82c9dd9ed.png)
> 
> If so, "it wont open KSP" is expected, that option creates a folder without the game for testing. To play KSP, you need to purchase and install it outside of CKAN.

I did do that but I've already purchased KSP before

HebaruSan:
OK, do **not** use the "fake" option, it's not what you want. Use "Add instance to CKAN" to tell CKAN where your game is installed.

https://github.com/KSP-CKAN/CKAN/wiki/User-guide#adding-game-folders

lordshitass:
> 

thank you

Issue #3177, opened 2020-10-15
Bug to launch a fake KSP instance


ToxicR11:
Background
----------
* Operating System: Windows 10
* CKAN Version:  1.28.0
* KSP Version:  1.3.0.1804

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
No

Problem
-------
After creating a fake ksp instance, i try to launch KSP and there is a message "Couldn't start KSP. Le fichier spécifié est introuvable" (traduction : the specified file cannot be found )

**Steps to reproduce**
* Go to create fake KSP instance
Create a fake ksp instance
* Click on launch KSP
* See error

**Expected behavior**
launch the game

**Screenshots (if applicable)**

https://postimg.cc/F7sbhp2f/3b493da8

**CKAN error code (if applicable):**

```
PLACE CODE HERE !
```

HebaruSan:
Duplicate of #2850

HebaruSan:
@ToxicR11, a "fake" instance is not a real instance, the game is not installed there. You're not supposed to be able to run it.

Issue #2850, opened 2019-08-22
[Bug] Error launching KSP


CorSPACE:
Background
----------
* Operating System: Windows 10
* CKAN Version: v1.26.4
* KSP Version: 1.7.3.2594

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**

I did a fresh install prior to installing CKAN

Problem
-------
CKAN will not launch KSP with the default command line argument or any other ones I found online

**CKAN error code (if applicable):**

```
Couldn't start KSP.

The system cannot find the file specified
```

HebaruSan:
What is the command line configuration under Settings &rarr; KSP command-line?

CorSPACE:
KSP_x64.exe -single-instance

HebaruSan:
Thanks. If you click File &rarr; Open KSP Directory, does it open a folder, and if so, is there a KSP_x64.exe file there?

CorSPACE:
No, This is a faked instance

HebaruSan:
Instances that don't have KSP can't run KSP. That should not be a surprise.

PR #2665, opened 2019-01-22
Implementation of clone and fake in GUI

DasSkelett:
My attempt to **implement the clone and fake functions in the GUI**.

Accessible via File > Select KSP Install >New KSP instance > Clone or fake new instance:
grafik

The window itself looks like this:
screenshot_20190122_214241 or on windows:
grafik
Won't win beauty prizes with that, but it works (hopefully).

Should be pretty self-explanatory (haha).
Has all the features of the CLI implementation:
* Clone:
  * Select instance by name or path
  * Enter path manually or open the folder browser dialog
* Fake
  * Select KSP version out of a drop down combo box (sorted from newest to oldest)
  * Enter DLC version manually or leave empty to create none
* Set new instance as default and/or switch to it right after

PR #3839, opened 2023-05-17
Hide fake instance creation in GUI

HebaruSan:
## Problem

Users somewhat frequently ask why they get an error when they try to launch the game for a fake instance (see #2665):

- #2850
- #3177
- #3658
- image
- image
- image
- image

## Cause

These are created via the "Fake new KSP instance" option:

image

I think we added this on the general principle that features should be available in CmdLine, ConsoleUI, and GUI when possible. But creating a non-functional game folder is not something that normal users need to do, and its availability causes confusion. Some seem to assume they can make it install an arbitrary game version, without considering why the option they're clicking on says "fake".

## Changes

Now instance faking is removed from GUI. The dropdown in Manage Game Instances now says "Clone instance" (translation updates attempted) and is disabled if no instance is selected in the grid:

image

Once you select an instance, it's enabled:

image

If you click it, the dialog appears, with all of the UI elements associated with "fake" removed:

image

The instance in the dropdown will match the row you selected in Manage Game Instances but can still be changed.

This should more firmly guide users towards cloning (which is useful to them) and away from faking (which is not). It also conveniently sidesteps the need to create faking functionality for KSP2 in GUI (which doesn't have KSP1's DLCs, meaning a more complex UI would be needed). The `ckan instance fake` command is still available for expert users and CI/CD, and has not been changed.


techman83 (approved):
Looks nice and clean, this has been a common source of confusion, so I'm super glad to see it cleaned up.

PR #3840, opened 2023-05-18
Show download errors for upgrades

HebaruSan:
## Problem

If an upgrade's download fails, the failed downloads popup from #3635 appears, but empty:

image

This dead-ends any investigation of such problems.

[Originally reported by forum user jost.](https://forum.kerbalspaceprogram.com/index.php?/topic/197082-ckan-the-comprehensive-kerbal-archive-network-v1320-kepler-ksp-2-support/&do=findComment&comment=4283056)

## Cause

The rows of that table are built based on the mods in the changeset, but only the mods to install were included. The mods to upgrade were not, so if one of them had an error, it wouldn't be shown.

## Changes

Now the mods we're upgrading are also passed to the form, so the error appears as expected:

image

While investigating this, I found that the popup could cause CKAN to freeze before it opened, and that passing itself to `Util.Invoke` instead of a reference to the `Main` form fixed it, so that's included here as well.

### DIFF
--- a/GUI/Main/MainInstall.cs
+++ b/GUI/Main/MainInstall.cs
@@ -194,7 +194,9 @@ namespace CKAN.GUI
                     {
                         // Get full changeset (toInstall only includes user's selections, not dependencies)
                         var crit = CurrentInstance.VersionCriteria();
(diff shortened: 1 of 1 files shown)


techman83 (approved):
Good find! Clean fix.

PR #3841, opened 2023-05-18
Set window title during startup

HebaruSan:
## Problem

After #3829, the GUI window title is blank at startup:

image

It's set later, once the instance is loaded.

## Cause

The window title depends on the name, version, and path of the game instance, so it's currently not set while the instance is loading.

## Changes

Now we set an initial window title in `GUI.Main.OnLoad` containing the pieces that are not dependent on having the instance loaded:

image

The full title is still set later when the instance is loaded.


techman83 (approved):
Nice, a good and simple piece of polish. I like it!

Issue #3826, opened 2023-04-17
[Bug]: CKAN Uninstalls Mods for KSP2 That Are Not Selected For Uninstall


Scarecrow-71:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Operating System

Windows 11

### CKAN Version

1.32.0

### KSP Version

0.1.2.0

### Did you make manual changes?

No

### Describe the bug

I have multiple mods installed for KSP2, but I wanted to add 2 new ones (K2-D2 and OAB Extensions), as well as checking the Auto-Update box for those mods I have installed.  Upon selecting Apply Changes, the 2 new mods were installed...and all existing mods were uninstalled.  So I went back through and reselected the mods I needed to have installed (which included both BepInEx and SpaceWarp)...and those were installed while K2-D2 and OAB were both uninstalled.  I had to uninstall everything and then select everything from scratch so that everything would be installed.

I'm not sure if Git supports youtube video for documentation, but I can replicate this and I will be more than happy to provide video to show what is happening if needed.  Find me on Discord or the Forums (Scarecrow71).

### Steps to reproduce

- Open CKAN as Administrator
- Filter for mods that are installed
- Check the box to Auto-Update any installed mods

ALSO

- Open CKAN as Administrator
- Filter for mods that are not installed
- Select one or more mods and apply changes

Scarecrow-71:
Yep, that's what I meant.  Sorry; sometimes the words I use don't always match what people see or understand.

I'm uploading a video to YT right now showing what happens.  Although processing up to HD, link is:  https://youtu.be/FdTDU9rrBNM

HebaruSan:
Thanks, and don't worry about uploading a video, there's no need for something that's working as intended.
Have a great day!

Issue #3146, opened 2020-08-26
[Bug] Uninstalling many mods as auto-installed


Seralyn:
Background
----------
* Operating System: Windows 10 x64 (legal, fwiw)
* CKAN Version: 1.28.0
* KSP Version: 1.10.1.2939

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
Yes, but not for over a month and my issue started out of the blue just now.

Problem
-------
**Describe the bug**
When I click to install a new mod and move to the Changeset tab, it CKAN tries to remove all of my installed mods. I am not new to CKAN, so this is highly unexpected behavior.

**Steps to reproduce**
Have CKAN installed.
Have some mods installed.
Try to install a new mod.

**Expected behavior**
Only the newly selected mod will try to install itself (and potential dependencies).

**Screenshots (if applicable)**
![screenshot](https://i.imgur.com/lU0dcHv.png)

(in this screenshot, I have all of my currently installed mods already clicked, as they are when CKAN loads. I click the checkmark on one mod that I want to install and click changeset, and this is what I see.

**CKAN error code (if applicable):**

```
PLACE CODE HERE !
```

HebaruSan:
Are all of those mods actually marked as auto-installed in the mod list?

HebaruSan:
Here's a guess... possibly a Duplicate of #3007? Did you install these via a .ckan metapackage in a CKAN version prior to v1.27.2?

Seralyn:
After reading these comments, I unclicked everything that had the "auto-installed" checkbox ticked and it seems to have fixed my issue! To answer your original question, though @HebaruSan , if it is true that this is part of some sort of metapackage, I can only say that it was not with my knowledge. I installed the game through Steam, and played on that file for quite some time, and only recently made a clone of my game so that my partner could play a stock version through Steam. Perhaps the process of cloning is rlated to metapackages? I clearly have no idea about that, but anyway, unclicked auto-installed, whatever that does, seems to have fixed my issue. Thank you!!

HebaruSan:
> if it is true that this is part of some sort of metapackage, I can only say that it was not with my knowledge

I guess I should have said "modpack" there. Did you install these mods by using the "Install from ckan..." option to install someone's mod list? Maybe that was how you cloned it?

![image](https://user-images.githubusercontent.com/1559108/91886011-36551e00-ec78-11ea-9c9b-e156ed3942ca.png)

Seralyn:
[...]

PR #3842, opened 2023-05-20
Tooltip for auto-installed checkboxes

HebaruSan:
## Motivation

We somewhat frequently get questions about the Auto-installed column from #2753:

- #3146
- #3826
- image
- image
- image
- image
- image
- image
- image
- image

Apparently the purpose of this checkbox is not clear. But on the bright side, everyone seems to "get it" when provided a brief explanation.

## Changes

Now these checkboxes have a tooltip that explains what they're for:

image


techman83 (approved):
Awesome, thanks @HebaruSan

PR #3843, opened 2023-05-20
Clarify remote swinfo network errors

HebaruSan:
## Background

#3817 added retrieval of remote `swinfo.json` files for KSP2 mods, which are very analogous to remote version files.

## Problem

If the `version_check` property contains a URL that's a broken link, the error message isn't clear. It's just a generic 404 error that could be coming from anywhere:

image

## Changes

Now that exception is caught and replaced with one indicating the context in which the error occurred:

```
$ netkan.exe --game KSP2 NetKAN/ManeuverNodeController.netkan
2233 [1] FATAL CKAN.NetKAN.Program (null) - Error fetching remote swinfo https://raw.githubusercontent.com/schlosrat/ManeuverNodeController/master/maneuver_node_controller/swinfo.json: The remote server returned an error: (404) Not Found.
```

I might self-review this since it only affects the bot and would be nice to have soon-ish.

### DIFF
--- a/Netkan/Transformers/SpaceWarpInfoTransformer.cs
+++ b/Netkan/Transformers/SpaceWarpInfoTransformer.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 
 using ICSharpCode.SharpZipLib.Zip;
 using log4net;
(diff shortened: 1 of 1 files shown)


techman83 (commented):
Makes sense to me!

Issue #3792, opened 2023-02-25
[Bug] Errors when trying to build on macOS Catalina


rob5408:
Background
----------
* Operating System: macOS Catalina
* CKAN Version: master branch, [#04034c00](https://github.com/KSP-CKAN/CKAN/commit/04034c000817d8159c02be96229797c6f2373144)
* KSP Version: NA

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** No

Problem
-------
**Describe the bug**
I was hoping to help solve [this issue](https://github.com/KSP-CKAN/CKAN/issues/3768) that I am also encountering. When trying to build on my Mac I am getting build errors. I read [building.md](https://github.com/KSP-CKAN/CKAN/blob/master/doc/building.md) and tried the build script `% ./build Ckan` or `% ./build osx`. 

```
~/CKAN/GUI/Controls/ModInfoTabs/Relationships.cs(17,51): error CS0234: The type or namespace name 'Resources' does not exist in the namespace 'CKAN.GUI.Properties' (are you missing an assembly reference?) [~/CKAN/GUI/CKAN-GUI.csproj]
~/CKAN/GUI/Controls/ModInfoTabs/Relationships.cs(21,51): error CS0234: The type or namespace name 'Resources' does not exist in the namespace 'CKAN.GUI.Properties' (are you missing an assembly reference?) [~/CKAN/GUI/CKAN-GUI.csproj]
~/CKAN/GUI/Controls/ModInfoTabs/Relationships.cs(25,51): error CS0234: The type or namespace name 'Resources' does not exist in the namespace 'CKAN.GUI.Properties' (are you missing an assembly reference?) [~/CKAN/GUI/CKAN-GUI.csproj]
~/CKAN/GUI/Controls/ModInfoTabs/Relationships.cs(29,51): error CS0234: The type or namespace name 'Resources' does not exist in the namespace 'CKAN.GUI.Properties' (are you missing an assembly reference?) [~/CKAN/GUI/CKAN-GUI.csproj]
~/CKAN/GUI/Controls/ModInfoTabs/Relationships.cs(33,51): error CS0234: The type or namespace name 'Resources' does not exist in the namespace 'CKAN.GUI.Properties' (are you missing an assembly reference?) [~/CKAN/GUI/CKAN-GUI.csproj]
[...]

PR #3847, opened 2023-06-06
fix: build with dotnet core on macOS

doinkythederp:
This pull request changes the build script to use Dotnet Core on macOS. Dotnet Framework (the default) is not available on Mac which causes build failures with a rather unintelligible error message.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,9 +53,9 @@ jobs:
           key: nuget-packref-modules-${{ hashFiles('**/packages.config') }}-${{ hashFiles('**/*.csproj') }}
 
       - name: Build ckan.exe and netkan.exe
-        run: ./build --configuration=${{ matrix.configuration}}
+        run: ./build ${{ matrix.configuration}}
       - name: Run tests
-        run: xvfb-run ./build test+only --configuration=${{ matrix.configuration }} --where="Category!=FlakyNetwork"
+        run: xvfb-run ./build ${{ matrix.configuration }} test+only --where="Category!=FlakyNetwork"
       - name: Run inflator container smoke test
         run: |
           cd _build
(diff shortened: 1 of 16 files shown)


doinkythederp:
Here's a link to the process of me solving the issue on the CKAN Discord server: https://discord.com/channels/601452466017665040/601456752906469388/1115099187416150036

doinkythederp:
Alternatively if this change is too disruptive I could update the documentation for building to make it clearer that you need to use the Debug_NetCore configuration on macOS

doinkythederp:
I'm looking forward to getting feedback on this issue and my proposed fix. Please let me know if you need more context or anything.

doinkythederp:
These changes seems to make it so that it can't build the entire EXE on macos. Anyone know how this might be fixed? I know it works on Linux (the dockerfile) so I'm thinking it would work on macOS too.

HebaruSan:
Hi @doinkythederp, thanks for wanting to contribute. Sorry to be the bearer of bad news, but as you've begun to notice, these changes don't fix anything.

- It's not true that the .NET Framework is unavailable on Mac. It's provided by Mono and its dev tools (but see below, the build will still fail!). The wiki may only link to an end-user runtime install, not sure, so exercise extra care and make sure you get the dev tools:
  https://github.com/KSP-CKAN/CKAN/wiki/Installing-CKAN-on-macOS#installing-manually
- You're not seeing the EXE because the `Debug_NetCore` configuration doesn't build it. It just builds the Core DLL, without the CmdLine, ConsoleUI, or GUI (or Netkan). This configuration was contributed by a user who wanted to use it to create a new GUI for MacOS but never finished it, see #2820 and #2799. So making it the default won't be useful. It also won't be a good idea to call it out in documentation, since just building the Core DLL isn't likely to be what anyone wants.
- As far as I know, even with Mono installed, MacOS lost the ability to build `ckan.exe` when Apple sunsetted support for 32-bit binaries, since that's the only way Mono provides the WinForms library (see mono/mono#6701), which is a dependency of GUI, which is a dependency of the EXE. (This is why it appeared to you at first that using the `Debug_NetCore` configuration was helping; you got past the compile errors because GUI wasn't being built. But that's not actually helpful.) See #3792 for the last time someone tried building on Mac.
- The ability to run `./build test` without specifying the configuration is a long established capability, and removing it would impose an inconvenience on the dev team.

I'm sorry that you clearly spent some time on this, but we won't be able to merge it.

doinkythederp:
Hey, thanks for getting back to me! It's unfortunate that macOS doesn't have the features needed to build CKAN. I'm still interested in getting a developer environment set up, though. While building on the system doesn't seem to be an option, interestingly, the Docker container (`docker build .`) can still build it perfectly. Knowing this, it's likely possible to set up a DevContainer (https://containers.dev) with a similar environment to facilitate the possibility of developing and building on a Mac. I'll try and take a look at this when I have the chance, but let me know if you think this seems feasible.

PR #3854, opened 2023-07-02
Handle missing KSP2 exe

HebaruSan:
## Problem

A user uninstalled KSP2 and reported this on Discord:

image

## Cause

`FileVersionInfo.GetVersionInfo` throws an exception when the file doesn't exist.

## Changes

Now we check `File.Exists` so the exception won't be thrown.


techman83 (approved):
That'll do it! Thanks @HebaruSan

Issue #3699, opened 2022-11-10
[Bug] Recent modules missing for some users


cumber:
Background
----------
* Operating System: Windows
* CKAN Version: 1.31.2, and probably previous versions based on how long ago the latest successful refresh was
* KSP Version: 1.12.4

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
No

Problem
-------
**Describe the bug**

First time playing KSP in some months, I opened CKAN to check whether there were any mod updates. None were found. However, on starting up KSP, the Addon Version Checker reports that updates are available for several mods.

I dug into one case further, the ClickthroughBlocker mod:

CKAN reports the latest available version is 0.1.10.17 (which the versions tab lists as `1:0.1.10.17` - I don't know what the `1:` prefix means or why it's shown differently in the actual metadata file and the versions tab than it is in the main GUI). This seems to correspond to this file in the CKAN-meta repository: https://github.com/KSP-CKAN/CKAN-meta/blob/master/ClickThroughBlocker/ClickThroughBlocker-1-0.1.10.17.ckan

Addon Version Checker tells me that ClickThroughBlocker has version 2.1.10.18 available.

I can also see that in the CKAN-meta repository there is this file: https://github.com/KSP-CKAN/CKAN-meta/blob/master/ClickThroughBlocker/ClickThroughBlocker-1-2.1.10.18.ckan, which contains `"version": "1:2.1.10.18"`. This does seem to be the version that Addon Version Checker is telling me about.

I don't know why CKAN isn't seeing this updated version (or of any of the other mods). It's like it's simply not updating the repository, but when I click refresh it claims that it is, I don't see any indication of any failure to update.

**Expected behavior**

I expect the CKAN client to show me mod versions available that are in the CKAN-meta repository, after refreshing.

HebaruSan:
Click "Refresh" in the toolbar to update your registry with the latest metadata.

HebaruSan:
Alternately, set all of 1.12 compatible:

https://github.com/KSP-CKAN/CKAN/wiki/User-guide#Choosing-compatible-game-versions

CKAN will detect an update as incompatible if you're on 1.12.4 and the update was released for 1.12.3.

cumber:
[...]

PR #3860, opened 2023-07-20
Include repo etags in transactions

HebaruSan:
## Problem

In the course of investigating why available modules were missing in #3699, users shared their registries, which revealed that the server etags (see #2682) were getting out of sync with the available modules; users were ending up with old module lists but up-to-date etags. This meant that if they tried to Refresh immediately, it would say the list was already up to date, even though it wasn't.

## Cause

I found this very confusing for quite a long while. That data is all part of the `Registry` object; shouldn't it all be saved to `registry.json` together or not at all? No! There is one specific scenario in which they can get out of sync:

1. Metadata retrieved from network successfully
2. Etags updated
3. Module list updated
4. **Exception thrown while still inside transaction** (any exception would cause this, but in practice it was `ReinstallModuleKraken` from #3344)
5. Transaction handling reverts to the most recently saved copy of the registry, which happened at the start of step 3, _after_ the etags were updated but _before_ the modules were updated

## Changes

In #3828 we eliminated this exception, so this is already functionally fixed, but architecturally it's still important to make the data objects behave properly.

Now the etags are saved by a new `Registry.SetETags` function, which calls `EnlistWithTransaction` first, which will save off a copy of the registry with the **original** etags. This will ensure that if the registry is reverted in a transaction, the etags will accurately reflect the state of the available modules list.

### DIFF
--- a/Core/Net/Repo.cs
+++ b/Core/Net/Repo.cs
@@ -72,16 +72,7 @@ namespace CKAN
                 .ToArray();
 
             // Loading done, commit etags
-            foreach (var kvp in savedEtags)
-            {
-                var repo = repos.Where(r => r.uri == kvp.Key).FirstOrDefault();
-                var etag = kvp.Value;
-                if (repo != null)
-                {
-                    log.DebugFormat("Setting etag for {0}: {1}", repo.name, etag);
-                    repo.last_server_etag = etag;
-                }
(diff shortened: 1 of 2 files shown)


techman83 (approved):
That's quite an edge case, good find! LGTM

PR #3864, opened 2023-07-22
Updated Chinese translation to reduce misunderstandings

Fierce-Cat:
The existing Chinese translations of the "back" and "cancel changes" buttons are very similar in Chinese, and these changes can help users distinguish the difference between the two buttons.


HebaruSan (approved):
Thanks for the update!

PR #3863, opened 2023-07-21
Fix exception at startup w/o default game inst

HebaruSan:
## Problems

If you have multiple game instances and no default game instance, the Manage Game Instances window appears at startup. In v1.33.0, this exception is thrown after you select an instance:
 
```
System.ArgumentException: Empty path name is not legal.
   at System.IO.StreamWriter..ctor(String path, Boolean append, Encoding encoding, Int32 bufferSize, Boolean checkHost)
   at System.IO.StreamWriter..ctor(String path)
   at CKAN.GUI.GUIConfiguration.SaveConfiguration(GUIConfiguration configuration)
   at CKAN.GUI.Main.CurrentInstanceUpdated(Boolean allowRepoUpdate)
   at System.ComponentModel.BackgroundWorker.OnRunWorkerCompleted(RunWorkerCompletedEventArgs e)
```

There's also a "Check for updates" prompt that will happen every time.

(Everything works fine if "Set as default" is checked for some instance.)

Initially reported by a Discord who prefers to remain anonymous, then [Poppa Wheelie on the forum](https://forum.kerbalspaceprogram.com/topic/197082-ckan-the-comprehensive-kerbal-archive-network-v1320-kepler-ksp-2-support/?do=findComment&comment=4304504).

## Causes

When you have no default instance, we use `GUIConfiguration`'s constructor to get the window geometry defaults from an instance of the class that doesn't correspond to a real file on disk, meaning its `path` field is null.

- `Main.CurrentInstanceUpdated` always saves the current `GUIConfiguration` object before loading the next one, but this causes an attempt to write to a null file path with the special instance described above
- The "Check for updates" prompt also uses the special `GUIConfiguration` instance, which can't save

## Changes

- Now `GUIConfiguration.Save` will abort if `GUIConfiguration.path` is null or empty, which will prevent the exception from occurring
- Now the "Check for updates" prompt is moved to `Main.CurrentInstanceUpdated`, where it will have a valid `GUIConfiguration` object and therefore always be able to save your choice

Fixes #3862.

We should do a hotfix release after this merged. I'm thinking v1.33.2, "Laplace" Δ.

PR #3878, opened 2023-08-10
Restore window position without default instance

### LINKED ISSUE #3870: [Bug]: [1.33.2] CKAN does not remember the last window position

### Operating System

Windows 10

### CKAN Version

1.33.2

### Game Version

1.12.5

_No response_

### Describe the bug

every time ckan is started, it opens with a fixed window. you have to change the size every time or switch to full screen.
the previous window position is not remembered.

that's a bit annoying

### Steps to reproduce

start CKAN

### Relevant log output

_No response_

HebaruSan:
## Problem

If you don't have a default game instance, the GUI window will always appear in its default location as if you have never run CKAN before. This can be annoying if you have a strong preference for having appear in a particular spot on the screen or at a particular size.

## Cause

The window geometry is saved per-instance, and as of #3829 and #3863, we display the GUI window as early as possible, and if there's no default instance, then the window appears before you choose an instance. There was no clearly best way to load the geometry without an instance.

## Changes

Now if you have no default instance, we scan your (valid) instances to find the one with the most recently written `CKAN/GUIConfig.xml` file, and then we load that configuration. This way the window will appear wherever it was when you most recently closed CKAN, even with no default instance. If you have no instances at all (or if all of your instances are corrupted), we fall back to the baseline defaults.

Fixes #3870.

### DIFF
--- a/GUI/Main/Main.cs
+++ b/GUI/Main/Main.cs
@@ -152,12 +152,9 @@ namespace CKAN.GUI
             }
 
             // We need a config object to get the window geometry, but we don't need the registry lock yet
-            configuration = CurrentInstance != null
-                ? GUIConfiguration.LoadOrCreateConfiguration(
-                    Path.Combine(CurrentInstance.CkanDir(), "GUIConfig.xml"),
-                    CurrentInstance.game)
-                // No instance, just get the defaults
-                : new GUIConfiguration();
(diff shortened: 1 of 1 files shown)


techman83 (approved):
Awesome, I like that logic! Thanks @HebaruSan

PR #2111, opened 2017-09-10
Fix crash in "ckan available" with curly braces in mod name

HebaruSan:
`mono ckan available` currently errors out halfway through the list:

```
...
* KSPDev-LogConsole (1:0.15) - KSPDev LogConsole

Unhandled Exception:
System.FormatException: Input string was not in a correct format.
  at System.Text.StringBuilder.AppendFormatHelper (System.IFormatProvider provider, System.String format, System.ParamsArray args) [0x000e9] in <8f2c484307284b51944a1a13a14c0266>:0 
  at System.String.FormatHelper (System.IFormatProvider provider, System.String format, System.ParamsArray args) [0x00026] in <8f2c484307284b51944a1a13a14c0266>:0 
  at System.String.Format (System.IFormatProvider provider, System.String format, System.Object[] args) [0x00029] in <8f2c484307284b51944a1a13a14c0266>:0 
  at System.IO.TextWriter.WriteLine (System.String format, System.Object[] arg) [0x00007] in <8f2c484307284b51944a1a13a14c0266>:0 
  at System.IO.TextWriter+SyncTextWriter.WriteLine (System.String format, System.Object[] arg) [0x00000] in <8f2c484307284b51944a1a13a14c0266>:0 
  at (wrapper synchronized) System.IO.TextWriter+SyncTextWriter:WriteLine (string,object[])
  at System.Console.WriteLine (System.String format, System.Object[] arg) [0x00016] in <8f2c484307284b51944a1a13a14c0266>:0 
  at CKAN.CmdLine.ConsoleUser.DisplayMessage (System.String message, System.Object[] args) [0x00001] in <b3a003dfcca84e4386996f3968e550bd>:0 
  at CKAN.NullUser.RaiseMessage (System.String message, System.Object[] url) [0x00001] in <b3a003dfcca84e4386996f3968e550bd>:0 
  at CKAN.CmdLine.Available.RunCommand (CKAN.KSP ksp, System.Object raw_options) [0x0008b] in <b3a003dfcca84e4386996f3968e550bd>:0 
  at CKAN.CmdLine.MainClass.RunAction (CKAN.CmdLine.Options cmdline, CKAN.CmdLine.CommonOptions options, System.String[] args, CKAN.IUser user, CKAN.KSPManager manager) [0x001b3] in <b3a003dfcca84e4386996f3968e550bd>:0 
[...]

PR #3880, opened 2023-08-14
Correctly print cmdline errors with braces

HebaruSan:
## Problem

While I was testing some edge case `install` stanzas for @JonnyOThan, this metadata:

```json
    "install": {
        "file": "Astrogator",
        "install_to": "GameRoot/KSP_x64_Data"
    },
```

gave me this error:

```
$ ckan.exe install -c Astrogator-v1.0.0.ckan

Unhandled Exception: System.FormatException: Input string was not in a correct format.
   at System.Text.StringBuilder.FormatError()
   at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.Format(IFormatProvider provider, String format, Object[] args)
   at System.IO.TextWriter.WriteLine(String format, Object[] arg)
   at System.IO.TextWriter.SyncTextWriter.WriteLine(String format, Object[] arg)
   at CKAN.CmdLine.ConsoleUser.RaiseError(String message, Object[] args)
   at CKAN.CmdLine.Install.RunCommand(GameInstance instance, Object raw_options)
   at CKAN.CmdLine.MainClass.RunSimpleAction(Options cmdline, CommonOptions options, String[] args, IUser user, GameInstanceManager manager)
   at CKAN.CmdLine.MainClass.Execute(GameInstanceManager manager, CommonOptions opts, String[] args)
```

## Cause

The root cause is that my .ckan file was formatted wrong; the `install` property has to be an array, but mine was an object, so the attempt to deserialize failed.

However, the error message about that was not printed because there are places in CmdLine where a message string from an exception thrown by core code or third party libraries is passed as the first parameter of `IUser.RaiseError`, which always interprets that parameter as a format string (see #2111). If such a string contains `{` and `}` (as is particularly likely for JSON deserialization problems), these will be interpreted as parameter interpolation tokens and the string formatting will fail.

## Changes

- Now when we want to print an exception message, the first parameter is always `{0}`, to prevent the message from being interpreted as a format string. The original root cause error is printed:
  ```
  $ install -c Astrogator-v1.0.0.ckan
  JSON deserialization error: Cannot deserialize the current JSON object (e.g. {"name":"value"}) into type 'CKAN.ModuleInstallDescriptor[]' because the type requires a JSON array (e.g. [1,2,3]) to deserialize correctly.
  To fix this error either change the JSON to a JSON array (e.g. [1,2,3]) or change the deserialized type so that it is a normal .NET type (e.g. not a primitive type like integer, not a collection type like an array or List<T>) that can be deserialized from a JSON object. JsonObjectAttribute can also be added to the type to force it to deserialize from a JSON object.
  Path 'install', line 67, position 17.: Cannot deserialize the current JSON object (e.g. {"name":"value"}) into type 'CKAN.ModuleInstallDescriptor[]' because the type requires a JSON array (e.g. [1,2,3]) to deserialize correctly.
  To fix this error either change the JSON to a JSON array (e.g. [1,2,3]) or change the deserialized type so that it is a normal .NET type (e.g. not a primitive type like integer, not a collection type like an array or List<T>) that can be deserialized from a JSON object. JsonObjectAttribute can also be added to the type to force it to deserialize from a JSON object.
  Path 'install', line 67, position 17.
  Usage: ckan install [--with-suggests] [--with-all-suggests] [--no-recommends] [--headless] Mod [Mod2, ...]
  ```
- An untranslated error message is internationalized.

### DIFF
--- a/Cmdline/Action/AuthToken.cs
+++ b/Cmdline/Action/AuthToken.cs
@@ -110,7 +110,7 @@ namespace CKAN.CmdLine
             }
             else
             {
-                user.RaiseError("Invalid host name: {0}", opts.host);
+                user.RaiseError(Properties.Resources.AuthTokenInvalidHostName, opts.host);
             }
             return Exit.OK;
         }
--- a/Cmdline/Action/GameInstance.cs
+++ b/Cmdline/Action/GameInstance.cs
@@ -637,13 +637,13 @@ namespace CKAN.CmdLine
             catch (BadInstallLocationKraken kraken)
             {
                 // The folder exists and is not empty.
-                User.RaiseError(kraken.Message);
+                User.RaiseError("{0}", kraken.Message);
                 return badArgument();
             }
             catch (WrongGameVersionKraken kraken)
             {
                 // Thrown because the specified game instance is too old for one of the selected DLCs.
-                User.RaiseError(kraken.Message);
+                User.RaiseError("{0}", kraken.Message);
                 return badArgument();
             }
             catch (NotKSPDirKraken kraken)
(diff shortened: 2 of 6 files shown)


techman83 (approved):
That'll do it! Thanks @HebaruSan

PR #3883, opened 2023-08-17
Mod list fixes and improvements

HebaruSan:
## Problems

- If you have no mods installed, the Auto-installed column is still displayed even though it serves no purpose:
  image
- A Discord user confused himself by scrolling the mod list to the right and forgetting that the checkbox to install a mod had scrolled off the edge of the grid:
  image

## Causes

- This column's visibility was managed in multiple places, and they conflicted with one another; an older one was hiding the Auto-installed column while a newer one was unhiding it
- The checkbox columns are small but essential UI elements, so allowing them to scroll out of view has more disadvantages than advantages

## Motivation

We have a very old `TODO` item here:

https://github.com/KSP-CKAN/CKAN/blob/1a884fe056bbca120167d89cfaba61f833113537/GUI/Model/ModList.cs#L172

This is _not_ a very good place for it (it would be architecturally better for this class to be a pure data model, and this function could be called with a changeset that isn't meant to be applied to a grid), but the idea makes sense because showing users which mods will be removed could be helpful.

## Changes

- Now column visibility management is consolidated and cleaned up such that the Auto-installed column is hidden if no installed mods are visible:
  image
- Now the mod list's checkbox columns are frozen at the left, so they'll always be visible even if you scroll to the right:
  image
- Now when your changeset includes removing some mods (including user-selected, removed due to missing dependencies, and removed due to no longer being needed), they will be shown with a ~~strikeout~~ font on the main mod list:
  image
- Another `TODO` regarding `ModList.HandleTooManyProvides` is resolved by deleting that delegate definition, which is no longer in use

### DIFF
--- a/GUI/Controls/ManageMods.Designer.cs
+++ b/GUI/Controls/ManageMods.Designer.cs
@@ -346,6 +346,7 @@ namespace CKAN.GUI
             // Installed
             //
             this.Installed.Name = "Installed";
+            this.Installed.Frozen = true;
             this.Installed.SortMode = System.Windows.Forms.DataGridViewColumnSortMode.Programmatic;
             this.Installed.DefaultCellStyle.Alignment = System.Windows.Forms.DataGridViewContentAlignment.MiddleCenter;
             this.Installed.Width = 50;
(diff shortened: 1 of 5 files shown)


techman83 (approved):
Awesome, thanks @HebaruSan

PR #3884, opened 2023-08-17
Deduce primary branch name in merge script

HebaruSan:
## Motivation

The commit script currently assumes a `master` branch, but the latest recommendation is to prefer `main`, as with the KSP2 repos.

## Changes

- Now the name of the primary branch is retrieved from the repo's refs, similarly to KSP-CKAN/NetKAN-Infra#288, which removes one obstacle to renaming CKAN's `master`
- I accumulated various trivial C# style changes that didn't belong in other pull requests, which are now dumped here

### DIFF
--- a/Core/Net/NetFileCache.cs
+++ b/Core/Net/NetFileCache.cs
@@ -153,17 +153,13 @@ namespace CKAN
         }
 
         // returns true if a url is already in the cache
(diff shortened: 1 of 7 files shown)


techman83 (approved):
Thanks @HebaruSan !

PR #3885, opened 2023-08-20
Multi-game labels

### LINKED ISSUE #3824: KSP 1 labels are applied to KSP 2 mods
If a label is applied to a mod for KSP1, CKAN auto-applies that same label to a similarly named mod for KSP2.

Specifically, I've applied the label "installed" to the KSP1 mod "KSP Resonant Orbit Calculator" by Linuxgurugamer.  When I added KSP2 as a game instance to CKAN, I found the label was already applied to the KSP2 mod "Resonant Orbit Calculator" by Schlosrat.

The name of the mods are similar but not identical (one is a substring of the other's) but even if they were identical, CKAN should recognize they are not the same mod.  These are different mods for different games by different authors.  

This is low priority right now, I would think, as the number of KSP2 mods is currently quite limited, but as the number grows, this could get annoying.  Hope it isn't a great deal of work to fix.  Thanks, and you guys are awesome!

HebaruSan:
## Problem

If you add a label to `ResonantOrbitCalculator` in KSP1, the different mod with the same identifier will also have that label added in KSP2, and vice versa. This is not ideal because these are two different mods, which you should be able to treat separately.

## Cause

The `labels.json` file just has one list of identifiers per label, with no distinctions between games, so when both games have a mod with the same identifier, they're treated as one mod.

## Changes

- Now `Meta.GetProductName()` provides a centralized way to get the string "CKAN", `CKANPathUtils.AppDataPath` uses it to provide a centralized way to get the path of the shared config folder, and code that uses either value is updated to use these properties (I did this initially because I was considering changing the path of the `labels.json` file, which I decided against, but this clean-up is still worth keeping)
- The property in the labels JSON file that formerly held a list of identifiers now holds an object with keys corresponding to game names and values corresponding to lists of identifiers:
  ```json
  "module_identifiers": {
      "KSP": [ "Mod1" ],
      "KSP2": [ "Mod2" ]
  }
  ```
  - For users who haven't added labels to mods before, these will start out empty and accumulate mods independently, without affecting the other game. For users who have already added some labels to mods, the existing list of mods will be copied to both games, so nothing in the UI will change immediately, but from that point forward the two lists will be maintained independently, so a mod can be removed from a label in one game without affecting the other game.
  - Since the old ckan.exe client is expecting this property to only be an array, it is renamed to `module_identifiers_by_game`. This means that you will lose all your label memberships if you temporarily switch back to an old client version, but at least it won't crash.

Fixes #3824.

## Considered and not done

Note that any mods that you have added to a label that are not actually available in that game (e.g., if you add `MechJeb2` to your Favorites for KSP1, then run these changes for the first time, then load a KSP2 instance) will be permanently stuck in those lists in the `labels.json` file behind the scenes (unless you edit it manually), since those mods can never appear in the mod list to be removed via the right click menu, and the mod counts in the label filter menu will be inflated by that amount.

I considered various options for trying to resolve this automatically, but none of them are safe. The problem is that there's no reliable way to determine the list of valid identifiers for a game, because users can add custom repos. For example, if someone adds the MechJeb2 dev build repo in one game instance, then the `MechJeb2-dev` module becomes available, and they could add it to a label. If they then later load a different game instance without that repo, it would be wrong to auto-remove that identifier just because we don't recognize it.

### DIFF
--- a/ConsoleUI/ExitScreen.cs
+++ b/ConsoleUI/ExitScreen.cs
@@ -48,7 +48,7 @@ namespace CKAN.ConsoleUI {
             Console.Clear();
 
             // Specially formatted snippets
-            var ckanPiece = new FancyLinePiece("CKAN", theme.ExitInnerBg, theme.ExitHighlightFg);
+            var ckanPiece = new FancyLinePiece(Meta.GetProductName(), theme.ExitInnerBg, theme.ExitHighlightFg);
             var ckanVersionPiece = new FancyLinePiece($"CKAN {Meta.GetVersion()}", theme.ExitInnerBg, theme.ExitHighlightFg);
             var releaseLinkPiece = new FancyLinePiece("https://github.com/KSP-CKAN/CKAN/releases/latest", theme.ExitInnerBg, theme.ExitLinkFg);
             var issuesLinkPiece = new FancyLinePiece("https://github.com/KSP-CKAN/CKAN/issues", theme.ExitInnerBg, theme.ExitLinkFg);
--- a/ConsoleUI/ExitScreen.cs
+++ b/ConsoleUI/ExitScreen.cs
@@ -43,7 +43,7 @@ namespace CKAN.ConsoleUI {
             }
             else
             {
-                Console.ResetColor();                
+                Console.ResetColor();
             }
             Console.Clear();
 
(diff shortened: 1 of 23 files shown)


HebaruSan:
... ahh, right. Tests don't like it when we reference `Main.Instance`...

techman83 (approved):
Awesome! Yes, I think that's a sensible approach and hopefully relatively edge case. Thanks @HebaruSan !

PR #3887, opened 2023-08-21
Fix schema issue with install_to: GameData/Mods

HebaruSan:
## Problem

KSP-CKAN/KSP2-NetKAN#81

```yaml
install:
  - find: BetterPartsManager
    install_to: GameData/Mods
```

https://github.com/KSP-CKAN/KSP2-NetKAN/actions/runs/5930964431/job/16081829297

image

## Cause

https://cswr.github.io/JsonSchema/spec/generic_keywords/

image

`GameData/Mods` is _too correct_; it matches both of these:

https://github.com/KSP-CKAN/CKAN/blob/858d71fc389910fb3ce357d70fa2d4ea0de82218/CKAN.schema#L440-L444
https://github.com/KSP-CKAN/CKAN/blob/858d71fc389910fb3ce357d70fa2d4ea0de82218/CKAN.schema#L460-L463

And the tooling apparently doesn't like that. How this could ever possibly be useful is quite beyond me.

## Changes

I want to keep the explicit definition of the KSP2 default folder independent of what happens elsewhere in the list, so now the `install_to` part uses `anyOf` instead of `oneOf`, which will allow more than one match.

image

FYI to @NathanPeake.

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -432,7 +432,7 @@
         },
         "install_to" : {
             "description" : "Where file should be installed to. As of v1.2, GameData may take a path",
-            "oneOf" : [
+            "anyOf" : [
                 {
                     "description" : "Spec version 1 targets",
                     "enum"        : [ "GameData", "Ships", "GameRoot", "Tutorial", "Scenarios" ]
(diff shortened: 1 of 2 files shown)


HebaruSan:
This should have a test.

HebaruSan:
Seems to work.

https://github.com/KSP-CKAN/KSP2-NetKAN/actions/runs/5930964431/job/16084267518

PR #3889, opened 2023-08-22
Parse quoted strings for `ckan prompt`

HebaruSan:
## Problems

- https://github.com/KSP-CKAN/KSP2-CKAN-meta/actions/runs/5938916126/job/16104294940
  image
- In `ckan prompt`, if you type `install --` and tab, `--ckanfiles` and the common options like `--instance` aren't included in the completion options
- In `ckan prompt`, if you type `install -` and tab, there are no completion options

## Cause

That module's version has a space in it, which ends up in the `.ckan` filename. This can already be handled from the raw command line by using your shell's quoting abilities, but the metadata validator does its work in a `ckan prompt` session, which doesn't do quoting.

https://github.com/KSP-CKAN/xKAN-meta_testing/blob/master/ckan_meta_tester/ckan_install_template.txt

```
install -c $ckanfile --headless
```

After parsing, the code to execute the `install` command just sees the filename up to the first space.

## Changes

- Now if you put double quotes around a string in `ckan prompt`, it will not be split on spaces, which will allow us to solve the original problem in a future pull request to `xKAN-meta_testing` by adding spaces around the filename
- Now the `install --ckanfiles` option can be tab-completed
- Now the common `--` options like `--instance` are included in tab-completion rather than just the verb-specific ones
- Now single-dash options like `-c` can be tab-completed as well
- Trivial code style updates in `Meta.cs`
- `Meta.cs.in` is deleted. This was added in #1589, I think as an intermediate stage of those changes in which version info would be populated into `Meta.cs` at compile time by the build system, but subsequently made unnecessary by retrieving those changes from the assembly attributes instead. As far as I can tell from examining the commit log, this file has never actually been used.


techman83 (approved):
All makes sense, thanks @HebaruSan !

I note the scanner has a bunch of notices about unreachable code? Benign or just related specifically .net core?

HebaruSan:
> I note the scanner has a bunch of notices about unreachable code? Benign or just related specifically .net core?

Ahh, neither—those are something I should fix...

techman83:
> Ahh, neither—those are something I should fix...

Awesome, if there are no major changes involved with them I'm happy enough with my review as is. So feel free to merge when you're sorted with them.

HebaruSan:
Unreachable code sorted, now let's see if we can do something about these. looks like `actions/cache` is up to v3 by now...

image

PR #3892, opened 2023-08-25
Show recommendations of full changeset with opt-out

### LINKED ISSUE #3838: [Feature]: Recommendations/Suggestions should be transitive; provide a way for mods to opt-out
### Problem

If ModA depends on ModB and I only select to install ModA, I don't see any of ModB's recommendations or suggestions.  For example if I install a sweet IVA mod that depends on RasterPropMonitor-Core, I don't get the option to install all the other cool stuff that RPM supports.

In discord discussions, it was brought up that RO/RP-1 folks explicitly asked for this, because they have such a sprawling network of mods that the number of recommendations and suggestions would be overwhelming (and likely conflicting I don't doubt).

### Suggestion

Create a new property that would prevent a related mod from transitively including its recommendations/suggestions (i.e. the current behavior):

```yaml
depends:
  - name: ProceduralParts
    suppress_recommendations: true
 ```
Note this should probably also be available for recommendations, not just dependencies...but that would probably mean that you might have to go through the "recommend/suggest" cycle multiple times...or selecting a mod in that dialog could repopulate the dialog with new transitive suggestions/recommendations.

Once this property exists, it could be added to the RO/RP-1 mods so they retain the current behavior.  And then finally switch CKAN's default behavior to include the transitive relationships.

### Alternatives

The "audit recommendations" feature can also be used to address this issue but I think it's not very discoverable.

### Additional context

_No response_

HebaruSan:
## Motivation

Currently we show only the recommendations and suggestions for the mods the user selects. The recommendations and suggestions of dependencies aren't shown, even though the user stands to benefit from seeing them.

## Changes

- Now recommendations and suggestions are shown for the full changeset, including both user-selected mods and all of their dependencies, except...
- The KSP-RO folks don't want that for their modules, because the list would be very long and some of it might not be applicable to an RSS install, so modules can do this to opt out:
  ```yaml
  depends:
    - name: AMod
      suppress_recommendations: true
  ```
  With that in place, AMod would be pulled in as a dependency, but its recommendations and suggestions (and those of _its_ dependencies) won't be shown. This will allow the RP-1 install's listings to be kept shorter.

This is done with a new relationship resolver option, `get_recommenders`, which stops searching the relationship tree when `suppress_recommendations` is found.

Fixes #3838.

As a courtesy, before the next release, we should submit pull requests adding `suppress_recommendations` to the `depends` relationships of these metanetkans:

- <https://github.com/KSP-RO/RealSolarSystem/blob/master/RealSolarSystem.netkan>
- <https://github.com/KSP-RO/RealismOverhaul/blob/master/RealismOverhaul.netkan>
- <https://github.com/KSP-RO/RP-1/blob/master/RP-1.netkan>
- <https://github.com/KSP-RO/RP-0/blob/master/RP-0.netkan>
- <https://github.com/KSP-RO/RP-1-ExpressInstall-Graphics/blob/main/RP-1-ExpressInstall-Graphics-Low.netkan>
- <https://github.com/KSP-RO/RP-1-ExpressInstall-Graphics/blob/main/RP-1-ExpressInstall-Graphics-Medium.netkan>
- <https://github.com/KSP-RO/RP-1-ExpressInstall-Graphics/blob/main/RP-1-ExpressInstall-Graphics-High.netkan>

FYI to @NathanKell that this may be on its way.

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -409,6 +409,10 @@
                             "choice_help_text": {
                                 "description" : "Optional help text shown when user has to choose among multiple modules",
                                 "type"        : "string"
+                            },
+                            "suppress_recommendations": {
+                                "description" : "If true, don't check this mod or its dependencies for recommendations or suggestions",
+                                "type"        : "boolean"
                             }
                         },
                         "required" : [ "name" ]
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -1251,6 +1251,20 @@ namespace CKAN
             }
         }
 
+        private static readonly RelationshipResolverOptions RecommenderOptions = new RelationshipResolverOptions()
+        {
+            // Only look at depends
+            with_recommends = false,
+
+            // Don't throw anything
+            without_toomanyprovides_kraken = true,
+            without_enforce_consistency    = true,
+            proceed_with_inconsistencies   = true,
+
+            // Skip relationships with suppress_recommendations==true
+            get_recommenders = true,
+        };
+
         /// <summary>
         /// Looks for optional related modules that could be installed alongside the given modules
         /// </summary>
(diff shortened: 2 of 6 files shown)


NathanKell:
Thank you for the headsup! I agree this is (in general) a great feature. It's just, as you say, dangerous for us. :D

The other netkans would be the Express Install ones ( https://github.com/KSP-RO/RP-1-ExpressInstall/blob/main/RP-1-ExpressInstall.netkan and the three graphics ones at https://github.com/KSP-RO/RP-1-ExpressInstall-Graphics ). Oh and RSSCanaveral-HD and CanaveralPads since those dep some stock mods that then rec other stock-KSC-focused mods.

I think that covers us? I don't think we have to worry about the ROMods because they don't dep much at all, but I might eventually do a pass on every mod of ours to suppress it (because most will dep, say, B9PS or TU or something like that).

As I understand it, this is how the netkan-inflater action runs anyway, or something like that? Certainly it pulls in more than depends.

techman83 (approved):
Really neat feature! Thanks @HebaruSan

PR #3893, opened 2023-08-25
Handle backslashes in ZIP paths

HebaruSan:
## Problem

KSP-CKAN/KSP2-NetKAN#80 has some errors in the validation checks:

https://github.com/KSP-CKAN/KSP2-NetKAN/pull/80/checks

image

An already indexed mod is having the same problem with a recent update:

image

## Cause

These mods use backslashes (`\`) instead of forward slashes (`/`) in their filenames. The ZIP spec says:

https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT

```
   4.4.17 file name: (Variable)

       4.4.17.1 The name of the file, with optional relative path.
       The path stored MUST NOT contain a drive or
       device letter, or a leading slash.  All slashes
       MUST be forward slashes '/' as opposed to
       backwards slashes '\' for compatibility with Amiga
       and UNIX file systems etc.  If input came from standard
       input, there is no file name field.  
```

There is a PowerShell tool that some modders use that doesn't obey this, and CKAN doesn't currently handle that well.

## Changes

There are already many places where backslashes are replaced by forward slashes when CKAN interprets ZIP files, so it looks like the intent all along was to allow for this, but it wasn't complete. Now one more substitution is made, and it actually works.

Note that the bot will be able to handle these mods immediately after merge, but the old client that people are using now will fail to install them. We will need to release a new client pretty quickly after this is merged, so it should wait until the other pending pull requests are done.


techman83 (approved):
Looks great! Thanks @HebaruSan

HebaruSan:
Did some more research into the PowerShell tool today. This bug was fixed several years ago, see PowerShell/Microsoft.PowerShell.Archive#48. For any mod authors landing here after having problems with backslashes in your ZIPs, run this in PowerShell to install the fix:

```powershell
Install-Module Microsoft.PowerShell.Archive -MinimumVersion 1.2.3.0 -Repository PSGallery -Force
Import-Module Microsoft.PowerShell.Archive
```

techman83 (approved):
It's annoying, but so is dealing with broken mods that "work" via other tools that allow the spec not to be adhered to. Thanks @HebaruSan !

HebaruSan:
A quick note: This doesn't need to be reviewed again. I'm holding off on merging it until shortly before the next release, to minimize the amount of time mods with backslash ZIPs will be indexed before a client that can install them is available.

PR #3903, opened 2023-09-13
Shorten toolbar button labels

HebaruSan:
## Problem

On Discord, a Linux user with a 1024x768 monitor reported not being able to see the Filter dropdown button even with the GUI maximized:

image

## Cause

Windows's toolbar has a dropdown feature to accommodate toolbar overflow, but Mono never implemented that.

image

## Changes

Now the following buttons' English titles are shortened to allow them to fit into a smaller space:

- Launch Game → Play
- Add available updates → Update all
- Apply changes → Apply

These should be clear enough to both new and existing users.

image


techman83 (approved):
Looks sensible to me!

PR #3907, opened 2023-09-15
Fix race condition for multiple very fast downloads

### LINKED ISSUE #3906: GitHub test runner doesn't like the new NetAsyncDownloader test
## Problem

#3904 made the `NetAsyncDownloader` set the `target.size` for the files it downloads as a conveniece for calling code (in this case, so the `RepositoryDataManger` could report updates for progress bars while loading) and added tests for it (with `file://` URLs so we could always run them despite a flaky network).

- Result on Windows: All tests pass, no errors or warnings.
- Result on Ubuntu VM: All tests pass with warning from the test runner:
  image
- Result on GitHub: All tests pass most of the time (7/10 in the [merge commit](https://github.com/KSP-CKAN/CKAN/actions/runs/6200671759/job/16835820096)), but once in a while it reports that some of the sizes aren't set in the multi-download test:
  image

I investigated this extensively yesterday and thought that we resolved it by upgrading `NUnit-ConsoleRunner` from 3.11.1 to 3.12.0, but apparently it still happens.

I am worried that this is somehow finding a real problem that could blow up once the release ships.

HebaruSan:
## Problem

See #3906 for details; to sum up, one of the new `NetAsyncDownloader` tests from #3904 was failing in GitHub workflows but nowhere else.

When downloading multiple URLs, the `DownloadTarget.size` values were 0 for some of the files when `NetAsyncDownloader.DownloadAndWait` completed.

## Cause

This new test found a real, pre-existing problem for us! I'm really glad I didn't give in to the temptation to add `[Category("FlakyNetwork")]` and move on. 🎉

There's a race condition with multiple extremely fast downloads (the test is using `file://` URLs to avoid spurious failures caused by server glitches, so the files are already on disk and possibly cached in memory). At the start, the main thread loops over the requested URLs and starts a thread for each one. This sets up the `downloads` list with the currently executing downloads, and the `queuedDownloads` list with the downloads that are waiting their turn. When a thread completes, the sizes of the `downloads` and `queuedDownloads` lists are compared to `completed_downloads` to determine whether there are any other downloads in progress or waiting.

If at any time the active download threads manage to finish their work before all of the downloads are _started_, then they will be comparing the completed download count to _incomplete_ `downloads` and `queuedDownloads` lists! This results in the completion signal being sent early, and when the main thread returns to `DownloadAndWait`, it will skip waiting and signal completion with whatever was finished up to that point.

## Changes

- Now the downloader's main thread holds the `dlMutex` lock (which the other threads acquire before they check for completion) while it starts all the downloads. This ensures that by the time any downloader thread is able to check for completion, the `downloads` and `queuedDownloads` lists will be fully populated, which will prevent the completion notification from being triggered prematurely.
- The test now uses 13 files instead of 4, which made the problem start happening in my Ubuntu VM where it hadn't before (there seems to be a sensitivity to the number of downloads, with more downloads making the problem more likely).
- Now the test performs two additional checks: comparing the real downloaded file sizes that it observes on disk to the target sizes and the sizes of the original files.
- Since that test was so helpful, I decided to try adding one for failed downloads. In the process, I discovered that the index values in the failed downloads exception can be incorrect, so some further changes are made to clean that up (specifically, never queueing `file://` URLs and looking up the failed targets in the original input, which is now an `IList` instead of an `ICollection` because we need to call `IndexOf` on it).

Fixes #3906.

### DIFF
--- a/Core/Net/Net.cs
+++ b/Core/Net/Net.cs
@@ -168,7 +168,7 @@ namespace CKAN
             return targets.First().filename;
         }
 
-        public static void DownloadWithProgress(ICollection<DownloadTarget> downloadTargets, IUser user = null)
+        public static void DownloadWithProgress(IList<DownloadTarget> downloadTargets, IUser user = null)
         {
             var downloader = new NetAsyncDownloader(user ?? new NullUser());
             downloader.onOneCompleted += (url, filename, error, etag) =>
--- a/Core/Net/NetAsyncDownloader.cs
+++ b/Core/Net/NetAsyncDownloader.cs
@@ -152,27 +152,30 @@ namespace CKAN
         /// Start a new batch of downloads
         /// </summary>
         /// <param name="targets">The downloads to begin</param>
-        public void DownloadAndWait(ICollection<Net.DownloadTarget> targets)
+        public void DownloadAndWait(IList<Net.DownloadTarget> targets)
         {
-            if (downloads.Count + queuedDownloads.Count > completed_downloads)
+            lock (dlMutex)
             {
-                // Some downloads are still in progress, add to the current batch
-                foreach (Net.DownloadTarget target in targets)
+                if (downloads.Count + queuedDownloads.Count > completed_downloads)
                 {
-                    DownloadModule(new NetAsyncDownloaderDownloadPart(target));
+                    // Some downloads are still in progress, add to the current batch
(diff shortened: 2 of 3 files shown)


techman83 (approved):
Oh awesome, this likely explains the errant failure reported here and there. Thanks @HebaruSan !

PR #3908, opened 2023-09-16
Fix duplicate key exception in CompatibilitySorter

HebaruSan:
## Problem

The metadata validator threw a duplicate key exception while checking KSP-CKAN/NetKAN#9800:

https://github.com/KSP-CKAN/NetKAN/actions/runs/6207962451/job/16853654203?pr=9800

```
  Unhandled Exception:
  System.ArgumentException: An item with the same key has already been added. Key: [kOS-EVA, CKAN.AvailableModule]
    at System.Collections.Generic.TreeSet`1[T].AddIfNotPresent (T item) [0x0001a] in <33b19a7ad5234d94abf4fd9b47566616>:0 
    at System.Collections.Generic.SortedSet`1[T].Add (T item) [0x00000] in <33b19a7ad5234d94abf4fd9b47566616>:0 
    at System.Collections.Generic.SortedDictionary`2[TKey,TValue].Add (TKey key, TValue value) [0x00020] in <33b19a7ad5234d94abf4fd9b47566616>:0 
    at CKAN.CompatibilitySorter.PartitionModules (System.Collections.Generic.IEnumerable`1[T] dicts, System.Collections.Generic.Dictionary`2[TKey,TValue] providers) [0x00114] in <f9f0ffbfa5e14bc18bf2f56116c5f6c1>:0 
    at CKAN.CompatibilitySorter..ctor (CKAN.Versioning.GameVersionCriteria crit, System.Collections.Generic.IEnumerable`1[T] available, System.Collections.Generic.Dictionary`2[TKey,TValue] providers, System.Collections.Generic.Dictionary`2[TKey,TValue] installed, System.Collections.Generic.HashSet`1[T] dlls, System.Collections.Generic.IDictionary`2[TKey,TValue] dlc) [0x0005b] in <f9f0ffbfa5e14bc18bf2f56116c5f6c1>:0 
```

## Cause

The `CompatibilitySorter` from #2963 was updated to handle multiple source repos in #3904, but multiple `AvailableModule`s with the same identifier weren't covered. This is essential for the metadata validator.

## Changes

- Now if the repositories contain multiple `AvailableModules` with the same identifier, the `CompatibilitySorter` merges them into one new `AvailableModule` containing the combined list of `CkanModule`s; if there are multiple `CkanModule`s with the same identifier and version, the module from the higher priority repo (defined as higher up in the list, so a lower number) takes precedence
- A test is added to exercise this case, which fails without the fix and passes with it
- `AvailableModule.Remove` wasn't being used and is deleted
  `AvailableModule.Add` is only being used internally so is now private

We may have to update the metadata validator after this to set the priority of the repositories to get the right modules to load; I think we were previously skating by on the coincidence that `default` comes before `local` alphabetically.

### DIFF
--- a/Core/Registry/CompatibilitySorter.cs
+++ b/Core/Registry/CompatibilitySorter.cs
@@ -1,6 +1,8 @@
 using System.Collections.Generic;
 using System.Linq;
+
 using log4net;
+
 using CKAN.Extensions;
 using CKAN.Versioning;
 
@@ -31,7 +33,12 @@ namespace CKAN
             this.installed = installed;
             this.dlls = dlls;
             this.dlc  = dlc;
-            PartitionModules(available, CompatibleProviders(crit, providers));
+            var merged = available.SelectMany(dict => dict)
+                                  .GroupBy(kvp => kvp.Key,
+                                           kvp => kvp.Value)
+                                  .ToDictionary(grp => grp.Key,
+                                                grp => AvailableModule.Merge(grp.ToList()));
+            PartitionModules(merged, CompatibleProviders(crit, providers));
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


techman83 (approved):
Looks great, and extra tests :heart_eyes:  thanks @HebaruSan !

Issue #3912, opened 2023-09-18
[Bug]: CKAN 1.33 producing *numerous* problems on Kubuntu 20.04


saxyomega90125:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Operating System

Kubuntu 20.04

### CKAN Version

1.33.2

### Game Version

1.12.5.3190

### Did you make any manual changes to your game folder (i.e., not via CKAN)?

Added Parallax Lifeless Eve folder

### Describe the bug

I installed CKAN 1.33.2 using the deb file, over an older version which worked fine as of May. (1.30.something, I can't view it now.) CKAN 1.33.2 opens fine, but the first issue I notice is that all mods show as manually installed, which is wrong as I know some of these are auto-installed dependencies.

When I click Add Available Updates, it seems like it temporarily freezes all mouse buttons on my system - I can still alt-tab between windows and move the cursor, but I cannot click or scroll anything. This lasts anywhere from 5 to 30 seconds, then it seems to resolve itself. However, if I attempt to close CKAN during this process, commanding the window to close and killing the process fail to close the window and to reduce CPU load, and I have to restart my system to achieve either of those. Other applications fail to close properly as well, with windows closing but the application still running in the background with full system resources.

Once it completes, when I click Apply Changes, it wants to update every mod I have, with most saying the reason for the action is "Reinstall (metadata changed)". When I then click Apply, it says "Please wait" for a good 20 seconds while my CPU load surges. Again, if I attempt to click Cancel or to close CKAN, it crashes and my entire system maintains load for no detectable reason.

Once CKAN completes this pause, it will begin working and seems fine until it reaches Parallax, at which point it stalls while trying to remove Parallax-StockTextures (about 29% progress for my mod list it seems). The logbox output is attached. After observing this issue repeatedly, I manually deleted the Parallax folders except for Lifeless Eve using the file manager. After refreshing CKAN and trying to update mods again, it appears to have worked correctly including reinstalling Parallax. It reported that Trajectories was a present and unused directory (which is strange since I have and use that mod) and I told it to delete the directory... which did nothing I can see in my file manager.

Now that I have completed this, on running CKAN and clicking Add Available Updates (which still behaves the same) it wants to update absolutely every installed mod with the reason once again being "Reinstall (metadata changed)". I did not attempt to do this again.

[...]

PR #3913, opened 2023-09-19
Fix auto-remove during upgrade

### LINKED ISSUE #3911: Auto-removal of MagiCore at upgrade of RP-1 caused RP-1 Express to be removed
Writing this down as an issue so it doesn't get lost on a mod's discord.

[Starting here on the RO discord](https://discord.com/channels/319857228905447436/512556137380315139/1153203205157687346) two users encountered an issue with the latest metadata update to RP-1. Here is the setup:

* They installed RP-1 via the express install, which deps a large number of mods, including RP-1 itself
* We released a new version of RP-1 today
* Concurrent with that, I merged [HebaruSan's PR here](https://github.com/KSP-RO/RP-1/pull/2233) and then removed the dependency on Magicore [here](https://github.com/KSP-RO/RP-1/commit/b9f085575debb6be46dfeed3f722f3110ce0d046)
* Netkan bot picked up the new release using the new netkan
* A couple hours later, I also merged [Qazerowl's PR here](https://github.com/KSP-RO/RP-1/pull/2205)
* A little later [this PR](https://github.com/KSP-CKAN/CKAN-meta/pull/3206) was merged to CKAN-meta, directly touching the .ckans of the RP-0 mod (the Express Install has an any_of dep on RP-1 or RP-0). This resulted in the there being no choice and RP-1 being auto-selected when resolving the any_of on new installs.
* A couple hours after that, the discussion began in the support channel of the discord.

The symptoms:
* The users were given the option to upgrade RO, RSS, and RP-1.
* On doing so, they were prompted to uninstall some, but not all, of the mods the Express Install depends on, as part of (it appears) the express install itself uninstalling.

Manually reselecting the express install in CKAN and reapplying brought it and its dependencies back successfully. In the past, when the _express install's_ metadata changed, users were prompted to upgrade it and thus to uninstall _but then CKAN reinstalled_ its dependencies; this seems to be a different situation.

I diffed the previous RP-1 .ckan file with the initial RP-1 v2.12 .ckan file, and the only changes were (a) the addition of the recommend-suppression on the dependencies, and the removal of Magicore as a dependency. I also diffed the 2.12 ckans from before and after netkan bot inflated the result of the second PR to change the netkan, and it showed only the expected reorganization of the recommends and suggests (which should not have been relevant at all to the Express Install on CKAN 1.33.2). There have been no changes to the express install or the graphics sub-install .ckans in the last 3 weeks, so this seems entirely due to the RP-1 change.

As a band-aid I've reverted the netkan changes to RP-1, in the hope that will solve the problem for other users and they will not also end up with broken installs. But I remain mystified how changing the dependencies of a dependency causes a parent mod (the only manually-selected, rather than auto-installed, mod, to boot!) to be uninstalled.

HebaruSan:
## Problem

In https://github.com/KSP-RO/RP-1/commit/b9f085575debb6be46dfeed3f722f3110ce0d046, `RP-1`'s dependency on `MagiCore` was removed from its metanetkan, and this change first reached users with the indexing of `RP-1` v2.12 in https://github.com/KSP-CKAN/CKAN-meta/commit/98a346180f649c277e026a18cd69d0c3b5055ac8. When they upgraded, `RP-1-ExpressInstall` and all of its auto-installed dependencies were uninstalled (which should not have happened!). The dependency was subsequently restored pending a CKAN fix.

### How to reproduce for testing

1. Use a fresh game instance of KSP1 1.12.5
2. In CKAN, go to the settings and replace the default repo with this URL to simulate using CKAN before the problematic upgrade (name doesn't matter):
   https://github.com/KSP-CKAN/CKAN-meta/archive/0451525c2c9cefd16914032a2ca38d69c9e3f21c.tar.gz
3. Click Refresh to update metadata (will happen automatically if using a current dev build)
4. Install `RP-1-ExpressInstall`
5. When prompted to choose a version of `RP-1`, choose the 2.x series version
   (the other choices you make don't matter, issue will happen regardless)
6. Note that with the current dev build, there will be an extremely long delay while it computes the recommendations on the extremely long `RP-1` dependency list because #3892 was merged, ~~KSP-RO/RP-1#2233 was reverted, and KSP-RO/RP-1-ExpressInstall-Graphics#4 and KSP-RO/RealismOverhaul#2913 have not yet been merged. I have sent a reminder about this.~~
   Those pull requests have now been merged, however I forgot that that will not help us here as we are intentionally reaching back into the archive to use old versions of the metadata for testing. But at least now it will not be as slow after the next client release.
7. Now go back to the CKAN settings and replace your repo with this URL to simulate using CKAN right after the release of RP-1 v2.12:
   https://github.com/KSP-CKAN/CKAN-meta/archive/98a346180f649c277e026a18cd69d0c3b5055ac8.tar.gz
8. Click Refresh to update metadata (will happen automatically if using a current dev build).
   You should see four mods ready to upgrade.
9. Choose to upgrade those mods and click Apply
   image
10. This is where `RP-1-ExpressInstall` (and many other mods) would have been removed previously. Instead you should see this with this PR's changes:
   image
11. Finish upgrading and confirm that `MagiCore` is gone and `RP-1-ExpressInstall` is still there

## Cause

- Upgrading RP-1 triggers auto-removal of MagiCore, as it should
- GUI always calls `ModuleInstaller.UninstallList` before `ModuleInstaller.InstallList` before `ModuleInstaller.Upgrade`:
  https://github.com/KSP-CKAN/CKAN/blob/341adf88886d6d86626bca0073421cf576af5081/GUI/Main/MainInstall.cs#L169-L184
  ... so the first change that happens is removal of MagiCore
- The CKAN core code performing that removal doesn't know anything about the pending upgrade, it just knows it's been asked to remove MagiCore and that the currently installed version of RP-1 depends on MagiCore
- When that code removes MagiCore, it also removes all mods that depend on it, including the currently installed (old) version of RP-1. This is mildly surprising and not ideal, but not necessarily a serious problem yet.
- However, that dependency search is **transitive**, so everything depending on RP-1 is also  removed, including RP-1 Express! **This is when things definitely are going wrong**.
- The removal of RP-1 Express also triggers the automatic removal of any mods marked as auto-installed that only RP-1 Express depended on. This is desirable in a narrow sense, but in the context of removing a mod that we don't really want to remove, it just makes a bigger mess.
- When the Remove step finishes, essentially everything is uninstalled
- Since we're not installing anything, the InstallList step does nothing
- Finally the Upgrade step happens, which causes all the "Update" mods in the changeset to be reinstalled:
  image
  ... so RP-1, RealismOverhaul, RealFuels, and KSPCommunityFixes.
- But RP-1 Express was never in the changeset, so it doesn't get reinstalled!

## Changes

- Now `ModuleInstaller.Upgrade` handles auto-removal internally
  (`ModuleInstaller.UninstallList` already handled auto-removal internally)
  - Tests are added for auto-removal
- Now GUI doesn't pass auto-removals to `ModuleInstaller.UninstallList` unless the changeset consists only of installs and auto-removals, because we want them to be handled by `Upgrade` and/or `UninstallList` whenever possible
- While working on this, I discovered that the refactor of the DLL scanning code from #3904 throws if you have multiple DLLs with the same filename (which the KSP-RO team is currently doing unintentionally by shipping `System.dll`, `System.Core.dll`, and `System.Xml.dll` in several of their ZIPs). Now this is fixed courtesy of `GroupBy`.
- While working on this, I discovered that `RP-1-ExpressInstall` has enough tags to wrap the display in GUI, but the panel containing them doesn't resize to fit. Now it does.
- While working on this, I noticed that if you select a very large changeset, it takes a long time to calculate the recommendations, during which CKAN shows the changeset tab and is responsive but doesn't look busy. Now the hourglass cursor appears during this to show that it is doing something.
- The right-click reinstall option now uses `Upgrade` instead of cobbling together its own custom changeset of uninstalls and installs, to take advantage of this smarter handling of auto-removal for that use case (fixes the part of #3912 where it's unnecessarily trying to uninstall Parallax-StockTextures). When this option is used, the changeset will say "Re-install (user requested)" in the English locale.

Fixes #3911.

### DIFF
--- a/Core/ModuleInstaller.cs
+++ b/Core/ModuleInstaller.cs
@@ -686,7 +686,7 @@ namespace CKAN
                     .Select(im => im.identifier))
                 .ToList();
 
-            // If there us nothing to uninstall, skip out.
+            // If there is nothing to uninstall, skip out.
             if (!goners.Any())
             {
                 return;
@@ -1063,14 +1063,15 @@ namespace CKAN
         public void Upgrade(IEnumerable<CkanModule> modules, IDownloader netAsyncDownloader, ref HashSet<string> possibleConfigOnlyDirs, RegistryManager registry_manager, bool enforceConsistency = true, bool resolveRelationships = false, bool ConfirmPrompt = true)
         {
             modules = modules.Memoize();
+            var registry = registry_manager.registry;
 
             if (resolveRelationships)
             {
                 var resolver = new RelationshipResolver(
                     modules,
-                    modules.Select(m => registry_manager.registry.InstalledModule(m.identifier)?.Module).Where(m => m != null),
+                    modules.Select(m => registry.InstalledModule(m.identifier)?.Module).Where(m => m != null),
                     RelationshipResolver.DependsOnlyOpts(),
-                    registry_manager.registry,
+                    registry,
                     ksp.VersionCriteria()
                 );
                 modules = resolver.ModList().Memoize();
@@ -1088,7 +1089,7 @@ namespace CKAN
             // Let's discover what we need to do with each module!
             foreach (CkanModule module in modules)
             {
-                InstalledModule installed_mod = registry_manager.registry.InstalledModule(module.identifier);
+                InstalledModule installed_mod = registry.InstalledModule(module.identifier);
 
                 if (installed_mod == null)
                 {
@@ -1161,6 +1162,26 @@ namespace CKAN
                 }
             }
 
+            var removingIdents = to_remove.Select(im => im.identifier).ToHashSet();
+            var autoRemoving = registry
+                .FindRemovableAutoInstalled(
+                    // Conjure the future state of the installed modules list after upgrading
+                    registry.InstalledModules
+                            .Where(im => !removingIdents.Contains(im.identifier))
+                            .Concat(modules.Select(m => new InstalledModule(null, m, new string[0], false)))
+                            .ToList(),
+                    ksp.VersionCriteria())
+                .ToList();
+            if (autoRemoving.Count > 0)
+            {
+                foreach (var im in autoRemoving)
+                {
+                    User.RaiseMessage(Properties.Resources.ModuleInstallerUpgradeAutoRemoving,
+                                      im.Module.name, im.Module.version);
+                }
+                to_remove.AddRange(autoRemoving);
+            }
+
             if (ConfirmPrompt && !User.RaiseYesNoDialog(Properties.Resources.ModuleInstallerContinuePrompt))
             {
                 throw new CancelledActionKraken(Properties.Resources.ModuleInstallerUpgradeUserDeclined);
@@ -1169,14 +1190,12 @@ namespace CKAN
             // Start by making sure we've downloaded everything.
             DownloadModules(modules, netAsyncDownloader);
 
-            AddRemove(
-                ref possibleConfigOnlyDirs,
-                registry_manager,
-                modules,
-                to_remove,
-                enforceConsistency,
-                true
-            );
+            AddRemove(ref possibleConfigOnlyDirs,
+                      registry_manager,
+                      modules,
+                      to_remove,
+                      enforceConsistency,
+                      true);
             User.RaiseProgress(Properties.Resources.ModuleInstallerDone, 100);
         }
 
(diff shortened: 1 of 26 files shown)


NathanKell:
Oh man we provided a _lot_ of problems, didn't we :D

HebaruSan:
@NathanKell, the KSP-RO project can be relied upon to probe formerly rare corner cases and sometimes turn them into core requirements overnight. 😜

techman83 (approved):
That looks like a sensible set of operations and an improvement of the logic. I shall leave the testing of the changes to @NathanKell / others.

I'd prefer the extra changes live their own commits, so context isn't lost if we have to look back through the history of why something was changed. But I'll leave it with your best judgement @HebaruSan

Great outcome!

PR #3927, opened 2023-11-12
Clean up Linux .desktop files

irasponsible:
Fixed a typo in `debian/ckan-cmdprompt.desktop` ('propmt' -> 'prompt'), and remove all 3 .desktop files from `Utilities`.
If it's to go anywhere outside of `Games`, it would probably be better in `Database` or `Development`, I think.

### DIFF
--- a/debian/ckan-cmdprompt.desktop
+++ b/debian/ckan-cmdprompt.desktop
@@ -2,7 +2,7 @@
 Name=CKAN Command Prompt
 Comment=Comprehensive Kerbal Archive Network Command Prompt
 Icon=/usr/share/icons/ckan.ico
-Categories=Game
+Categories=PackageManager;ConsoleOnly;
 Exec=/usr/bin/ckan prompt
 Terminal=true
 Type=Application
--- a/debian/ckan-cmdprompt.desktop
+++ b/debian/ckan-cmdprompt.desktop
@@ -1,8 +1,8 @@
 [Desktop Entry]
-Name=CKAN Command Propmt
+Name=CKAN Command Prompt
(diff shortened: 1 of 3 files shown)


HebaruSan:
Thanks for the fixes, @irasponsible! I think `Utility` was added because CKAN itself isn't actually a _game_ per se so it didn't quite feel right adding it only there. `Game Utility` or a `Utility` category *within* `Game` would have been perfect. But yeah, it probably doesn't belong alongside disk space graphers or calculators.

Fortunately while looking for an official guide on this question, I found the .desktop file specification has some official lists:

- https://specifications.freedesktop.org/menu-spec/latest/apa.html
- https://specifications.freedesktop.org/menu-spec/latest/apas02.html

`PackageManager` may be more appropriate than `Game`, what do you think?

There's a `ConsoleOnly` that we could add to both of the non-GUI ones as well.

HebaruSan (approved):
Thanks again! I've moved it to the `PackageManager` category as suggested so a user finding it in "Game" doesn't feel disappointed when it isn't fun after launching it.

irasponsible:
> But yeah, it probably doesn't belong alongside disk space graphers or calculators.

I was looking at the same spec, trying to see if it defines `Utilities`, since that was my exact thought too.

I think having it in `Game` as well is a good idea - putting it in only `PackageManager` (at least with Cinnamon) relegates it to "other programs", hiding it from the menu. Maybe in both `PackageManager` and `Game`, or `PackageManager` and `Network`? If anyone disagrees too much, they can always edit the .desktop themselves! (like I did)

HebaruSan:
OK, added it back in fe366fee4440575fb6d1c2ed33c50b95fbb1d973.

PR #3931, opened 2023-11-22
Make where-CKAN-would-install logic consistent

HebaruSan:
## Problem

In an extremely obscure and unlikely case, you could get an error similar to this while installing a mod with CKAN if you had previously installed another mod manually:

screenshot

(This exact error has a different cause, but it uses the same error message template.)

Specifically, it could happen if:

- You're installing a module containing a file with a name of the form `identifier-something-else.dll`, where "identifier" is the mod's identifier and "something-else" is a string indicating that it's not the real DLL
- You have manually installed a file in game data with a name of the form `identifier.dll`

## Cause

Where the code to detect manually installed mods checks that the filename starts with `identifier.` with a dot, the loop over the installing files checked whether the file started with the identifier, but it didn't check what the next character was. In effect very slightly different logic was being used for the files on disk and the files being installed. This is very unlikely to cause a real problem because the identifier still has to match the file on disk, but it's still better to keep it tidy when something like this is spotted.

## Changes

- Now we use the exact same logic for installing files as for already installed unmanaged files
- Now three new test cases exercise this logic, one of which fails under the old logic and passes under the new. To make this work:
  - The `DogeCoinPlugin` ZIP file is updated to contain `GameData/DogeCoinPlugin/Plugins/DogeCoinPlugin.dll` instead of `GameData/DogeCoinFlag/plugin/dogecoin.dll`
  - A new `DogeCoinPluginAddonFerram` ZIP file is created containing `GameData/DogeCoinPlugin/Plugins/DogeCoinPlugin-Addon-Ferram.dll` so we can test the case that is fixed by the change in Core

Adding multiple reviewers based on familiarity with the issue. Since we only need one approval, don't feel obligated to look at this if you're disinclined.

(Related to KSP-CKAN/NetKAN#9856 and giuliodondi/kOS-Ferram#3 but not a fix for either one.)


techman83 (approved):
Awesome! Concise, clear, and super sensible to use the same implementation, with tests! Love your work @HebaruSan

PR #3933, opened 2023-11-30
Modernize administrator and Mono version checks

HebaruSan:
## Motivations

- CKAN has a check for whether the user is privileged and an `--asroot` flag to bypass this, but it's Linux-only. Windows users are just as likely to benefit from this guardrail. It's also implemented entirely within CmdLine, even though it could in principle apply more generally.
- There's a check that prints a warning if the user is on a version of Mono before 3.1.0, which duplicates platform-specific logic from Core in CmdLine, and I'm pretty sure you need a much newer version of Mono to run current CKAN
- `Platform.IsMonoFourOrLater` and `Platform.IsOnMonoFourOrLater` have been unused since #3041

## Changes

- Now the root user check is moved to `Platform.IsAdministrator()`
  - Now it has an `IsWindows` section that checks for administrator or system roles
  - Now the messages for the root user check are platform agnostic
  - I wanted to add an `--asadmin` alias for the `--asroot` flag, but I could not find a mechanism for that in the command line parser library, and I don't want to break existing scripts or shortcuts by changing it
- Now the duplicative Mono versioning logic from `Platform.IsOnMonoFourOrLater` and `CmdLine.Options.CheckMonoVersion` is moved to a `Platform.MonoVersion` field
  - Now the Mono warning is printed if they're on a version before 5.0.0 (stored in `Platform.RecommendedMonoVersion`), which is still pretty old but not as guaranteed ancient as 3.1.0
  - This uses a new `RegexExtensions.TryMatch` function to check a regex for a match and use its output within a functional-style expression

### DIFF
--- a/Cmdline/Main.cs
+++ b/Cmdline/Main.cs
@@ -7,7 +7,6 @@ using System;
 using System.Net;
 using System.Diagnostics;
 using System.Linq;
-using System.Runtime.InteropServices;
 #if NET5_0_OR_GREATER
 using System.Runtime.Versioning;
 #endif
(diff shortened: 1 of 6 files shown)


techman83 (approved):
This is super clean, and removing unused code is always a win :tada: thanks @HebaruSan !

PR #3938, opened 2023-12-06
Improve file deletion error while the game is running

HebaruSan:
## Motivation

When you launch KSP1 on Windows, it locks either many or all of the mods' files. If you then try to uninstall those mods with CKAN while the game is still running, you get a hard to understand message that mentions the file but gives no information about why the problem happened or what you can do about it:

before

Brought up by @JonnyOThan on Discord.

## Changes

- Now the message is reformatted to be more understandable:
  > Unable to overwrite or delete files for MyMod:
  >
  > GameData\MyMod\MyMod.dll
  >
  > If the game is still running, close it and try again. Otherwise check the permissions.
  - The top line states simply that we were trying to delete files and could not
  - The bottom line says to close the game and check permissions
  - The file list is now relative to game root and uses local OS path separators (backslash on Windows)
- Now the changeset warns the user to close the game before applying changes that involve deletion:
  changeset after
- At some point, .NET added an `Enum.HasFlag()` function for checking flag-style enums; previously you had to use bitwise AND and a comparison. Now we use the new function.

### DIFF
--- a/ConsoleUI/Toolkit/ConsoleButton.cs
+++ b/ConsoleUI/Toolkit/ConsoleButton.cs
@@ -72,7 +72,7 @@ namespace CKAN.ConsoleUI.Toolkit {
         {
             switch (k.Key) {
                 case ConsoleKey.Tab:
-                    Blur((k.Modifiers & ConsoleModifiers.Shift) == 0);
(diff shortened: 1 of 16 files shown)


techman83 (approved):
Awesome, this has been a common one to confuse people. Thanks @HebaruSan !

PR #3939, opened 2023-12-07
Search by supports relationship and other search improvements

HebaruSan:
## Motivation

- The mod search can search all relationships except `supports`. @JonnyOThan suggested adding `supports`.

## Problems

- If you click the search dropdown button on Windows and then click it again, the dropdown disappears but you have to click the button two more times to get it to reappear. (On Mono this already works perfectly.)
- Ctrl+Backspace is a common hotkey for deleting whole words in a textbox (with a dubious history as a "rogue feature" starting with early Internet Explorer releases), but the CKAN GUI currently only supports it on Mono (where it's built in to the runtime)
- While working on the rest of this, I noticed that the "Help &rarr; Report an issue with mod metadata" link assumed the game was KSP1, even if you were working with a KSP2 instance

## Causes

- The search dropdown was set up to be a separate floating window, but clicking outside of it didn't make it invisible! Instead the main window was moved in front of it. The second click on Windows then toggled the button and officially hid the already non-visible dropdown, and the third click re-displayed it.

before

## Changes

- Now there's a Supports field in the search dropdown, corresponding to a `sup:` search prefix in the English locale
  image
- Now the search dropdown is properly set up as always-on-top, and we hide it when you click outside of it. A special case is needed when you click on the expand/collapse button again to avoid hiding it and then immediately re-displaying it. `Platform.IsWindows` is used because Windows and Mono differ drastically in how they handle this.
- Now Ctrl+Backspace works in our textboxes, using the de facto standard Windows API.
  Closes #3398.
- Now the mod metadata link is game-specific and opens KSP2-NetKAN for KSP2

### DIFF
--- a/Core/Games/IGame.cs
+++ b/Core/Games/IGame.cs
@@ -45,7 +45,8 @@ namespace CKAN.Games
         string[]          BuildIDFiles { get; }
 
         // How to get metadata
-        Uri DefaultRepositoryURL { get; }
-        Uri RepositoryListURL    { get; }
+        Uri DefaultRepositoryURL  { get; }
+        Uri RepositoryListURL     { get; }
+        Uri MetadataBugtrackerURL { get; }
     }
 }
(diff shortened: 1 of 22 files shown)


HebaruSan:
I want to get this merged so we can prep for a new release before KSP2 0.2.0.0 comes out, and it looks solid in testing on Windows and Linux. Self-reviewing...

PR #3943, opened 2023-12-12
Fix two small typos in Resources.pt-BR.resx at line 182

idrkwhattoput:
Fixed small typos in line 182 (appears when the user tries to install a mod in a non-supported KSP version).

accidentaly discovered this when i was trying to install a mod, so here it is 👍

### DIFF
--- a/GUI/Properties/Resources.pt-BR.resx
+++ b/GUI/Properties/Resources.pt-BR.resx
@@ -179,7 +179,7 @@ Tentando mover {2} de {3} e reiniciar o CKAN.</value></data>
   <data name="MainTSV" xml:space="preserve"><value>Arquivo separado por tabulação (*.tsv)</value></data>
   <data name="MainNotFound" xml:space="preserve"><value>Não encontrado.</value></data>
   <data name="MainCantInstallDLC" xml:space="preserve"><value>O CKAN não pode instalar a expansão{0}!</value></data>
-  <data name="AllModVersionsInstallPrompt" xml:space="preserve"><value>{0} não é suportado pela versão atual do seu jogo e pode não funcionar. Se você tiver qualquer problema com isso, você NÃO deve pedir ajuda aos colaboradores.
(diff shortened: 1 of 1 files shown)


HebaruSan:
Thanks for the update! Was the other change intentional? The old text "mantenedores" seems to be a closer fit to the English "maintainers" than "colaboradores".

Also FYI that there's an easier way to edit these nowadays:  https://crowdin.com/project/ckan

idrkwhattoput:
> Hi @idrkwhattoput, thanks for the update! Was the other change intentional? The old text "mantenedores" seems to be a closer fit to the English "maintainers" than "colaboradores".
> 
> Also FYI that there's an easier way to edit these nowadays: https://crowdin.com/project/ckan

hello, yes the other change was intentional it maked more sense for me to put "colaboradores" instead of "mantenedores" im also looking out to help with the translation so thanks for the link

HebaruSan:
> it maked more sense for me to put "colaboradores" instead of "mantenedores"

@idrkwhattoput, thanks, would you mind saying why you think that? As I mentioned earlier, "mantenedores" gives an exact match to the English translation when I put it into a translator.

idrkwhattoput:
> > it maked more sense for me to put "colaboradores" instead of "mantenedores"
> 
> @idrkwhattoput, thanks, would you mind saying why you think that? As I mentioned earlier, "mantenedores" gives an exact match to the English translation when I put it into a translator.

Hmmm, well then sorry i never heard the word "mantenedores" when speaking Brazilian Portuguese 😅 so i thought it was incorrect.

idrkwhattoput:
finished

HebaruSan:
Please don't close this. We need it to be open so we can merge your fixes into the main branch.

idrkwhattoput:
ph sorry didnt know

HebaruSan (approved):
Thanks for the fix!

PR #3951, opened 2023-12-14
Conflict highlight for selected row

### LINKED ISSUE #3947: SelectionBackColor of conflict rows should be a different color (dark red?)
If you mark a mod to install that is in conflict (e.g. install Grannus and then Scatterer), you might not notice the red background of the conflict row because selected rows still use the default blue background color.

It should be possible to set the background color of selected conflict rows to a darker red:

https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.datagridviewrow.defaultcellstyle?view=windowsdesktop-8.0#system-windows-forms-datagridviewrow-defaultcellstyle

HebaruSan:
## Problems / Motivations

- If you select to install two mods that conflict, their background color properties are set to a reddish color, but this is completely overridden by the blue row highlight color. If the other conflicting mod's row is off-screen, this usually meant that you couldn't see any red highlights at all.
- Conflict descriptions are missing from the status bar in some cases
- If you apply multiple labels to a mod, only one of their background colors is applied

## Causes

- `RelationshipResolver.ConflictDescriptions` only returned descriptions of conflicts for the mod with fewer indirectly-depending mods, intended to choose the conflict closest to what the user selected, but when these values were equal, neither would appear.

## Changes

- Now if two conflicting mods have equal numbers of indirectly-depending mods, both of their description conflicts are shown in the status bar
- Now if you select a row with a conflict, the red color is blended 75%/25% with the highlight color, producing a reddish purple color that can still be recognized as conflict-y:
  image
  The red conflict color has had its green and blue channels roughly halved to make this work a bit better.
- Similar to the above, a background color from a label is also blended 75%/25% with the highlight color so you can still tell a label has been applied while the row is highlighted
- Now if you apply multiple labels, their colors are blended together, and the labels menu shows the colors of the labels:
  image

Fixes #3947.

### DIFF
--- a/Core/Relationships/RelationshipResolver.cs
+++ b/Core/Relationships/RelationshipResolver.cs
@@ -632,7 +632,7 @@ namespace CKAN
         public IEnumerable<string> ConflictDescriptions
             => conflicts.Where(kvp => kvp.Value == null
                                       // Pick the pair with the least distantly selected one first
-                                      || totalDependers(kvp.Key) < totalDependers(kvp.Value))
+                                      || totalDependers(kvp.Key) <= totalDependers(kvp.Value))
                         .Select(kvp => string.Format(
                             Properties.Resources.RelationshipResolverConflictsWith,
                             conflictingModDescription(kvp.Key,   null),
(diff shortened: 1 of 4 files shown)


HebaruSan:
Self-reviewing so we can include this in the hotfix release...

PR #3958, opened 2023-12-16
Update and optimize meta tester and inflator images

HebaruSan:
## Motivations

- After KSP-CKAN/NetKAN-Infra#322, our Docker images are updated to Python 3.11, except for the meta tester. We don't want to fall too far behind the current versions of Python.
- The meta tester contains a lot of stuff it doesn't need, weighing in at a hefty 1.23GB:
  - C/C++ compilers and libraries
  - Mono compilers
  - Mono libraries we don't need
  - Python build tools
  - Flask and gunicorn
  - APT and pip download caches
  - Old logs
  - Etc.

```
REPOSITORY         TAG       IMAGE ID       CREATED        SIZE
kspckan/metadata   latest    df95b1b2804f   20 hours ago   1.23GB
kspckan/inflator   latest    165d396574f3   20 hours ago   792MB
```

## Changes

Now instead of using `mono:latest`, which is an old OS that installs Python 3.7, we start with the much more current `ubuntu:latest` and install the latest Mono and Python. This gives the following software versions as of this writing:

```
Ubuntu 22.04.3 LTS
git 2.40.1
Python 3.10.12
Mono 6.12.0.200
```

These will be updated to the current versions for `ubuntu:latest` each time we build the image. Using a newer base OS also allows us to retire the certificate hackery from #3457, since the expired certificate has been removed from the distro. The image now includes only the files it needs at runtime (I built it and confirmed it can inflate netkans from GitHub and SpaceDock, create fake instances, and install mods) and is much smaller:

```
REPOSITORY            TAG       IMAGE ID       CREATED          SIZE
meta_tester_testing   latest    cecc228c5517   41 minutes ago   364MB
inflator_testing      latest    7f20aaa5f3cc   27 seconds ago   200MB
```

This might allow it to load faster when we use it on GitHub. :crossed_fingers:

The inflator also now excludes the Mono build tools and is a fraction of its former size.

To make the inflator's smoke test work without installing `curl` in the image, `netkan.exe` can now accept a URL to a netkan as its command line parameter.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,8 +76,7 @@ jobs:
           sh get-docker.sh
           docker build --tag inflator --file ../Dockerfile.netkan .
           docker run --rm --name inflator --entrypoint /bin/bash inflator -c "
-            curl -O https://raw.githubusercontent.com/KSP-CKAN/NetKAN/master/NetKAN/ZeroMiniAVC.netkan && \
-            mono netkan.exe ZeroMiniAVC.netkan
+            mono netkan.exe https://raw.githubusercontent.com/KSP-CKAN/NetKAN/master/NetKAN/ZeroMiniAVC.netkan
           "
         if: matrix.configuration == 'release'
 
--- a/Dockerfile.metadata
+++ b/Dockerfile.metadata
@@ -1,15 +1,57 @@
-FROM mono:latest
+# Everything we need in both the build and prod images
+FROM ubuntu:latest as base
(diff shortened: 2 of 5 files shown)


techman83 (approved):
Looks awesome, thanks for spending some time on this! There is probably a few more opportunities to shave more off. It might be worth combining all the apt pieces into one run action. It might not be significant, docker is a lot more efficient by default than the earlier days.

PR #3961, opened 2023-12-18
System Linq

techman83:
Linq was missing, causing the AWS libraries to fail loading. This installs it during build


HebaruSan (approved):
Confirmed this is what we discussed. :rocket:

Issue #3965, opened 2023-12-19
Cannot open CKAN


The-DapAttack:
Hello, I am attempting to open ckan.exe and it flashes a command line box then it disappears and nothing. 
I ran an update roughly 3 days ago and after the update I have not been able to open it successfully.

Any help is appreciated.
Apologies in advance if this is a duplicate question. I have looked all over and can find nothing.

HebaruSan:
Hi @The-DapAttack, thanks for the report. Can you try the suggestions in #3956 to see if your issue is the same? This sounds exactly the same as what was reported there so far.

The-DapAttack:
This worked, thank you!

HebaruSan:
You're welcome, glad you got it working! :tada:

HebaruSan:
Duplicate of #3956

Armour9:
I am having this error as well.  I have uninstalled CKAN and have reinstalled it and I am getting the immediate popup and CKAN closes. When I did a quick screenshot is says because I am trying to run as Admin.  I have ensured "run as admin" is unchecked.  Now CKAN does not work at all.

Issue #3956, opened 2023-12-14
Trying to run CKAN as an administrator


mnh713:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Operating System

Windows 10 home

### CKAN Version

version before v. 1.33.2 update

### Games

KSP 1

### Game Version

1.12.5

### Did you make any manual changes to your game folder (i.e., not via CKAN)?

no

### Describe the bug

After selecting update Ckan to v.1.33.2 from the Ckan main page. Ckan closed and no longer will open

### Steps to reproduce

removed Ckan from computer and re-download Ckan v. 1.33.2 and launch. Ckan still does not launch.

HebaruSan:
Hi @mnh713, which version of .NET Framework do you have installed? 4.8 is required to run the latest CKAN release, so you'll need to update that if you haven't already.

mnh713:
Yes I have v. 4.8

HebaruSan:
OK. And when you say it will no longer open, is there an error message?

mnh713:
No, after opening (launching) Ckan, the program launching window (looks
like a DOS window) then immediately closes. No error message is shown

HebaruSan:
Hmm, are you comfortable working from the command line? We might be able to capture more information like this:

1. Start → run
2. cmd
3. `cd` to the folder where ckan.exe is
4. `ckan.exe --show-console`

That should cause the output from the terminal window to be shown in the window you opened. If any messages were printed before it was closed, it would remain visible instead of disappearing too quickly to read.

mnh713:
Sure, I'll give it a try

mnh713:
Okay, I had to use the --asroot--show-console

HebaruSan:
Ahh, if you're running as administrator, can you try not doing that?

mnh713:
Okay, I'll try again

mnh713:
I typed (see attached),

HebaruSan:
Hi @mnh713, attachments don't come through GitHub's email interface. Can you share what you found in some other way?

mnh713:
I typed,

d:\SteamLibrary\steamapps\common\ckan.exe --show-console

The replay is "You are trying to run as CKAN as an administrator"

HebaruSan:
Thanks. Do you have a non-administrator user on your PC that you could try this as?

mnh713:
No, would need to create a non-administrative user, can try and do this now

[...]

PR #3966, opened 2023-12-19
Pause after administrator error message

HebaruSan:
## Problem

Users attempting to run GUI as a Windows administrator have reported that the console window appears and disappears before they have a chance to see what it says (see #3956 and #3965).

## Changes

- Now we use a "Press any key to exit" prompt to keep the window visible and only close it after the user presses a key.
- The message now describes the risks involved in more detail.

This should allow users to understand what's happening and make an informed decision.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -619,6 +619,7 @@ All notable changes to this project will be documented in this file.
 - [GUI] Fix 'ManageKspInstances' dialog logic (#2787 by: DasSkelett; reviewed: HebaruSan)
 
 ### Internal
+
 - [Build] Update packages (#2775 by: Olympic1; reviewed: DasSkelett, HebaruSan)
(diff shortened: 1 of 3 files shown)


techman83 (approved):
Awesome, I think this'll be good enough for the purpose. Thanks @HebaruSan

cjc123456789:
How does windows11 run CKAN with only one administrator account?

cjc123456789:
I have only one administrator account in Windows 11, this window appears when I run CKAN, and the window shows that I can't start after I press any key to exit. How do I start the program?

JonnyOThan:
@cjc123456789 this is the discussion thread for a pull request.  There’s more discussion about the administrator error here: https://github.com/KSP-CKAN/CKAN/issues/3956

cjc123456789:
@JonnyOThan Thank you

PR #3993, opened 2024-01-05
Fix #3518 rewrite de-indexing policy

### LINKED ISSUE #3518: [Policy] De-indexing policy still contradictory
Problem
-------
The de-indexing policy was amended a number of years ago, controversially. The outcome of this amendment is that the policy contradicts itself. 

The policy currently contains this text:

>   When an author gives us explicit and irrevocable permission via a legal document to redistribute their work, we cannot in good faith de-index the mod and still act in the interests of our users. Whenever possible, mods with permissive licenses will be preserved by the CKAN, especially when alternate downloads to such mods exist, including those cached by the CKAN's own automated systems.

>   In some cases it's not possible to install mods via the CKAN and maintain the goodwill of the author(s). The CKAN will not install any mod when the primary maintainer (or a significant group of maintainers when no primary is identifiable) requests the mod be removed from the CKAN.

These two paragraphs directly contradict each other. 

For those who came in late: This was last addressed in #2147, which was closed without merging. The issue was first raised in #1793. 

From the PR:

>   Delisting permissively-licensed mods is a workaround until we can implement a system to allow listing with manual installation.

It seems clear this "temporary workaround" is in fact the official policy by default. 

Suggestions
-----------
I submitted a PR already, which was closed (#2147). This could be adopted as it stands. 

Alternatives exist. A discussion by current contributors about CKAN's values and policies could generate a more up to date and inclusive version which clarifies the policy as it stands. 

Musings
----------
It seems clear that one must be more important - either the requirement to act in the interests of CKAN users, or the requirement to maintain the goodwill of the authors. At present, we prioritise the latter, and this seems unlikely to change anytime soon. Therefore it seems likely that any change to the document must come to the former paragraph - at its simplest, by deleting the first line. I suggest that simplicity of change is not a desirable outcome here.

JonnyOThan:


HebaruSan:
Thanks for taking this on! This is a huge improvement over the old policy.

I made a bunch of edits, largely tweaking the phrasing slightly in a few places, especially the "takes licenses seriously" part that evokes the old policy and it's "your license lets us do whatever we want" attitude. I also felt it was important to add a "fix problems if we can" step to the de-indexing section, so people aren't surprised when we ask them what's wrong.

JonnyOThan:
Hah, I didn't interpret the "takes licenses seriously" thing to imply "we can do what we want with mods that have permissive licenses."  I actually intended it in the opposite manner (as I hoped the body would make clear), we act to protect the license rights of mod authors against infringement by others.

IMO CKAN is *not* distributing mods, so the llicense of the mod has nothing to do at all with whether it can be indexed on CKAN.  It *is* relevant to the archive.org backup thing of course.

This all looks pretty good to me.

JonnyOThan:
Yeah to be honest the change to "but does not use them as an excuse to do things mod authors do not want" sounds kind of weird if you don't know the history here (as I suspect most people reading this probably won't).  I'm a little concerned it could be used against us at some point.  The original language, on its face, doesn't really promise any action one way or another - it merely states a principle we think is important.  I agree that keeping the old language could be weird to people who DO know the history, but I think in context it really changes the meaning.

I don't feel *that* strongly about it, so whatever you decide is fine with me.

HebaruSan:
In [the original policy](https://github.com/KSP-CKAN/CKAN/blob/82f971adf935954bf754852e6f2726686035cda9/policy/de-indexing.md), most of the document was focused on explaining how and why most de-indexing requests would be denied, using permissive licenses as justification. That phrase was there to lead in to the explanation of why the team back then would not do what authors asked.

This argument was both legally dubious (as you say, we don't "distribute" mods any more than Firefox and Google do) and a distraction from the question of what kind of policy would be best for all parties.

My problem with "takes licenses very seriously" is not that it _clearly means_ that same policy holds, but that its meaning is _ambiguous_ and subtly refers back to that policy where an explicit clean break is preferable.

JonnyOThan:
Right, principles are simple (sometimes ambiguous) statements of values "X is important."  The body of the document goes into what it actually means.

I know we're splitting hairs here, but something about the new language just feels a little weird to me.  The "very" in the old one is also a little odd.  How about:

> CKAN takes mod authors' licensing rights seriously

HebaruSan:
That's a nice compromise. It focuses on what licenses do _for_ authors rather than making them a liability. Go for it.

HebaruSan (approved):
Looks good, might as well merge now. If other team members have feedback, we can always make further changes.

PR #3997, opened 2024-01-10
Support dev builds for auto updater

### LINKED ISSUE #997: Ability to point the CKAN autoupdater to CKAN-nightlies rather than releases
rabble rabble bleeding edge rabble rabble cooler than release rabble rabble hipster rabble rabble

On a much more serious note though sometimes we get errors reported that are hotfixed in the code (mono 4 incompatible gui comes to mind) and leads to a nightly release but the real release might be anywhere from a few hours to a few days away due to a number of reasons.

In times like these one could point users to a nigthly release and ask them to install that but upon running a nightly the user is always prompted by the autoupdater to update to the latest official version, presumably due to < logic >.

Being able to point a user to a guisetting/cli command which switched the autoupdater over to nightlies could be helpful in those situations. I'd hope that such a change would prompt the user that the change  they're doing switches them over to bleeding-edge with all the potential unstability which might come out of such a thing.

### LINKED ISSUE #3518: [Policy] De-indexing policy still contradictory

HebaruSan:
## Motivation

- Fixes land in the dev builds before releases are made, so it would be nice for users experiencing problems to have easier access to them
- The v1.34.0 release had a lot of big changes that I knew would have bugs, but very few people were trying to test it. Unsurprisingly, we've now done two subsequent fix releases to catch up with the issues that made it into the release. Even if we could find people to help with testing, it would still be cumbersome for them to get the dev builds and stay up to date.

## Changes

Now interested users can opt in to the bleeding edge dev builds to get changes as they are merged into the `master` branch, which will hopefully allow us to receive earlier bug reports and reduce bug counts of releases. To make this work:

- Now the build piece of CKAN's version string is no longer part of the git commit hash but instead the git commit date in YYDDD format (2-digit year concatenated with 0-padded day of year, e.g. 1.34.5.24009, which we already use for the .rpm and .deb packages), to make dev build versions comparable
  - This is displayed in the CKAN title bar as well, to aid in identifying precise dev builds from screenshots
    image
- Now when we upload the dev build to S3, we also upload a new `version.json` file containing the client version (including the above build number) and the dev build's changelog entries, generated by a new Python script in `bin/version_info.py`:
  ```json
  {
      "version": "v1.34.5.24010",
      "changelog": "### Bugfixes\n\n- [GUI] Suppress admin user check for URL handler registration (#3996 by: HebaruSan)\n\n### Internal\n\n- [Policy] Fix #3518 rewrite de-indexing policy (#3993 by: JonnyOThan; reviewed: HebaruSan)"
  }
  ```
- Now the core config JSON file has a boolean `DevBuilds` property, and if it's true:
  - We retrieve that `version.json` file from S3 to determine whether the dev build is newer than the current client as well as get the changelog
  - We get `ckan.exe` and `AutoUpdater.exe` from S3 as well to update to the dev build
- Now the GUI settings dialog contains a "Use dev builds" checkbox in the Auto-Update section that toggles the `DevBuilds` setting
  - If you toggle this, it switches to the other kind of build and the labels show the version
  - The install button is enabled if the version is older in case you want to downgrade from dev builds to regular builds (otherwise there would be no way to "undo" switching to the dev builds)
  - The settings dialog is now wider and less tall to make vertical room for this new checkbox, with two groupboxes per row. The large height of this window was an occasional problem previously, so this should be a good change in general.
    image
- Now if you have auto-updates enabled, GUI will ask you whether you want to use dev builds (but this prompt will only appear once), so existing users can find out about this option:
  image
- `AutoUpdate.cs` is moved from `Core/Net` to `Core/AutoUpdate` and accompanied by a variety of other autoupdate-related files
- For CmdLine, the `ckan upgrade ckan` command now has `--dev-build` and `--stable-release` flags that toggle the setting before upgrading. If neither flag is present, then the previous value of the setting is used.

While I was working on the above, I noticed some minor refactoring opportunities:

- `Net` has several members that are only used by/for `NetAsyncDownloader`, which are now moved into that class to reduce the coupling of these two classes:
  - `DownloadTarget` (which is now also moved to its own file to make it easier to find)
  - `DownloadWithProgress`
- `NetAsyncDownloader.NetAsyncDownloaderDownloadPart` is a long redundant name, so now it's shortened to just `DownloadPart` as well as moved to its own file to make it easier to find
- `LeftRightRowPanel` was accidentally left out of the `GUI` namespace and is now moved into it
- The buttons in the `YesNoDialog` are now auto-sized (`GrowOnly` mode) and automatically positioned by a `LeftRightRowPanel`
- We now pass more of the objects the settings dialog needs so it can eliminate some of its `Main.Instance` references

Fixes #997 (just before its 10-year anniversary!).


HebaruSan:
`actions/checkout` isn't actions checkoutting anymore? And it wants us to control which git version it uses?

image

I'll look at this after my tolerance for nonsense has recharged.

PR #4007, opened 2024-01-23
Sort mods satisfying the same recommendation by download count

### LINKED ISSUE #4004: Recommendations are not sorted by download count
image

If you install EnvironmentalVisualEnhancements (eve redux) directly, you get this window recommending a config.  For some reason Whirligig World gets auto-selected, even though it has far fewer downloads than AVP or the stock configs.

HebaruSan:
## Motivation

In #3934, we sorted the providing mods prompt by download count to make it easier for users to find the mods that they're most likely to want.

EnvironmentalVisualEnhancements recommends rather than depends on its config module, so when you install it, instead of the providing mods prompt, it uses the recommendations prompt, which is sorted by the related mod column and otherwise tries to reflect the ordering from the mods' relationships, but doesn't do any particular sorting within a group of mods satisfying a virtual module recommendation:

image

## Changes

Now `Registry.LatestAvailableWithProvides`, which is used to satisfy virtual module recommendations, sorts its returned list by download count, which puts the most commonly used EVE configs at the top:

image

@JonnyOThan, does that list look right to you? I noticed a few spots where the download counts appear to be slightly out of order, but I'm not sure whether those are actual problems or just some harmless quirk elsewhere in the sorting logic.

Fixes #4004.

### DIFF
--- a/Core/Registry/Registry.cs
+++ b/Core/Registry/Registry.cs
@@ -814,6 +814,10 @@ namespace CKAN
                     .Select(am => am.Latest(gameVersion, relationship_descriptor,
                                             installed, toInstall))
                     .Where(m => m?.ProvidesList?.Contains(identifier) ?? false)
+                    // Put the most popular one on top
+                    .OrderByDescending(m => repoDataMgr.GetDownloadCount(Repositories.Values,
+                                                                         m.identifier)
+                                            ?? 0)
                     .ToList();
             }
             else


JonnyOThan:
I’m a bit surprised that GPP would be higher than spectra but having BoulderCo (heh) and AVP at the top certainly looks right.

HebaruSan:
To reassure myself, I dug up the identifiers to get the actual numbers:

```sh
jq '{"EnvironmentalVisualEnhancements-HR","AstronomersVisualPack","GPPCloudsHighRes","GPPCloudsLowRes","Spectra","RSSVE-HR","RSSVE-LR","JNSQ","PhotoRealisticVisualEnhancement","PhotoRealisticVisualEnhancement-LowRes","WhirligigWorld","StockPlanetaryUniversalRevamp","SciFiVisualEnhancements"}' download_counts.json
```

```json
{
  "EnvironmentalVisualEnhancements-HR": 2699354,
  "AstronomersVisualPack": 1122970,
  "GPPCloudsHighRes": 220474,
  "GPPCloudsLowRes": 220458,
  "Spectra": 163811,
  "RSSVE-HR": 146983,
  "RSSVE-LR": 146911,
  "JNSQ": 51429,
  "PhotoRealisticVisualEnhancement": 9430,
  "PhotoRealisticVisualEnhancement-LowRes": 9384,
  "WhirligigWorld": 7945,
  "StockPlanetaryUniversalRevamp": 4826,
  "SciFiVisualEnhancements": 22
}
```

The download counting algorithm could probably be improved here, but at a glance, I think the sort itself is working.

PR #4023, opened 2024-02-12
Better version specific relationships at install and upgrade

### LINKED ISSUE #3849: [Feature]: Upgrade to latest compatible version when a version conflict on absolute latest is encountered
### Problem

So for example I have an RP-1 v1.x install which I'd like to upgrade *but keep on v1.x* while installing the available compatible upgradable mods.

So for example I have:

```
% ckan list | egrep RealismOverhaul
^ RealismOverhaul v14.23.0.0
```

When I try to upgrade that I get:

```
% ckan upgrade RealismOverhaul
* RP-0 v1.13.2.2 missing dependency RealismOverhaul v15.99.0.0 or earlier
```

This is correct if I wanted to upgrade to RealismOverhaul v16.x, however CKAN-meta indicates that RealismOverhaul-v15.0.1.0.ckan is available to be installed.  Really I'm trying to upgrade to that v15.x version which is compatible.

I can get that to work manually by becoming my own depsolver with:

```
% ckan upgrade RealismOverhaul=v15.0.1.0
About to upgrade:

 * Upgrade: Realism Overhaul v14.23.0.0 to v15.0.1.0 (github.com, 18.1 MiB)

Continue? [Y/n]
```

### Suggestion

Probably create some flag which would allow doing more complicated depsolving to upgrade only to the latest compatible versions with all over currently installed mods.

### Alternatives

_No response_

### Additional context

I'm not sure how to do this with `--all` and what the correct workflow should be.

It may be possible that old versions of RP-1 1.x didn't have the constraint on <= v15.99 -- I'm not sure if all the RP-1 metadata got retconned or not to have that constraint added.

But certainly the situation could happen where CKAN decides to pick a lower version of RP-1 without the constraint and it installs RealismOverhaul v16.x, instead of the other way around and picking the latest RP-1 and an earlier version of RealismOverhaul.

I don't know exactly what to suggest about the ambiguity in that case, but I'd mostly be happy if I could upgrade RP-1 in isolation and thereby disallow downgrades of that mod, then it would make it easier on the depsolver to find the latest non-conflicting version of RealismOverhaul.

I know that this is how bundler/rubygems works (also that it is somewhat difficult to implement and this is probably getting close to where depsolvers become NP-complete)

### LINKED ISSUE #3945: [Feature]: Latest compatible version should be selected when installing a new mod where there are version conflicts with already installed mods
### Problem

1. Install Grannus Expansion Pack
2. try to install scatterer

Result:
```
Inconsistencies were found:
* Scatterer 3:v0.0838 conflicts with GrannusExpansionPack 1.2.8
```

Further if you try to install the latest scatterer from the versions tab, you get:

image

Which is just all sorts of weird.

### Suggestion

This sounds a bit like https://github.com/KSP-CKAN/CKAN/issues/3849 and might be the same issue at the core.

When trying to install new mods from the main mod list (not the versions tab since obviously that is a bit more direct) and the latest version is in conflict with whatever is currently installed, it should find the latest version that *doesn't* conflict and install that instead.

### Alternatives

_No response_

### Additional context

_No response_

HebaruSan:
I've been sitting on these changes for a while because I wasn't quite happy with the implementation, but I think it's about as good as it's going to get for now, and I want to get it off my plate to focus on other issues.

## Problem

If you install a mod that has a version-specific dependency on an older version of another mod, CKAN will appear to treat that other mod as upgradeable, but the upgrade will fail if attempted. Intermediate versions that could satisfy all requirements will not be discovered automatically, but may be selected manually by the user.

E.g., `RP-0` recently did this:

```yaml
depends:
  - name: RealismOverhaul
    max_version: v15.99.0.0
```

Versions after that are meant to be used only with `RP-1`. But `RealismOverhaul` will still show up as upgradeable, and trying to upgrade it will fail.

## Cause

- `AvailableModule.DependsAndConflictsOK` doesn't check version requirements of other modules that depend on the given module
- In general, mod upgradeability is treated as a property of the individual mod, when it really needs to be a collective property of groups of mods that are installed or being installed.

## Changes

- Now reverse dependencies of installed mods are considered when looking for available versions
- `IRegistryQuerier.HasUpdate` now accepts a list of other modules that are used to limit the compatible versions of the given mod
- A new function `IRegistryQuerier.CheckUpgradeable` is created to provide more complex upgrading logic; it first finds the latest available versions of all installed mods, then uses their relationship constraints to roll back one after another until a consistent changeset is generated. This function is now used by:
  - GUI's upgrade checkboxes, via setting `GUIMod.HasUpdate` from outside `GUIMod`
  - The changeset is updated based on what upgrades are actually allowed
  - `ckan upgrade --all`
  - `ckan list`'s status characters
- Mods with a "held" label are treated as not upgradeable, so they're excluded from that initial step of looking for the latest versions. When you add or remove labels or edit a label, we recalculate which upgrade checkboxes are shown.
- Several tests are added to ensure `ckan upgrade` works as intended

Fixes #3849.
Fixes #3945.

### DIFF
--- a/Cmdline/Action/List.cs
+++ b/Cmdline/Action/List.cs
@@ -1,6 +1,7 @@
 using System;
 using System.IO;
 using System.Collections.Generic;
+using System.Linq;
 
 using log4net;
 
@@ -53,6 +54,11 @@ namespace CKAN.CmdLine
             if (exportFileType == null)
             {
                 var installed = new SortedDictionary<string, ModuleVersion>(registry.Installed());
+                var upgradeable = registry
+                                  .CheckUpgradeable(instance.VersionCriteria(), new HashSet<string>())
+                                  [true]
+                                  .Select(m => m.identifier)
+                                  .ToHashSet();
 
                 foreach (KeyValuePair<string, ModuleVersion> mod in installed)
                 {
@@ -102,7 +108,7 @@ namespace CKAN.CmdLine
                                     }
                                 }
                             }
-                            else if (latest.version.IsEqualTo(current_version) && !registry.HasUpdate(mod.Key, instance.VersionCriteria()))
+                            else if (!upgradeable.Contains(mod.Key))
                             {
                                 // Up to date
                                 log.InfoFormat("Latest {0} is {1}", mod.Key, latest.version);
@@ -119,7 +125,7 @@ namespace CKAN.CmdLine
                                     }
                                 }
                             }
-                            else if (latest.version.IsGreaterThan(mod.Value) || registry.HasUpdate(mod.Key, instance.VersionCriteria()))
+                            else
                             {
                                 // Upgradable
                                 bullet = "^";
--- a/Cmdline/Action/Prompt.cs
+++ b/Cmdline/Action/Prompt.cs
@@ -202,7 +202,7 @@ namespace CKAN.CmdLine
             CKAN.GameInstance inst = MainClass.GetGameInstance(manager);
             var registry = RegistryManager.Instance(inst, repoData).registry;
             return registry.Installed(false, false)
-                           .Select(kvp => kvp.Key)
+                           .Keys
                            .Where(ident => ident.StartsWith(prefix, StringComparison.InvariantCultureIgnoreCase)
                                            && !registry.GetInstalledVersion(ident).IsDLC)
                            .ToArray();
@@ -214,7 +214,7 @@ namespace CKAN.CmdLine
 
         private string[] GetGameInstances(string prefix)
             => manager.Instances
-                      .Select(kvp => kvp.Key)
+                      .Keys
                       .Where(ident => ident.StartsWith(prefix, StringComparison.InvariantCultureIgnoreCase))
                       .ToArray();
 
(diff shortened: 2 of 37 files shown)


HebaruSan:
Hi @lamont-granquist, @NathanKell, and @JonnyOThan, would you have time to test these changes in the next few days? The build is under the Artifacts dropdown on the Checks tab. I'd appreciate knowing about any problems you discover.

NathanKell:
Oooh, that's cool! Sadly I'm very swamped with non-KSP stuff atm so I make no promises. :( (Especially sucks given how desirable I find this, and how grateful I am!)

Gonna ping @siimav and @capkirk123 as other folks who might be able to provide help here.

HebaruSan:
This has a couple of bugs that need to be addressed before merge...

HebaruSan:
Finally got this into a state I'm happy with. The `ManageMods.MarkModForInstall` / `GUIMod.SetInstallChecked` / `ManageMods.MarkModForUpdate` / `GUIMod.SetUpgradeChecked` mess is eliminated to reduce coupling between the data model and the UI. `ManageMods` now listens for property change notifications of `GUIMod.SelectedMod` like the Versions tab already did, so the universal way to tweak the changeset is now to set that property. (The exception is replacing, which isn't represented in the data model at all.)

We did not manage to assemble a volunteer testing corps for this, so might as well merge now...

PR #4044, opened 2024-02-28
Support Windows KSP1 instances on Linux

### LINKED ISSUE #4036: [Bug]: Number must be either non-negative and less than or equal to Int32.MaxValue

### Operating System

Linux

### CKAN Version

1.34.4

### Games

KSP 1

### Game Version

KSP v1.12.5

_No response_

### Describe the bug

When you select and start CKAN, it crashes. I tried to try the Linux version, but it doesn't see ksp.exe

### Steps to reproduce

Run CKAN

### Relevant log output

```
System.ArgumentOutOfRangeException: Number must be either non-negative and less than or equal to Int32.MaxValue or -1.
Parameter name: timeout
at System.Threading.WaitHandle.WaitOne (System.TimeSpan timeout, System.Boolean exitContext) [0x00026] in <006aeda5450e44e2a7c0b3f3e62d74fd>:0
at System.Transactions.Transaction.DoPreparePhase () [0x00055] in <321ca8c1b88347dcb99ed658286b1b33>:0
at System.Transactions.Transaction.DoCommit () [0x00076] in <321ca8c1b88347dcb99ed658286b1b33>:0
at System.Transactions.Transaction.CommitInternal () [0x00022] in <321ca8c1b88347dcb99ed658286b1b33>:0
```

### LINKED ISSUE #4038: [Bug]: [LINUX]: " *directory* is not a valid game directory. " message when trying to create new game instance

- [x] I have searched the existing issues

### Operating System

Linux Mint

### CKAN Version

1.34.4

### Games

KSP 1

### Game Version

1.12.5.3190 (With Proton)

No.

### Describe the bug

When I run CKAN, a window pops up for managing game instances. I press the "New Game Instance" button, followed by "Add instance to CKAN". This opens up a window, where I can manually go to the correct directory, the one that contains the game's contents. Unfortunately this window doesn't let me select (or see for that matter) any non folder files, which makes me unable to select the correct file. Attempting to go to the right directory and write the filename in the correct box results in the error message "*directory* is not a valid game directory.". I have attempted to use the CKAN Console UI to circumvent this but I get the "Path does not correspond to a game folder!" error message.
Screenshot from 2024-02-18 15-31-41
Screenshot from 2024-02-18 15-29-47

### Steps to reproduce

*Open CKAN
*Attempt to create a new instance
*Go to the game directory

### Relevant log output

_No response_

### LINKED ISSUE #4041: [Feature]: Allow using Proton instance for significantly improved speed over Linux native
### Problem

In #4038 it was brought up that on Linux, CKAN won't find instances under Proton, but the issue was closed because there's a native version, however, the GE-Proton version runs _significantly_ faster than the native version, by up to 1.5x*, and I would vastly prefer to use it for heavy graphical modding. 

### Suggestion

Would detecting/using Proton instances be considered in light of this?

### Alternatives

_No response_

### Additional context

*When heavily modded, on one of my space stations I saw about 67-71 fps with GE-Proton 8-32 vs ~53 FPS with Linux Native. Mods included EVE Redux, Scatterer, SVE, Parallax, Planetshine, and TUFX

HebaruSan:
## Background

Originally, CKAN looked for `buildID.txt` and `buildID64.txt` to identify KSP1 instances. For KSP2, no equivalent file exists (but luckily we won't need it anymore, see #4034), so we look for the game executable instead.

It was fairly common for confused KSP1 users to ask how to add an instance, because they didn't know what `buildID.txt` was and so clicking it was a non-obvious action. Meanwhile, nobody seemed the least bit confused by choosing the KSP2 executable. So in #3964 we changed the prompting for KSP1 to also use the executable, which seems to have succeeded in improving usability so far.

Later, a Linux user submitted #3984 to _run_ the KSP2 Windows executable from Linux, which apparently can work thanks to Proton, which somehow can reach out from inside however Steam installs it to launch EXE files similar to how Mono does (I'm still not sure about the details of this).

## Motivation

Some KSP1 users on Linux also prefer to run via Proton, because it performs better for them, or because DirectX makes some graphics-heavy mods like Parallax work better than they do on OpenGL. Currently such users have to create a `KSP.x86_64` file to be able to add such an instance to CKAN, which is a very non-obvious workaround, and they also have to configure the command line manually.

## Changes

Now CKAN will look for Windows KSP1 executables in addition to the Linux ones when running on Linux. This applies both to adding the instance and to launching the game. Whichever executable is present in the folder will be treated as the right one to run, as per #3964.

Fixes #4036.
Fixes #4038.
Fixes #4041.

### DIFF
--- a/Core/Games/KerbalSpaceProgram.cs
+++ b/Core/Games/KerbalSpaceProgram.cs
@@ -222,7 +222,9 @@ namespace CKAN.Games.KerbalSpaceProgram
         public string[] InstanceAnchorFiles =>
             // KSP.app is a directory :(
               Platform.IsMac     ? new string[] { "buildID64.txt", "buildID.txt" }
-            : Platform.IsUnix    ? new string[] { "KSP.x86_64",    "KSP.x86" }
+            : Platform.IsUnix    ? new string[] { "KSP.x86_64",    "KSP.x86",
+                                                  // Windows EXEs via Proton on Linux
+                                                  "KSP_x64.exe",   "KSP.exe" }
             :                      new string[] { "KSP_x64.exe",   "KSP.exe" };
 
         public Uri DefaultRepositoryURL => new Uri("https://github.com/KSP-CKAN/CKAN-meta/archive/master.tar.gz");


OboTheHobo:
I downloaded the nightly build to try this and it didn't work perfectly, the game launched but it seemed to be using the default wine prefix on my computer rather than proton. The solution was to put the following in a bash script and get ckan to launch that (for some reason it didn't work if I just put this line in the CKAN start line):

`STEAM_COMPAT_CLIENT_INSTALL_PATH=/path/to/steam/install/Steam/ STEAM_COMPAT_DATA_PATH=~/.proton/ /path/to/steamapps/common/Proton\ -\ Experimental/proton run ./KSP_x64.exe `

Not sure if this can be easily implemented into the program given that the paths to all those are highly dependent on the system. Still, thought I'd add this here in case someone else wanted help getting ckan to work with proton.

HebaruSan:
Hi @OboTheHobo, thanks for that contribution. That explains my puzzlement about how simply running the KSP1 Windows executable could possibly connect up with Steam's Proton setup (i.e., _it doesn't_).

Would you mind putting that info somewhere on [the CKAN wiki](https://github.com/KSP-CKAN/CKAN/wiki) to make it more findable? Our pages about Linux are all distro-specific, so there may not be a perfect spot, but something would still be better than nothing, and we can always move it around later if needed.

> (for some reason it didn't work if I just put this line in the CKAN start line)

Yes, setting environment variables wouldn't work because Mono's `Process.Start` call just starts the given executable without starting a shell first. You might have more success if you try `bash -c "your command"` to force it to start a shell within which those variables could be set.

OboTheHobo:
> That explains my puzzlement about how simply running the KSP1 Windows executable could possibly connect up with Steam's Proton setup (i.e., _it doesn't_).

I'm thinking it would if proton is the only version of wine installed, it might. For me it started with the default wineprefix from installing wine normally, but I imagine if proton is the only version there is then it could default for that.

> Also, the dev build should have an option in the Play button's dropdown menu to launch via Steam:

I don't see this option on my end (is the dev build different from nightly?) but I would think this wouldn't work when having multiple instances of the game installed. Still, nice to have another option.

As for the wiki, do you think maybe I should just make a new page for proton? seems a little unnecessary for a relatively small workaround, but as you said the linux guides are all spread out across different pages for distros so it might be the best option to just make a new page.

OboTheHobo:
Or, I could also add it to the troubleshooting page maybe, since the linux pages are more for installation than setup.

HebaruSan:
> > Also, the dev build should have an option in the Play button's dropdown menu to launch via Steam:
> 
> I don't see this option on my end (is the dev build different from nightly?) but I would think this wouldn't work when having multiple instances of the game installed. Still, nice to have another option.

Ahh yes, that would only appear for instances that are actually _in_ Steam. If you made a copy of the game dir, then it would only show the executable option by default for that instance. You could try Steam's "non-Steam game" option, which the dev build should detect:

image

But I guess in that case you'd still have to tell _Steam_ how to run the game...?

> As for the wiki, do you think maybe I should just make a new page for proton?
> Or, I could also add it to the troubleshooting page maybe, since the linux pages are more for installation than setup.

Either of those options sounds workable to me. Thanks for being willing to help out!

OboTheHobo:
Alright, I added a section to the troubleshooting page, hopefully it's clear in what to do. Glad I could help!

PR #4051, opened 2024-03-06
Use better console hiding API

### LINKED ISSUE #3902: [Bug] A Windows 11 update broke console hiding
### Operating System

Windows 11 22H2 or later

### CKAN Version

v1.33.3

### Game Version

1.12.5

## Problem

Multiple users on Discord reported that the terminal window was appearing and not going away on Windows 11 (and the latest one stuck around long enough for us to investigate). Turns out a Windows 11 update broke `GUI.Util.HideConsoleWindow` by swapping out the default terminal app to one that doesn't support `GetConsoleWindow`:

- <https://support.microsoft.com/en-us/windows/command-prompt-and-windows-powershell-for-windows-11-6453ce98-da91-476f-8651-5c14d5777c20>
- <https://stackoverflow.com/questions/74384519/hiding-the-new-windows-terminal>

Apparently it was deprecated for a while:

<https://learn.microsoft.com/en-us/windows/console/getconsolewindow>

image
image

`ckan.exe` is built as a console app because it has command line functionality built-in, so Windows insists on opening a terminal for it. Linux has the ability to launch a console app in the background:

https://github.com/KSP-CKAN/CKAN/blob/ff291b1820e859069594cdc9bd44c8bbb1a8a328/debian/ckan.desktop#L7

### Steps to reproduce

- Install Windows 11 22H2 or later
- Launch `ckan.exe`
- The terminal remains visible instead of hiding

## Suggestions

- The user on Discord was satisfied with a shortcut that launches the terminal window minimized. Could we do that by default somehow?
- Build separate EXEs for GUI and console? Would be a loss of generality that has served CKAN well for many years. More confusion about what to download.
- Build only the GUI app? Would completely break the metadata validator action and be a big disruption for anyone still using cmdline.
- Build a GUI app that can still read and write from the terminal if launched from one? Don't know if this is possible, would have to investigate.

HebaruSan:
## Background

Windows 11 recently introduced a new virtual terminal to replace the Windows console. It behaves much more like standard Unix terminals:

- <https://learn.microsoft.com/en-us/windows/console/classic-vs-vt>

## Problems

- If you open a console (old style) in Windows and then run ckan.exe without the `--show-console` flag (and none of the CmdLine commands like `list` or `install`), your console windows closes. This is not typical behavior for a command that you launch from a console and can be very surprising.
- If you double click ckan.exe in Windows 11, the virtual terminal stays open the entire time you're using the GUI

## Causes

Since ckan.exe is both a console application and a GUI application, Windows always automatically opens a console for it at launch. This is not desirable when running in GUI mode, so the GUI closes its console by finding it with `GetConsoleWindow` and then hiding it with `ShowWindow`.

- This approach is too aggressive because it doesn't care whether that console is still in use by other applications
- The new virtual terminal doesn't support `GetConsoleWindow`, so the whole approach breaks down anyway

## Changes

Now we use [`FreeConsole`](https://learn.microsoft.com/en-us/windows/console/freeconsole) instead of `ShowWindow` and `GetConsoleWindow`. This API just _detaches_ the current process from its console or terminal, so it won't aggressively close consoles that are still in use by other processes (such as a shell), but will auto-close any terminal that is _only_ associated with ckan.exe. It is also supported by the new virtual terminals (see microsoft/terminal#14544), so console hiding should now work on Win11.

Fixes #3902.

### DIFF
--- a/GUI/Util.cs
+++ b/GUI/Util.cs
@@ -55,17 +55,14 @@ namespace CKAN.GUI
 
         // hides the console window on windows
         // useful when running the GUI
-        [DllImport("kernel32.dll")]
-        private static extern IntPtr GetConsoleWindow();
-
-        [DllImport("user32.dll")]
-        private static extern bool ShowWindow(IntPtr hWnd, int nCmdShow);
+        [DllImport("kernel32.dll", SetLastError=true)]
+        private static extern int FreeConsole();
 
         public static void HideConsoleWindow()
         {
             if (Platform.IsWindows)
             {
-                ShowWindow(GetConsoleWindow(), 0);
+                FreeConsole();
             }
         }
 


errolkokbas:
I'm on win11, tested the beta is it is correctly killing the terminal window again. Thank you.

PR #4076, opened 2024-04-10
Add OpenSUSE zypper command for rpm repo

pastadudes:
added zypper command, tested and works just like normal
im not really good at markdown so feel free to make changes


HebaruSan (approved):
Hi @pastadudes, thanks for the contribution!

PR #4082, opened 2024-05-04
Refactor + Modernise Actions

techman83:
This is a complete refactor and overhaul of our actions, and either updating or changing any outdated actions we have used within our workflows.

### Building
The build workflow is now a self contained, and runs on master or via 'workflow_call'. Making a single source of truth for building CKAN/NetKAN.

Note: The build on master is intentional, I expect there we can do a small amount of work to enable [dotnet caching](https://github.com/actions/setup-dotnet?tab=readme-ov-file#caching-nuget-packages). For a cache to be available on branches, it *must* have been previously built on the default branch.

### Testing
There is a new test file that calls build first. It is unnecessary to build inside the Mono containers, as we still use dotnet to build. This however ensures we are keeping the same validation as before. We can likely look to considering if this is still required.

### Deploy
Deployment now calls build first, followed by versioning, and then the rest of the steps.
- All S3 actions now use the official AWS tools
- Only RPM needs an extra tool outside what comes by default

Example here: https://github.com/KSP-CKAN/CKAN/actions/runs/8948785797
Example with containers (fails as I omitted auth): https://github.com/KSP-CKAN/CKAN/actions/runs/8948756697/job/24582270638

### Release
Release has gone through a similar tidy up to deploy. Arguably a shared workflow could be built for this, but it felt prudent to keep it separate for now. When signing is fully available we can take a look at it then.
- All S3 actions now use the official AWS tools
- Only RPM + OSX needs an extra tool outside what comes by default

Example here: https://github.com/KSP-CKAN/CKAN/actions/runs/8936371852

## Other things of note
- Signing is waiting upon verification of the GitHub Action, it'll be a small amount of work to switch to the signed binaries
- I used a `/test` path for all S3 uploads, which look to have worked in the same manner as the previous action.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Setup .NET Core
-        uses: actions/setup-dotnet@v3
+        uses: actions/setup-dotnet@v4
         with:
           dotnet-version: '7'
       - name: Install runtime dependencies
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
           apt-get install -y apt-transport-https
       - name: Installing checkout/build dependencies
         run: apt-get update && apt-get install -y git
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Setup .NET Core
         uses: actions/setup-dotnet@v3
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,10 +12,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Setup .NET Core
-        uses: actions/setup-dotnet@v4
-        with:
-          dotnet-version: '7'
       - name: Build ckan.exe and netkan.exe
         run: ./build --configuration=${{ inputs.configuration }}
       - uses: actions/upload-artifact@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,101 +1,37 @@
 name: Build CKAN + NetKAN
 
 on:
-  push:
-    branches:
-      - master
-  pull_request:
-    types:
-      - opened
-      - synchronize
-      - reopened
+  workflow_call:
+    inputs:
+      configuration:
+        type: string
+        default: Debug
 
 jobs:
   build:
     runs-on: ubuntu-latest
-
-    strategy:
-      fail-fast: false
-      matrix:
-        mono:
-          - latest
-          - '6.12'
-          - '6.10'
-          - '6.8'
-        configuration:
-          - Debug
-          - Release
-
-    container:
-      image: mono:${{ matrix.mono }}
-
     steps:
-      - name: Adding HTTPS support to APT for old Mono images
-        run: |
-          apt-get update || true
-          apt-get install -y apt-transport-https
-      - name: Installing checkout/build dependencies
-        run: apt-get update && apt-get install -y git
       - uses: actions/checkout@v4
-
       - name: Setup .NET Core
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: '7'
-      - name: Install runtime dependencies
-        run: apt-get install -y xvfb
-      - name: Restore cache for _build/tools
-        uses: actions/cache@v3
-        with:
-          path: _build/tools
-          key: build-tools-${{ hashFiles('build', 'build.ps1', 'build.cake') }}
-      - name: Restore cache for _build/cake
-        uses: actions/cache@v3
-        with:
-          path: _build/cake
-          key: build-cake-${{ hashFiles('build.cake') }}
-      - name: Restore cache for _build/lib/nuget
-        uses: actions/cache@v3
-        with:
-          path: _build/lib/nuget
-          key: nuget-oldref-modules-${{ hashFiles('**/packages.config') }}-${{ hashFiles('**/*.csproj') }}
-      - name: Restore cache for ~/.nuget/packages
-        uses: actions/cache@v3
-        with:
-          path: ~/.nuget/packages
-          key: nuget-packref-modules-${{ hashFiles('**/packages.config') }}-${{ hashFiles('**/*.csproj') }}
-
       - name: Build ckan.exe and netkan.exe
-        run: ./build --configuration=${{ matrix.configuration}}
-      - name: Run tests
-        run: xvfb-run ./build test+only --configuration=${{ matrix.configuration }} --where="Category!=FlakyNetwork"
-      - name: Run inflator container smoke test
-        run: |
-          cd _build
-          curl -fsSL https://get.docker.com -o get-docker.sh
-          sh get-docker.sh
-          docker build --tag inflator --file ../Dockerfile.netkan .
-          docker run --rm --name inflator --entrypoint /bin/bash inflator -c "
-            mono netkan.exe https://raw.githubusercontent.com/KSP-CKAN/NetKAN/master/NetKAN/ZeroMiniAVC.netkan
-          "
-        if: matrix.configuration == 'release'
-
+        run: ./build --configuration=${{ inputs.configuration }}
+      - uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.configuration }}-repack-unsigned
+          path: _build/repack/
+          retention-days: 7
+      - uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.configuration }}-out
+          path: _build/out/
+          retention-days: 1
       - name: Upload ckan.exe artifact
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: ckan.exe
           path: _build/repack/Release/ckan.exe
           retention-days: 7
-        if: matrix.configuration == 'release' && matrix.mono == 'latest' && github.event_name == 'pull_request'
-
-      - name: Send Discord Notification
-        env:
-          JOB_STATUS: ${{ job.status }}
-          WEBHOOK_URL: ${{ secrets.DISCORD_WEBHOOK }}
-          HOOK_OS_NAME: ${{ runner.os }}
-          WORKFLOW_NAME: ${{ github.workflow }}
-        if: ${{ failure() && env.WEBHOOK_URL }}
-        run: |
-          git clone https://github.com/DiscordHooks/github-actions-discord-webhook.git webhook
-          bash webhook/send.sh $JOB_STATUS $WEBHOOK_URL
-        shell: bash
+        if: inputs.configuration == 'Release'
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -24,7 +24,7 @@ jobs:
           apt-get update
           apt-get install -y --no-install-recommends \
             git make sed gzip fakeroot lintian dpkg-dev gpg gpg-agent createrepo python3-pip
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Check version
         id: check_version
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -37,7 +37,7 @@ jobs:
           fi
 
       - name: Setup .NET Core
-        uses: actions/setup-dotnet@v3
+        uses: actions/setup-dotnet@v4
         with:
           dotnet-version: '7'
       - name: Installing runtime dependencies
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -11,107 +11,162 @@ on:
 
 concurrency: deploy
 
+env:
+  AWS_S3_BUCKET: ksp-ckan
+
 jobs:
-  deploy:
+  test-release:
+    uses: ./.github/workflows/test.yml
+
+  smoke-inflator:
+    uses: ./.github/workflows/smoke.yml
+
+  upload-release-s3:
+    needs:
+      - test-release
+      - smoke-inflator
     runs-on: ubuntu-latest
-
-    container:
-      image: mono:latest
-
+    outputs:
+      odd-build: ${{ steps.check-version.outputs.odd-build }}
+      credentials: ${{ steps.credentials.outputs.credentials }}
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
     steps:
-      - name: Installing checkout/build dependencies
-        run: |
-          apt-get update
-          apt-get install -y --no-install-recommends \
-            git make sed gzip fakeroot lintian dpkg-dev gpg gpg-agent createrepo python3-pip
       - uses: actions/checkout@v4
-
       - name: Check version
-        id: check_version
+        id: check-version
         shell: bash
         run: |
           VERSION=$(egrep '^\s*\#\#\s+v.*$' CHANGELOG.md | head -1 | sed -e 's/^\s*\#\#\s\+v//' -e 's/-.*$//')
           if [[ $VERSION =~ [13579]$ ]]
           then
-            echo 'odd_build=true' >> $GITHUB_OUTPUT
+            echo 'odd-build=true' >> $GITHUB_OUTPUT
           fi
+      - uses: actions/download-artifact@v4
+        with:
+          name: Release-repack-unsigned
+          path: _build/repack/
+      - name: Credentials
+        id: credentials
+        run: echo 'credentials=false' >> $GITHUB_OUTPUT
+        if: ${{ env.AWS_ACCESS_KEY_ID && env.AWS_SECRET_ACCESS_KEY }}
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-access-key-id: ${{ env.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ env.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+        if: steps.credentials.outputs.credentials
+      - name: Push deb to S3
+        run: aws s3 sync _build/repack/Release s3://${AWS_S3_BUCKET} --follow-symlinks
+        if: steps.credentials.outputs.credentials
 
+  upload-deb:
+    needs: upload-release-s3
+    runs-on: ubuntu-latest
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    if: needs.upload-release-s3.outputs.odd-build && needs.upload-release-s3.outputs.credentials
+    steps:
+      - uses: actions/checkout@v4
       - name: Setup .NET Core
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: '7'
-      - name: Installing runtime dependencies
-        run: apt-get install -y xvfb
-
-      - name: Restore cache for _build/tools
-        uses: actions/cache@v1
+      - uses: actions/download-artifact@v4
         with:
-          path: _build/tools
-          key: build-tools-${{ hashFiles('build', 'build.ps1', 'build.cake') }}
-      - name: Restore cache for _build/cake
-        uses: actions/cache@v1
-        with:
-          path: _build/cake
-          key: build-cake-${{ hashFiles('build.cake') }}
-      - name: Restore cache for _build/lib/nuget
-        uses: actions/cache@v1
-        with:
-          path: _build/lib/nuget
-          key: nuget-oldref-modules-${{ hashFiles('**/packages.config') }}-${{ hashFiles('**/*.csproj') }}
-      - name: Restore cache for ~/.nuget/packages
-        uses: actions/cache@v1
-        with:
-          path: ~/.nuget/packages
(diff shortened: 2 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -4,98 +4,32 @@ on:
...
-          sh get-docker.sh
-          docker build --tag inflator --file ../Dockerfile.netkan .
-          docker run --rm --name inflator --entrypoint /bin/bash inflator -c "
-            mono netkan.exe https://raw.githubusercontent.com/KSP-CKAN/NetKAN/master/NetKAN/ZeroMiniAVC.netkan
-          "
-        if: matrix.configuration == 'release'
-
-      - name: Upload ckan.exe artifact
-        uses: actions/upload-artifact@v3
+        run: xvfb-run ./build test+only --configuration=${{ inputs.configuration }} --where="Category!=FlakyNetwork"
```
HebaruSan, line 25: I feel like `build.yml` should no longer run the tests here if we have an explicit `test.tml` file now.

.github/workflows/build.yml
```diff
@@ -4,98 +4,32 @@ on:
...
-      - name: Installing checkout/build dependencies
-        run: apt-get update && apt-get install -y git
-      - uses: actions/checkout@v3
-
+      - uses: actions/checkout@v4
       - name: Setup .NET Core
-        uses: actions/setup-dotnet@v3
+        uses: actions/setup-dotnet@v4
         with:
           dotnet-version: '7'
```
HebaruSan, line 21: Any thoughts on removing this "Setup .NET Core" step, since we discovered that the dotnet tools are already installed in the container by default?
(and ".NET Core" isn't even the name anymore)

techman83: I tried that first and it complained. And whilst they are installed in the actions container by default, they *not* installed in the mono container.

.github/workflows/build.yml
techman83: This tests it within our build step, and ensures that any time we need to consume it elsewhere we don't need to remember to test it.

What we could do is setting this to only run when the configuration is `Release`. As any `Debug` is expected to be run via the test workflow.

techman83: Disabling it shaves off about 30 seconds, on balance I'm not sure if it's worth it. Whilst potentially unnecessary, it would shortcut the build by failing *before* doing the extended tests. Along with providing a strong guarantee that a test was run *before* being used for the release/deploy assets.

.github/workflows/build.yml
HebaruSan: What was the complaint? Remember, it's a no-op as is:

image

techman83: Lemme push to re-check the assumption, it might be a path / job setup thing. If it's basically a no-op there are no consequences for leaving it in, though it does add a point of maintenance, so if it's just pathing maybe we can export an path or something.

techman83: Pathing probably
```
Run xvfb-run ./build test+only --configuration=Debug --where="Category!=FlakyNetwork"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 8224k  100 8224k    0     0   125M      0 --:--:-- --:--:-- --:--:--  125M
./build: 43: ./build: dotnet: not found
./build: 64: ./build: dotnet: not found
```

techman83: It's possible these tests aren't really super relevant, but it's definitely not a no-op inside the containers

https://github.com/KSP-CKAN/CKAN/actions/runs/8949111709/job/24583068605#step:7:1

What are we testing for inside these containers? Is it even relevant? Can we actually test different mono versions?

Happy to make an adjustment, but when it came to the test cases, I mostly tidied them up and brought them across with the existing configuration.

HebaruSan: Wait. Why is a change to `build.yml` affecting `test.yml`? I thought the only things that transfer over were the artifacts?

HebaruSan: ... oh! You updated `test.yml` instead! Yeah, that's definitely not what I was suggesting. I'll fix that...

.github/workflows/test.yml
```diff
@@ -0,0 +1,97 @@
...
+          docker run --rm --name inflator --entrypoint /bin/bash inflator -c "
+            mono netkan.exe https://raw.githubusercontent.com/KSP-CKAN/NetKAN/master/NetKAN/ZeroMiniAVC.netkan
+          "
+      - name: Upload ckan.exe artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: ckan.exe
+          path: _build/repack/Release/ckan.exe
+          retention-days: 7
+        if: github.event_name == 'pull_request'
```
HebaruSan, line 78: Uploading a ckan.exe artifact isn't logically part of smoke testing. Should this be relocated or just removed?

HebaruSan:
Regarding the current test failure, I think something may actually be up with GitHub's cert. I get this locally:

```
$ netkan.exe https://raw.githubusercontent.com/KSP-CKAN/NetKAN/master/NetKAN/ZeroMiniAVC.netkan
FATAL CKAN.NetKAN.Program - The remote certificate is invalid according to the validation procedure.
```

.github/workflows/build.yml
HebaruSan: There isn't really a point to having `test.yml` at all if this stays here. Any test failure will be caught here before `test.yml` runs; as far as I know, we've never had tests that only pass on some versions of Mono, so `test.yml` will not achieve anything for us.

The point of handling multiple versions of Mono was the **compilation/build** step, because it's possible that some build techniques may fail on older Mono versions. We're losing that now, which is fine with me because even the newest versions of Mono are old now, but when I see `build.yml` and `test.yml`, it seems pretty clear where the testing should happen.

techman83: I don't fully agree with you here, but I'll ponder, because I get your point. I will counter though, part of it is to ensure that the built output (that we consume elsewhere) is guaranteed to have had tests run. IMO, those CPU cycles are worth the cost to have that guarantee, that we can depend that the artifacts are built successfully. Because I can assure, I have experienced the assumption that they passed, and lost way more time than the what an extra 30 seconds in a workflow has (4 hours time = 480 x 30 seconds).

Personally even if it's belts and bracers, I'd prefer to have the extra step. It's going to pretty obvious that the build failed either way, and we will know the build failed a step earlier than we would have otherwise.

HebaruSan: Let's just get rid of `test.yml`, then, if you want to **always** run the tests after building. It no longer serves a purpose.

.github/workflows/test.yml
techman83: Yeah, valid. We could break that into a separate job or drop it entirely. I was carrying over the use case from our previous workflow (having a build of ckan automatically available in a PR).

.github/workflows/build.yml
techman83: So no tests in the mono containers? If we're fine to drop them, that's ok with me. To me, those tests are for the artifacts we need elsewhere, rather extended testing, so it made sense to break them out separately and depend on a pre-validated build.

HebaruSan: See above; Mono version mattered for _**compiling/building**_ CKAN, not the tests.

.github/workflows/build.yml
HebaruSan: Looks like this came back in the latest changes. Note to self to purge it again.

.github/workflows/deploy.yml
```diff
@@ -11,185 +11,221 @@ on:
...
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    if: needs.upload-release-s3.outputs.odd-build && needs.upload-release-s3.outputs.credentials
+    steps:
+      - uses: actions/checkout@v4
       - name: Setup .NET Core
-        uses: actions/setup-dotnet@v3
+        uses: actions/setup-dotnet@v4
         with:
           dotnet-version: '7'
```
HebaruSan, line 77: Looks like another unnecessary .NET setup step that we can remove.

.github/workflows/deploy.yml
```diff
@@ -11,185 +11,221 @@ on:
...
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    if: needs.upload-release-s3.outputs.odd-build && needs.upload-release-s3.outputs.credentials
+    steps:
+      - uses: actions/checkout@v4
+      - name: Installing rpm build dependencies
+        run: sudo apt-get install -y createrepo-c
+      - name: Setup .NET Core
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '7'
```
HebaruSan, line 124: Another .NET setup to remove.

.github/workflows/deploy.yml
```diff
@@ -11,185 +11,221 @@ on:
...
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    if: needs.upload-release-s3.outputs.credentials
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup .NET Core
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '7'
```
HebaruSan, line 165: Another one!

.github/workflows/deploy.yml
```diff
@@ -11,185 +11,221 @@ on:
...
+      - test-release
+      - smoke-inflator
+    runs-on: ubuntu-latest
+    if: false
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup .NET Core
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '7'
```
HebaruSan, line 194: There are a lot of these!

.github/workflows/build.yml
HebaruSan: Done.

.github/workflows/deploy.yml
```diff
@@ -11,185 +11,205 @@ on:
...
         run: |
           echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+          cp -v _build/repack/Release/netkan.exe _build/.
           ./build docker-inflator --exclusive
+
+  upload-metadata-tester:
+    needs:
+      - test-release
+      - smoke-inflator
+    runs-on: ubuntu-latest
```
HebaruSan, line 175: Why does `upload-inflator` depend on `upload-release-s3` and this one depends on the test steps? The two docker images are pretty similar, so I would expect the dependencies to be the same.

.github/workflows/deploy.yml
```diff
@@ -11,185 +11,205 @@ on:
...
+        run: aws s3 sync _build/rpm/repo s3://${AWS_S3_BUCKET}/rpm/nightly --follow-symlinks
+
+  upload-inflator:
+    needs: upload-release-s3
+    runs-on: ubuntu-latest
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    if: needs.upload-release-s3.outputs.credentials
+    steps:
```
HebaruSan, line 152: Can't the Inflator can be built and uploaded without AWS credentials?

.github/workflows/deploy.yml
```diff
@@ -11,185 +11,205 @@ on:
...
           echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+          cp -v _build/repack/Release/netkan.exe _build/.
           ./build docker-inflator --exclusive
+
+  upload-metadata-tester:
+    needs:
+      - test-release
+      - smoke-inflator
+    runs-on: ubuntu-latest
+    if: false
```
HebaruSan, line 176: Is this some leftover debugging logic? Presumably we would want this step to run some of the time.

.github/workflows/deploy.yml
```diff
@@ -11,185 +11,205 @@ on:
...
+          path: _build/repack/
       - name: Generate metadata tester Docker image and publish to Hub
         env:
           DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
         if: ${{ env.DOCKERHUB_USERNAME && env.DOCKERHUB_PASSWORD }}
         run: |
           echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+          cp -v _build/repack/Release/*kan.exe _build/.
           ./build docker-metadata --exclusive
```
HebaruSan, line 191: I don't think we need this. The `./build docker-metadata` command does this itself.

.github/workflows/deploy.yml
```diff
@@ -11,185 +11,205 @@ on:
...
+          name: Release-repack-unsigned
+          path: _build/repack/
       - name: Generate metadata tester Docker image and publish to Hub
         env:
           DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD }}
         if: ${{ env.DOCKERHUB_USERNAME && env.DOCKERHUB_PASSWORD }}
         run: |
           echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+          cp -v _build/repack/Release/*kan.exe _build/.
```
HebaruSan, line 190:
```suggestion
```

The `./build docker-metadata` command copies these files itself.

.github/workflows/release.yml
```diff
@@ -58,115 +106,105 @@ jobs:
...
+    steps:
+      - uses: actions/download-artifact@v4
         with:
-          upload_url: ${{ steps.release_data.outputs.upload_url }}
-          asset_path: _build/rpm/RPMS/noarch/ckan-${{ steps.release_data.outputs.rpm_version }}-1.noarch.rpm
-          asset_name: ckan-${{ steps.release_data.outputs.version }}-1.noarch.rpm
-          asset_content_type: application/x-rpm
+          name: Release-repack-unsigned
+          path: _build/repack/
+      - run: ls -lah _build/repack/
```
HebaruSan, line 182:
```suggestion
```

This looks like some debugging. Do we need it?

.github/workflows/release.yml
```diff
@@ -5,30 +5,47 @@ on:
...
+  smoke-inflator:
+    uses: ./.github/workflows/smoke.yml
 
+  sign-release:
+    runs-on: ubuntu-latest
+    needs:
+      - test-release
+      - smoke-inflator
+    outputs:
+      artifact-url: ${{steps.sign.outputs.signing-request-id }}
```
HebaruSan, line 24: I'm trying to figure out where this `artifact-url` output is used. Any clues?

.github/workflows/release.yml
```diff
@@ -58,115 +106,105 @@ jobs:
...
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  build-rpm:
+    runs-on: ubuntu-latest
+    needs:
+      - test-release
+      - smoke-inflator
+    env:
+      RPM_VERSION: ${{ needs.version.output.rpm-version }}
+    steps:
```
HebaruSan, line 131: It looks like `rpm-version` isn't set anywhere.

.github/workflows/release.yml
```diff
@@ -58,115 +106,105 @@ jobs:
...
-        uses: actions/upload-release-asset@v1.0.1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  build-rpm:
+    runs-on: ubuntu-latest
+    needs:
+      - test-release
+      - smoke-inflator
+    env:
+      RPM_VERSION: ${{ needs.version.output.rpm-version }}
```
HebaruSan, line 130: It looks like `rpm-version` isn't set anywhere.

.github/workflows/release.yml
```diff
@@ -5,30 +5,47 @@ on:
...
 
+  upload-nuget:
+    runs-on: ubuntu-latest
+    needs:
+      - test-release
+      - smoke-inflator
+    outputs:
+      artifact-url: ${{steps.sign.outputs.signing-request-id }}
+    steps:
+      - uses: actions/upload-artifact@v4
```
HebaruSan, line 45:
```suggestion
      - uses: actions/download-artifact@v4
```

Is this using the right action? I would expect downloading here, not uploading.

.github/workflows/release.yml
```diff
@@ -58,115 +106,105 @@ jobs:
...
-          asset_content_type: application/vnd.microsoft.portable-executable
+        run: gh release upload ${{ github.event.release.tag_name }} _build/repack/Release/AutoUpdater.exe
 
+  notify-discord:
+    runs-on: ubuntu-latest
+    needs:
+      - build-dmg
+      - build-deb
+      - build-rpm
+      - upload-binaries
```
HebaruSan, line 194:
```suggestion
    needs:
      - build-dmg
      - build-deb
      - build-rpm
      - upload-binaries
      - upload-nuget
```

Should `upload-nuget` be listed here so we get notified if it fails?

.github/workflows/release.yml
```diff
@@ -58,115 +106,105 @@ jobs:
...
+    env:
+      JOB_STATUS: failure
+    if: always()
+    steps:
+      - name: Set Success
+        run: echo "JOB_STATUS=success" >> $GITHUB_ENV
+        if: contains('failure', join(needs.*.result, ' ')) == false
```
HebaruSan, line 201:
```suggestion
        if: contains(join(needs.*.result, ' '), 'failure') == false
```

These parameters to `contains` are backwards. The left is the haystack, the right is the needle.

- <https://docs.github.com/en/actions/learn-github-actions/expressions#contains>

.github/workflows/deploy.yml
HebaruSan, line 176:
```suggestion
```

Is this some leftover debugging logic? Presumably we would want this step to run some of the time.

HebaruSan (changes requested):
Found a couple of things that look like they need more attention.

But I love the overall approach here, and I'm really looking forward to using these nicely refactored workflows!

.github/workflows/release.yml
techman83: I tried without the join, couldn't get it to work. This does work.

HebaruSan: No, this doesn't work, becaues the arguments are swapped. That's probably why your test without the `join` didn't work, either.

.github/workflows/release.yml
techman83: It will be used when we want to download it when we have the real certificate available.

.github/workflows/release.yml
techman83: Ahh, yes, that's leftover

.github/workflows/release.yml
```diff
@@ -5,30 +5,47 @@ on:
...
+          artifact-configuration-slug: release
+          wait-for-completion: true
 
+  upload-nuget:
+    runs-on: ubuntu-latest
+    needs:
+      - test-release
+      - smoke-inflator
+    outputs:
+      artifact-url: ${{steps.sign.outputs.signing-request-id }}
```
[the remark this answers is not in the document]
techman83: Yup, I'll clean it up!

techman83:
> Found a couple of things that look like they need more attention.
> 
> But I love the overall approach here, and I'm really looking forward to using these nicely refactored workflows!

Awesome! I note the dotnet action was removed, now I'm aware it's a no-op, because they are available by default. But that may not always be the case, and we are setting version 7 specifically. So it wasn't an accident that they appeared again. 

I guess if it stops working in the future we can just add it back in, but it is useful for using things like 'act' for local testing, that may not have all actions cached locally.

HebaruSan:
> Awesome! I note the dotnet action was removed, now I'm aware it's a no-op, because they are available by default.

Yup, trying to review these changes made me particularly awaer how much mental bandwidth I was dedicating to, "Oh, that's unnecessary, I can just ignore those 4 lines." Better to not have them there at all.

.github/workflows/release.yml
HebaruSan: Cool, I figured it would be something like that, but I wanted to make sure I wasn't missing some nuance in what's already there. Thanks.

.github/workflows/release.yml
HebaruSan: Linking the quick test showing that this suggestion actually works:

https://github.com/KSP-CKAN/CKAN/actions/runs/9105861890

.github/workflows/deploy.yml
techman83: It's about the credentials, I think it can be done a heap better. But would rather leave that to an overhaul of the deploy workflows.

HebaruSan: OK, thanks for the explanation. Marking resolve because this was just a question for my own understanding. 👍

.github/workflows/release.yml
techman83: We do generate a version in the workflow, and in the makefile. I'm going to make a mental note of it as using the github cli we can use a wildcard instead of the actual name of the file. I wanted to however keep things reasonably close in this iteration.

techman83:
Alright, I think that's everything!

.github/workflows/release.yml
HebaruSan: Not sure why this is marked resolved.

.github/workflows/release.yml
HebaruSan: As of #3997, the compiled assembly itself has the same version in it that we were computing for deb and rpm. When we eventually refactor this, we could get them all from the same source.

HebaruSan (approved):
Agree, looks good to me. Let's go!

PR #4083, opened 2024-05-06
Ubuntu 24.04 breaks build

techman83:
There are a number of changes within 24.04 which upset the apple cart, for now 22.04 is still supported for another year, so we can pin while we give ourselves some time to update things.


techman83:
Given netkan is currently broken and is just reverting the existing container version, I shall self review!

HebaruSan (commented):
Thanks!

PR #4088, opened 2024-05-18
Consolidate Discord notification steps

HebaruSan:
## Motivation

After #4082, I wanted to look for ways to set the first parameter of the Discord notification script (`success` or `failure`) without environment variables, extra pseudo-steps ("Set Success"), or side effects. Gradually I realized that writing a `git clone` command in every place we want to notify was a bit duplicative and could use better encapsulation.

## Changes

- Now our own `notify.yml` action wrapper provides the nice API that `DiscordHooks/github-actions-discord-webhook` should have.
  - Inputs `name` and `success` are accepted, and the `DISCORD_WEBHOOK` secret is required.
  - The somewhat awkward `git clone` logic required to use it is now centralized in one place. If the optimal notification method changes in the future, we will be able to update it in just one spot.
- Now what [the action documentation](https://docs.github.com/en/actions/learn-github-actions/expressions#example) calls "ternary operator like behaviour" is used to choose between `success` and `failure` based on the value of the boolean `success` input, with no environment variable twiddling involved.
- Now all spots that previously sent their own notifications are updated to go through the new `notify.yml`, and the `contains` expression usees the `!` not operator instead of `== false`.
- The `notify` jobs in `test.yml` and `smoke.yml` now depend on their respective `build-*` steps to ensure that failure can be detected by checking `contains(needs.*.result, 'failure')` (if you only `needs` the intermediate step, its status will be `skipped` when a previous step fails).

This should be a bit more maintainable, and it gave me some more practice in action coding.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -188,25 +188,14 @@ jobs:
           ./build docker-metadata --configuration=Release --exclusive
 
   notify-discord:
-    runs-on: ubuntu-latest
     needs:
       - upload-deb
       - upload-rpm
       - upload-inflator
       - upload-metadata-tester
-    env:
-      JOB_STATUS: failure
     if: always()
-    steps:
-      - name: Set Success
-        run: echo "JOB_STATUS=success" >> $GITHUB_ENV
-        if: contains(needs.*.result, 'failure') == false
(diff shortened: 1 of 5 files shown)


techman83:
Love this! It was on my hit list to look at 🙂

PR #4089, opened 2024-05-19
Use checkout action for notify workflow

HebaruSan:
## Motivation

#4088 seems to work, but a `git clone` command line is not the best level of abstraction for a workflow.

## Changes

- Now we use `actions/checkout` instead.
- Now the merge script is updated to prevent merging if any of the checks fail or are still pending.
- Now the notify jobs in `release.yml` and `deploy.yml` depend on their respective build/test jobs to ensure that any failures will be available in `needs.*.result`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -189,6 +189,8 @@ jobs:
 
   notify-discord:
     needs:
+      - test-release
+      - smoke-inflator
       - upload-deb
       - upload-rpm
       - upload-inflator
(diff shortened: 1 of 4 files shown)


HebaruSan:
Classic!

```
$ .venv/bin/python bin/ckan_merge_pr.py --self-review 4089
Request GET /repos/KSP-CKAN/CKAN failed with 403: rate limit exceeded
Setting next backoff to 1012.615892s
```

PR #4092, opened 2024-05-20
Linux: Improve desktop entries

mmvanheusden:
I added keywords to the `.desktop` files of CKAN, so that they can be found easier. For example when a user forgets the name of the program they just type in `Kerbal` and it shows up as well next to the game.

See: https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html
image
image


HebaruSan (approved):
Cool idea. Thanks for the submission!

PR #4093, opened 2024-05-21
Pass image from smoke to deploy, restore build caching

HebaruSan:
## Motivation

Currently we build the Inflator image twice when pushing to `master`, once for smoke testing and again for pushing to Docker Hub. This is a small but not tiny amount of extra time and repeated work.

Another few extra seconds are spent re-downloading build dependencies that formerly were cached in the old workflows.

## Changes

- The Inflator image build / push / redeploy steps are moved out of `build.cake` and into `smoke.yml` and `deploy.yml` using the standard docker actions and commands, in order to allow the steps to be separated and smoke testing inserted before push.
  <https://github.com/docker/build-push-action>
- `smoke.yml` now uploads a tar of the image as an artifact after the smoke test passes
  <https://docs.docker.com/build/ci/github-actions/share-image-jobs/>
  <https://docs.docker.com/build/ci/github-actions/test-before-push/>
- `deploy.yml` now downloads the tar of the image and loads it instead of building it
- The caching of the tools and dependencies is restored in `build.yml`

This should eliminate some redundant work.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,23 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Restore cache for _build/tools
+        uses: actions/cache@v3
+        with:
+          path: _build/tools
+          key: build-tools-${{ hashFiles('build', 'build.ps1', 'build.cake') }}
+      - name: Restore cache for _build/cake
+        uses: actions/cache@v3
+        with:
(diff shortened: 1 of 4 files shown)


HebaruSan:
Two learnings:

1. It seems to still rebuild the image between load and push. I guess the export doesn't include caching.
2. The redeploy commands need `AWS_REGION` rather than `AWS_DEFAULT_REGION` :face_with_spiral_eyes:

PR #4094, opened 2024-05-21
Only redeploy the metadata tester on repo dispatch

HebaruSan:
## Background

#3575, KSP-CKAN/NetKAN-Infra#267, and KSP-CKAN/xKAN-meta_testing#86 made it so the main CKAN repo would run its `deploy.yml` workflow anytime we pushed changes to the Python code from the other two repos.

## Motivation

Currently the entire `deploy.yml` worklow is run, including some unnecessary jobs:

- Uploading `ckan.exe` and `netkan.exe` to S3
- Building the rpm/deb packages and uploading them to S3
- Rebuilding, pushing, and redeploying the Inflator

This takes a little more time and may unnecessarily mark some files as changed when they really aren't.

## Changes

Now when `deploy.yml` is triggered from `NetKAN-Infra` or `xKAN-meta_testing`, we only rebuild and push the metadata tester image (and run its dependencies). This is the only job that depends on the Python code from those other repos.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -25,6 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       dev-build: ${{ steps.check-version.outputs.dev-build }}
+    if: github.event_name != 'repository_dispatch'
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


HebaruSan:
image

PR #4111, opened 2024-06-26
Store GitHub Discussions links and display in UIs

HebaruSan:
## Motivation

The KSP franchise's future status has become a bit uncertain, and the forum web site has been somewhat unreliable lately.

If the forums became unavailable, GitHub Discussions would be one possible alternate place where particular mods might be discusssed.

image

- <https://github.com/HebaruSan/Astrogator/discussions>

## Changes

- Now the spec and schema allow `resources.discussions` to be a URL
  (But no `spec_version` limitations are imposed because it's easy for the old client to just ignore the new value)
- Now Netkan checks whether discussions are enabled for GitHub repos, and if so, it sets `resources.discussions` to that repo's discussions URL, just like we do for `resources.bugtracker`
- Now CmdLine, ConsoleUI, and GUI include the `resources.discussions` link in their respective mod info displays

This way if mod authors start enabling discussion for their repos, either now or upon the sudden disappearance of the forum, the URLs will begin appearing in CKAN so users can find out about them.

### Considered but not done

The obvious question here is: What about `resources.homepage`, which for most mods will be set to a forum thread? It would be amazing to automatically retire those links and switch `resources.homepage` over to discussions automatically, but after some consideration, I don't think that's a viable option:

- The forum is not yet dead, and Netkan won't know programmatically if that happens
- Some other forum preservation option may become available for migrating homepages, such as via the Internet Archive
- Some authors may prefer to keep the old links available for nostalgic purposes
- Some authors may _want_ to use both a hompage _and_ a discussions link (for example, when the homepage is non-interactive like a wiki)

For these reasons, the discussions link is added separately and homepage is left alone.

### DIFF
--- a/CKAN.schema
+++ b/CKAN.schema
@@ -201,6 +201,11 @@
                     "type"        : "string",
                     "format"      : "uri"
                 },
+                "discussions" : {
+                    "description" : "Mod discussions",
+                    "type"        : "string",
+                    "format"      : "uri"
+                },
                 "license" : {
                     "description" : "Mod license",
                     "type"        : "string",
--- a/Cmdline/Action/Show.cs
+++ b/Cmdline/Action/Show.cs
@@ -281,6 +281,11 @@ namespace CKAN.CmdLine
                     user.RaiseMessage(Properties.Resources.ShowBugTracker,
                                       Net.NormalizeUri(module.resources.bugtracker.ToString()));
                 }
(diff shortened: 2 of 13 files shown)


HebaruSan:
Huh, a surprising number of mod repos are already using it.

- https://github.com/KSP-CKAN/CKAN-meta/compare/d7f6ba2...f5cac92

PR #4117, opened 2024-07-05
Update Artifact Signing

techman83:
Version 4 of [SignPath/github-action-submit-signing-request](https://github.com/SignPath/github-action-submit-signing-request/tree/v0.4) required the following to be addressed:
> Id of the Github Actions artifact. Must be uploaded using the actions/upload-artifact v4+ action before it can be signed. Use {{ steps.<step-id>.outputs.artifact-id }} from the preceding actions/upload-artifact action step.

We now have outputs to be able to pass that artifact id through the workflows to the signing action. Results of this working can be seen here -> https://github.com/KSP-CKAN/CKAN/actions/runs/9804590624

## Extra
- Bumped `actions/cache` to v4

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,24 +6,30 @@ on:
       configuration:
         type: string
         default: Debug
+    outputs:
+      repack-artifact-id:
+        description: "Artifact ID of the repack"
+        value: ${{ jobs.build.outputs.repack-artifact-id }}
 
 jobs:
   build:
     runs-on: ubuntu-latest
+    outputs:
+      repack-artifact-id: ${{ steps.upload-repack-artifact.outputs.artifact-id }}
     steps:
       - uses: actions/checkout@v4
       - name: Restore cache for _build/tools
-        uses: actions/cache@v3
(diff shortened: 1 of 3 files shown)


HebaruSan:
How do we attach the signed artifact to the release?

HebaruSan (approved):
I guess this is just an API-compatibility update with no new or changed functionality.
Looks fine so far. :+1:

techman83:
There is either an url we can call, but we'd just set an output path
```yaml
    output-artifact-directory: './_build-signed'
```
Then attach it to the release
```
        run: gh release upload ${{ github.event.release.tag_name }} _build-signed/Release/ckan.exe
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
We'd probably also upload it as an artifact we can use for the rest of release artifacts to consume, and just switch everything out. It's the same directory structure as the repack-unsigned, so in theory if we upload it as 'repack-signed', we can just change everything from '-unsigned' to '-signed' with no other changes.

For the nuget package, we'll have to sign that separately, but I think that's reasonable in the scheme of things.

PR #4118, opened 2024-07-06
Add Translation of Language Classical Chinese

zhouyiqing0304:
First, I would like to tell you, why I named it zh-HX(Classical Chinese).
The pinyin for “华夏” is “Huá Xià”, and the combination of the first letters of the pinyin is “HX”.
"华夏" means China, is as same as "中国"
华夏 will be translated to HuaXia in google translate, but 中国 will be translated to China.

Next, I'm going to tell you the importance of this language.
Long long time ago(more than 150 years), people was using this language. 
Although we don't use it now, understanding it is still important.
Now the teachers are still teaching them.
You probably will say “Why do I need to add the translation of Classical Chinese?”
Now Minecraft has it. So we have to have it too.
image

I only translated 5 files. Still working.
Translated: 
SettingsDialog.zh-HX.resx
UnmanagedFiles.zh-HX.resx
Versions.zh-HX.resx
Wait.zh-HX.resx
YesNoDialog.zh-HX.resx


zhouyiqing0304:
@HebaruSan Have a check.
@050644zf u2

HebaruSan:
Hi @zhouyiqing0304, thanks for contributing! Here are the compile errors for these changes:

`         /usr/lib/mono/msbuild/Current/bin/Microsoft.Common.CurrentVersion.targets(3177,5): error MSB3577: Two output file names resolved to the same output path: "../_build/out/CKAN-GUI/Release/obj/Release/net48/CKAN.GUI.AboutDialog.resources" [/home/runner/work/CKAN/CKAN/GUI/CKAN-GUI.csproj]`

Unfortunately, it looks like `zh-HX` is not a standard locale code, so we can't use it. These strings are used by standard tools like the crowdin page and the .NET compiler, so we don't get to make up whatever strings we like. Even some standard codes like `es-419` sometimes aren't supported by these tools. This _might_ be the cause of the above error.

The available `zh-*` options seem to be:

- zh-CN
- zh-HK
- zh-MO
- zh-SG
- zh-TW
- zu-ZA

zhouyiqing0304:
I used VS

zhouyiqing0304:
I'll have a check

Issue #3770, opened 2023-02-01
[Bug] Transaction failed during `GameInstance.Scan` on Mac


ofbakirci:
Background
----------
* Operating System: macOS 13.1 Ventura 
* CKAN Version: v1.31.2
* KSP Version: 1.25.5.3190

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
No, I have not. 

Problem
-------
When I try to run CKAN, first screen on terminal shows up, says "Press any key to continue", after that error shows up and CKAN crashed back.

Install Mono
Check if Mono install is working using directions [here](https://www.mono-project.com/docs/getting-started/mono-basics/)
Copy CKAN to KSP main folder
Run CKAN
Press any key.

**Expected behavior**
Terminal UI of CKAN showing up

**Screenshots (if applicable)**
[Error screen](https://imgur.com/F9gV4Xd
)

**CKAN error code (if applicable):**

```
Unhandled Exception:
System.Transactions.TransactionAbortedException: Transaction failed ---> System.ArgumentOutOfRangeException: Number must be either non-negative and less than or equal to Int32.MaxValue or -1.
Parameter name: timeout
  at System.Threading.WaitHandle.WaitOne (System.TimeSpan timeout, System.Boolean exitContext) [0x00026] in <e0a3c60063684e309e609c6019ed1f88>:0 
  at System.Transactions.Transaction.DoPreparePhase () [0x00055] in <8bca7f503b7943c0aceeedfbf6372e05>:0 
  at System.Transactions.Transaction.DoCommit () [0x00076] in <8bca7f503b7943c0aceeedfbf6372e05>:0 
  at System.Transactions.Transaction.CommitInternal () [0x00022] in <8bca7f503b7943c0aceeedfbf6372e05>:0 
   --- End of inner exception stack trace ---
  at System.Transactions.Transaction.CommitInternal () [0x00039] in <8bca7f503b7943c0aceeedfbf6372e05>:0 
  at System.Transactions.TransactionScope.Dispose () [0x001ec] in <8bca7f503b7943c0aceeedfbf6372e05>:0 
  at CKAN.GameInstance.Scan () [0x0013d] in <79ae9046bd9c43dabfd2f7be4b883034>:0 
  at CKAN.ConsoleUI.ModListScreen.ScanForMods () [0x0000b] in <79ae9046bd9c43dabfd2f7be4b883034>:0 
[...]

PR #4119, opened 2024-07-09
Fix default transaction timeout

romi2002:
Fixes GameInstance scanning described in #3770, switching over to a default `timeout` value of `Int32.MaxValue`.

### DIFF
--- a/Core/CkanTransaction.cs
+++ b/Core/CkanTransaction.cs
@@ -26,7 +26,7 @@ namespace CKAN
         }
 
         // System.ArgumentOutOfRangeException : Time-out interval must be less than 2^32-2. (Parameter 'dueTime')
-        private const double timeoutMs = 4294967294d;
+        private const double timeoutMs = Int32.MaxValue;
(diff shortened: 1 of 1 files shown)


HebaruSan:
Thanks! That makes sense. I must have assumed that `Int32.MaxValue` was equal to the value we had there without looking it up, but I guess one is signed and the other is unsigned.

Hey @ofbakirci or @jfinkbei or anyone else having the issue from #3770, would you mind trying out this PR's build and confirming whether that issue is fixed?

- <https://github.com/KSP-CKAN/CKAN/suites/25783394137/artifacts/1683351635>

For posterity, it looks like the new max value is now ~25 days, down from the 49 days documented in #3512.

romi2002:
I was getting a similar issue as the one @jfinkbei was getting in #3770 on Arch Linux, and CKAN was hanging on game instance scanning. This PR seems to have fixed it for me.

HebaruSan (approved):
Awesome, thanks @romi2002! I just wanted to make sure this had been stress tested to catch any unknown unknowns.

PR #4129, opened 2024-07-25
Option to clone smaller instances with junction points (Windows) or symbolic links (Unix)

### LINKED ISSUE #4087: [Feature]: Symlinks for GameData
### Problem

When using multiple KSP instances, the squad & squadexpansions folders, along with some others, are duplicated, wasting space.

### Suggestion

My idea is for that all mods in a specific instance, and any save files etc, would be saved outside KSP. Then, when you click "Launch Game", the data for that instance would be symlinked in.

For example, I have a KSP installation in C:/Games/Kerbal Space Program. I create a instance using this in CKAN, named "KSP test". When I install a mod, it goes to %appdata%/Local/CKAN/KSP test/GameData. Inside the CKAN/KSP test you would also have saves and ships. 

Now, I click launch game in CKAN. Symlinks are created from %appdata%/Local/CKAN/KSP test & its folders to the respective folders in C:/Games/Kerbal Space Program. When I play the game, KSP sees exactly what it usually would and runs accordingly.

The advantage of this is saving 3-4 GB of space per installation, which is quite significant. It also isolates the actual KSP folder, which   might be useful to prevent needing a reinstallation (unsure of this part).

I've been doing something similar with .bat files myself, and so far it works well!

### Alternatives

_No response_

### Additional context

 I believe most Skyrim mod installers do something similar to this, which could be useful for some help.

HebaruSan:
## Background

It's common for users to make extra copies of KSP1 so they can install different mod lists in each one. CKAN supports this by tracking multiple instances and providing a clone operation, which copies an existing instance to a new folder and adds it to CKAN.

## Motivation

- Each game instance contains about 5.5 GiB of files that come with the game and are necessary for it to run. Copying these takes a long time, and having multiple copies of them wastes a lot of space.
- Each game instance also contains files generated during gameplay and while CKAN is operating that are relevant to that instance but not really to clones. For example, screenshots are relevant to the instance where they're created, but having multiple copies of them isn't generally desirable. Copying these into cloned instances also wastes some time and space.

## Changes

- Now there is an option to share the stock files between the clone and the original instance, either via `ckan instance clone --share-stock` or a new "Share stock files" checkbox in GUI:
  image
  which is done by making junction points (Windows) or symbolic links (Unix) for the following directories instead of copying them:
  - `KSP_x64_Data` (the main stock game contents folder)
  - `KSP_Data` (same but on Linux)
  - `GameData/Squad`
  - `GameData/SquadExpansion`
  - Some stuff from the now defunct launchers, just in case those are present
- Any folders that are _already_ symbolic links (e.g., if you're cloning a previous thin clone or if you're a mod developer with symbolic links from GameData to your source folders) will be duplicated in the cloned instance as symbolic links to the same targets
- Now the following directories' contents aren't copied into cloned instances of KSP1; the directories are created, but left empty:
  - `CKAN/history` (old mod lists)
  - `CKAN/downloads` (deprecated but might still contain files in very old instances)
  - `saves`
  - `Screenshots`
  - `thumbs`
  - `Missions`
  - `Logs`
- Equivalent updates are made for KSP2 instances

This will allow the creation of game instances that are much smaller but still playable (a thin clone of my Steam instance shrinks from 7 GiB to 81 MiB, or about 1% of its original size, and is very quick to generate).

Note that the CKAN registry and the currently installed mods will still be copied across, so users will still be able to set up some mods in one instance and then clone them elsewhere as a starting point for a different mod list.

Fixes #4087.

### DIFF
--- a/Cmdline/Action/GameInstance.cs
+++ b/Cmdline/Action/GameInstance.cs
@@ -97,6 +97,9 @@ namespace CKAN.CmdLine
 
     internal class CloneOptions : CommonOptions
     {
+        [Option("share-stock", DefaultValue = false, HelpText = "Use junction points (Windows) or symbolic links (Unix) for stock dirs instead of copying")]
+        public bool shareStock { get; set; }
+
         [ValueOption(0)] public string nameOrPath { get; set; }
         [ValueOption(1)] public string new_name { get; set; }
         [ValueOption(2)] public string new_path { get; set; }
@@ -341,7 +344,7 @@ namespace CKAN.CmdLine
                         if (instance.Name == instanceNameOrPath)
                         {
                             // Found it, now clone it.
-                            Manager.CloneInstance(instance, newName, newPath);
+                            Manager.CloneInstance(instance, newName, newPath, options.shareStock);
                             break;
                         }
                     }
@@ -350,7 +353,7 @@ namespace CKAN.CmdLine
                 // If it's valid, go on.
                 else if (Manager.InstanceAt(instanceNameOrPath) is CKAN.GameInstance instance && instance.Valid)
                 {
-                    Manager.CloneInstance(instance, newName, newPath);
+                    Manager.CloneInstance(instance, newName, newPath, options.shareStock);
                 }
                 // There is no instance with this name or at this path.
                 else
--- a/Core/DirectoryLink.cs
+++ b/Core/DirectoryLink.cs
@@ -1,0 +1,167 @@
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Text;
+using Microsoft.Win32.SafeHandles;
+
+using ChinhDo.Transactions.FileManager;
+
+namespace CKAN
+{
+    /// <summary>
+    /// Junctions on Windows, symbolic links on Unix
+    /// </summary>
+    public static class DirectoryLink
+    {
+        public static void Create(string target, string link, TxFileManager txMgr)
+        {
+            if (!CreateImpl(target, link, txMgr))
+            {
+                throw new Kraken(Platform.IsWindows
+                    ? $"Failed to create junction at {link}: {Marshal.GetLastWin32Error()}"
+                    : $"Failed to create symbolic link at {link}");
+            }
+        }
+
+        private static bool CreateImpl(string target, string link, TxFileManager txMgr)
+            => Platform.IsWindows ? CreateJunction(link, target, txMgr)
+                                  : symlink(target, link) == 0;
+
+        [DllImport("libc")]
+        private static extern int symlink(string target, string link);
+
+        private static bool CreateJunction(string link, string target, TxFileManager txMgr)
+        {
+            // A junction is a directory with some extra magic attached
+            if (!txMgr.DirectoryExists(link))
+            {
+                txMgr.CreateDirectory(link);
+            }
+            using (var h = CreateFile(link, GenericWrite, FileShare.Read | FileShare.Write, IntPtr.Zero,
+                                      FileMode.Open, BackupSemantics | OpenReparsePoint, IntPtr.Zero))
+            {
+                if (!h.IsInvalid)
+                {
+                    var junctionInfo = ReparseDataBuffer.FromPath(target, out int byteCount);
+                    return DeviceIoControl(h, FSCTL_SET_REPARSE_POINT,
+                                           ref junctionInfo, byteCount + 20,
+                                           null, 0,
+                                           out _, IntPtr.Zero);
+                }
+            }
+            return false;
+        }
+
+        public static bool TryGetTarget(string link, out string target)
+        {
+            target = null;
+            var fi = new DirectoryInfo(link);
+            if (fi.Attributes.HasFlag(FileAttributes.ReparsePoint))
+            {
+                if (Platform.IsWindows)
+                {
+                    var h = CreateFile(link, 0, FileShare.Read, IntPtr.Zero,
+                                       FileMode.Open, BackupSemantics | OpenReparsePoint, IntPtr.Zero);
+                    if (!h.IsInvalid)
+                    {
+                        if (DeviceIoControl(h, FSCTL_GET_REPARSE_POINT,
+                                            null, 0,
+                                            out ReparseDataBuffer junctionInfo, Marshal.SizeOf(typeof(ReparseDataBuffer)),
+                                            out _, IntPtr.Zero))
+                        {
+                            if (junctionInfo.ReparseTag == IO_REPARSE_TAG_MOUNT_POINT)
+                            {
+                                target = junctionInfo.PathBuffer.TrimStart("\\\\?\\");
+                            }
+                        }
+                        h.Close();
+                    }
+                }
+                else
+                {
+                    var bytes = new byte[1024];
+                    var result = readlink(link, bytes, bytes.Length);
+                    if (result > 0)
+                    {
+                        target = Encoding.UTF8.GetString(bytes);
+                    }
+                }
+            }
+            return !string.IsNullOrEmpty(target);
+        }
+
+        private const uint GenericWrite               = 0x40000000u;
+        private const uint BackupSemantics            = 0x02000000u;
+        private const uint OpenReparsePoint           = 0x00200000u;
+        private const uint FSCTL_SET_REPARSE_POINT    = 0x000900A4u;
+        private const uint IO_REPARSE_TAG_MOUNT_POINT = 0xA0000003u;
+        private const uint FSCTL_GET_REPARSE_POINT    = 0x000900A8u;
+
+        [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Auto)]
+        private static extern bool DeviceIoControl(SafeFileHandle        hDevice,
+                                                   uint                  IoControlCode,
+                                                   ref ReparseDataBuffer InBuffer,
+                                                   int                   nInBufferSize,
+                                                   byte[]                OutBuffer,
+                                                   int                   nOutBufferSize,
+                                                   out int               pBytesReturned,
+                                                   IntPtr                Overlapped);
+
+        [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Auto)]
+        private static extern bool DeviceIoControl(SafeFileHandle        hDevice,
+                                                   uint                  IoControlCode,
+                                                   byte[]                InBuffer,
+                                                   int                   nInBufferSize,
+                                                   out ReparseDataBuffer OutBuffer,
+                                                   int                   nOutBufferSize,
+                                                   out int               pBytesReturned,
+                                                   IntPtr                Overlapped);
+
+        [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Auto)]
+        private struct ReparseDataBuffer
+        {
+            public           uint   ReparseTag;
+            public           ushort ReparseDataLength;
+            private readonly ushort Reserved;
+            public           ushort SubstituteNameOffset;
+            public           ushort SubstituteNameLength;
+            public           ushort PrintNameOffset;
+            public           ushort PrintNameLength;
+            [MarshalAs(UnmanagedType.ByValTStr, SizeConst = 8184)]
+            public           string PathBuffer;
+
+            public static ReparseDataBuffer FromPath(string target, out int byteCount)
+            {
+                var fullTarget = $@"\??\{Path.GetFullPath(target)}";
+                byteCount = Encoding.Unicode.GetByteCount(fullTarget);
+                return new ReparseDataBuffer
+                {
+                    ReparseTag           = IO_REPARSE_TAG_MOUNT_POINT,
+                    ReparseDataLength    = (ushort)(byteCount + 12),
+                    SubstituteNameOffset = 0,
+                    SubstituteNameLength = (ushort)byteCount,
+                    PrintNameOffset      = (ushort)(byteCount + 2),
+                    PrintNameLength      = 0,
+                    PathBuffer           = fullTarget,
+                };
+            }
+        }
+
+        [DllImport("libc")]
+        private static extern int readlink(string link, byte[] buf, int bufsize);
+
+        [DllImport("kernel32.dll", CharSet = CharSet.Auto, SetLastError = true)]
+        private static extern SafeFileHandle CreateFile([MarshalAs(UnmanagedType.LPTStr)] string    filename,
+                                                        [MarshalAs(UnmanagedType.U4)]     uint      access,
+                                                        [MarshalAs(UnmanagedType.U4)]     FileShare share,
+                                                        IntPtr securityAttributes,
+                                                        [MarshalAs(UnmanagedType.U4)]     FileMode  creationDisposition,
+                                                        [MarshalAs(UnmanagedType.U4)]     uint      flagsAndAttributes,
+                                                        IntPtr templateFile);
+
+        private static string TrimStart(this string orig, string toRemove)
+            => orig.StartsWith(toRemove) ? orig.Remove(0, toRemove.Length)
+                                         : orig;
+
+    }
+}
--- a/Core/GameInstanceManager.cs
+++ b/Core/GameInstanceManager.cs
@@ -239,7 +239,10 @@ namespace CKAN
         /// <exception cref="DirectoryNotFoundKraken">Thrown by CopyDirectory() if directory doesn't exist. Should never be thrown here.</exception>
         /// <exception cref="PathErrorKraken">Thrown by CopyDirectory() if the target folder already exists and is not empty.</exception>
         /// <exception cref="IOException">Thrown by CopyDirectory() if something goes wrong during the process.</exception>
-        public void CloneInstance(GameInstance existingInstance, string newName, string newPath)
+        public void CloneInstance(GameInstance existingInstance,
+                                  string       newName,
+                                  string       newPath,
+                                  bool         shareStockFolders = false)
         {
             if (HasInstance(newName))
             {
@@ -252,11 +255,13 @@ namespace CKAN
             }
 
             log.Debug("Copying directory.");
-            Utilities.CopyDirectory(existingInstance.GameDir(), newPath, true);
+            Utilities.CopyDirectory(existingInstance.GameDir(), newPath,
+                                    shareStockFolders ? existingInstance.game.StockFolders
+                                                      : Array.Empty<string>(),
+                                    existingInstance.game.LeaveEmptyInClones);
 
             // Add the new instance to the config
-            GameInstance new_instance = new GameInstance(existingInstance.game, newPath, newName, User);
-            AddInstance(new_instance);
+            AddInstance(new GameInstance(existingInstance.game, newPath, newName, User));
         }
 
         /// <summary>
--- a/Core/Games/IGame.cs
+++ b/Core/Games/IGame.cs
@@ -24,10 +24,11 @@ namespace CKAN.Games
         string         PrimaryModDirectoryRelative     { get; }
         string[]       AlternateModDirectoriesRelative { get; }
         string         PrimaryModDirectory(GameInstance inst);
-        string[]       StockFolders      { get; }
-        string[]       ReservedPaths     { get; }
-        string[]       CreateableDirs    { get; }
-        string[]       AutoRemovableDirs { get; }
+        string[]       StockFolders       { get; }
+        string[]       LeaveEmptyInClones { get; }
+        string[]       ReservedPaths      { get; }
+        string[]       CreateableDirs     { get; }
+        string[]       AutoRemovableDirs  { get; }
         bool           IsReservedDirectory(GameInstance inst, string path);
         bool           AllowInstallationIn(string name, out string path);
         void           RebuildSubdirectories(string absGameRoot);
--- a/Core/Games/KerbalSpaceProgram.cs
+++ b/Core/Games/KerbalSpaceProgram.cs
@@ -65,6 +65,17 @@ namespace CKAN.Games.KerbalSpaceProgram
             "PDLauncher",
         };
 
+        public string[] LeaveEmptyInClones => new string[]
+        {
+            "saves",
+            "Screenshots",
+            "thumbs",
+            "Missions",
+            "Logs",
+            "CKAN/history",
+            "CKAN/downloads",
+        };
+
         public string[] ReservedPaths => new string[]
         {
             "GameData", "Ships", "Missions"
--- a/Core/Games/KerbalSpaceProgram2.cs
+++ b/Core/Games/KerbalSpaceProgram2.cs
@@ -59,6 +59,12 @@ namespace CKAN.Games.KerbalSpaceProgram2
             "PDLauncher",
         };
 
+        public string[] LeaveEmptyInClones => new string[]
+        {
+            "CKAN/history",
+            "CKAN/downloads",
+        };
+
         public string[] ReservedPaths => Array.Empty<string>();
 
         public string[] CreateableDirs => new string[]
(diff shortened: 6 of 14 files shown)


HebaruSan:
Hey @ProgrammerFailure, this is the pull request for cloning with symlinks. You can [click the Checks tab](https://github.com/KSP-CKAN/CKAN/pull/4129/checks), then the Artifacts dropdown, to download a test build with the "Share stock files" checkbox in the cloning window.

image

ProgrammerFailure:
Hey, this didn't actually create the symlinks for me. As expected, stock files weren't copied, but symlinks also weren't created. Tested with three preexisting CKAN instances, one mod-free, one dev install, and one heavily modded game. No symlinks for Squad were created. 

I used the CKAN you provided, clicked share stock files, and cloned an instance to no effect. I also had to disable the windows batac warning.

ProgrammerFailure:
Windows 11, if that helps

HebaruSan:
Ahh, that's too bad. OK, I'll have to reboot into Windows and do some debugging. Thanks for checking it out!

ProgrammerFailure:
No problem, still looking forward to it.

HebaruSan:
:vomiting_face:, apparently Windows requires you to be an admin to create symlinks?!?!

@ProgrammerFailure, if you have time, can you please try making a shortcut that runs `ckan --asroot`, and then run it as administrator and try cloning with this option again? If it works, then we have some really dumb OS-level design to work around here, otherwise there would remain some hope that it's just a bug that I can actually fix....

ProgrammerFailure:
I can confirm it works with asroot.

HebaruSan:
Thanks! Well, that's tragic. Might have to make this feature Linux-only then, unless I can find a way to check security on Windows. Bookmarking this as the closest thing I've found so far to documentation of the specific thing users need to have to make symbolic links.

https://gist.github.com/filimonov/cd9c41f8eff1f0900312d90e51fbb081

ProgrammerFailure:
For what its worth, I'm fine with granting CKAN admin once per clone operation, and considering it saves 5-6 GB, I imagine others will be too.

HebaruSan:
Sure, that's fine for you, but it's awful for discoverability, and there is no fricking way we can hand-hold the entire user base through some Windows security dance.

ProgrammerFailure:
True. Plus, I suppose there are still people without admin access to their system?

JonnyOThan:
CKAN actually tries very hard to make sure that it’s NOT running as admin.  This causes countless problems with users who run elevated all the time.  No one understands this stuff.

There’s gotta be a way you can request UAC elevation?  I’m kinda surprised that CKAN wouldn’t have that permission level considering that it just created the directories, right? Maybe there’s some option for setting the permission when creating.  Or can the clone be created elsewhere and then moved into position?

JonnyOThan:
Oh interesting, you have to start a new process: https://stackoverflow.com/a/573150

Maybe just run console ckan with all the parameters and —as root?

ProgrammerFailure:
No, I am not. Should I test with it enabled?

HebaruSan:
Not yet, it won't do anything with the current build. I'll let you know when the new one is ready.

HebaruSan:
On second reading, that developer mode requirement is stated to only be required for UWP, which... I would _guess_ doesn't apply to us? Anyway, pushed a couple of changes:

- I tried passing that allow-unprivileged create flag, in case it actually helps
- We're now checking the return value of `CreateSymbolicLink` and throwing our own exception on failure, for consistency with every other IO operation in C#

I don't envision this as the final state we would put in a release, but it's a step that will allow us to figure out what's possible.

@ProgrammerFailure if you re-download the latest build from the same place in the Checks tab, a test as a non-admin user would be helpful. Thanks!

HebaruSan:
Asking Windows whether the currently logged in user can create a symbolic link appears to be an extremely arduous process entailing multiple screens of code, sigh:

- <https://www.codeproject.com/Articles/5341307/Win32-Security-Retrieving-Privileges-and-Groups-fo>

HebaruSan:
MS said they changed this 7.5 years ago, but it still works that way?

<https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/>

image

HebaruSan:
New bookmark, actual code to check the special security token:

https://github.com/andykimpe/cygwin-setup/commit/449640c40241439272704d91a1854513a72d8ae0

HebaruSan:
I pushed a few more updates:

- The new checkbox is now only visible if CKAN is running as a user that is allowed to create symbolic links, including:
  - Everybody on Unix
  - The "development mode" flag
  - Administrators
  - The specific symbolic link creation right in Windows policy

This way the option should only be available if it'll succeed.

At this point I am still working exclusively in Linux based on documentation and sample code, so it's possible there are bugs in some of those checks.

ProgrammerFailure:
My windows machine had both the symlink policy and admin access enabled. However, symlinks were only created with admin access. So I'm unsure how useful that symlink right is. I've struggled with that thing in the past, and usually I just end up using admin.

HebaruSan:
> My windows machine had both the symlink policy and admin access enabled. However, symlinks were only created with admin access.

Depending on exactly when you tested and with which iteration of the build, that could be explained by the `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` flag, which I originally wasn't passing but have subsequently added. If your test ckan.exe is any older than 2 hours, I'd recommend replacing it from the same source.

ProgrammerFailure:
The build I was testing was yesterday's one. I'll try to test in 4-5 hours.

ProgrammerFailure:
What permissions do I need for the checkbox to appear? There's no share symbolic links checkbox when running as an unelevated user.

HebaruSan:
Either one of these is supposed to do it:

- <https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development#activate-developer-mode>
- <https://portal.perforce.com/s/article/3472>

ProgrammerFailure:
I think I know what the issue was; I was running on an admin account that ALSO had the create symlink right, so windows wouldn't allow using that right without the standard UAC popup. Will try on my non-admin, symlink account.

ProgrammerFailure:
That was not the issue; I'll try later with the command line method.

HebaruSan:
Thanks for trying it out; like I said, it's very possible that those checks simply aren't working due to one or more bugs. The API calls to implement them are extremely painful C++-in-C# stuff that would be very unlikely to be perfect on the first attempt. I'll take a look in Windows in a day or two or three.

ProgrammerFailure:
Could you send a version without the checks? I could at least test whether symlink creation works. Or would the command line override checks?

HebaruSan:
Alright, just pushed an update to make the checkbox always visible. I'll roll that back once I investigate the checks.

ProgrammerFailure:
What are all the cases I should check under? I can think of:
> Admin user with symlink rights
> > Both as root and not as root

> Standard user with symlink rights 

> Standard user without symlink rights 
> > That's probably just gonna fail. 

Will test all these cases tomorrow.

ProgrammerFailure:
Latest CKAN from the checks tab, and when the share stock files button is checked, it still copies stock files in their entirety. No symlinks in sight. Not running as admin.

ProgrammerFailure:
Running as admin, a bunch of .symlink files are created, but they don't actually function as symlinks. Screenshot below.
image

Using edit in notepad on these files gives the following message:
image

ProgrammerFailure:
Note the Squad item without an icon in picture 1. That's what I tried to edit in notepad for picture 2.

HebaruSan:
Alright, I am finally back in Windows! My user apparently isn't allowed to make symlinks.

image

```
ERROR_PRIVILEGE_NOT_HELD
1314 (0x522)
A required privilege is not held by the client.
```

Sorry, but Windows's catastrophically incompetent implementation of symlinks seems to be a huge barrier to delivering this feature. Hiding the checkbox for most Windows users looks like the right move; this feature may well end up being Linux-only.

HebaruSan:
> Running as admin, a bunch of .symlink files are created, but they don't actually function as symlinks. 

Yeah, I'm seeing the same. Strike Two for being able to deliver this on Windows. 🙄

ProgrammerFailure:
Maybe this could work: CKAN generates a .bat file, that when run by the user, will create appropriate symlinks. My experience with .bat files has been much better than this programmatic approach.

HebaruSan:
No.

HebaruSan:
Aha, found a bug! Directory symlinks were being created with the `SymbolicLinkFlags.File` flag instead of the `SymbolicLinkFlags.Directory` flag, which would explain the weird .symlink files. Hopefully fixing that helps...🤞

ProgrammerFailure:
Ah, makes sense. The symlinks were targeting a FILE called Squad as opposed to a folder? Is that correct?

HebaruSan:
Upon giving myself the "Create symbolic link" privilege, the actual link creation works, but the detection of the privilege doesn't. `GetTokenInformation` supposedly returns 88 privileges, but none of them are found by `LookupPrivilegeName` except for `SeUndockPrivilege`. I'll need to explore this API a bit more to figure out how to use it for this...

HebaruSan:
In the latest test build, the temporary hack to make the checkbox always visible is removed, plus:

- The file vs dir thing from before is fixed; symlinks on Windows should now actually work
- The "Create symbolic link" privilege check should now work, or at least it did for me
- The "developer mode" check was already working, but has now actually been tested in Windows

JonnyOThan:
Just realized something: I never actually use symlinks on windows when setting up references; I use _junctions_ (`mklink /j`).  I've never had to switch to an admin account or elevate privileges.  I'm not exactly sure how to do that programmatically but maybe we can just invoke mklink directly on windows?

https://stackoverflow.com/questions/3615129/how-to-create-a-junction-point-using-winapi

HebaruSan:
@JonnyOThan, thanks for the idea! That does indeed look like the better approach on Windows. We're only making symlinks to the absolute paths of directories, so the only difference for us would be the security stuff.

[image](https://stackoverflow.com/a/48586946)

HebaruSan:
@JonnyOThan, do junctions work across separate physical disks? I.e., can I `mklink /j` from `C:` to `D:`? I would test this myself, but my Windows system only has one disk.

JonnyOThan:
I'm like 90% sure they can...I recall doing that at a previous job.

techman83:
@HebaruSan yes they do. It's how I used to point some games on a different drive when stream didn't support install locations. I've probably still got the PowerShell script for the task somewhere.

https://en.m.wikipedia.org/wiki/NTFS_links#:~:text=While%20NTFS%20junction%20points%20support,support%20for%20cross%2Dfilesystem%20links.

HebaruSan:
Update: I'm struggling to get the junction API to work:

image

> ERROR_INVALID_REPARSE_DATA
>
> 4392 (0x1128)
>
> The data present in the reparse point buffer is invalid.

I'm sure it will work fine eventually, but right now I have no idea what's it doesn't like about my data buffer.

HebaruSan:
I think the problems I was having were caused by thinking that `string`s consist of 1-byte chars like in C, but they're actually 2-byte chars (the API requires byte counts to be specified in several places).
Latest push:

- Junctions instead of symlinks on Windows
- The checkbox is always visible and checked by default

Should be a lot more useful now that all that security craziness is gone. I'll probably merge this version unless I or anyone else finds an issue shortly.

ProgrammerFailure:
@HebaruSan ,

I wanted to download this but it says that these artifacts have expired. Is there any way to access it without building CKAN myself?

HebaruSan:
@ProgrammerFailure yes, you can try the dev build here:
<https://ksp-ckan.s3-us-west-2.amazonaws.com/ckan.exe>

ProgrammerFailure:
Thanks! Will that link always go to the current dev build?

HebaruSan:
Yup, it gets built and replaced every time we merge something to the master branch.

ProgrammerFailure:
On trying to make a symlinked instance, I get this error:

`

CKAN.InvalidKSPInstanceKraken: Exception of type 'CKAN.InvalidKSPInstanceKraken' was thrown.
   at CKAN.GameInstanceManager.SetCurrentInstance(String name)
   at CKAN.GUI.CloneGameInstanceDialog.<buttonOK_Click>d__5.MoveNext()


************** Loaded Assemblies **************
mscorlib
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9261.0 built by: NET481REL1LAST_C
    CodeBase: file:///C:/Windows/Microsoft.NET/Framework64/v4.0.30319/mscorlib.dll
ckan
    Assembly Version: 1.34.0.0
    Win32 Version: 1.34.5.24253
    CodeBase: file:///D:/Games%20and%20co/Other%20game%20apps/ckan.exe
System.Core
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9261.0 built by: NET481REL1LAST_C
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Core/v4.0_4.0.0.0__b77a5c561934e089/System.Core.dll
System
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9261.0 built by: NET481REL1LAST_C
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System/v4.0_4.0.0.0__b77a5c561934e089/System.dll
System.Configuration
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Configuration/v4.0_4.0.0.0__b03f5f7f11d50a3a/System.Configuration.dll
System.Xml
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Xml/v4.0_4.0.0.0__b77a5c561934e089/System.Xml.dll
Anonymously Hosted DynamicMethods Assembly
    Assembly Version: 0.0.0.0
    Win32 Version: 4.8.9261.0 built by: NET481REL1LAST_C
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_64/mscorlib/v4.0_4.0.0.0__b77a5c561934e089/mscorlib.dll
System.Numerics
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Numerics/v4.0_4.0.0.0__b77a5c561934e089/System.Numerics.dll
System.Runtime.Serialization
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9241.0 built by: NET481REL1LAST_C
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Runtime.Serialization/v4.0_4.0.0.0__b77a5c561934e089/System.Runtime.Serialization.dll
System.Data
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9214.0 built by: NET481REL1LAST_B
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_64/System.Data/v4.0_4.0.0.0__b77a5c561934e089/System.Data.dll
System.Windows.Forms
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9251.0 built by: NET481REL1LAST_C
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Windows.Forms/v4.0_4.0.0.0__b77a5c561934e089/System.Windows.Forms.dll
System.Drawing
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Drawing/v4.0_4.0.0.0__b03f5f7f11d50a3a/System.Drawing.dll
System.Transactions
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_64/System.Transactions/v4.0_4.0.0.0__b77a5c561934e089/System.Transactions.dll
Microsoft.GeneratedCode
    Assembly Version: 1.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Xml/v4.0_4.0.0.0__b77a5c561934e089/System.Xml.dll
Accessibility
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/Accessibility/v4.0_4.0.0.0__b03f5f7f11d50a3a/Accessibility.dll
System.EnterpriseServices
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_64/System.EnterpriseServices/v4.0_4.0.0.0__b03f5f7f11d50a3a/System.EnterpriseServices.dll

************** JIT Debugging **************
To enable just-in-time (JIT) debugging, the .config file for this
application or computer (machine.config) must have the
jitDebugging value set in the system.windows.forms section.
The application must also be compiled with debugging
enabled.

For example:

<configuration>
    <system.windows.forms jitDebugging="true" />
</configuration>

When JIT debugging is enabled, any unhandled exception
will be sent to the JIT debugger registered on the computer
rather than be handled by this dialog box.
`

And it doesn't make the instance. Did I mess up something? Reputation based protection is disabled.

HebaruSan:
> Did I mess up something

Nope, an unrelated bug caused by a later change is causing the "copy from" instance's path to not fill in to this box:

image

I just pushed a fix for this, which our GitHub workflows will process and eventually generate a new dev build, which the previous dev build should prompt you to install when you launch once it's ready.

ProgrammerFailure:
I'm now getting this error:

<details>
<summary>Details</summary>
```

CKAN.InvalidKSPInstanceKraken: Exception of type 'CKAN.InvalidKSPInstanceKraken' was thrown.
   at CKAN.GameInstanceManager.SetCurrentInstance(String name)
   at CKAN.GUI.CloneGameInstanceDialog.<buttonOK_Click>d__5.MoveNext()


************** Loaded Assemblies **************
mscorlib
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9261.0 built by: NET481REL1LAST_C
    CodeBase: file:///C:/Windows/Microsoft.NET/Framework64/v4.0.30319/mscorlib.dll
ckan
    Assembly Version: 1.34.0.0
    Win32 Version: 1.34.5.24254
    CodeBase: file:///D:/Games%20and%20co/Other%20game%20apps/ckan.exe
System.Core
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9261.0 built by: NET481REL1LAST_C
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Core/v4.0_4.0.0.0__b77a5c561934e089/System.Core.dll
System
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9261.0 built by: NET481REL1LAST_C
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System/v4.0_4.0.0.0__b77a5c561934e089/System.dll
System.Configuration
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Configuration/v4.0_4.0.0.0__b03f5f7f11d50a3a/System.Configuration.dll
System.Xml
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Xml/v4.0_4.0.0.0__b77a5c561934e089/System.Xml.dll
Anonymously Hosted DynamicMethods Assembly
    Assembly Version: 0.0.0.0
    Win32 Version: 4.8.9261.0 built by: NET481REL1LAST_C
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_64/mscorlib/v4.0_4.0.0.0__b77a5c561934e089/mscorlib.dll
System.Numerics
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Numerics/v4.0_4.0.0.0__b77a5c561934e089/System.Numerics.dll
System.Runtime.Serialization
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9241.0 built by: NET481REL1LAST_C
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Runtime.Serialization/v4.0_4.0.0.0__b77a5c561934e089/System.Runtime.Serialization.dll
System.Data
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9214.0 built by: NET481REL1LAST_B
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_64/System.Data/v4.0_4.0.0.0__b77a5c561934e089/System.Data.dll
System.Windows.Forms
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9251.0 built by: NET481REL1LAST_C
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Windows.Forms/v4.0_4.0.0.0__b77a5c561934e089/System.Windows.Forms.dll
System.Drawing
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Drawing/v4.0_4.0.0.0__b03f5f7f11d50a3a/System.Drawing.dll
System.Transactions
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_64/System.Transactions/v4.0_4.0.0.0__b77a5c561934e089/System.Transactions.dll
Microsoft.GeneratedCode
    Assembly Version: 1.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/System.Xml/v4.0_4.0.0.0__b77a5c561934e089/System.Xml.dll
Accessibility
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_MSIL/Accessibility/v4.0_4.0.0.0__b03f5f7f11d50a3a/Accessibility.dll
System.EnterpriseServices
    Assembly Version: 4.0.0.0
    Win32 Version: 4.8.9032.0 built by: NET481REL1
    CodeBase: file:///C:/WINDOWS/Microsoft.Net/assembly/GAC_64/System.EnterpriseServices/v4.0_4.0.0.0__b03f5f7f11d50a3a/System.EnterpriseServices.dll

************** JIT Debugging **************
To enable just-in-time (JIT) debugging, the .config file for this
application or computer (machine.config) must have the
jitDebugging value set in the system.windows.forms section.
The application must also be compiled with debugging
enabled.

For example:

<configuration>
    <system.windows.forms jitDebugging="true" />
</configuration>

When JIT debugging is enabled, any unhandled exception
will be sent to the JIT debugger registered on the computer
rather than be handled by this dialog box.


```


</details>

ProgrammerFailure:
Latest dev version, and the box does get autofilled

HebaruSan:
Thanks again, found and fixed another recent issue with that popup.
The next dev build you try should work.

PR #4147, opened 2024-08-06
Recommendation suppression for modpacks

### LINKED ISSUE #4145: [Feature] .ckan files need a way to suppress recommendations (allows for accurate backups/replication)
## Motivation

Currently, CKAN is not able to accurately save and restore a set of installed mods with minimal user attention.  This is important for multiplayer situations where everyone needs the exact same mods installed.

The issue specifically is around recommendations: If you install from a .ckan file, you must opt-out of all of the recommendations in order to restore the exact set of mods that were installed previously (because if they had been installed, they would have been listed in the .ckan file).

The `suppress_recommendations` flag does not work in .ckan files because all of the dependencies are promoted to top-level changeset members.

## Suggestions

Either allow `suppress_recommendations` to work like it normally does in a modpack .ckan file, or add a top-level `suppress_recommendations` flag that applies to all items in the .ckan file.

In the "export modpack" UI screen, there should be a checkbox next to the "save mod versions" box that sets the `suppress_recommendations` flag (whether top-level or in each dependency).

At the end of the day, you should be able to export a modpack with "save mod versions" and "suppress recommendations" enabled, and importing that will ALWAYS get you the exact set of mods that had been installed.

HebaruSan:
## Motivation

Currently, in order to install the exact list of mods in a modpack, the user has to deselect the recommendations of all the mods. It would be nice to be able to skip this step.

It looks like setting `suppress_recommendations` for each item in the `depends` list should work, but it doesn't because those items get added as first class members of the changeset, which causes that property to be ignored in the modpack.

A similar thing can happen in a normal changeset; if A and B are in the changeset, and both depend on C but only A has `suppress_recommendations` set, B's lack of the property will mean that C's recommendations are still shown. That probably frustrates the author of A, especially if they're from KSP-RO.

## Changes

- Now `suppress_recommendations` is obeyed if it is set for _any_ `depends` relationship in the changeset, even if other mods have the same dependency without that property or if the dependency is in the changeset itself.
  - Since modpacks are included in changesets, this means that `depends[].suppress_recommendations` in a modpack will now work.
  - It also means that in the above A/B/C scenario, C's recommendations will no longer be shown, which the KSP-RO folks will probably like.
- Now the export modpack screen in GUI has an "Include optional relationships" checkbox that sets `suppress_recommendations` for every relationship in `depends` if unchecked, to make it easy to generate such metadata.

Fixes #4145.

### DIFF
--- a/Core/Relationships/RelationshipResolver.cs
+++ b/Core/Relationships/RelationshipResolver.cs
@@ -229,6 +229,7 @@ namespace CKAN
                 if (options.get_recommenders && descriptor.suppress_recommendations)
                 {
                     log.DebugFormat("Skipping {0} because get_recommenders option is set", descriptor.ToString());
+                    suppressedRecommenders.Add(descriptor);
                     continue;
                 }
                 options = orig_options.OptionsFor(descriptor);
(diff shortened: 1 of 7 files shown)


JonnyOThan:
Seems to work!

PR #4162, opened 2024-08-18
Update Signpath Action

techman83:
The release version is now available for signpath, though there are still a couple of steps to go through, this will ensure signing is working correctly with the release action.

- Successful run: https://github.com/KSP-CKAN/CKAN/actions/runs/10438536973/job/28905992320
- Signed artifact: [Release-repack-unsigned.zip](https://github.com/user-attachments/files/16647463/Release-repack-unsigned.zip)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,7 +23,7 @@ jobs:
     outputs:
       artifact-url: ${{steps.sign.outputs.signing-request-id }}
     steps:
-      - uses: signpath/github-action-submit-signing-request@v0.4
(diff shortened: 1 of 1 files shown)


HebaruSan (approved):
Exciting to get closer to the finish line!

Could we move this value to a secret? I know it's not technically "_secret_" per se in the sense of being sensitive, but if we want to change it at some point in the future, updating a value in the repo's secrets seems like a better way than editing the workflow.

```yaml
          organization-id: 0cd9fc3f-b78d-4214-b152-b2e93c952e14
```

PR #4174, opened 2024-09-05
Stabilize SourceForge mirror, Cmdline export fixes and modpack format

HebaruSan:
## Motivation

Since #4172 added support for SourceForge krefs and KSP-CKAN/KSP2-NetKAN#183 indexed one, KSP2 mod `KSRe`'s `download` property has been switching from one mirror host to another at every inflation:

image

## Changes

Now Netkan always uses the `psychz` mirror for SourceForge downloads since it is in a central, densely populated location.

### Side fixes

While working on the above, a few additional small things came up:

- The SHA1 and SHA256 download hashes aren't considered in `CkanModule.MetadataEqual`, but the `install_size` is. This could mean that a user would not be prompted to reinstall a re-uploaded mod if the new ZIP had the same file size but different hashes.
  Now the hashes are also checked.
- `ckan list --export` had several quirks:
  - The `--help` info said it accepts a format parameter without saying what the allowed values were.
  Now it mentions the allowed values.
  - There was no way to export a .ckan modpack file even though this is the most popular format.
    Now `--export ckan` does this.
  - The `csv` and `tsv` formats were missing three columns because we were using a format string that had to be updated every time a column was updated, and it wasn't.
    Now we use `string.Join` instead and it works better.


HebaruSan:
I believe the smoke test error is because SpaceDock is down.

PR #4178, opened 2024-09-08
Inflation error for multi-hosted mods with mismatched versions

HebaruSan:
## Motivation

Authors of multi-hosted mods don't always upload the same versions to all hosts. Currently this results in an `Out-of-order version found on unreliable server` error (about 8 of these are active currently); this message does not accurately report what the actual problem is, and but it's possible for a partially-inflated mod to be indexed first.

## Changes

Now if we get more modules than requested, an inflation error is thrown that describes which versions were found on which hosts:

```
$ netkan.exe NetKAN/ZTheme.netkan --releases 3
6461 [1] FATAL CKAN.NetKAN.Program (null) - Generated 4 modules but only 3 requested: v1.1.3.1 on github.com; v1.1.3 on github.com, spacedock.info; v1.1.2 on github.com, spacedock.info; v1.1.1 on spacedock.info
```

### DIFF
--- a/Netkan/Model/Metadata.cs
+++ b/Netkan/Model/Metadata.cs
@@ -34,6 +34,18 @@ namespace CKAN.NetKAN.Model
         public bool           Staged          { get; private set; }
         public string?        StagingReason   { get; private set; }
 
+        public string[]       Hosts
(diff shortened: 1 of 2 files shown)


HebaruSan:
image

PR #4186, opened 2024-09-19
Dev builds for Mac

HebaruSan:
## Motivation

Dev builds have been a handy way of getting quick fixes out to specific users while [waiting for the cryptographic stars to align for a new release](https://github.com/KSP-CKAN/CKAN/pull/4162), but they're currently only available as .exe, .deb, or .rpm. To use a dev build the same way they use the usual `CKAN.dmg` release, a Mac user would have to download the .exe and insert it into the `CKAN.app` bundle him- or herself, which is not an accessible task for all users.

## Changes

Now a new `upload-dmg` step is added to `deploy.yml` that plops the already generated `ckan.exe` into a `CKAN.app` into a `CKAN.dmg` and uploads the latter to s3. After this is merged, a Mac-packaged dev build will be available at this location:

- <https://ksp-ckan.s3-us-west-2.amazonaws.com/CKAN.dmg>

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -73,6 +73,36 @@ jobs:
       - name: Push ckan.exe, netkan.exe, and version.json to S3
         run: aws s3 sync _build/repack/Release s3://${AWS_S3_BUCKET} --follow-symlinks
 
+  upload-dmg:
+    needs:
+      - check-dev-build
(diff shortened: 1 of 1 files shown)


HebaruSan:
I guess `aws s3 sync` can only upload directories, not individual files? 😵‍💫 

image

PR #4189, opened 2024-09-21
Sign releases, dev builds, and nuget package

HebaruSan:
## Motivation

SignPath's action is ready, we have a certificate, and #4188 created a helper build target to collect assets for signing. The next step toward signing releases is updating our workflows.

## Changes

- Now `build.yml` uses the build target from #4188 to upload an artifact containing the assets we need to sign for the Release configuration
- Now `sign.yml` is created to retrieve that artifact, get it signed, and upload a new `signed` artifact with the results
- Now `release.yml` calls `sign.yml` to generate signed assets for releases
- Now `deploy.yml` calls `sign.yml` to generate signed assets for uploading dev builds

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,15 +7,15 @@ on:
         type: string
         default: Debug
     outputs:
-      repack-artifact-id:
-        description: "Artifact ID of the repack"
-        value: ${{ jobs.build.outputs.repack-artifact-id }}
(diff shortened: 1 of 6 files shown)


techman83 (changes requested):
Only change will be, can we update the Readme. Otherwise all looks good to me!

techman83 (approved):
Love it! Thanks @HebaruSan

PR #4194, opened 2024-09-25
Tests for safety of string format calls

HebaruSan:
## Motivation

My first code fix for CKAN was to remove a superfluous `String.Format` call that was causing exceptions to be thrown because externally-generated strings were being used as part of a format string (see #2111).

This problem generalizes; _any_ call to `IUser.RaiseMessage` or `IUser.RaiseError` can throw an exception if the first parameter isn't a valid format string. We can ensure that a string literal is safe by looking at it, and we can ensure that an i18n resource is safe by inspecting our resx files, but anytime a value is externally sourced, it isn't safe to use it as a format string because curly braces could turn up any time and ruin our day.

In principle, the C# compiler _could_ check this and emit warnings similar to what it does for nullable references, but currently it doesn't. The closest thing is the [`StringSyntax` attribute](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.stringsyntaxattribute), which can be used to mark a parameter as a format string, but which doesn't have any built-in functionality associated with it.

## Changes

- Now a new `MonoCecilAnalysisTestsBase` base class is factored out of the test from #3914
- Now a new `StringFormatSafetyTests` class inherits from `MonoCecilAnalysisTestsBase` and scans all code in all of our assemblies for calls to functions with the `StringSyntax` attribute. These calls are deemed safe if the argument list is non-empty, if the format string is a string literal, or if the format string is from `Properties.Resources`; otherwise a test failure is logged.
- Now our format string parameters are tagged with the `StringSyntax` attribute
- Now the 67 existing places that fail the new test are updated to use a format string of `"{0}"`, or to remove a superfluous call to `string.Format`, or other such solutions

This guard rail will help us to write safer code.

### DIFF
--- a/Cmdline/Action/Available.cs
+++ b/Cmdline/Action/Available.cs
@@ -22,8 +22,9 @@ namespace CKAN.CmdLine
                 .Where(m => !m.IsDLC)
                 .OrderBy(m => m.identifier);
 
-            user.RaiseMessage(string.Format(Properties.Resources.AvailableHeader,
-                instance.game.ShortName, instance.Version()));
+            user.RaiseMessage(Properties.Resources.AvailableHeader,
+                              instance.game.ShortName,
+                              instance.Version()?.ToString() ?? "");
             user.RaiseMessage("");
 
             if (opts.detail)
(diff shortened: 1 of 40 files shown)


techman83 (approved):
Safer code and tests! :heart_eyes:

PR #4227, opened 2024-10-09
Tests for Newtonsoft.Json's handling of octal literals in version files

HebaruSan:
## Motivation

A mod uploaded a release today with this in its version file:

```json
		"PATCH":09,
```

This causes an inflation error:

> `New inflation error for <mod name>: Error parsing version file <path to version file>.version: Input string '09' is not a valid number. Path 'VERSION.PATCH', line 9, position 12.`

Digging into this, apparently a `0` prefix for numbers isn't technically allowed at all in strict JSON (and a few online validators confirm this), but some parsers (including the most popular one for C#, Newtonsoft.Json) decided to be lenient and parse them as C-style octal literals. But `08` and `09` are not valid octal literals because only the digits `01234567` can be used in octal.

## Changes

Now several new test cases are added to check and document exactly what is and is not allowed by Newtonsoft.Json's numeric parser, for future reference.

Note that the `QuotedInvalidOctalNumber` test is technically using an exploit, as the [version file schema](https://github.com/linuxgurugamer/KSPAddonVersionChecker/blob/3cca622167dabcb24f1267a865abc13ee6d96c6e/KSP-AVC.schema.json#L122-L145) requires these fields to be [integers](https://json-schema.org/understanding-json-schema/reference/numeric#integer), not strings, but CKAN does not enforce this.

### DIFF
--- a/Netkan/Sources/Avc/JsonAvcToVersion.cs
+++ b/Netkan/Sources/Avc/JsonAvcToVersion.cs
@@ -1,20 +1,20 @@
 using System;
+using System.Collections.Generic;
 
-using log4net;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using log4net;
 
 using CKAN.Versioning;
 
 namespace CKAN.NetKAN.Sources.Avc
 {
     /// <summary>
-    /// Converts AVC style KSP versions into CKAN ones.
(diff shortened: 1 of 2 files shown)


JonnyOThan:
for reference here is the official json grammar: https://www.json.org/json-en.html

PR #4240, opened 2024-10-24
Cache migration and other fixes

### LINKED ISSUE #4086: [Feature]: Download cache change management
### Problem

When updating the cache folder to a new location, existing contents of the cache are not moved to the new location, and the existing contents are not deleted.
This means that the existing contents effectively become orphaned files.
CKAN still appears to have them marked as being downloaded, and allows the "purge from cache" option to be selected, however when clicking on this option it does not delete the cached download from what is now the old location.

### Suggestion

When changing the download cache location, the existing cache data should be moved to the new cache location, potentially unless an option is ticked not to override one of these defaults.

### Alternatives

Alternatively, the old cache should be purged automatically when selecting a new location.

Another alternative would be to take no actions with the old cache, but prompt the user that the old location still contains data, e.g.
"Old cache location contains xxx MiB of data.  Do you wish to purge?"

### Additional context

_No response_

HebaruSan:
## Background

- In #2535, the cache path setting was introduced, always fully editable, with automatic migration of cached files to new locations as you typed
- In #2538, the automatic file migration was removed because it deleted and scrambled the developers' folders and was generally unsafe
- In #3804, the updating of the setting was moved from when you edit the box to when you close the form, to work around problems with attempting to set it to a drive letter temporarily on Windows

The core issue was that the settings dialog primarily uses a live-editing paradigm (check a checkbox and the setting is updated instantly), but the cache path needs an edit/cancel/accept paradigm where the user explicitly chooses to validate and commit the input when it's ready and can address any errors or further choices at that time.

## Motivation / Initial Problems

Users want a cache migration option.

Some related issues were also reported:

- Since the original functionality assumed auto-migration, the GUI still treats mods as cached or uncached according to the old cache rather than the new
- After changing the settings, the GUI still considers a mod to be cached after purging, and vice versa for downloading and uncached

### Causes

- The cache object raises events when mods are added and purged, but GUI wasn't re-subscribing to those events after a new cache object replaced an old one
- `NetModuleCache` wasn't handling mods with multiple download URLs correctly; one failed purge was treated as the mod not being cached at all when one of the other URLs might still be cached

## Problems

While working on the above, a few small things came up:

- A new Steam exception
- A tooltip in the Settings dialog (for the "refresh modlist every" text box) isn't internationalized, so it will always appear in English regardless of locale
- The button in Manage Mods to add a new search was not vertically aligned well with the search text box
- Purging a mod can make the Contents tab flicker

### Causes

- Apparently the `shortcuts.vdf` file itself can be corrupted
- The `ToolTip.SetToolTip` call was hard-coded in `SettingsDialog.Designer.cs` and so got missed when looking for strings
- Purging a mod purges _all_ historical versions of that mod, each of which was refreshing the Contents tab even if it wasn't cached

## Changes

- Now editing the cache path works differently:
  - The text box is read-only by default, with an "Edit" button to enable it:
  image
  - When you click Edit, the text box becomes editable, but the rest of the form becomes read-only, the close X is hidden, and the Edit button is replaced by Cancel and Save buttons:
  image
  - Clicking the Cancel button or pressing Escape reverts the setting to what it was before you clicked Edit
  - Clicking the Save button or pressing Enter tries to set the setting to the value in the text box. If it doesn't exist or can't be used for some reason, an error is presented and you can continue editing.
  - If you change the setting to a different, valid folder, a prompt appears asking you what to do next: move the files (the default if there's enough space), delete the files, open both folders in a file browser for manual partial migration (the default if there isn't enough space), nothing, or cancel the change and keep using the old folder:
  image
  - A progress bar appears while a cache move is in progress and disappears when it completes:
  image
  - The selection dialog has many usability improvements, including internationalization of the ` -- Default` string as `{0} (DEFAULT)`, resizability with a minimum size, auto-sizing to fit, centering on the parent window, auto-selecting the default option, accepting on double-click, accepting on Enter, and cancelling on Escape
  - The settings window's buttons to change the cache via folder selection popup, reset the cache to default, open the cache in file browser, and purge the cache now have tooltips
  - `NetModuleCache.MoveFrom` and `NetFileCache.MoveFrom` are updated to support an `IProgress<int>` object and recursive moving of files in subdirectories (to handle the `downloading` folder)
- The settings dialog now uses `#region`/`#endregion` to organize its many groups of functions
- The settings dialog's auth tokens list box now auto-sizes its columns to fit
- Many parameters and return values are changed from a path `string` to `FileInfo` because it's better and because we were using those `string`s to make `FileInfo`s much of the time anyway
- Now the settings dialog raises a `CacheChanged` event after it replaces the main cache object, which `Main` uses to re-subscribe to the mod added/purged events
- Now `NetFileCache.Remove` has an overload acccepting a sequence of `Uri`s, which attempts to remove all of them and returns true if any was present, and `NetModuleCache.Purge` uses this to purge a mod
- Now `NetModuleCache.Purge` has an overload accepting a sequence of `CkanModules`, which attempts to remove all of them and only raises the `ModPurged` event once
- Now if you right click to download a mod to the cache, the Contents tab is shown afterwards to cue users in that it exists
- Now we ignore corrupted `shortcuts.vdf` files
- Now the tooltip is internationalized
- Now the search button's `Anchor` property has `Top` removed and `Bottom` added, so it'll move up and down as its container resizes, and the container is initialized to the same height as the first search box, which will keep the button aligned
- Now `GUIUser.RaiseSelectionDialog` no longer uses a reference in `Main.selectionDialog` because this makes the dialog crash when the settings dialog tries to use it

Fixes #4086.

### DIFF
--- a/Cmdline/Action/Cache.cs
+++ b/Cmdline/Action/Cache.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 
 using CommandLine;
@@ -177,7 +178,9 @@ namespace CKAN.CmdLine
 
             if (manager != null)
             {
-                if (manager.TrySetupCache(options.Path, out string? failReason))
+                if (manager.TrySetupCache(options.Path,
+                                          new Progress<int>(p => {}),
+                                          out string? failReason))
                 {
                     IConfiguration cfg = ServiceLocator.Container.Resolve<IConfiguration>();
                     user?.RaiseMessage(Properties.Resources.CacheSet, cfg.DownloadCacheDir ?? "");
@@ -205,7 +208,9 @@ namespace CKAN.CmdLine
         {
             if (manager != null)
             {
-                if (manager.TrySetupCache("", out string? failReason))
+                if (manager.TrySetupCache("",
+                                          new Progress<int>(p => {}),
+                                          out string? failReason))
                 {
                     IConfiguration cfg = ServiceLocator.Container.Resolve<IConfiguration>();
                     user?.RaiseMessage(Properties.Resources.CacheReset, cfg.DownloadCacheDir ?? "");
--- a/Core/CKANPathUtils.cs
+++ b/Core/CKANPathUtils.cs
@@ -114,7 +114,9 @@ namespace CKAN
             return NormalizePath(Path.Combine(root, path));
         }
 
-        public static void CheckFreeSpace(DirectoryInfo where, long bytesToStore, string errorDescription)
+        public static void CheckFreeSpace(DirectoryInfo where,
+                                          long          bytesToStore,
+                                          string        errorDescription)
         {
             if (bytesToStore > 0)
             {
@@ -130,6 +132,11 @@ namespace CKAN
                                                    : "unknown bytes");
             }
         }
+
+        public static bool PathEquals(this FileSystemInfo a,
+                                      FileSystemInfo      b)
+            => NormalizePath(a.FullName).Equals(NormalizePath(b.FullName),
+                                                Platform.PathComparison);
 
     }
 }
--- a/Core/GameInstanceManager.cs
+++ b/Core/GameInstanceManager.cs
@@ -33,6 +33,7 @@ namespace CKAN
         public GameInstance?  CurrentInstance { get; set; }
 
         public NetModuleCache? Cache { get; private set; }
+        public event Action<NetModuleCache>? CacheChanged;
 
         public readonly SteamLibrary SteamLibrary = new SteamLibrary();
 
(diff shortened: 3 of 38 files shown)


HebaruSan:
I need to test this in Windows before merging.

HebaruSan:
Hi @gsf600y, there's a `ckan.exe` test build under the Artifacts dropdown here, if you'd like to take these changes for a spin:

https://github.com/KSP-CKAN/CKAN/pull/4240/checks

gsf600y:
Thanks @HebaruSan !
Downloaded, tested, and appears to be working perfectly!

One question - what's the different to selecting "Revert back to old folder" and hitting the Cancel button?

HebaruSan:
Cool, thanks for taking a look!

> One question - what's the different to selecting "Revert back to old folder" and hitting the Cancel button?

They're the same. I added "Revert back to old folder" because I felt it would not necessarily be clear what the Cancel button did, and this makes it an explicit, visible option.

gsf600y:
Ah I see, understood!
I first thought "Revert" might flip back to the previous folder I was using, having stored that folder name somewhere.
That's how I read "Revert" anyway, rather than "cancel" which I understand to be "don't do anything, get me out of this menu".

PR #4250, opened 2024-11-06
Exclude DLLs being upgraded to full modules

### LINKED ISSUE #4225: Trying to upgrade a manually installed mod fails if it conflicts with another mod
Repro:

1. Install Deferred using CKAN
2. Manually install RasterPropMonitor (any version)
3. Hit Refresh in CKAN so that it auto-detects RPM
4. Check the "upgrade" box for RPM and hit Apply

Result:

image

image

This may have something to do with the fact that RPM is incorrectly auto-detected; the DLL name is RasterPropMonitor but the module that contains the DLL is RasterPropMonitor-Core.

HebaruSan:
## Problem

If you manully install RPM alongside a CKAN install of Deferred and then try to upgrade RPM to CKAN-managed, it doesn't let you:

image

## Cause

As of #4067, a DLL can be upgraded to a full module, but the relationship resolver's consistency check still checks conflicts against the pre-upgrade DLL list.

In the case of Deferred, it conflicts with older versions of RPM, but versions can't be checked with DLLs, so the registry starts out in a conflicted state and remains that way despite the planned upgrade to a non-conflicting module.

## Changes

Now the relationship resolver computes a post-install set of DLLs based on the list from the registry minus the identifiers of modules being installed. This allows upgrading a version-specifically-conflicted DLL to a module with a version that doesn't match the conflict. A test is included to catch any future regressions.

Fixes #4225.

### DIFF
--- a/Core/Properties/Resources.fr-FR.resx
+++ b/Core/Properties/Resources.fr-FR.resx
@@ -596,7 +596,7 @@ Pensez à ajouter un jeton d'authentification pour augmenter la limite avant bri
   <data name="KrakenAlreadyRunning" xml:space="preserve">
     <value>Fichier verrou avec un ID de processus actif à {0}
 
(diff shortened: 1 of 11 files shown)


HebaruSan:
Hmm, this probably deserves a test...

PR #4258, opened 2024-11-11
fix: `NetKAN.schema` not rejecting invalid pattern for `$kref` and `$vref` in VSCode

Danny2462:
Optional capture groups allowed invalid patterns like
`"$kref": "#/ckan/github/foo/foo-bar/asset_WARN_match/foo-bar-baz.zip"`
to not result in a warning by VSCode, now they are rejected correctly:

before:
image

after:
image

Note that it warns about the spacedock pattern, which is the first of the possible patterns - this unhelpful message is a limitation of VSCode / just this regex approach of JSON schemas in general 🤷‍♂️

Also added escape slashes where needed, which is superfluous for the C# flavor, but required in the JavaScript flavor of regex, just to be safe.

### DIFF
--- a/NetKAN.schema
+++ b/NetKAN.schema
@@ -42,39 +42,39 @@
             "type": "string",
             "oneOf": [
                 {
-                    "pattern": "^#\/ckan\/spacedock\/[0-9]+$",
+                    "pattern": "^#/ckan/spacedock/[0-9]+$",
                     "description": "Indicates that data should be fetched from SpaceDock"
                 },
                 {
-                    "pattern": "^#\/ckan\/curse\/[A-Za-z0-9_-]+$",
+                    "pattern": "^#/ckan/curse/[A-Za-z0-9_-]+$",
                     "description": "Deprecated format, no longer supported"
                 },
                 {
-                    "pattern": "^#\/ckan\/github\/[A-Za-z0-9_-]+\/[A-Za-z0-9_-]+(\/asset_match\/.+|\/version_from_asset\/.+)$",
+                    "pattern": "^#/ckan/github/[A-Za-z0-9_-]+/[A-Za-z0-9_-]+(/asset_match/.+|/version_from_asset/.+)?$",
                     "description": "Indicates that data should be fetched from the GitHub user and repo provided"
                 },
                 {
-                    "pattern": "^#\/ckan\/gitlab\/[A-Za-z0-9_-]+\/[A-Za-z0-9_-]+$",
+                    "pattern": "^#/ckan/gitlab/[A-Za-z0-9_-]+/[A-Za-z0-9_-]+$",
                     "description": "Indicates that data should be fetched from the GitLab the user and repo provided"
                 },
                 {
-                    "pattern": "^#\/ckan\/sourceforge\/.+$",
+                    "pattern": "^#/ckan/sourceforge/.+$",
                     "description": "Indicates that data should be fetched from SourceForge using the repo provided"
                 },
                 {
-                    "pattern": "^#\/ckan\/jenkins\/.+$",
+                    "pattern": "^#/ckan/jenkins/.+$",
                     "description": "Indicates that data should be fetched from a Jenkins CI server using the joburl provided"
                 },
                 {
-                    "pattern": "^#\/ckan\/http\/.+$",
+                    "pattern": "^#/ckan/http/.+$",
                     "description": "Indicates data should be fetched from an HTTP server, using the URL provided"
                 },
                 {
-                    "pattern": "^#\/ckan\/ksp-avc\/.+$",
+                    "pattern": "^#/ckan/ksp-avc/.+$",
                     "description": "Indicates that data should be fetched from a KSP-AVC .version file at the URL provided"
                 },
                 {
-                    "pattern": "^#\/ckan\/netkan\/.+$",
+                    "pattern": "^#/ckan/netkan/.+$",
                     "description": "Indicates that data should be fetched from the .netkan file at the given URL"
                 }
             ]
@@ -84,11 +84,11 @@
             "type": "string",
             "oneOf": [
                 {
-                    "pattern": "^#\/ckan\/ksp-avc(\/.*)$",
+                    "pattern": "^#/ckan/ksp-avc(/.*)?$",
                     "description": "Version information should be retrieved from an embedded KSP-AVC .version file inside the downloaded file"
                 },
                 {
-                    "pattern": "^#\/ckan\/space-warp(\/.*)$",
+                    "pattern": "^#/ckan/space-warp(/.*)?$",
                     "description": "Version information should be retrieved from an embedded SpaceWarp swinfo.json file inside the downloaded file"
                 }
             ]
--- a/NetKAN.schema
+++ b/NetKAN.schema
@@ -42,39 +42,39 @@
             "type": "string",
             "oneOf": [
                 {
-                    "pattern": "^#/ckan/spacedock/[0-9]+",
+                    "pattern": "^#\/ckan\/spacedock\/[0-9]+$",
                     "description": "Indicates that data should be fetched from SpaceDock"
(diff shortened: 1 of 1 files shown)


NetKAN.schema
```diff
@@ -42,39 +42,39 @@
...
                     "description": "Indicates that data should be fetched from SpaceDock"
                 },
                 {
-                    "pattern": "^#/ckan/curse/[A-Za-z0-9_-]+",
+                    "pattern": "^#\/ckan\/curse\/[A-Za-z0-9_-]+$",
                     "description": "Deprecated format, no longer supported"
                 },
                 {
-                    "pattern": "^#/ckan/github/[A-Za-z0-9_-]+/[A-Za-z0-9_-]+(/asset_match/.+|/version_from_asset/.+)?",
+                    "pattern": "^#\/ckan\/github\/[A-Za-z0-9_-]+\/[A-Za-z0-9_-]+(\/asset_match\/.+|\/version_from_asset\/.+)$",
```
HebaruSan, line 53:
```suggestion
                    "pattern": "^#/ckan/github/[A-Za-z0-9_-]+/[A-Za-z0-9_-]+(/asset_match/.+|/version_from_asset/.+)?$",
```

Removing the `?` makes the capture group **required**, which is wrong; it's perfectly allowed to leave out both `asset_match` and `version_from_asset`. E.g., this should be considered valid:

```json
{
  "identifier": "ReclaimKspDres",
  "$kref": "#/ckan/github/reclaim-ksp/reclaim-ksp-monorepo"
}
```

Luckily I don't think `?$` will confuse the parser.

NetKAN.schema
```diff
@@ -84,11 +84,11 @@
             "type": "string",
             "oneOf": [
                 {
-                    "pattern": "^#/ckan/ksp-avc(/.*)?",
+                    "pattern": "^#\/ckan\/ksp-avc(\/.*)$",
```
HebaruSan, line 87:
```suggestion
                    "pattern": "^#/ckan/ksp-avc(/.*)?$",
```

This capture group is optional, not required (which is why it's a group in the first place). This should be treated as valid:

```json
    "$vref": "#/ckan/ksp-avc",
```

NetKAN.schema
```diff
@@ -84,11 +84,11 @@
...
                     "description": "Version information should be retrieved from an embedded KSP-AVC .version file inside the downloaded file"
                 },
                 {
-                    "pattern": "^#/ckan/space-warp(/.*)?",
+                    "pattern": "^#\/ckan\/space-warp(\/.*)$",
```
HebaruSan, line 91:
```suggestion
                    "pattern": "^#/ckan/space-warp(/.*)?$",
```

This capture group is optional, not required. This should be treated as valid:

```json
    "$vref": "#/ckan/space-warp",
```

NetKAN.schema
```diff
@@ -42,39 +42,39 @@
             "type": "string",
             "oneOf": [
                 {
-                    "pattern": "^#/ckan/spacedock/[0-9]+",
+                    "pattern": "^#\/ckan\/spacedock\/[0-9]+$",
```
HebaruSan, line 45:
```suggestion
                    "pattern": "^#/ckan/spacedock/[0-9]+$",
```

I did look up the backslash forward slash thing before submitting the previous PR, and it seems to not be needed, since unlike with `sed` or Perl, there are no slashes around the outside of the regex.

- <https://stackoverflow.com/questions/1580647/json-why-are-forward-slashes-escaped>

I confirmed that by testing at <https://www.jsonschemavalidator.net/>.

Let's revert these to single forward slashes for concision and ease of reading (everywhere).

NetKAN.schema
```diff
@@ -42,39 +42,39 @@
...
                     "description": "Indicates that data should be fetched from SpaceDock"
                 },
                 {
-                    "pattern": "^#/ckan/curse/[A-Za-z0-9_-]+",
+                    "pattern": "^#\/ckan\/curse\/[A-Za-z0-9_-]+$",
```
HebaruSan, line 49:
```suggestion
                    "pattern": "^#/ckan/curse/[A-Za-z0-9_-]+$",
```

NetKAN.schema
```diff
@@ -42,39 +42,39 @@
...
                     "description": "Deprecated format, no longer supported"
                 },
                 {
-                    "pattern": "^#/ckan/github/[A-Za-z0-9_-]+/[A-Za-z0-9_-]+(/asset_match/.+|/version_from_asset/.+)?",
+                    "pattern": "^#\/ckan\/github\/[A-Za-z0-9_-]+\/[A-Za-z0-9_-]+(\/asset_match\/.+|\/version_from_asset\/.+)$",
                     "description": "Indicates that data should be fetched from the GitHub user and repo provided"
                 },
                 {
-                    "pattern": "^#/ckan/gitlab/[A-Za-z0-9_-]+/[A-Za-z0-9_-]+",
+                    "pattern": "^#\/ckan\/gitlab\/[A-Za-z0-9_-]+\/[A-Za-z0-9_-]+$",
```
HebaruSan, line 57:
```suggestion
                    "pattern": "^#/ckan/gitlab/[A-Za-z0-9_-]+/[A-Za-z0-9_-]+$",
```

NetKAN.schema
```diff
@@ -42,39 +42,39 @@
...
                     "description": "Indicates that data should be fetched from the GitLab the user and repo provided"
                 },
                 {
-                    "pattern": "^#/ckan/sourceforge/.+",
+                    "pattern": "^#\/ckan\/sourceforge\/.+$",
```
HebaruSan, line 61:
```suggestion
                    "pattern": "^#/ckan/sourceforge/.+$",
```

NetKAN.schema
```diff
@@ -42,39 +42,39 @@
...
                     "description": "Indicates that data should be fetched from SourceForge using the repo provided"
                 },
                 {
-                    "pattern": "^#/ckan/jenkins/.+",
+                    "pattern": "^#\/ckan\/jenkins\/.+$",
```
HebaruSan, line 65:
```suggestion
                    "pattern": "^#/ckan/jenkins/.+$",
```

NetKAN.schema
```diff
@@ -42,39 +42,39 @@
...
                     "description": "Indicates that data should be fetched from a Jenkins CI server using the joburl provided"
                 },
                 {
-                    "pattern": "^#/ckan/http/.+",
+                    "pattern": "^#\/ckan\/http\/.+$",
```
HebaruSan, line 69:
```suggestion
                    "pattern": "^#/ckan/http/.+$",
```

HebaruSan (changes requested):
Thanks for this! You're right, we need `$` to prevent the user from appending random extraneous text to an otherwise valid string in most cases.

NetKAN.schema
Danny2462: Ah, entirely true.
And, turns out, having `$` close off the regex is enough for the actual fix I opened this PR for: with `$` present, my example case above still works and the warning squiggle in VSCode is there.

So yes this is the correct change!

NetKAN.schema
Danny2462: Same as above, `$` fixes what I wanted to fix with `?`, lgtm

NetKAN.schema
Danny2462: Same as above, `$` fixes what I thought removing `?` would fix, lgtm.

NetKAN.schema
Danny2462: I based my superfluous backslashes entirely on https://regex101.com/ reporting faulty regex for JS (but it does work in VSCode anyway 🤷‍♂️), these would just make it consumable for more regex engines without change.

But this is not an important consideration and was just personal taste, so I will gladly defer to your pick!

HebaruSan (approved):
Thanks again! :bow_and_arrow:

PR #4299, opened 2025-02-06
Handle GitHub API rate limit errors appropriately

HebaruSan:
## Motivation

The bot has occasionally hit the GitHub API rate limit recently. When this happens, it keeps querying over and over regardless:

image

... which isn't what we're supposed to do:

<https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api#handle-rate-limit-errors-appropriately>

image

## Changes

Now if we get a rate limiting response from the GitHub API, we pause inflation for the time it requests and log that time span in the current mod's inflation error. This should allow the rate limit budget to recharge and keep us safer from being banned.

### Known limitations

Due to how the logging works, the messages about pausing will only appear _after_ the pause occurs. The Discord messages are sent by the Infra Python code, which receives them from AWS queue messages, which are batched in groups of 10. So if the third mod of a batch of 10 hits the limit, we'll pause for however long, then continue with mods 4–10 and finally send the message that alerts us to the pause.

I looked for a way around this, but did not find one. Maybe we can figure it out in the future.

### DIFF
--- a/Cmdline/Action/Install.cs
+++ b/Cmdline/Action/Install.cs
@@ -213,9 +213,9 @@ namespace CKAN.CmdLine
                     user.RaiseError("{0}", kraken.ToString());
                     return Exit.ERROR;
                 }
-                catch (DownloadThrottledKraken kraken)
+                catch (RequestThrottledKraken kraken)
                 {
-                    user.RaiseError("{0}", kraken.ToString());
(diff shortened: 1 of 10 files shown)


HebaruSan:
So far so good:

image

:crossed_fingers:

PR #4341, opened 2025-04-03
Add some missing Polish translations

3meraldK:
Doesn't cover 100% of missing translations, but these are 99% "most visited/read" ones, I guess


HebaruSan (approved):
Thanks so much for contributing! ☄️

PR #4344, opened 2025-04-06
Spec: clarify some info about how version is populated

JonnyOThan:


HebaruSan (approved):
Thanks for the update. 👍

PR #4363, opened 2025-05-06
Add download count to Metadata Tab

averageksp:
# Changes
- Added `DownloadCountTextBox.Text = gui_module.DownloadCount` into `Metadata.cs`
- Added `<data name="DownloadCountTextBox.Text" xml:space="preserve"><value>Downloads:</value></data>` for the Download Count to show up in the Metadata Tab
- Added `this.DownloadCountTextBox = new System.Window.Forms.Label();`
- Added `private System.Windows.Forms.TextBox DownloadCountTextBox;`
- Added `private CKAN.GUI.TransparentTextBox DownloadCountTextBox;`
- Some other changes were done too, but I didn't list them for now.

# Image
image

### DIFF
--- a/GUI/Controls/ModInfoTabs/Metadata.Designer.cs
+++ b/GUI/Controls/ModInfoTabs/Metadata.Designer.cs
@@ -38,6 +38,8 @@ namespace CKAN.GUI
             this.MetadataModuleLicenseTextBox = new CKAN.GUI.TransparentTextBox();
             this.AuthorLabel = new System.Windows.Forms.Label();
             this.AuthorsPanel = new System.Windows.Forms.FlowLayoutPanel();
+            this.DownloadCountLabel = new System.Windows.Forms.Label();
+            this.DownloadCountTextBox = new CKAN.GUI.TransparentTextBox();
             this.ReleaseLabel = new System.Windows.Forms.Label();
             this.MetadataModuleReleaseStatusTextBox = new CKAN.GUI.TransparentTextBox();
             this.GameCompatibilityLabel = new System.Windows.Forms.Label();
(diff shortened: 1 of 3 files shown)


GUI/Controls/ModInfoTabs/Metadata.cs
```diff
@@ -72,6 +72,18 @@ public void UpdateModInfo(GUIMod gui_module)
                     ReplacementTextBox.Text = module.replaced_by.ToString();
                 }
 
+                if (gui_module.DownloadCount == null)
+                {
+                    DownloadCountTextBox.Visible = false;
+                    MetadataTable.LayoutSettings.RowStyles[0].Height = 0;
```
SofieBrink, line 78: Are you sure this is the right row and height?

averageksp: Not sure, what should it be?

HebaruSan: The row should match the row you used for the new textbox, here:

https://github.com/KSP-CKAN/CKAN/blob/712db57fabc4df5d3268ca7b4064160085b51085/GUI/Controls/ModInfoTabs/Metadata.Designer.cs#L71

GUI/Controls/ModInfoTabs/Metadata.cs
```diff
@@ -72,10 +72,10 @@ public void UpdateModInfo(GUIMod gui_module)
                     ReplacementTextBox.Text = module.replaced_by.ToString();
                 }
 
-                if (gui_module.DownloadCount == null)
+                if (gui_module.DownloadCount == null or gui_module.DownloadCount == 0)
```
SofieBrink, line 75: `or` isn't valid C# syntax, you need ` || ` instead.
Are you using an IDE like Visual Studio or JetBrains Rider? it'd show these kinds of issues by underlining them in red before you post them.

GUI/Controls/ModInfoTabs/Metadata.cs
```diff
@@ -72,10 +72,10 @@ public void UpdateModInfo(GUIMod gui_module)
...
                 {
                     DownloadCountTextBox.Visible = false;
-                    MetadataTable.LayoutSettings.RowStyles[0].Height = 0;
+                    MetadataTable.LayoutSettings.RowStyles[0].Height = 3;
```
SofieBrink, line 78: I'm pretty sure the value being 0 when setting visble to false was correct, but you need the index of the DownloadCount row, which i believe would be 7 but do not quote me on that

GUI/Controls/ModInfoTabs/Metadata.cs
HebaruSan: If you want to use `or`, you can do this:

```suggestion
                if (gui_module.DownloadCount is null or 0)
```

GUI/Controls/ModInfoTabs/Metadata.cs
HebaruSan: The row is 3, and it goes between the brackets; the height goes after the equals sign:

```suggestion
                    MetadataTable.LayoutSettings.RowStyles[3].Height = 0;
```

GUI/Controls/ModInfoTabs/Metadata.cs
SofieBrink: > If you want to use `or`, you can do this:
What's this fancy new syntax lol, you learn something new every day.

HebaruSan: > What's this fancy new syntax lol, you learn something new every day.

Patterns! They can be confusing at first, but I've grown to love them:

- <https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/patterns>

GUI/Controls/ModInfoTabs/Metadata.Designer.cs
```diff
@@ -66,15 +68,17 @@ private void InitializeComponent()
...
-            this.MetadataTable.Controls.Add(this.ReplacementLabel, 0, 6);
-            this.MetadataTable.Controls.Add(this.ReplacementTextBox, 1, 6);
+            this.MetadataTable.Controls.Add(this.ReleaseLabel, 0, 4);
+            this.MetadataTable.Controls.Add(this.MetadataModuleReleaseStatusTextBox, 1, 4);
+            this.MetadataTable.Controls.Add(this.GameCompatibilityLabel, 0, 5);
+            this.MetadataTable.Controls.Add(this.MetadataModuleGameCompatibilityTextBox, 1, 5);
+            this.MetadataTable.Controls.Add(this.IdentifierLabel, 0, 6);
+            this.MetadataTable.Controls.Add(this.MetadataIdentifierTextBox, 1, 6);
+            this.MetadataTable.Controls.Add(this.ReplacementLabel, 0, 7);
+            this.MetadataTable.Controls.Add(this.ReplacementTextBox, 1, 7;
```
HebaruSan: A close parenthesis was lost here:

```suggestion
            this.MetadataTable.Controls.Add(this.ReplacementTextBox, 1, 7);
```

averageksp: Yes, I saw that. Has been fixed :)

GUI/Controls/ModInfoTabs/Metadata.cs
```diff
@@ -72,6 +72,19 @@ public void UpdateModInfo(GUIMod gui_module)
...
+                {
+                    DownloadCountLabel.Visible = false;
+                    DownloadCountTextBox.Visible = false;
+                    MetadataTable.LayoutSettings.RowStyles[3].Height = 0;
+                }
+                else
+                {
+                    DownloadCountLabel.Visible = true;
+                    DownloadCountTextBox.Visible = true;
+                    MetadataTable.LayoutSettings.RowStyles[3].Height = 0;
```
HebaruSan, line 85: The height of the row should be 30 when the count is supposed to be visible:

```suggestion
                    MetadataTable.LayoutSettings.RowStyles[3].Height = 30;
```

GUI/Controls/ModInfoTabs/Metadata.cs
```diff
@@ -31,22 +31,31 @@ public void UpdateModInfo(GUIMod gui_module)
                 MetadataTable.SuspendLayout();
                 MetadataModuleVersionTextBox.Text = gui_module.LatestVersion.ToString();
                 MetadataModuleLicenseTextBox.Text = string.Join(", ", module.license);
-
+                DownloadCountTextBox.Text = gui_module.DownloadCount.ToString() { Value = $"{mod.DownloadCountTextBox.Text:N0}" };
```
HebaruSan, line 34: `ToString()` doesn't accept a `{}` code block after it, `mod` isn't defined here, `GUIMod` doesn't have a `DownloadCountTextBox` property, and `Text` is the property we're trying to _set_, so we shouldn't reference it.

Try this:

```suggestion
                DownloadCountTextBox.Text = $"{gui_module.DownloadCount:N0}";
```

averageksp: Okay, but doesn't `DownloadCountTextBox.Text` be converted from `int` to `string` since that is what has been the issue before where `ToString()` wasn't there. Or will it be done automatically?

HebaruSan: The `$"{expression}"` syntax automatically converts the value of `expression` to a string.

averageksp: Awesome!

HebaruSan (approved):
Looks good! 👍

PR #4365, opened 2025-05-07
Use setter instead of reflection to set max transaction timeout

doinkythederp:
.NET 7 added a setter for the max transaction timeout, which means it's no longer necessary to use reflection on that runtime.


HebaruSan (approved):
Thanks! 🎉

PR #4381, opened 2025-05-28
Show recommendations of chosen recommendations in GUI, readable foreground colors for labels

### LINKED ISSUE #4031: [Feature]: Install Recommended Dependencies Recursively
### Problem

I am not sure if I am missing something obvious here, but lets say i want to install Realism Overhaul for the most recent KSP version.
If I select it, then click apply, it shows the mods that will be installed (RO, its required dependencies, and any recursively required dependenices).  That is fine.
Then It asks about installing recommended dependencies for RO.  I select all (barring conflicts).
Any recursively required dependencies needed for RO's recommended dependencies are also installed
But there is no opportunity to ask that the selected recommend dependencies also recursively install their respective recommended dependencies.
Why isn't this an option?

### Suggestion

Make it so that during the installation process, if a program has recommended dependencies, list them as you do now, but also have a tree like structure that shows the dependencies own recommended dependencies.  I feel like these should all be selected by default (since that is the behavior of the parent mod).

### Alternatives

_No response_

### Additional context

_No response_

### LINKED ISSUE #4379: [Feature]: Change label text color
### Problem

When I choose a dark background colour for my LABELS, the label text becomes impossible to read.

### Suggestion

It should be possible to choose a white or other 'light' colour for the text for that label, as the default black text becomes impossible to read.

### Alternatives

_No response_

### Additional context

_No response_

HebaruSan:
## Motivation

- Currently GUI only prompts to choose among recommendations, suggestions, and supporters once per install, for the initial changeset. However, if you choose to install any of them, _those_ mods might also have recommendations, suggestions, or supporters that you'd like to install, so it would be nice to show them as well. For comparison:
  -  CmdLine installs all recommendations-of-recommendations by default and has a `--no-recommends` option to turn that behavior off and a `--with-all-suggests` option to enable it for suggestions.
  - ConsoleUI shows the recommendations screen multiple times until the whole relationship tree is exhausted.
- Currently we only set the background color for labels, and the foreground color remains the default of the mod list or menu where the label is being shown. If you choose a color that's too similar to those default foreground colors, the text may not be readable.

## Changes

- Now after you choose to install recommendations, suggestions, or supporters in GUI, the mods you chose are then checked for further recommendations, suggestions, or supporters, and if there are any, the recommendations list is shown again, just like in ConsoleUI. Note that each mod will only be shown once across all such passes; once you decide you don't want a particular mod, it won't appear again.
  Fixes #4031.
- Now the foreground color of labels is set to white when the background color's brightness is less than 0.5 and black otherwise. Tests are added to ensure that several light and dark colors use a sensible foreground color (see https://learn.microsoft.com/en-us/dotnet/media/art-color-table.png).
  Fixes #4379.
- Now when you add or remove a mod from a label, the mod info header refreshes to reflect the change.
- Now the configured label color is shown in the mod info header, hidden label list, and the edit labels dialog. The foreground will be set according to the above brightness-based logic.
- Now if you right-click a label link, a context menu appears with an option to toggle whether the currently selected mod is included in that label.
- Now if you right-click a tag link, a context menu appears with an option to toggle whether mods with that tag are hidden in the mod list.

image
image
image
image image

## Known limitations

That "Label with a space" label's background color is `#8000FF`, sometimes called "[Electric Indigo](https://www.htmlcsscolor.com/hex/8000FF)". Its brightness happens to be 0.5 or slightly greater, so it is shown with black text, but it looks dark enough to my eye that white text would be better. I investigated this for a while and was not able to determine a way to programmatically distinguish such colors. (Increasing the threshold above 0.5 isn't feasible because there are several other 0.5-brightness colors that definitely should have black text.) I'm leaving it with black text for now because medium-brightness colors are still readable either way, just not optimally so.


HebaruSan:
I did find a function to calculate that while I was looking into this:

https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.platformui.colorutilities.getluminance

... but unfortunately it's in `Microsoft.VisualStudio.PlatformUI` instead of in the regular .NET runtime or something on Nuget we could redistribute. I don't love the idea of embedding more specialized algorithms and magic numbers into CKAN, but I suppose it's an option...

SofieBrink:
Hmm yeah that’s a bummer, i can’t find a real alternative to it either, at least not quickly

Issue #4353, opened 2025-04-19
[Feature]: Remove the Mono dependency on macOS


doinkythederp:
### Problem

Right now users on macOS need to install Mono independently of CKAN to use it. This is undesirable because it complicates the install process and doesn't support newer ARM devices without a translation layer (i.e. Rosetta 2).

### Suggestion

CKAN could switch to a self-contained .NET 8 build for Mac because it doesn't need the WinForms feature of Mono. This seems to be as simple as using `dotnet build` with the right flags:

```shell
dotnet publish -r osx-arm64 -f net8.0 --self-contained Cmdline/CKAN-Cmdline.csproj
dotnet publish -r osx-x64 -f net8.0 --self-contained Cmdline/CKAN-Cmdline.csproj

file ./_build/out/CKAN-CmdLine/Release/bin/net8.0/osx-arm64/publish/CKAN-CmdLine consoleui
# --> "Mach-O 64-bit executable arm64"

./_build/out/CKAN-CmdLine/Release/bin/net8.0/osx-arm64/publish/CKAN-CmdLine consoleui
```

I needed to add this to each csproj to prevent the duplicate `log4net.xml` files being an issue:
```xml
<ErrorOnDuplicatePublishOutputFiles>false</ErrorOnDuplicatePublishOutputFiles>
```

The entire `publish` directory would have to be shipped because the output isn't single file, but that's probably okay because Mac apps are folders anyways.

There might be some differences between Mono and .NET 8 that could cause bugs, so it'd likely be worth making sure everything is working first.

HebaruSan:
Interesting. The .dmg is built here, if you want to try a PR:

https://github.com/KSP-CKAN/CKAN/blob/4fc8117ccc0688837369a8bfb800ef408f64b9d3/macosx/Makefile#L1-L46

Clayell:
If and when this is fixed, the [KSPBurst Compiler description](https://github.com/KSP-CKAN/NetKAN/blob/aedc1ee22910ee2cc54f751f9d323a32dcfe63a7/NetKAN/KSPBurst.netkan#L5) should be updated. (and possibly [Harmony](https://github.com/KSP-CKAN/NetKAN/blob/aedc1ee22910ee2cc54f751f9d323a32dcfe63a7/NetKAN/Harmony2.frozen)?)
> Requires Mono on Linux and macOS (you have it if you're running CKAN).

PR #4387, opened 2025-06-07
Switch Inflator from Mono to dotnet

HebaruSan:
## Motivation

While auditing Netkan code to try to make progress on the array bounds errors that occasionally show up in the status page, I noticed that:

- `netkan.exe --releases all` wouldn't work because `all` was translated into `null` which later was defaulted to `1` in the chain of processing of that option.
- It would be nice to be able to pass multiple netkans on the command line of `netkan.exe`.
- #4353's idea of switching from Mono to dotnet, while problematic for the Mac build because I don't have a Mac on which to test it, could be piloted in the Inflator container much more easily.

## Problem

See KSP-CKAN/NetKAN#10588, the progress bar for removing HideEmptyTechNodes looks weird, showing negative bytes remaining and greater than 100% completion:

image

Reported by Discord user Shadlock.

## Cause

If a module's `install_size` doesn't match up with what's on disk, then we end up removing a different number of bytes than expected. Bytes remaining is calculated naïvely by subtracting bytes completed from `install_size`, which can be negative if there are too many bytes. The percent completed works similarly.

## Changes

- Now the Inflator container uses dotnet instead of Mono
  - The build scripts now run `dotnet publish` for Netkan to generate a self-contained app for net8.0 and linux-x64
  - `Dockerfile.netkan` now uses `ubuntu:latest` again because it no longer installs Mono, it installs the entire net8.0 `publish` folder for Netkan into `/home/netkan`, and its `ENTRYPOINT` now runs the dotnet build of `CKAN-NetKAN`
  - The workflow that builds the container now retrieves the artifact containing `_build/out` instead of `_build/repack`, and runs the Docker build in `_build/out/CKAN-NetKAN/Release/bin/net8.0/linux-x64/publish` instead of `_build/repack/Release`
  - One potential advantage of dotnet is ["trimming"](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trim-self-contained), which can shrink the container by excluding assemblies and even _parts of_ assemblies that aren't needed (including system assemblies). I experimented with this, but it consistently broke the app. Instead, a `.dockerignore` file contains all the DLLs that I was able to confirm are not needed, as a halfway measure until we are able to get real trimming to work.
- Now the metadata tester container installs a config file at `/usr/local/etc/metadata.ini` as specified in KSP-CKAN/xKAN-meta_testing#101, which we can use later to make it use dotnet instead of Mono (which is **not** attempted here).
- Now the `--releases all` option will properly inflate all releases.
- Now you can pass as many netkans as you like to `netkan.exe`, and all will be inflated.
- Now `ModuleInstaller` sanity checks the numbers that it sends out for status updates, so negative bytes remaining will no longer be seen for mods with bigger than expected files.
- The Netkan error about "Generated 2 modules but only 1 requested" will now append "(pre-release)" after versions which are pre-releases
- TLS1.3 is now force-allowed in addition to TLS1.2, just in case it becomes necessary at some point in the future (see #2297).

Note that we may not get a new dev build after merging this, depending on whether the SignPath action continues to error out as in SignPath/github-action-submit-signing-request#9.


HebaruSan:
All errors are being filed as "Permission denied" because the `log4net.xml` file is there despite being in the `.dockerignore` file, which seems to be completely missing in action somehow when built on GitHub, but very much present when I try it locally. :facepalm: 

image

HebaruSan:
... and this is why.

<https://github.com/actions/upload-artifact/blob/v4/docs/MIGRATION.md#hidden-files>

(In Unix, files that begin with `.` are considered "hidden" because `ls` won't display them without the `-a` or `-A` flags.)

PR #4388, opened 2025-06-08
Fix 'netkan' user in Inflator

techman83:
ubuntu:latest has an ubuntu user by default, which changes the uid of the user we create. This causes issues with shared volume permissions. This renames the ubuntu user so we can be aligned with the rest of our containers.


HebaruSan (approved):
:bow:

techman83:
Spacedock wonky, changes fix currently broken container.

PR #4392, opened 2025-06-14
Writethrough when saving files, add Netkan tests

HebaruSan:
## Motivations

- Since f33b9011e7cce669865e8b4426910b4948f6544b, we have known about the possibility that `GUIConfig.xml` could become corrupted. Generally this takes the form of the entire file being populated by `NUL` characters on Windows, and we've seen it a few times for JSON files as well. Some browsing of StackExchange indicated that this may be due to filesystem caching and buffering of various sorts delaying the file from being written past the point of something that stops it.
- Now that we have coverage tracking as of #4383, [Coveralls makes it possible to browse the source tree](https://coveralls.io/github/KSP-CKAN/CKAN) looking for untested code. It would be nice to get our overall coverage as close to 100% as possible, and currently it's hovering around a fairly pathetic 30.3%. There are several files in `Netkan/Sources` and `Netkan/Transformers` that have no coverage at all currently and can be tested fairly easily.
- Coveralls currently includes `_build`, `Tests`, and `ConsoleUI` in the coverage stats, which is not optimal:
  -  `_build` isn't a real code folder.
  -  `Tests` does not contain any end-user app code.
  - `ConsoleUI` is purely UI and not testable because the difference between good or bad outcomes are things like, does that look right? does it make sense to a human?
    (A similar argument could be made for GUI, but there we do actually have a few meaningful tests and could conceivably add more, especially of the `Model` classes.)

## Changes

- Now we use `FileOptions.WriteThrough` to _strongly encourage_ Windows to actually write our various JSON files to disk when we ask it to. This might conceivably slow down response times slightly in some cases, but that trade-off is worth the benefit of greater robustness.
- New tests are added for several Netkan transformer classes.
- Tests involving hosts other than SpaceDock are removed from the `FlakyNetwork` category to increase coverage. Of all the servers we use, SpaceDock is currently the only one truly deserving of that label.
- Some existing tests of Netkan transformer classes are refactored to increase their coverage, chiefly by replacing the mockup of API objects with mockups of `IHttpService` objects that are then _used_ by the real API objects.
- The `ConsoleUI`, `Tests`, and `_build` folders are now excluded from Coveralls (hopefully).

### DIFF
--- a/Core/Configuration/JsonConfiguration.cs
+++ b/Core/Configuration/JsonConfiguration.cs
@@ -295,8 +295,8 @@ namespace CKAN.Configuration
         // </summary>
         private void SaveConfig()
         {
-            File.WriteAllText(configFile,
-                              JsonConvert.SerializeObject(config, Formatting.Indented));
+            JsonConvert.SerializeObject(config, Formatting.Indented)
+                       .WriteThroughTo(configFile);
         }
 
         /// <summary>
--- a/Core/Configuration/StabilityToleranceConfig.cs
+++ b/Core/Configuration/StabilityToleranceConfig.cs
@@ -5,6 +5,8 @@ using System.ComponentModel;
 
 using Newtonsoft.Json;
 
+using CKAN.Extensions;
+
 namespace CKAN.Configuration
 {
     [JsonObject(MemberSerialization.OptIn)]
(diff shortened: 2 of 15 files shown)


HebaruSan:
I'm sorry that you're mad, @coveralls, but the previous number was inflated due to the former inclusion of `Tests`.
27% is a more accurate baseline, and the effort to increase coverage in specific areas of Netkan was mildly successful.

image  &rArr;  image

PR #4410, opened 2025-07-30
Migrate Mac app bundle from Mono to dotnet

### LINKED ISSUE #4353: [Feature]: Remove the Mono dependency on macOS

HebaruSan:
## Motivation

- .NET Framework 4.8 was released in April 2019 and updated to 4.8.1 in August 2022. No further non-security updates are planned. We depend on it for the WinForms GUI, but for anything else it would be better to rely on code that's still supported and maintained.
- Mono, which runs .NET Framework code on Linux and Mac (and is still the _only_ way to run WinForms code on those platforms), has been all but abandoned and donated to WineHQ, who have been [maintaining it somewhat](https://gitlab.winehq.org/mono/mono/-/releases) but [not providing releases via the established channels](https://gitlab.winehq.org/mono/mono/-/issues/12).
- The last version of Mac OS X that was able to run the GUI (because it could run Mono in 32-bit mode) was 10.14, released September 2018 and ended support in November 2021. It has not received security updates in almost 4 years. It is very unlikely that any users can currently run the CKAN GUI on Mac, so our reason for sticking with Mono no longer applies. It would be nice to eliminate the requirement to install Mono.
- .NET (the post-"Framework" framework) continues marching along with LTS version 8.0 released in November 2023, 9.0 out in November of 2024, both of them updated recently, and future LTS version 10.0 available in a preview status. Each of these releases represents fixes, performance improvements, and new features that might be useful to us, while Framework and Mono just sit there bit-rotting.
- We have already migrated the Inflator (see #4387) and the metadata tester (see #4390) from Mono to dotnet, and it's working OK so far.

## Changes

- Now instead of a ~5 MiB .NET Framework executable to be run with Mono, `CKAN.app` contains a ~160 MiB native Mac build (for two architectures) that needs no external dependencies to run
  - Now `./build.sh osx` runs `dotnet publish` for the `osx-arm64` and `osx-x64` runtimes before it builds `CKAN.dmg`
  - `macosx/Makefile` now copies the `osx-arm64` and `osx-x64` directories into `_build/osx/dmg/CKAN.app/Contents/MacOS` and replaces duplicated files with relative symlinks
  - The Mono-specific logic is removed from the `CKAN` script for Mac and replaced by commands to launch the `CKAN-CmdLine` executable for the correct local system architecture
  - The `ErrorOnDuplicatePublishOutputFiles` property is set in the project files for CmdLine and Netkan so we don't have to exclude `log4net.xml` from being copied into the output dir
- Now the `build.yml` workflow builds `CKAN.dmg` and uploads it as an artifact with 7 retention days so Mac users can test this pull request, and the steps to build it in `deploy.yml` and `release.yml` are updated to download the artifact instead
  - Retention days for intermediate build artifacts are changed from 7 to 1 because they're not needed after the build finishes

Fixes #4353.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,13 +37,18 @@ jobs:
           key: nuget-oldref-modules-${{ hashFiles('**/packages.config') }}-${{ hashFiles('**/*.csproj') }}
       - name: Build ckan.exe and netkan.exe
         run: ./build.sh --configuration=${{ inputs.configuration }}
+      - name: Build CKAN.dmg
+        if: inputs.configuration == 'Release'
+        run: |
+          sudo apt-get install -y libplist-utils xorriso rdfind symlinks
+          ./build.sh osx --configuration=${{ inputs.configuration }} --exclusive
       - name: Upload repack artifact
         id: upload-repack-artifact
         uses: actions/upload-artifact@v4
         with:
           name: ${{ inputs.configuration }}-repack-unsigned
           path: _build/repack/
-          retention-days: 7
+          retention-days: 1
       - name: Upload out artifact
         uses: actions/upload-artifact@v4
         with:
@@ -61,11 +66,18 @@ jobs:
         with:
           name: unsigned
           path: _build/signpath/Release
-          retention-days: 7
+          retention-days: 1
       - name: Upload ckan.exe artifact
         if: inputs.configuration == 'Release'
         uses: actions/upload-artifact@v4
         with:
           name: ckan.exe
           path: _build/repack/Release/ckan.exe
+          retention-days: 7
+      - name: Upload CKAN.dmg artifact
+        if: inputs.configuration == 'Release'
+        uses: actions/upload-artifact@v4
+        with:
+          name: CKAN.dmg
+          path: _build/osx/CKAN.dmg
           retention-days: 7
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,11 +37,25 @@ jobs:
           key: nuget-oldref-modules-${{ hashFiles('**/packages.config') }}-${{ hashFiles('**/*.csproj') }}
       - name: Build ckan.exe and netkan.exe
         run: ./build.sh --configuration=${{ inputs.configuration }}
+      - name: Build CKAN.app
+        if: inputs.configuration == 'Release'
+        run: |
+          sudo apt-get install -y libplist-utils rdfind symlinks genisoimage
+          ./build.sh osx-app --configuration=${{ inputs.configuration }} --exclusive
+      - name: Ad hoc codesign x86_64 binary
+        if: inputs.configuration == 'Release'
+        uses: indygreg/apple-code-sign-action@v1
+        with:
+          input_path: _build/osx/dmg/CKAN.app/Contents/MacOS/x86_64/CKAN-CmdLine
+      - name: Ad hoc codesign arm64 binary
+        if: inputs.configuration == 'Release'
+        uses: indygreg/apple-code-sign-action@v1
+        with:
+          input_path: _build/osx/dmg/CKAN.app/Contents/MacOS/arm64/CKAN-CmdLine
       - name: Build CKAN.dmg
         if: inputs.configuration == 'Release'
         run: |
-          sudo apt-get install -y libplist-utils xorriso rdfind symlinks
-          ./build.sh osx --configuration=${{ inputs.configuration }} --exclusive
+          ./build.sh osx-dmg --configuration=${{ inputs.configuration }} --exclusive
       - name: Upload repack artifact
         id: upload-repack-artifact
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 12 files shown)


HebaruSan:
Needs testing by Mac users; requested help on Discord.

HebaruSan:
Thanks to @lewisfm's testing efforts, we discovered that we need to figure out Apple code signing for this to be usable:

```
kernel    AMFI: hook..execve() killing zsh (pid 56647): Attempt to execute completely unsigned code (must be at least ad-hoc signed).
kernel    ASP: Sleep interrupted: ref 25242, signal 0x100
kernel    ASP: Security policy would not allow process: 56647, /Users/user/Downloads/CKAN.app/Contents/MacOS/arm64/CKAN-CmdLine
```

We might be able to use https://github.com/indygreg/apple-code-sign-action for that:

- https://gregoryszorc.com/docs/apple-codesign/stable/apple_codesign_github_actions.html

Signpath has some documentation about Mac apps, but it looks much more involved than the Windows signing that we're already doing with their tools:

- https://about.signpath.io/blog/2024/11/29/building-trusted-software-for-macos

HebaruSan:
image

It's trying to sign the bash script; `23212f62` are the hex codes for `#!/b`. So we need to specify the paths to the binaries.

HebaruSan:
@lewisfm, I took your PR as a jumping off point to implement ad hoc codesigning using the GitHub Action noted earlier, and it's finally to the point of uploading the artifact (the failed job here is an unrelated complaint from Coveralls about being overloaded):

https://github.com/KSP-CKAN/CKAN/actions/runs/16920418421/artifacts/3749012406

Can you please check whether this build works for you?

lewisfm:
This worked as expected for me! However, the system is caching the app as needing `x86_64` emulation because this is the first version to have an `arm64` version. I was able to clear the cache by running this command and then rebooting:
```sh
/usr/libexec/PlistBuddy -c 'Delete "Architectures for arm64":org.ksp-ckan.ckan' ~/Library/Preferences/com.apple.LaunchServices/com.apple.LaunchServices.plist
```

Alternatively, CKAN could add these lines to `Info.plist` to make systems that support arm64 use that version by default, removing the need for clearing the cache:
```xml
<key>LSArchitecturePriority</key>
<array>
    <string>arm64</string>
</array>
```

lewisfm:
I think Keys.cs needs to be updated to check for Mono, not Windows:
```cs
        public static readonly ConsoleKeyInfo Enter = new ConsoleKeyInfo(
            Platform.IsMono ? LinuxEnter : WindowsEnter,
            ConsoleKey.Enter, false, false, false
        );
```

HebaruSan:
> However, the system is caching the app as needing `x86_64` emulation because this is the first version to have an `arm64` version.

Does that actually matter somehow? As long as it _runs,_ a poorly managed cache seems like Apple's problem to solve rather than ours.

lewisfm:
> Does that actually matter somehow? As long as it _runs,_ a poorly managed cache seems like Apple's problem to solve rather than ours.

I haven't tested it, but my [guess](https://apple.stackexchange.com/q/452070) is that the current setup will ask users to install the Rosetta library even though they don't need it. Also, setting this key will improve performance with nearly no downsides.

HebaruSan:
> Also, setting this key will improve performance with nearly no downsides.

Even on non-arm64 architectures? Setting one architecture as a higher priority just seems like the opposite of what we should do, when what we really want is for their priorities to be equal.

lewisfm:
> Even on non-arm64 architectures? Setting one architecture as a higher priority just seems like the opposite of what we should do, when what we really want is for their priorities to be equal.

Since there's no arm64-to-x86_64 translation on macOS, this key will just be ignored on x86_64, making it fairly harmless. This might be an issue if this feature was added in the future, but my guess is that it's unlikely since Apple is phasing out full x86_64 support after the next operating system release.

HebaruSan:
Also [this person](https://apple.stackexchange.com/a/473929) said it didn't work.

lewisfm:
> Also [this person](https://apple.stackexchange.com/a/473929) said it didn't work.

Oh. In that case it's probably not worth the trouble. If you think this is worth it: we could put the command for clearing the cache somewhere in an FAQ or something.

PR #4426, opened 2025-09-01
A few more tests

HebaruSan:
## Motivation

[Coverage](https://coveralls.io/github/KSP-CKAN/CKAN) says we're up to 67%, which is an improvement, but not comfortable enough to be an end state.

## Changes

New tests are added for:

- Deduplication during installation
- `ckan import`
- `NameComparer`
  (used only for printing mods after `repo update`, and required sorting them alphanumerically)
- `GUI.EmbeddedImages`
- `GUI.ReleaseStatusItem`
- `NetKAN.CmdLineOptions`
- `NetKAN.ModuleService.GetSpaceWarpInfo`
  (FYI to @JohnsterSpaceProgramOfficial, we're borrowing your `BurnController-0.8.1.zip` for testing, since it is MIT-licensed. We'll just be using it to confirm that we can extract and parse the `swinfo.json` correctly.)

More untestable code is excluded from coverage:

- `GUI.ControlFactory`
- `GUI.FormCompatibility`
- `GUI.WinFormsExtensions`
- `GUI.GUI`
- `GUI.X11`

Unused code is deleted:

- `InconsistentKraken` handler in `CmdLine.Main`
- `ServiceLocator.Container`'s setter
- `GUI/Properties/Settings.*`


HebaruSan:
> [Coverage Status](https://coveralls.io/builds/75336350)

Nice! 🎉

PR #4439, opened 2025-09-17
Even more tests

HebaruSan:
## Motivation

Another batch of updates to try to boost our [Coveralls](https://coveralls.io/github/KSP-CKAN/CKAN) stats.

## Changes

New or expanded tests:

- `ckan stability set --mod`
- `ckan replace --all`
- `ckan show` with more metadata in the test mod
- `GUI.ModChange`
- `NetKAN.GithubRelease.SourceArchiveAsset`
- Uninstalling a mod that installs crafts into `Ships`
- `SpacedockTransformer`
- `SpaceWarpInfoValidator`

Untestable code excluded from coverage:

- CmdLine help text printing code
- URL launching code
- `GUI.PluginController` because initializing a plugin implementing the `IGUIPlugin` from `ckan.exe` isn't possible outside of `ckan.exe`
  (I had to fork and update <https://github.com/KSP-CKAN/CKAN-SamplePlugins> to discover this.)

Side changes:

- `GUIMod` has an `IsCKAN` property that's always `true`, which makes `ToCkanModule()` redundant with `ToModule()`.
  Now `IsCKAN` is removed and the others are replaced by a single `Module` property.
- `Logging.Initialize()` was being called in `DisposableKSP`'s constructor, which meant that whether `log4net.xml` was applied depended on whether we were running a test that needs a fake game instance, which confounded some of my investigations using `--where=` to run only certain tests.
  Now the `Logging.Initialize()` is moved to a global test setup function tagged with `[OneTimeSetUp]` inside a class tagged with `[SetUpFixture]`.
- `OverrideOptions` had a `ToString` override that just serialized the object, which wasn't used.
  Now this is removed.
- `UnixDateTimeMillisecondsConverter` (used for handling Jenkins) had a `WriteJson` override that wasn't used.
  Now this is removed.

### DIFF
--- a/Cmdline/Action/AuthToken.cs
+++ b/Cmdline/Action/AuthToken.cs
@@ -152,6 +152,7 @@ namespace CKAN.CmdLine
             return ht;
         }
 
+        [ExcludeFromCodeCoverage]
         public static IEnumerable<string> GetHelp(string verb)
         {
             // Add a usage prefix line
--- a/Cmdline/Action/Cache.cs
+++ b/Cmdline/Action/Cache.cs
@@ -41,6 +41,7 @@ namespace CKAN.CmdLine
             return ht;
         }
 
+        [ExcludeFromCodeCoverage]
         public static IEnumerable<string> GetHelp(string verb)
         {
             // Add a usage prefix line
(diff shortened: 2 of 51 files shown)


HebaruSan:
The tests are passing, but the workflows report failure because of lemurheavy/coveralls-public#1757.

PR #4446, opened 2025-10-03
Refactor and enhance tests

### LINKED ISSUE #4445: [Bug]: Surprising recommendations when switching from parallax to parallax continued

### Operating System

Windows 10

### CKAN Version

1.36.1.25271

### Games

KSP 1

### Game Version

1.12.5

no

### Describe the bug

I accidentally installed parallax 2, and then when I tried to switch to parallax continued I was prompted with recommendations related to 1thousandspecialplanetpack

And I didn't realize until just now, but I'm blocked from continuing at this point:

Image

### Steps to reproduce

1. fully install parallax 2 with ckan
2. uncheck parallax 2, check parallax continued
3. hit apply

Image

Image

What is this doing here??

Image

### Relevant log output

_No response_

HebaruSan:
## Copilot's summary

This update improves the handling of module relationships, refines the import tests for KSP2, and removes unused code. It also standardizes string handling and enhances the configuration interface.

## Motivation

[Coveralls](https://coveralls.io/github/KSP-CKAN/CKAN) still shows a few bits of low hanging test fruit.

## Problem

If you install Parallax, then try to remove it and install ParallaxContinued in the same changeset, recommendations of 1ThousandSpecialPlanetPack will be shown, which is confusing because you aren't installing it and it isn't shown anywhere in the UI up to that point.

## Cause

The recommendations logic doesn't factor in mods that are being removed, so when it runs, Parallax and the Parallax texture modules are treated as still installed. The ParallaxContinued texture modules conflict with all of the Parallax modules, so the recommendations logic concludes they can't be installed, which is wrong. However, 1ThousandSpecialPlanetPack's textures provide all of the ParallaxContinued texture modules, and there are no conflicts blocking it, so it gets pulled into the list of recommenders.

## Changes

- Now `ModuleInstaller.FindRecommendations` has a `toRemove` parameter that it uses to exclude mods from the changeset (plus autoremovable mods), so the ParallaxContinued texture modules will no longer be excluded from the changeset, and 1ThousandSpecialPlanetPack won't be pulled in.
  A test is created that fails with the old code and passes with the new fixed code.
  Fixes #4445.

Tests are added or improved for:

- `HardLink.CreateOrCopy` - was making hardlinks into `Tests/Data`, which could trip over themselves when tests ran in parallel for `net481`, `net8.0`, and `netstandard2.0`, so now the test copies the file into a (thread-specific) temp dir before linking
- `EnumerableExtensions.Sum<TimeSpan>`
- `AvcKrefTransformer` with `GITHUB` object defined in the version file
- `ckan show` with partial match of installed modules (previously not supported)
- Importing modules for KSP2 using their internal .ckan files and `swinfo.json`
- `RelationshipResolver` with dependencies that need the same identifier with different versions. This previously had some code that became unreachable when `ResolvedRelationshipsTree` was added, which is now removed and replaced by some enhancements to the tree class that report the full chain of relationships

Some existing code is moved to facilitate the above:

- `DependsAndConflictsOK` from static in `AvailableModule` to instance method of `CkanModule`, and refactored into `CkanModule.BadRelationships` that tells you what went wrong
- `RelationshipType` from `GUI` to `Core` (used in new `Relationship` class needed for `CkanModule.BadRelationships`)

Some unused code is removed:

- `EnumerableExtensions.ZipMany`
- `EnumerableExtensions.WithMatches`
- `CkanModule.serialise`
- `ModuleInstallDescriptor`'s `ICloneable` implementation
- The `returnValue:` parameter prefix of `NotNullWhen`, which is the _only_ parameter


HebaruSan:
I want to look at the details of coverage for `ModuleImporter` because it looks like the test for the `SpaceWarpInfo` applicator may not be running, but lemurheavy/coveralls-public#1757 is still happening.

... Coveralls is still broken, so I finally got coverage data collection set up on Windows. Yes, the `SpaceWarpInfo` stuff wasn't running because the internal .ckan had no `$vref`, so I fixed that.

PR #4449, opened 2025-10-13
Build on Windows, upload multi-platform coverage

HebaruSan:
## Motivation

Currently we test and upload coverage data only for Linux. This means that our Windows-specific code isn't tested and falsely reported as not covered.

## Changes

- Now the steps in `build.yml` that build `CKAN.dmg` are refactored into a separate job with a consolidated `Release` configuration check
- Now `test.yml` uses a matrix strategy to run tests and upload coverage on both Windows and Linux, using [parallel builds support for Coveralls](https://docs.coveralls.io/parallel-builds). Since AltCover seems unable to handle assemblies built on a different OS, or a different Windows container when running on Windows, the Windows job does a complete build+test rather than trying to re-use the artifacts from `build.yml`.
- A new `coveralls-finish` job finalizes the coverage upload with `parallel-finished: true`.
  - This job then runs the [`rerun_build` workaround](https://github.com/search?q=https%3A%2F%2Fcoveralls.io%2Frerun_build&type=code) from lemurheavy/coveralls-public#1629 to try to get Coveralls to generate accurate numbers for the merged results (untested as of yet because the secrets are not available in the current fork, so I will be replacing this PR with a new one shortly)
- The [environment variable `DOTNET_CLI_TELEMETRY_OPTOUT`](https://learn.microsoft.com/en-us/dotnet/core/tools/telemetry) is set to `true` to prevent `dotnet` from "collect[ing] usage data and send[ing] it to Microsoft"
- The [`<IsTestProject>` property](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#istestproject) is set to `true` in `Tests.csproj`
- The check that exempts Linux containers from the admin user check is extended to also work in Windows containers (at least the GitHub ones)
- The Mac DMG is only generated in `Release` configuration on Linux, but formerly the needed files were _published_ in `Debug` configuration as well. Now both steps are `Release`-only.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,11 +23,6 @@ jobs:
         with:
           path: _build/tools
           key: build-tools-${{ hashFiles('build.sh', 'build.ps1', 'build/*') }}
-      - name: Restore cache for _build/cake
-        uses: actions/cache@v4
-        with:
-          path: _build/cake
-          key: build-cake-${{ hashFiles('build/*') }}
       - name: Restore cache for _build/lib/nuget
         uses: actions/cache@v4
         with:
@@ -37,25 +32,6 @@ jobs:
           key: nuget-oldref-modules-${{ hashFiles('**/packages.config') }}-${{ hashFiles('**/*.csproj') }}
       - name: Build ckan.exe and netkan.exe
         run: ./build.sh --configuration=${{ inputs.configuration }}
-      - name: Build CKAN.app
-        if: inputs.configuration == 'Release'
-        run: |
-          sudo apt-get install -y libplist-utils rdfind symlinks genisoimage
-          ./build.sh osx-app --configuration=${{ inputs.configuration }} --exclusive
-      - name: Ad hoc codesign x86_64 binary
-        if: inputs.configuration == 'Release'
-        uses: indygreg/apple-code-sign-action@v1
-        with:
-          input_path: _build/osx/dmg/CKAN.app/Contents/MacOS/x86_64/CKAN-CmdLine
-      - name: Ad hoc codesign arm64 binary
-        if: inputs.configuration == 'Release'
-        uses: indygreg/apple-code-sign-action@v1
-        with:
-          input_path: _build/osx/dmg/CKAN.app/Contents/MacOS/arm64/CKAN-CmdLine
-      - name: Build CKAN.dmg
-        if: inputs.configuration == 'Release'
-        run: |
-          ./build.sh osx-dmg --configuration=${{ inputs.configuration }} --exclusive
       - name: Upload repack artifact
         id: upload-repack-artifact
         uses: actions/upload-artifact@v4
@@ -88,8 +64,41 @@ jobs:
           name: ckan.exe
           path: _build/repack/Release/ckan.exe
           retention-days: 7
+
+  build-osx:
+    needs:
+      - build
+    runs-on: ubuntu-22.04
+    if: inputs.configuration == 'Release'
+    steps:
+      - uses: actions/checkout@v4
+      - name: Restore cache for _build/tools
+        uses: actions/cache@v4
+        with:
+          path: _build/tools
+          key: build-tools-${{ hashFiles('build.sh', 'build.ps1', 'build/*') }}
+      - name: Download out artifact
+        uses: actions/download-artifact@v5
+        with:
+          name: ${{ inputs.configuration }}-out-unsigned
+          path: _build/out/
+      - name: Build CKAN.app
+        run: |
+          chmod a+x _build/out/Build/*/bin/net8.0/Build _build/out/Build/*/obj/*/net8.0/apphost
+          sudo apt-get install -y libplist-utils rdfind symlinks genisoimage
+          ./build.sh osx-app --configuration=${{ inputs.configuration }} --exclusive
+      - name: Ad hoc codesign x86_64 binary
+        uses: indygreg/apple-code-sign-action@v1
+        with:
+          input_path: _build/osx/dmg/CKAN.app/Contents/MacOS/x86_64/CKAN-CmdLine
+      - name: Ad hoc codesign arm64 binary
+        uses: indygreg/apple-code-sign-action@v1
+        with:
+          input_path: _build/osx/dmg/CKAN.app/Contents/MacOS/arm64/CKAN-CmdLine
+      - name: Build CKAN.dmg
+        run: |
+          ./build.sh osx-dmg --configuration=${{ inputs.configuration }} --exclusive
       - name: Upload CKAN.dmg artifact
-        if: inputs.configuration == 'Release'
         uses: actions/upload-artifact@v4
         with:
           name: CKAN.dmg
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,37 +12,83 @@ on:
 jobs:
   build-debug:
     uses: ./.github/workflows/build.yml
-  test-build:
+  run-tests:
     needs: build-debug
-    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - windows-latest
+          - ubuntu-22.04
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
-      - name: Install runtime dependencies
-        run: sudo apt-get install -y xvfb
-      - name: Download out artifact
-        uses: actions/download-artifact@v4
+      - name: Restore cache for _build/tools
+        uses: actions/cache@v4
+        with:
+          path: _build/tools
+          key: build-tools-${{ hashFiles('build.sh', 'build.ps1', 'build/*') }}
+      - name: Download out artifact on Linux
+        if: matrix.os == 'ubuntu-22.04'
+        uses: actions/download-artifact@v5
         with:
           name: Debug-out-unsigned
           path: _build/out/
       - name: Restore permissions discarded by zip/unzip of artifact
+        if: matrix.os == 'ubuntu-22.04'
         run: |
           chmod a+x _build/out/Build/*/bin/net8.0/Build _build/out/Build/*/obj/*/net8.0/apphost
-      - name: Download repack artifact
-        uses: actions/download-artifact@v4
+      - name: Download repack artifact on Linux (for smoke tests)
+        if: matrix.os == 'ubuntu-22.04'
+        uses: actions/download-artifact@v5
         with:
           name: Debug-repack-unsigned
           path: _build/repack/
-      - name: Run tests
-        run: xvfb-run ./build.sh test+only --configuration=Debug --where="Category!=FlakyNetwork"
+      - name: Run tests on Linux
+        if: matrix.os == 'ubuntu-22.04'
+        run: |
+          sudo apt-get install -y xvfb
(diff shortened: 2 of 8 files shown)


HebaruSan:
Summing up the errors that are being reported...

- The Windows job fails at smoke testing, apparently due to admin user, even though we added `--asroot`:
  image
  Looking more closely, it says "Warning", so maybe the problem is just some extra output.
  Actually, we should probably extend this `.dockerenv` check to work on Windows, [however its containers are flagged](https://stackoverflow.com/questions/43002803/detect-if-process-executes-inside-a-windows-container):
  https://github.com/KSP-CKAN/CKAN/blob/d4cfe90a8b57286e1dbd79b6d27f809f401cd4f1/Cmdline/Options.cs#L252-L254
- The Linux job runs the tests, which all pass, but no coverage data is collected, and the call to Coveralls fails:
  image
  image

HebaruSan:
Got past the admin user stuff on Windows. Now it runs for 2 minutes, only 4 seconds of which was spent in the actual build:

image

Just Windows being slow? Presumably the tests did not actually run, since they typically take a minute per target framework. Why no output explaining why? Does the "Workload updates..." message relate to the problem, or is it just noise?

I think the Windows container is showing us what is printed to `stderr` but not `stdout`, which would be very surprising and extremely undesirable.

HebaruSan:
Finally got everything to finish successfully, but it's not in a usable state:

- To get the Windows tests to run, I had to remove the `NoRestore` and `NoBuild` flags, which means it's duplicating all the work of the build step. Need to figure out why copying the `_build/out` folder from one job to the next isn't sufficient.
  On closer inspection, the Windows data is truly bizarre. E.g., the `Main/` folder shows 0.0 coverage, but inside of it, every file has two entries, one with 0.0 coverage and one with 100.0 coverage:
  image
  Obviously this is beyond useless.
- To get the Linux tests to collect coverage data, I had to remove the `LocalSource` flag, which means it's collecting a bunch of garbage data from AltCover and Humanizer (although with zero actual hits?). No idea why it's not collecting the data from CKAN that it's supposed to. Even stranger is that we didn't really change the Linux testing logic in a meaningful way.
  image

On the bright side, Coveralls successfully merged these two sources of absolutely trash data. 🎉

HebaruSan:
Now all the tests are running and passing, but no coverage data is being collected on either platform.

- The instrumented assemblies are generated and used for net481 testing on Linux, but it's still false reporting zero hits anywhere:
  image
- On Windows, nothing coverage-related is reported; possible that AltCover isn't running at all:
  image

Here's the verbose output on Linux, which says `Failed one of : symbols true/ pass filter false/ ILOnly true`.
Determining which of `true` or `false` indicates a failure will probably require auditing [AltCover's code](https://github.com/SteveGilham/altcover/blob/master/AltCover.Engine/Main.fs), but this is finally looking like an actual clue.

... I think `pass filter false/` is indicating the failure, but we have 6 assembly exclude filters, 2 type filters, 6 path filters, 1 attribute filter, and the local source filter, and I have no idea which one of those might be misbehaving.

... no change after disabling **all** of those filters except for the local source filter. So `pass filter` apparently refers to something other than the actual "filters" we can control.

```
/home/runner/work/CKAN/CKAN/_build/out/CKAN.Tests/Debug/bin/net481/CKAN.dll : beginning process
Assembly symbol GUID = d387aad0-6fd8-46a5-88e8-058f6b3570c2 mvid = 796550cb-dc64-4994-901f-88d8de22466c
Symbol file /home/runner/work/CKAN/CKAN/_build/out/CKAN.Tests/Debug/bin/net481/CKAN.pdb GUID = d387aad0-6fd8-46a5-88e8-058f6b3570c2
Assembly /home/runner/work/CKAN/CKAN/_build/out/CKAN.Tests/Debug/bin/net481/CKAN.dll symbols from folder '"/home/runner/work/CKAN/CKAN/_build/out/CKAN.Tests/Debug/bin/net481/CKAN.pdb"'
/home/runner/work/CKAN/CKAN/_build/out/CKAN.Tests/Debug/bin/net481/CKAN.dll : ** Failed one of : symbols true/ pass filter false/ ILOnly true **
```

With the local source filter turned off, the failure goes away:

```
/home/runner/work/CKAN/CKAN/_build/out/CKAN.Tests/Debug/bin/net481/CKAN.dll : beginning process
Assembly symbol GUID = 802c32d6-f30c-45ca-990c-dee1c40df656 mvid = 5a6012fd-65b7-4235-853d-d3d04918133f
Symbol file /home/runner/work/CKAN/CKAN/_build/out/CKAN.Tests/Debug/bin/net481/CKAN.pdb GUID = 802c32d6-f30c-45ca-990c-dee1c40df656
Assembly /home/runner/work/CKAN/CKAN/_build/out/CKAN.Tests/Debug/bin/net481/CKAN.dll symbols from folder '"/home/runner/work/CKAN/CKAN/_build/out/CKAN.Tests/Debug/bin/net481/CKAN.pdb"'
   => /home/runner/work/CKAN/CKAN/_build/out/CKAN.Tests/Debug/bin/net481/CKAN.dll
```

Somehow it thinks the CKAN.dll from the Windows build isn't compiled from local source, so now I need to figure out how that logic is implemented in AltCover...

It looks for some kind of `Url` property in some "module documents" associated with the assembly, then checks whether they exist as files, so I guess it's looking for paths from the Windows container inside the Linux container, where they wouldn't exist:

```fsharp
    member nameProvider.LocalFilter: bool =
      match nameProvider with
      | :? AssemblyDefinition as a ->
        (CoverageParameters.local.Value)
        && a.MainModule
           |> ProgramDatabase.getModuleDocuments
           |> Seq.map _.Url
           |> Seq.exists File.Exists
           |> not
      | _ -> false
```

```fsharp
  let internal getModuleDocuments (``module``: ModuleDefinition) =
    ``module``.GetAllTypes()
    |> Seq.collect _.GetMethods()
    |> Seq.collect _.DebugInformation.SequencePoints
    |> Seq.map _.Document
    |> Seq.distinctBy _.Url
    |> Seq.toList
```

Turning on all the options in ILSpy, I think this is what those "sequence points" look like, and yes, they contain the file path:

image

And here's what they look like in the files from the container:

image

I guess for the "local source" filter to work, the assemblies have to be built at least on the same OS, if not the same container. Too bad MS didn't use relative paths here.

HebaruSan:
With the local source filter removed and replaced by additional `AssemblyFilter`s, tests are now running and passing on both platforms, but Windows generates no coverage data files, and Linux's all have `skippedDueTo="File"` for things that absolutely should not be skipped:

```xml
            <Method skippedDueTo="File" visited="false" cyclomaticComplexity="1" nPathComplexity="0" sequenceCoverage="0" branchCoverage="0" isConstructor="false" isStatic="false" isGetter="false" isSetter="false" crapScore="2">
              <MetadataToken>100663429</MetadataToken>
              <Name>System.String CKAN.GameInstance::DllPathToIdentifier(System.String)</Name>
              <SequencePoints />
              <BranchPoints />
              <MethodPoint vc="0" uspid="100663429" ordinal="0" offset="0" />
            </Method>
```

No guess what that's about yet. AltCover's maintainer has been impressively helpful on its issue tracker in the past, so if he's still active, I might file an issue in hopes of getting some clues about this.

HebaruSan:
This is a bit suspicious on Windows:


> `       Command line arguments = "C:\Program Files\dotnet\sdk\9.0.305\MSBuild.dll -maxcpucount -verbosity:m -tlp:default=auto -nologo -nodereuse:false -target:VSTest -nologo -property:VSTestTestCaseFilter="TestCategory!=FlakyNetwork" -property:VSTestNoBuild=true -property:VSTestResultsDirectory="D:\a\CKAN\CKAN\_build\test\nunit" -property:VSTestNoLogo=true -property:Configuration=Debug -restore:false -verbosity:detailed D:/a/CKAN/CKAN/CKAN.sln /p:AltCover=true /p:AltCoverAssemblyFilter=Microsoft|NUnit3|testhost|IndexRange|OxyPlot|Humanizer|AltCover\.Monitor|CKAN-ConsoleUI|CKAN\.Tests| /p:AltCoverTypeFilter=System|Microsoft| /p:AltCoverAttributeFilter=ExcludeFromCodeCoverage| /p:AltCoverPathFilter=_build|GUI/Dialogs|GUI/Controls|GUI/Main|GUI\\Dialogs|GUI\\Controls|GUI\\Main| /p:AltCoverReport=D:/a/CKAN/CKAN/_build/test/coverage/coverage.xml /p:AltCoverSourceLink=true /p:AltCoverLcovReport=D:/a/CKAN/CKAN/_build/test/coverage/lcov.info /p:AltCoverCobertura=D:/a/CKAN/CKAN/_build/test/coverage/cobertura.xml /p:AltCoverSummaryFormat=O /p:AltCoverVerbosity=Verbose /p:AltCoverShowSummary=O /p:AltCoverForce=true -property:VSTestVerbosity=detailed -property:VSTestArtifactsProcessingMode=collect -property:VSTestSessionCorrelationId=2680_2ae00bf7-590c-4162-92af-36bdf55cae7e -distributedlogger:Microsoft.DotNet.Tools.MSBuild.MSBuildLogger,C:\Program Files\dotnet\sdk\9.0.305\dotnet.dll*Microsoft.DotNet.Tools.MSBuild.MSBuildForwardingLogger,C:\Program Files\dotnet\sdk\9.0.305\dotnet.dll"`

Specifically, all of the filters have been processed to end with a `|` character!

> `/p:AltCoverAssemblyFilter=Microsoft|NUnit3|testhost|IndexRange|OxyPlot|Humanizer|AltCover\.Monitor|CKAN-ConsoleUI|CKAN\.Tests| /p:AltCoverTypeFilter=System|Microsoft| /p:AltCoverAttributeFilter=ExcludeFromCodeCoverage| /p:AltCoverPathFilter=_build|GUI/Dialogs|GUI/Controls|GUI/Main|GUI\\Dialogs|GUI\\Controls|GUI\\Main|`

The way string processing typically works, that would mean that every one of these ends with an empty string piece, which likely would match any input string, meaning that _everything_ would be filtered out! If that's correct, this might actually be a bug that we can [file with AltCover](https://github.com/SteveGilham/altcover/tree/master/AltCover.Cake).

EDIT: The problem with this theory is that the exact same arguments work perfectly well with locally-built assemblies. If the filter format was a problem, that would break all runs, not just those with remotely-built assemblies.

HebaruSan:
Asked for help at SteveGilham/altcover#249, but the author has not been super active lately. I'll wait a while and hope for a response, and then I suppose we'll have to consider hacks like always building and testing in the same container.

HebaruSan:
After reverting all of the changes that affected Ubuntu and doing a full build+test inside the Windows container instead of trying to share the assemblies, we now get relatively reasonable results for each platform:

[image](https://coveralls.io/jobs/173065215)

[image](https://coveralls.io/jobs/173065214)

... except that the files are now weirdly doubled on ~~Ubuntu~~ both platforms:

image

But then Coveralls completely flubs the merge of the data from Linux and Windows, with gigantic jumps in coverage across the board totalling about 9.8% according to the autogenerated comment above, far in excess of the prevalence of our platform-specific code:

[image](https://coveralls.io/builds/76097246)

HebaruSan:
Looks like I'm going to have to move this branch from my fork to the main fork to be able to send the Coveralls repo token to the `rerun_build` API.

HebaruSan:
Replaced by #4454.

PR #4454, opened 2025-10-23
Build on Windows, upload multi-platform coverage

HebaruSan:
## Motivation

Currently we test and upload coverage data only for Linux. This means that our Windows-specific code isn't tested and falsely reported as not covered.

This PR replaces #4449, which did not have access to the repo secret required to run the `rerun_build` workaround for Coveralls (which wasn't needed after all).

## Changes

- Now the steps in `build.yml` that build `CKAN.dmg` are refactored into a separate job with a consolidated `Release` configuration check
- Now `test.yml` uses a matrix strategy to run tests and upload coverage on both Windows and Linux, using [parallel builds support for Coveralls](https://docs.coveralls.io/parallel-builds). Since AltCover seems unable to handle assemblies built on a different OS, or a different Windows container when running on Windows, the Windows job does a complete build+test rather than trying to re-use the artifacts from `build.yml`.
- A new `coveralls-finish` job finalizes the coverage upload with `parallel-finished: true`.
  - ~~This job then runs the [`rerun_build` workaround](https://github.com/search?q=https%3A%2F%2Fcoveralls.io%2Frerun_build&type=code) from lemurheavy/coveralls-public#1629 to try to get Coveralls to generate accurate numbers for the merged results (untested as of yet because the secrets are not available in the current fork, so I will be replacing this PR with a new one shortly)~~
    Removed because it wasn't helping.
- The [environment variable `DOTNET_CLI_TELEMETRY_OPTOUT`](https://learn.microsoft.com/en-us/dotnet/core/tools/telemetry) is set to `true` to prevent `dotnet` from "collect[ing] usage data and send[ing] it to Microsoft"
- The [`<IsTestProject>` property](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#istestproject) is set to `true` in `Tests.csproj`
- The check that exempts Linux containers from the admin user check is extended to also work in Windows containers (at least the GitHub ones)
- The Mac DMG is only generated in `Release` configuration on Linux, but formerly the needed files were _published_ in `Debug` configuration as well. Now both steps are `Release`-only.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,8 +12,9 @@ dotnet_diagnostic.IDE0010.severity = none
 # Don't "simplify" if statements that throw
 dotnet_diagnostic.IDE0016.severity = none
 
-# Expression bodies are good for methods
+# Expression bodies are good for methods and coversion operators
 dotnet_diagnostic.IDE0022.severity = none
+dotnet_diagnostic.IDE0023.severity = none
 
 # Expression bodies are fine for operators
 dotnet_diagnostic.IDE0024.severity = none
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,11 +23,6 @@ jobs:
         with:
           path: _build/tools
           key: build-tools-${{ hashFiles('build.sh', 'build.ps1', 'build/*') }}
-      - name: Restore cache for _build/cake
-        uses: actions/cache@v4
-        with:
-          path: _build/cake
-          key: build-cake-${{ hashFiles('build/*') }}
       - name: Restore cache for _build/lib/nuget
         uses: actions/cache@v4
         with:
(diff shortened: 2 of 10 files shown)


HebaruSan:
[Last build before this PR](https://coveralls.io/builds/76093675):

image

[Current Ubuntu build](https://coveralls.io/jobs/173099783):

image

It's still doubling the files, and the data looks really weird. The copy with fewer covered lines and more missed lines has a _higher_ coverage percentage than the copy with more covered lines and fewer missed lines. That's exactly backwards from what common sense would suggest. And the branch hits/misses data seems to have disappeared.

HebaruSan:
Meanwhile [the Windows build](https://coveralls.io/jobs/173100875) is telling us that there's a copy of that file with 100% coverage:

image

The details page for that file indicates that [Coveralls thinks none of the lines are relevant](https://coveralls.io/jobs/173100875/source_files/8949415741), so they don't _need_ to be tested. Needless to say, that's dramatically wrong. Notably, the 88.54% in the _other_ entry exactly matches one of the ones from the Ubuntu build; could Coveralls be mixing data from different builds where it isn't meant to go?

The combined stats are also weird:

image

_**Zero**_ missed lines, with a "-105" change annotation unlike any of the other rows. I have no idea from where it could possibly be getting 737 covered lines for this file; [the details page](https://coveralls.io/builds/76112461/source?filename=Core%2FIO%2FModuleInstaller.cs) indeed shows coverage of lines that aren't covered in either of the specific builds, and there are even some red uncovered lines that somehow escaped being counted as uncovered in the summary stats. It is starting to look like Coveralls parallel builds are just extremely buggy.

At least the four entries from the upstream builds are represented by just one in the merge. 🎉

HebaruSan:
Temporarily excluding the Linux coverage data, we should get a "merged" build that is identical to the Windows data in every way, since all it has to do is copy the data over. At a glance, Coveralls's parallel builds implementation looks completely and utterly broken.

Windows build: https://coveralls.io/jobs/173106504

image

Merged: https://coveralls.io/builds/76114443

image

That's... kind of shocking. This feature was developed and released into the wild and has existed presumably for years, yet it essentially _does not work at all_. I guess we can't capture coverage data from multiple platforms. Sigh.

PR #4474, opened 2025-12-23
Update download count image for Minkowski²

Clayell:
Possibly the most trivial/pointless PR ever, heh.

Original:
image

Fixed:
image


HebaruSan (approved):
Thanks for the fix! 🏗️

PR #4480, opened 2025-12-28
Updated version string documentation and parsing.

codeandsolder:
Changed spec to match implementation (good ole' DDT), added some examples to clarify and effectively fixed possible overflow in int parsing (one dead mod affected).

#6 borrowed some of the spec from Debian, but the "all the letters sort earlier than all the non-letters" part was just never implemented.

### DIFF
--- a/Core/Versioning/ModuleVersion.cs
+++ b/Core/Versioning/ModuleVersion.cs
@@ -54,7 +54,7 @@ namespace CKAN.Versioning
             // If we have an epoch, then record it.
             if (match.Groups["epoch"].Value.Length > 0)
             {
-                _epoch = Convert.ToInt64( match.Groups["epoch"].Value );
+                _epoch = Convert.ToUInt64( match.Groups["epoch"].Value );
             }
 
             _version = match.Groups["version"].Value;


codeandsolder:
image
I don't make the rules, I just describe them.

Core/Versioning/ModuleVersion.cs
```diff
@@ -54,7 +54,7 @@ public ModuleVersion(string version)
             // If we have an epoch, then record it.
             if (match.Groups["epoch"].Value.Length > 0)
             {
-                _epoch = Convert.ToInt32( match.Groups["epoch"].Value );
+                _epoch = Convert.ToInt64( match.Groups["epoch"].Value );
```
HebaruSan: `_epoch` is still just an int, though? Wouldn't this still overflow, just in the assignment to `int` rather than inside the `ToInt32/64` call?

```csharp
        private readonly int _epoch;
```

Spec.md
```diff
@@ -237,40 +237,51 @@ The comparison behavior of the package management system with respect to the
...
+any other character.
+If one non-numerical segment runs out before the other it's considered smaller.
+
+We're well aware this system has flaws, changing it hovewer would break backwards
+compatibility without providing material benefits.
+
+###### Examples:
+    "2:1.0" > "1:999.9"    Epoch is most significant
+    "0:1.0" == "1.0"       Missing epoch is treated as 0
+    "1.10" > "1.2"         v
```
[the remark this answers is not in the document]
codeandsolder: That's supposed to be an arrow down as the whole block is examples of the same behavior in different contexts, I can change formatting if you want.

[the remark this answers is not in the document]
HebaruSan: Yeah, I figured that out eventually. A spec shouldn't require mental effort just to parse.

Spec.md
```diff
@@ -237,40 +237,53 @@ The comparison behavior of the package management system with respect to the
...
-These two steps (comparing and removing initial non-digit strings and initial
-digit strings) are repeated until a difference is found or both strings are
-exhausted.
-
-Note that the purpose of epochs is to allow us to leave behind mistakes in
-version numbering, and to cope with situations where the version numbering
-scheme changes. It is not intended to cope with version numbers containing
-strings of letters which the package management system cannot interpret (such as
-ALPHA or pre-), or with silly orderings.
+When comparing two version numbers, first the `epoch` (continous integers at
```
HebaruSan: The format of an epoch is already defined above in the `###### epoch` section. Repeating it here is redundant, which is bad for readability and maintainability.

Spec.md
```diff
@@ -237,40 +237,53 @@ The comparison behavior of the package management system with respect to the
...
+the start of the version string followed by the character ":" or 0 if none is
+found), then the `mod_version`if both `epoch` are equal. The `mod_version`
```
HebaruSan: Looks like "`epoch` is compared numerically` got deleted here (and the remaining text ran together without spaces).
Let's bring it back.

Spec.md
```diff
@@ -237,40 +237,53 @@ The comparison behavior of the package management system with respect to the
...
+
+Both strings are split into alternating segments consisting of consecutive digits
+and consecutive non-digits, starting with non-digits.
+Numeric segments are compared as integers, non-numeric segments are compared as
+ASCII strings, with the exception of a dot which is considered larger than 
+any other character.
+If one non-numerical segment runs out before the other it's considered smaller.
+
+We're well aware this system has flaws, changing it hovewer would break backwards
+compatibility without providing material benefits.
```
HebaruSan: Let's remove this paragraph. Anything can have flaws, but a spec should just describe what it _is_.

codeandsolder: Are you sure? I'd assume it would help greatly reduce the number of people seeing the section above and below, thinking "that's stupid" and going to leave feedback. Maybe a more neutral phrasing, something about frozen specification?

Spec.md
```diff
@@ -237,40 +237,53 @@ The comparison behavior of the package management system with respect to the
...
+
```
HebaruSan: Why was the paragraph explaining the purpose of epochs deleted?

codeandsolder: I thought the part in epoch definition ("It is provided to allow mistakes in the version numbers of older versions of a package, and also a package's previous version numbering schemes, to be left behind.") was sufficient, I can bring it back if you prefer.

Core/Versioning/ModuleVersion.cs
```diff
@@ -34,7 +34,7 @@ public partial class ModuleVersion
 
     public partial class ModuleVersion
     {
-        private readonly int _epoch;
+        private readonly ulong _epoch;
```
HebaruSan: We should also update the type of `Metadata.Epoch` in `Netkan/Model/Metadata.cs`, if `long`s are going to be supported here.

Spec.md
HebaruSan: Considering that for about ten years that number of people has been zero, yeah, I'm pretty sure. 😛

Spec.md
HebaruSan: OK, let me re-read it with that in mind...

HebaruSan: ... yeah, that makes sense. They both cover the same two key points, errors and changes in schemes. Thanks.

Spec.md
```diff
@@ -244,10 +244,9 @@ are encouraged to restrict version names to letters, digits, and the
 
 ###### Version ordering
 
-When comparing two version numbers, first the `epoch` (continous integers at
-the start of the version string followed by the character ":" or 0 if none is
-found), then the `mod_version`if both `epoch` are equal. The `mod_version` 
-is compared using the following algorithm:
+When comparing two version strings, first the `epoch` are compared numerically,
+then the, if both `epoch` are equal, the `mod_version` is compared
```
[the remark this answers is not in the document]
codeandsolder: There shouldn't be a the, correct, I'm tired lol

Core/Versioning/ModuleVersion.cs
```diff
@@ -54,7 +54,7 @@ public ModuleVersion(string version)
             // If we have an epoch, then record it.
             if (match.Groups["epoch"].Value.Length > 0)
             {
-                _epoch = Convert.ToInt32( match.Groups["epoch"].Value );
+                _epoch = Convert.ToUInt64( match.Groups["epoch"].Value );
```
HebaruSan: This doesn't actually address the "one dead mod", does it? 

> CParts - v1.1.0.202206081400

That mod _has no epoch_; the 32-bit overflow would be in the `int.TryParse` calls in `numComp`.

codeandsolder: That is also correct, I'll take a proper look at refactoring the version section tomorrow.

HebaruSan:
Hi @codeandsolder, are you still planning to finish this PR? If so, please do so and let us know, and we can re-open it.

PR #4502, opened 2026-01-29
Download errors dialog improvements

HebaruSan:
## Motivation

While reviewing KSP-CKAN/NetKAN#10922, I got some download errors for `TURD-BDArmory` because the ZIP was replaced, which gave me a chance to nitpick the download errors popup:

- The abort button says "Abort whole changeset" even when there is no changeset (i.e., downloading from the Contents tab or the mod list right click menu)
- The SHA256 error message starts with the full path to the downloaded ZIP file, so if the path is long, you have to scroll horizontally to see what the error is about
- In addition to being hard to see, the only way to copy the download error messages to share elsewhere is by taking a graphical screenshot

## Changes

- Now the caption of the abort button is "Abort download" when downloading without a changeset
- Now the default (en-GB) translation of the SHA256 error message is reformatted to put the file path at the end, so it doesn't obscure the important part (other translations will have to wait for changes on Crowdin because I don't know the needed grammar)
- Now the Error column wraps on Windows, to make it easier to see the whole thing.
  (Mono doesn't obey `CellStyle.WrapMode`, so this will have no effect on Linux, and I still need to test this on Windows to make sure it works and resolve any remaining glitches.)
- Now the download errors grid has a right click menu with a "Copy errors to clipboard" option to make it easier to capture the text for pasting into issues

### DIFF
--- a/GUI/Dialogs/DownloadsFailedDialog.Designer.cs
+++ b/GUI/Dialogs/DownloadsFailedDialog.Designer.cs
@@ -47,11 +47,13 @@ namespace CKAN.GUI
             //
             // ExplanationLabel
             //
+            this.ExplanationLabel.AutoSize = true;
+            this.ExplanationLabel.MaximumSize = new System.Drawing.Size(490, 0);
             this.ExplanationLabel.Dock = System.Windows.Forms.DockStyle.Top;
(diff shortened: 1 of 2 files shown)


HebaruSan:
> (Mono doesn't obey `CellStyle.WrapMode`, so this will have no effect on Linux, and I still need to test this on Windows to make sure it works and resolve any remaining glitches.)

image

🎉

PR #4512, opened 2026-02-13
Remove redundant duplicate check

geromet:
.Distinct() already removes the duplicates.
This change makes sure users don't have to manually make sure they don't enter duplicates.

PR #4514, opened 2026-02-13
Cache instance-specific installation filters in memory see https://forum.kerbalspaceprogram.com/topic/229774-list-of-mods-and-parts-marked-as-compatible-by-ckan-while-being-known-to-be-broken/page/2/#findComment-4501168

geromet:
I have cached the installation filters so they are no longer JSONDeserialized on each get.
This has improved my modpack installation time from 1 minute 47 seconds to 46 seconds.
I also still have my other fix in here.

### DIFF
--- a/Core/GameInstance.cs
+++ b/Core/GameInstance.cs
@@ -203,17 +203,31 @@ namespace CKAN
 
         private string SuppressedCompatWarningIdentifiersFile
             => Path.Combine(CkanDir, "suppressed_compat_warning_identifiers.json");
-
+        private List<string> _installFilters = new ();
         public string[] InstallFilters
         {
-            get => (File.Exists(InstallFiltersFile)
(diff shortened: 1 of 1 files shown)


drewcassidy:
Did you clear the download cache between running your benchmarks? did you profile the code to make sure the functions you patched are actually in the hot path?

HebaruSan:
Hi @geromet, your measurement is highly suspect. There is no way that reading a few strings out of a JSON file could take a full minute.

~~And this change would break the case where the user edits the installation filters and then installs a mod; with the old code, the latest change would be applied, but with your change they would have to restart CKAN.~~
EDIT: On second look, that would work OK, because the cached values are updated when the filters are saved. What _would_ break is if you were running multiple copies of CKAN for different game instances and editing the installation filters (which is probably a very rare case, if it's ever happened at all).

Thanks for wanting to contribute, but this change isn't appropriate.
(And the other part was already rejected in #4512 / #4513 and isn't going to be sneaked in.)

PR #4441, opened 2025-09-19
Don't clone with symlinks on Linux

HebaruSan:
## Problem

If you clone a KSP1 instance using the "Share stock files" option from #4129, the game throws exceptions like this:

```
[LOG 11:21:03.627] ------------------- initializing flight mode... ------------------
[LOG 11:21:03.627] No save file found for path: /media/DataTrove/Games/KSP-thin-clone-5/KSP_Data/../saves/Testing symlinks/persistent.sfs
[EXC 11:21:03.628] NullReferenceException: Object reference not set to an instance of an object
	FlightDriver.Start () (at <be370b73275e49439ea5e41ceef6700f>:0)
[EXC 11:21:03.635] NullReferenceException: Object reference not set to an instance of an object
	KSP.UI.Screens.EditorActionGroups.CreateGroups_Action () (at <be370b73275e49439ea5e41ceef6700f>:0)
	KSP.UI.Screens.EditorActionGroups.ConstructGroupList () (at <be370b73275e49439ea5e41ceef6700f>:0)
	KSP.UI.Screens.EditorActionGroups.Start () (at <be370b73275e49439ea5e41ceef6700f>:0)
[EXC 11:21:03.642] NullReferenceException: Object reference not set to an instance of an object
	ManeuverNodeEditorTabOrbitBasic.UpdateUIElements () (at <be370b73275e49439ea5e41ceef6700f>:0)
	ManeuverNodeEditorTabOrbitBasic.SetInitialValues () (at <be370b73275e49439ea5e41ceef6700f>:0)
	ManeuverNodeEditorTab.Setup (UnityEngine.Transform parent) (at <be370b73275e49439ea5e41ceef6700f>:0)
	ManeuverNodeEditorManager.setupAllTabs () (at <be370b73275e49439ea5e41ceef6700f>:0)
	ManeuverNodeEditorManager.Start () (at <be370b73275e49439ea5e41ceef6700f>:0)
[EXC 11:21:03.703] NullReferenceException: Object reference not set to an instance of an object
	VesselAutopilotUI.LateUpdate () (at <be370b73275e49439ea5e41ceef6700f>:0)
[EXC 11:21:03.712] NullReferenceException: Object reference not set to an instance of an object
	FlightGlobals.GetFoR (FoRModes mode) (at <be370b73275e49439ea5e41ceef6700f>:0)
	FlightCamera.GetCameraFoR (FoRModes mode) (at <be370b73275e49439ea5e41ceef6700f>:0)
[...]

PR #4519, opened 2026-03-01
Fix symlinks being generated on Linux

HebaruSan:
## Problem

#4441 intended to suppress symlinks when cloning instances on Linux, but it didn't work; symlinks are still being generated.

Reported by @PewPewCricket on Discord and <https://github.com/KSP-CKAN/CKAN/pull/4441#issuecomment-3978053277>.

## Cause

I updated the default for the `Checked` property, but `CheckedState` is set to the opposite value on the next line, so the hidden checkbox was still always checked. 🤦

## Changes

- Now `CheckedState` defaults to `Unchecked`, so there will actually not be symlinks on Linux.
- [`ini-parser`](https://www.nuget.org/packages/ini-parser)'s actual latest version is 2.5.2; 3.4.0 was a typo'd duplicate of 2.4.0 that was de-listed from Nuget. However, Nuget does not allow fully deleting releases, because it would break builds.
  Our reference is now updated from 3.4.0 to 2.5.2.

### DIFF
--- a/GUI/CKAN-GUI.csproj
+++ b/GUI/CKAN-GUI.csproj
@@ -34,7 +34,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Autofac" Version="4.9.4" />
-    <PackageReference Include="ini-parser" Version="3.4.0" />
+    <PackageReference Include="ini-parser" Version="2.5.2" />
     <PackageReference Include="log4net" Version="2.0.15" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageReference Include="Nullable" Version="1.3.1" PrivateAssets="all" />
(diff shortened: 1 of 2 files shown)


hpax:
I recently ran into this issue and looked at it in some detail. If it is desirable to create symlinks for the purpose of saving storage, it does work correctly if the symlink tree is created one level further down, i.e. instead of linking KSP_Data linking the contents of KSP_Data. I "repaired" the instances on my system by using the attached shell script which did exactly this.
[kspfix.sh](https://github.com/user-attachments/files/27634607/kspfix.sh)

That being said, this is not an issue if the data is on the same file system and hard links are used instead.

(I did not mess with KSPLauncher_Data since it didn't seem to have the same problem as KSP_Data.)

PR #4543, opened 2026-04-05
Auto-size new repo popup to fit rows in list

### LINKED ISSUE #4541: [Feature]: Increase the height of the add metadata repository list
### Problem

It seems quite silly to me that the add metadata repository list requires you to scroll, when there's currently only 3 entries.

Image

### Suggestion

Increase the [size](https://github.com/KSP-CKAN/CKAN/blob/a2f45090f12cd20b9b1738ba9fa44eb58d717e91/GUI/Dialogs/NewRepoDialog.Designer.cs#L71) of the list to make it so you don't have to scroll with so few repositories. Maybe something like 3x the current height?

### Alternatives

_No response_

### Additional context

_No response_

HebaruSan:
## Motivation

- KSP-CKAN/CKAN-meta#3383 added a new default metadata repo option, but the popup's default size is too small to display it without scrolling
  image
- While installing Sol, I got a lot of `provides` prompts, all of which were sorted lexicographically by name because we don't know their download counts. This meant the default was the 16K graphics, but I wanted to install the smallest ones. When I uninstalled and installed over and over for testing, I had to keep re-selecting the mod I picked last time.

## Problems

- Discord user `santana` reported that trying to remove TweakScale and install TweakScaleRescaled gets the user stuck at the recommendations screen with a disabled Continue button
  image
- After installing Sol's custom version of ParallaxContinued, I noticed that there were three empty folders that showed up with red text in the Contents tab, even though they weren't missing
- The export modpack screen displays DLCs, but if you click them, the mod info pane disappears instead of displaying the DLC's info
- If you install a mod with a lot of `provides` prompts to resolve, the screen flickers noticeably in between prompts when you click Continue

## Causes

- #3913 removed auto-removable mods from the `toUninstall` list, which meant auto-removable mods were no longer being passed to the recommendations screen. When that screen's conflict checker ran, it discovered that TweakScaleRescaled conflicted with Recall, since it didn't know that Recall was going to be removed.
- The red text was always applied to any non-file node with no child nodes, intending to catch folders where the user had deleted the contents
- The mod info handler for export modpack was depending on finding rows in the main mod list, which don't exist for DLCs
- When you clicked Continue, the recommendations tab was removed without activating any other tab, which caused the leftmost tab, Manage Mods, to be activated, which is slow to refresh and which loads mod info with its most recently selected mod

## Changes

- Now the new repo popup auto-sizes to fit its contents
  Fixes #4541.
  - It also centers on its parent window rather than the screen
  - The columns auto resize to fit the space, so you can see the whole URLs
  - While working on that, I saw some weird behavior having to do with DPI scaling, so several changes are made regarding that:
    - An `app.manifest` is added and our `App.config` is updated with properties that various online tutorials say should enable proper scaling (but `DeviceDpi` is still always 96, so I have my doubts; luckily `Graphics.DpiX` and `Graphics.DpiY` seem to be accurate)
    - We now call `SetProcessDPIAware` to tell Windows to turn off its blurry auto-scaling and let us handle it
    - Our dialogs and controls are switched from font-based to DPI-based scaling
    - Some fonts that were hard-coded to use pixel-based sizes are switched to point-based, which accounts for DPI (a "point" is 1/72 of an inch, so a higher DPI translates to more pixels per point)
    - Icons are now scaled to fit the DPI as needed
- Now the `provides` prompt sorts cached mods above uncached mods, so the default will be whatever you picked last time in most cases when reinstalling
- Now all uninstalling mods, auto- or not, are passed to the recommendations screen, so it can correctly identify that there won't be a conflict after Recall is removed.
- Now the Contents tab checks whether an empty folder exists on disk instead of always marking it as missing
- Now DLCs' mod info is properly displayed in the export modpack screen
- Now the recommendations tab explicitly switches to the installing/progress screen before hiding itself, so the flicker will be lessened and mod info won't briefly appear


HebaruSan:
Yikes! Mono apparently does not allow anything in `App.config` other than `dllmap` and `location`:

https://github.com/elijahtaylor/mono/blob/13b742d0646bcffac41b7c03d3eeaef443a573c1/mcs/class/System.Configuration/System.Configuration/SectionGroupInfo.cs#L330

Might have to roll that part back...

PR #4591, opened 2026-04-27
Layout and scaling related fixes

### LINKED ISSUE #4579: Duplicated modules with epochs stripped off of versions from alternate metadata repo

### Operating System

Windows 10

### CKAN Version

Latest Dev (v1.36.2.26110), language of error message changes from Politas but still persists

### Games

KSP 1

### Game Version

1.12.5.3190, latest

Several, because of the same issue, which led me to manually install a few mods: 
TweakScale Rescaled (may have fixed its metadata since then)
Astrogator
FreeIva
Probe Control Room Recontrolled
Reviva

Also manually installed Scatterer/Blackrack Vol Clouds v5 and some other manual config changes here n there

### Describe the bug

Expected: move through install menus past suggested or recommended screens, have the mods be installed.

What happens: before getting to suggestion screen, the operation fails and gives error message from title, depending on CKAN version. 

### Steps to reproduce

0. First, add this alternate metadata repository, without which you will not be able to reproduce any of these problems:
   Image

Example: I'm trying to install Kerbin Side Remastered - The Life Aquatic. Note the duplicated version of the suggested mod.

Image

Click apply, get to this screen, all good so far.

Image

Apply again, and we hit the error. 

Image

The error message is more descriptive in the latest dev version that I installed today, which led me to check the suggested mod metadata, and sure enough there's that duplication from above. Seems to me that CKAN is trying to add the same mod twice to the suggested mods screen, which it freaks out at, and won't continue. I'd investigate this mod as well as the mods I have listed above as manually installed. For instance, a very similar suspicious suggested mod that causes similar errors from Astrogator:

Image

I uninstalled it and tried to have CKAN install it just to verify, and because I love you guys. Thanks for the great software, I'll help if I can. This bug is buggin me tho.

Image

### Relevant log output

```
Scanning for DLCs and manually installed modules...
System.ArgumentException: Already added KerbinSideRemasteredGAP 1:v3.0.6, can't add KerbinSideRemasteredGAP v3.0.6 (Suggested by KerbinSideRemasteredGapExtras v1.0.6)
   at CKAN.RelationshipResolver.Add(CkanModule module, SelectionReason reason)
   at CKAN.RelationshipResolver.ResolveStanza(List`1 stanza, SelectionReason reason, RelationshipResolverOptions options, Boolean soft_resolve, IEnumerable`1 old_stanza)
   at CKAN.RelationshipResolver.Resolve(CkanModule module, RelationshipResolverOptions options, IEnumerable`1 old_stanza)
   at CKAN.RelationshipResolver.ResolveStanza(List`1 stanza, SelectionReason reason, RelationshipResolverOptions options, Boolean soft_resolve, IEnumerable`1 old_stanza)
   at CKAN.RelationshipResolver.Resolve(CkanModule module, RelationshipResolverOptions options, IEnumerable`1 old_stanza)
   at CKAN.RelationshipResolver.AddModulesToInstall(CkanModule[] modules)
   at CKAN.RelationshipResolver..ctor(IEnumerable`1 modulesToInstall, IEnumerable`1 modulesToRemove, RelationshipResolverOptions options, IRegistryQuerier registry, IGame game, GameVersionCriteria versionCrit)
   at CKAN.IO.ModuleInstaller.FindRecommendations(GameInstance instance, IReadOnlyCollection`1 sourceModules, IReadOnlyCollection`1 toInstall, IReadOnlyCollection`1 toRemove, IReadOnlyCollection`1 exclude, Registry registry, Dictionary`2& recommendations, Dictionary`2& suggestions, Dictionary`2& supporters)
   at CKAN.GUI.Main.InstallMods(Object sender, DoWorkEventArgs e)
   at System.ComponentModel.BackgroundWorker.OnDoWork(DoWorkEventArgs e)
Your game folder has been returned to its original state.
Error during installation!
...
```

Astrogator:

`System.ArgumentException: Already added GravityTurnContinued 3:1.8.4.2, can't add GravityTurnContinued 1.8.4.2 (Suggested by Astrogator v1.0.2)
   at CKAN.RelationshipResolver.Add(CkanModule module, SelectionReason reason)`

HebaruSan:
## Problems and motivations

- In #4579, a user saw multiple versions of the same mod treated oddly as if they were sometimes different mods. From a support perspective, the fix was to remove the extra metadata repo that contained a bunch of modules with bad `version` properties. But that really should have worked better, since multiple repos providing versions of the same mods is supposed to be supported.
  ImageImage
- When a .netkan has multiple sections, certain metadata properties are automatically mirrored among all the sections, and some are not. Anything pertaining exclusively to the _contents_ of the ZIP file makes sense to mirror, since all sections deal with the same contents; `$vref` is one such property but is not mirrored, which is inconvenient for maintenance.
- The PlayTime grid lets you focus uneditable cells, which isn't useful and gives a bad UI cue for how it works.
- The command line options popup has a Reset button that changes you rsettings to the default according to CKAN and your Steam library. Currently it is always enabled, so you can click it when it will do nothing, which may create confusion about what it does.
- Many users join the CKAN Discord to ask non-CKAN related questions
- Opening the download statistics screen twice in the same session would cause the pie chart to not appear
- The mod list columns' default widths don't scale with the DPI
- The rightmost mod list column with the abstract can be hard to read, and the description isn't shown at all
- The resource links in mod info aren't always rendered nicely
- The compatible versions popup looks pretty messed up

Many thanks to @KerballOne for reporting many of the layout problems on Discord.

## Causes

- Unlike `CompatibilitySorter` (which drives the main mod list rows and many other things), `Registry.LatestAvailableWithProvides` did not account for de-duplicating `AvailableModule`s with the same identifier.
- The mirroring used `StripNetkanMetadataTransformer.Strip` to determine which properties to mirror, which unfortunately caused `$vref` to be excluded.
- The Help menu lists the CKAN Discord prominently near the top, so users may think it is the go-to for all possible questions they may have
- The pie chart control appears not to like re-using the same `PlotModel` with different collections of `Series`
- #4543 made a number of changes related to scaling, which can be complex and need some massaging to reach full maturity.
- Somehow the initial sizing we give to the resource link cells isn't good enough and only the full resize-rows call fixes them
- The compatible versions popup was originally done to a kind of weird scaling, so the DPI-based scaling messes up various aspects of it

## Changes

- Now `Registry.LatestAvailableWithProvides` only returns the module with the highest version per identifier.
  Fixes #4579.
- Now `Inflator.IsHostRelated` defines which properties are mirrored, and it allows `$vref` to be mirrored. This will be easier for maintenance.
- Now if you click on one of the left column cells, the focus goes to the right cell on the same column, in case you want to edit. This also improves the initial focus.
- Now the Reset button on the command line options popup is disabled when your settings are the same as the defaults and enabled when different.
- Now a new game-related Discord menu option is the first option in the Help menu, followed by the link to the support forum, and then a separator before the CKAN-specific options. Hopefully this will steer people to the most useful places to ask their questions.
  image image
- Now we create a new `PlotModel` every time we load new data for the download statistics screen
- Now the mod list columns are auto-sized to fit their contents or to distribute the space among columns that are too wide for that
- Now the description column in the mod list has a tooltip that shows the abstract followed by a blank line and a word-wrapped rendering of the description
- Now the mod info resource links look better
- The mod name in mod info is slightly bigger
- Several instances of subtle misalignments in padding are fixed
- Now the scaling of the compatible versions popup is fixed

### DIFF
--- a/Core/Extensions/Polyfills.cs
+++ b/Core/Extensions/Polyfills.cs
@@ -56,6 +56,38 @@ namespace System.Linq
 
         #if NETFRAMEWORK || NETSTANDARD2_0
 
+        public static TSource? MaxBy<TSource, TKey>(this IEnumerable<TSource> source,
+                                                    Func<TSource, TKey>       keySelector,
+                                                    IComparer<TKey>?          comparer = null)
+        {
+            using (var iterator = source.GetEnumerator())
+            {
+                if (!iterator.MoveNext())
+                {
+                    if (default(TSource) == null)
+                    {
+                        return default;
+                    }
+                    throw new InvalidOperationException("Sequence contains no elements");
+                }
+
+                comparer ??= Comparer<TKey>.Default;
+                var maxElement = iterator.Current;
+                var maxKey     = keySelector(maxElement);
+                while (iterator.MoveNext())
+                {
+                    var currentElement = iterator.Current;
+                    var currentKey     = keySelector(currentElement);
+                    if (comparer.Compare(currentKey, maxKey) > 0)
+                    {
+                        maxElement = currentElement;
+                        maxKey     = currentKey;
+                    }
+                }
+                return maxElement;
+            }
+        }
+
         /// <summary>
         /// Eliminate duplicate elements based on the value returned by a callback
         /// </summary>
--- a/Core/Games/IGame.cs
+++ b/Core/Games/IGame.cs
@@ -54,6 +54,7 @@ namespace CKAN.Games
         Uri DefaultRepositoryURL  { get; }
         Uri RepositoryListURL     { get; }
         Uri MetadataBugtrackerURL { get; }
+        Uri DiscordURL            { get; }
         Uri ModSupportURL         { get; }
     }
 }
--- a/Core/Games/KerbalSpaceProgram.cs
+++ b/Core/Games/KerbalSpaceProgram.cs
@@ -299,6 +299,8 @@ namespace CKAN.Games.KerbalSpaceProgram
 
         public Uri MetadataBugtrackerURL => new Uri("https://github.com/KSP-CKAN/NetKAN/issues/new/choose");
 
+        public Uri DiscordURL => new Uri("https://discord.gg/65hp7G7");
+
         public Uri ModSupportURL => new Uri("https://forum.kerbalspaceprogram.com/forum/70-ksp1-technical-support-pc-modded-installs/");
 
         private static string Missions(GameInstance inst)
--- a/Core/Games/KerbalSpaceProgram2.cs
+++ b/Core/Games/KerbalSpaceProgram2.cs
@@ -245,6 +245,8 @@ namespace CKAN.Games.KerbalSpaceProgram2
 
         public Uri MetadataBugtrackerURL => new Uri("https://github.com/KSP-CKAN/KSP2-NetKAN/issues/new/choose");
 
+        public Uri DiscordURL => new Uri("https://discord.gg/ZbMp7RjVhU");
+
         public Uri ModSupportURL => new Uri("https://forum.kerbalspaceprogram.com/forum/137-ksp2-technical-support-pc-modded-installs/");
 
         private const string DataDir = "KSP2_x64_Data";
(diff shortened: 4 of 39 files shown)


HebaruSan:
The smoke test failure is because GitHub isn't working right now (which is also causing about 299 inflation errors on the status page):

- <https://www.githubstatus.com/incidents/ql942tw29yl6>

PR #4594, opened 2026-04-29
Windows dark mode in .NET 10 build

HebaruSan:
## Background

- [Windows 10 added a dark mode setting](https://support.microsoft.com/en-us/windows/personalize-your-colors-in-windows-3290d30f-d064-5cfe-6470-2fe9c6533e37)
- [.NET 9 added experimental dark mode support for WinForms](https://learn.microsoft.com/en-us/dotnet/desktop/winforms/whats-new/net90#dark-mode)
- [.NET 10 promoted that support to non-experimental](https://learn.microsoft.com/en-us/dotnet/desktop/winforms/whats-new/net100#dark-mode)

## Motivation

CKAN users on Windows want a dark mode. (Mono's WinForms implementation provides a dark mode automatically if the user's system theme is dark, so Linux users have had this all along.) We've provided some [custom registry keys to make the old WinForms system colors dark](https://github.com/KSP-CKAN/CKAN/wiki/Installing-CKAN-on-Windows#dark-theme), but it would be better if this was automatic and looked nicer, which the .NET 10 dark mode achieves.

## Changes

- The net8.0 build is updated to net10.0
  - Some references are updated to address [supply chain warnings](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/10.0/nugetaudit-transitive-packages), but many others cannot be addressed
  - Some changes are made to handle updates to nullability support in libraries
  - Apparently [the `Form.Closing` event is now replaced by `FormClosing`](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.form.onclosing?view=windowsdesktop-10.0&redirectedfrom=MSDN#System_Windows_Forms_Form_OnClosing_System_ComponentModel_CancelEventArgs_), so we've made that switch
- The net10.0-windows build is now published to a `ckan-windows.exe` file, which is attached to commits as an artifact and uploaded to releases and S3 for dev builds.
  To make this work:
  - We now get built files from the Windows build instead of the Ubuntu build (since `ckan-windows.exe` can only be built on Windows)
  - Since AltCover cannot handle testing in a different Windows container than the one that built the DLLs, `test.yml` is now eliminated and `build.yml` handles both building and testing in one step for both platforms
  - Now tests are run in Release mode instead of Debug
  - The autoupdater now looks for a remote file with the same name as the running process rather than just `ckan.exe`, which will still be the net481 build, since the `ckan-windows.exe` build would need to replace itself with the latest version of that same build
  - The SignPath config is updated to sign `ckan-windows.exe`
- When the net10.0-windows `ckan-windows.exe` build launches the GUI, if the registry key `HKEY_CURRENT_USER\SOFTWARE\Microsoft\CurrentVersion\Themes\Personalize\AppsUseLightTheme` is set to a value other than `1` (which is controlled by the display settings dark mode option), then GUI will start in dark mode:
  image
  - The title bar is also made dark when in dark mode (this is a separate and independent API call :facepalm:)
  - `ListView`s with groups will have a somewhat lighter background than is the default in dark mode, to make the hard-coded-blue group headers readable:
    image
- `LinkLabel`s will use a new `Util.LinkColorForBackColor()` function to set their text color to a blend of blue and white when the background is dark (the default is a harder-to-read blue)
- The black-on-transparent icons used in various toolbars and treeviews such as in Unmanaged Files and the Relationships tab are now inverted when the background is dark:
  Relationships | Contents
  :-- | :--
  image | image
- The flat toolstrip renderer is updated to work better by setting the gradient colors in our own custom color table, such that we can now also use it in Windows, eliminating the annoying gradient in menus and toolbars
- The borders of the main mod list column headers are restored so users can see where to click to resize them
- While working on the above, I noticed that while the "Check All" and "Check Recommendations" buttons on the choose recommendations screen disable and enable themselves when all applicable mods are or are not checked, the "Check Suggestions" button didn't. Now it does.

### Important caveats

- **In order to get Windows dark mode, users MUST use the `ckan-windows.exe` build, NOT the `ckan.exe` build!**
  `ckan.exe` is built for .NET Framework 4.8.1, which does not have dark mode support for WinForms, but is possible to run with Mono, so this is the build that will be used for Linux.
  `ckan-windows.exe` is built for .NET 10, which _does_ have dark mode support for WinForms, but is **not** possible to run with Mono.
- There will be no automatic switchover; the autoupdater will update users of previous releases to the `ckan.exe` version only (because this is already pre-determined and hard-coded by the code in the old releases), and they will have to download `ckan-windows.exe` on their own
  (But once users are on the `ckan-windows.exe` track, the autoupdater should keep them on it.)
  Maybe we can figure out a smoother way to handle this in the future.
- The new `ckan-windows.exe` build uses [single-file mode](https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/overview), but not [self-contained](https://learn.microsoft.com/en-us/dotnet/core/deploying/?pivots=visualstudio#self-contained-deployment) or [trimmed](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trim-self-contained). This allows us to still ship just one file (I do **not** want to receive support requests because somebody didn't copy one DLL out of fifty), but without it being 100+ MB.
  (The .NET Framework builds have always used ILRepack to produce one .exe file, but this does not work with .exe files on post-Framework .NET.)
  Trimming would be nice to shrink the build further, but when I tried enabling it, there were some very thorny errors to fix related to how to avoid dropping dependencies that are accessed via reflection. Leaving this as a possible future project.


HebaruSan:
The ILRepack error on Linux is:

```
Mono.Cecil.ResolutionException: Failed to resolve System.Diagnostics.Tracing.EventKeywords
  at Mono.Cecil.Mixin.CheckedResolve (Mono.Cecil.TypeReference self) [0x0000f] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.MetadataBuilder.GetConstantType (Mono.Cecil.TypeReference constant_type, System.Object constant) [0x000a4] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.MetadataBuilder.AddConstant (Mono.Cecil.IConstantProvider owner, Mono.Cecil.TypeReference type) [0x00007] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.MetadataBuilder.AddField (Mono.Cecil.FieldDefinition field) [0x00079] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.MetadataBuilder.AddFields (Mono.Cecil.TypeDefinition type) [0x00013] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.MetadataBuilder.AddType (Mono.Cecil.TypeDefinition type) [0x00093] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.MetadataBuilder.AddNestedTypes (Mono.Cecil.TypeDefinition type) [0x0001c] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.MetadataBuilder.AddType (Mono.Cecil.TypeDefinition type) [0x000ed] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.MetadataBuilder.AddTypes () [0x00018] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.MetadataBuilder.BuildTypes () [0x00014] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.MetadataBuilder.BuildModule () [0x000ad] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.MetadataBuilder.BuildMetadata () [0x00000] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.ModuleWriter.BuildMetadata (Mono.Cecil.ModuleDefinition module, Mono.Cecil.MetadataBuilder metadata) [0x00008] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.ModuleWriter.Write (Mono.Cecil.ModuleDefinition module, Mono.Disposable`1[T] stream, Mono.Cecil.WriterParameters parameters) [0x00121] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.ModuleWriter.WriteModule (Mono.Cecil.ModuleDefinition module, Mono.Disposable`1[T] stream, Mono.Cecil.WriterParameters parameters) [0x00002] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.ModuleDefinition.Write (System.String fileName, Mono.Cecil.WriterParameters parameters) [0x00017] in <49977043b607420d927e893db6be00d8>:0 
  at Mono.Cecil.AssemblyDefinition.Write (System.String fileName, Mono.Cecil.WriterParameters parameters) [0x00000] in <49977043b607420d927e893db6be00d8>:0 
  at ILRepacking.ILRepack.RepackCore (System.String tempOutputDirectory) [0x005b2] in <49977043b607420d927e893db6be00d8>:0 
  at ILRepacking.ILRepack.Repack () [0x0005d] in <49977043b607420d927e893db6be00d8>:0 
  at ILRepacking.Application.Main (System.String[] args) [0x00080] in <49977043b607420d927e893db6be00d8>:0 
```

(There should not be any change here because we're still repacking the build we generate with Mono, as before.)

HebaruSan:
Test build available here for curious users:

- <https://github.com/KSP-CKAN/CKAN/suites/66970353869/artifacts/6739321297>

It's not code-signed, so you will have to tell Windows that you trust it anyway.

PR #4603, opened 2026-05-05
Fix ckan://identifier

### LINKED ISSUE #4633: [Bug]: ckan-handler.desktop is written with UTF-8 BOM

### Operating System

Arch Linux KDE Plasma

### CKAN Version

1.36.4

### Games

KSP 1

### Game Version

N/A

_No response_

### Describe the bug

`.local/share/applications/ckan-handler.desktop` is created by ckan on startup but it seems to be malformed as KDE fails to read it properly.
This seems to be because the file was written with a [Byte order mark](https://en.wikipedia.org/wiki/Byte_order_mark). Removing the BOM with a hex editor fixes the problem, but only until the next launch when the file is recreated.

### Steps to reproduce

1. `$ rm .local/share/applications/ckan-handler.desktop`
2. Run Ckan
3. `$ desktop-file-validate .local/share/applications/ckan-handler.desktop`
See errors reported below

### Relevant log output
```
$ desktop-file-validate .local/share/applications/ckan-handler.desktop
.local/share/applications/ckan-handler.desktop: error: file contains line "﻿[Desktop Entry]", which is not a comment, a group or an entry
.local/share/applications/ckan-handler.desktop: error: file contains entry "Version=1.0" before the first group, but only comments are accepted before the first group
.local/share/applications/ckan-handler.desktop: error: file contains entry "Type=Application" before the first group, but only comments are accepted before the first group
.local/share/applications/ckan-handler.desktop: error: file contains entry "Exec=mono "/usr/lib/ckan/ckan.exe" gui %u" before the first group, but only comments are accepted before the first group
.local/share/applications/ckan-handler.desktop: error: file contains entry "Icon=ckan" before the first group, but only comments are accepted before the first group
.local/share/applications/ckan-handler.desktop: error: file contains entry "StartupNotify=true" before the first group, but only comments are accepted before the first group
.local/share/applications/ckan-handler.desktop: error: file contains entry "NoDisplay=true" before the first group, but only comments are accepted before the first group
.local/share/applications/ckan-handler.desktop: error: file contains entry "Terminal=false" before the first group, but only comments are accepted before the first group
.local/share/applications/ckan-handler.desktop: error: file contains entry "Categories=Utility" before the first group, but only comments are accepted before the first group
.local/share/applications/ckan-handler.desktop: error: file contains entry "MimeType=x-scheme-handler/ckan" before the first group, but only comments are accepted before the first group
.local/share/applications/ckan-handler.desktop: error: file contains entry "Name=CKAN Launcher" before the first group, but only comments are accepted before the first group
.local/share/applications/ckan-handler.desktop: error: file contains entry "Comment=Launch CKAN" before the first group, but only comments are accepted before the first group
```

Halbann:
- this.focusIdent was hidden by local variable focusIdent. Renamed focusIdent to focusIdentArg.

- When stripping focusIdent, StartsWith("ckan://") condition  was unreachable due to prior Length > 2 condition.

These issues seemed to be introduced in https://github.com/KSP-CKAN/CKAN/pull/4176.

The context is that I'm working on an in-game helper for finding recolour mods to reduce the number of people getting stuck trying to use Lazy Painter.

image

For future reference, it would be good to look at how to have the URL handler talk to the running process instead of starting a new process every time.

Fixes #4633.

### DIFF
--- a/GUI/Main/Main.cs
+++ b/GUI/Main/Main.cs
@@ -68,8 +68,9 @@ namespace CKAN.GUI
             log.Info("Starting the GUI");
             if (//cmdlineArgs is [_, string focusIdent, ..]
                 cmdlineArgs.Length > 1
-                && cmdlineArgs[1] is string { Length: >= 2 } focusIdentArg
+                && cmdlineArgs[1] is string { Length: >= 2 } focusIdentArg)
             {
+                // Strip any leading `//` or any leading `ckan://`.
                 if (//focusIdentArg is ['/', '/', .. var rest]
                     focusIdentArg.Length > 2
                     && focusIdentArg.StartsWith("//"))
@@ -82,10 +83,13 @@ namespace CKAN.GUI
                 {
                     focusIdent = focusIdentArg[7..];
                 }
+
+                // Strip any trailing forward slashes.
                 if (//focusIdent is [.. var start, '/']
-                    focusIdentArg.EndsWith("/"))
+                    focusIdent is { Length: > 1 }
+                    && focusIdent.EndsWith("/"))
                 {
-                    focusIdent = focusIdentArg[..^1];
+                    focusIdent = focusIdent.TrimEnd('/');
                 }
             }
 
--- a/GUI/Main/Main.cs
+++ b/GUI/Main/Main.cs
@@ -70,6 +70,8 @@ namespace CKAN.GUI
                 cmdlineArgs.Length > 1
                 && cmdlineArgs[1] is string { Length: >= 2 } focusIdentArg)
             {
+                focusIdent = focusIdentArg;
+
                 // Strip any leading "//" or any leading "ckan://".
                 if (//focusIdentArg is ['/', '/', .. var rest]
                     focusIdentArg.Length > 2
--- a/GUI/Main/Main.cs
+++ b/GUI/Main/Main.cs
@@ -68,22 +68,23 @@ namespace CKAN.GUI
             log.Info("Starting the GUI");
             if (//cmdlineArgs is [_, string focusIdent, ..]
                 cmdlineArgs.Length > 1
-                && cmdlineArgs[1] is string focusIdent)
+                && cmdlineArgs[1] is string focusIdentArg
+                && focusIdentArg.Length >= 2)
             {
                 if (//focusIdent is ['/', '/', .. var rest]
-                    focusIdent.Length > 2)
+                    focusIdentArg.StartsWith("//"))
                 {
-                    focusIdent = focusIdent[2..];
+                    focusIdent = focusIdentArg[2..];
                 }
                 else if (//focusIdent is ['c', 'k', 'a', 'n', ':', '/', '/', .. var rest2]
-                    focusIdent.StartsWith("ckan://"))
+                    focusIdentArg.StartsWith("ckan://"))
                 {
-                    focusIdent = focusIdent[7..];
+                    focusIdent = focusIdentArg[7..];
                 }
                 if (//focusIdent is [.. var start, '/']
-                    focusIdent.EndsWith("/"))
+                    focusIdentArg.EndsWith("/"))
                 {
-                    focusIdent = focusIdent[..^1];
+                    focusIdent = focusIdentArg[..^1];
                 }
             }
 
--- a/GUI/Main/Main.cs
+++ b/GUI/Main/Main.cs
@@ -68,16 +68,17 @@ namespace CKAN.GUI
             log.Info("Starting the GUI");
             if (//cmdlineArgs is [_, string focusIdent, ..]
                 cmdlineArgs.Length > 1
-                && cmdlineArgs[1] is string focusIdentArg
-                && focusIdentArg.Length >= 2)
+                && cmdlineArgs[1] is string { Length: >= 2 } focusIdentArg
             {
-                if (//focusIdent is ['/', '/', .. var rest]
-                    focusIdentArg.StartsWith("//"))
+                if (//focusIdentArg is ['/', '/', .. var rest]
+                    focusIdentArg.Length > 2
+                    && focusIdentArg.StartsWith("//"))
                 {
                     focusIdent = focusIdentArg[2..];
                 }
-                else if (//focusIdent is ['c', 'k', 'a', 'n', ':', '/', '/', .. var rest2]
-                    focusIdentArg.StartsWith("ckan://"))
+                else if (//focusIdenArg is ['c', 'k', 'a', 'n', ':', '/', '/', .. var rest2]
+                    focusIdentArg.Length > 7
+                    && focusIdentArg.StartsWith("ckan://"))
                 {
                     focusIdent = focusIdentArg[7..];
                 }
--- a/GUI/Main/Main.cs
+++ b/GUI/Main/Main.cs
@@ -70,7 +70,7 @@ namespace CKAN.GUI
                 cmdlineArgs.Length > 1
                 && cmdlineArgs[1] is string { Length: >= 2 } focusIdentArg)
             {
-                // Strip any leading `//` or any leading `ckan://`.
+                // Strip any leading "//" or any leading "ckan://".
                 if (//focusIdentArg is ['/', '/', .. var rest]
                     focusIdentArg.Length > 2
                     && focusIdentArg.StartsWith("//"))
--- a/GUI/CKAN-GUI.csproj
+++ b/GUI/CKAN-GUI.csproj
@@ -35,7 +35,6 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Autofac" Version="9.1.0" />
-    <PackageReference Include="ini-parser" Version="2.5.2" />
     <PackageReference Include="log4net" Version="3.3.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
     <PackageReference Include="Nullable" Version="1.3.1" PrivateAssets="all" />
--- a/GUI/Controls/ManageMods.cs
+++ b/GUI/Controls/ManageMods.cs
@@ -1341,7 +1341,8 @@ namespace CKAN.GUI
                     bool row_match;
                     if (exactMatch)
                     {
-                        row_match = mod?.Name == key || mod?.Identifier == key;
+                        row_match = mod?.Name == key
+                                    || string.Equals(mod?.Identifier, key, StringComparison.OrdinalIgnoreCase);
                     }
                     else
                     {
--- a/GUI/Controls/ManageMods.cs
+++ b/GUI/Controls/ManageMods.cs
@@ -1350,7 +1350,8 @@ namespace CKAN.GUI
                     bool row_match;
                     if (exactMatch)
                     {
-                        row_match = mod?.Name == key || mod?.Identifier == key;
+                        row_match = mod?.Name == key
+                                    || string.Equals(mod?.Identifier, key, StringComparison.OrdinalIgnoreCase);
                     }
                     else
                     {
--- a/GUI/URLHandlers.cs
+++ b/GUI/URLHandlers.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Text;
 #if !NET5_0_OR_GREATER
 using System.Reflection;
 #endif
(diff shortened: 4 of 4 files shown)


GUI/Main/Main.cs
```diff
@@ -68,22 +68,23 @@ public Main(string[]             cmdlineArgs,
             log.Info("Starting the GUI");
             if (//cmdlineArgs is [_, string focusIdent, ..]
                 cmdlineArgs.Length > 1
-                && cmdlineArgs[1] is string focusIdent)
+                && cmdlineArgs[1] is string focusIdentArg
+                && focusIdentArg.Length >= 2)
             {
                 if (//focusIdent is ['/', '/', .. var rest]
-                    focusIdent.Length > 2)
+                    focusIdentArg.StartsWith("//"))
```
HebaruSan:
```suggestion
                    focusIdentArg.Length > 2
                    && focusIdentArg.StartsWith("//"))
```

Good catch! I think we should keep the length check and add the `StartsWith` to it, so an argument of just `//` doesn't confuse it.

GUI/Main/Main.cs
```diff
@@ -68,22 +68,23 @@ public Main(string[]             cmdlineArgs,
...
                 {
-                    focusIdent = focusIdent[2..];
+                    focusIdent = focusIdentArg[2..];
                 }
                 else if (//focusIdent is ['c', 'k', 'a', 'n', ':', '/', '/', .. var rest2]
-                    focusIdent.StartsWith("ckan://"))
+                    focusIdentArg.StartsWith("ckan://"))
```
HebaruSan:
```suggestion
                    focusIdentArg.Length > 7
                    && focusIdentArg.StartsWith("ckan://"))
```

Similar to the above, I think we should avoid setting the var if the arg is just `ckan://` with no identifier.

GUI/Main/Main.cs
```diff
@@ -68,22 +68,23 @@ public Main(string[]             cmdlineArgs,
...
                 {
-                    focusIdent = focusIdent[7..];
+                    focusIdent = focusIdentArg[7..];
                 }
                 if (//focusIdent is [.. var start, '/']
-                    focusIdent.EndsWith("/"))
+                    focusIdentArg.EndsWith("/"))
                 {
-                    focusIdent = focusIdent[..^1];
+                    focusIdent = focusIdentArg[..^1];
```
HebaruSan:
```suggestion
                    focusIdent is { Length: > 1 }
                    && focusIdent.EndsWith("/"))
{
                    focusIdent = focusIdent[..^1];
```

This one is actually supposed to be using the original variable name; it's stripping a trailing `/` off of the values set by the above assignments, if any. But that requires a null and length check.

Maybe this would be clearer as `.TrimEnd()` calls in the above assignments, feel free to do that instead if you prefer.

GUI/Main/Main.cs
```diff
@@ -68,22 +68,23 @@ public Main(string[]             cmdlineArgs,
             log.Info("Starting the GUI");
             if (//cmdlineArgs is [_, string focusIdent, ..]
                 cmdlineArgs.Length > 1
-                && cmdlineArgs[1] is string focusIdent)
+                && cmdlineArgs[1] is string focusIdentArg
+                && focusIdentArg.Length >= 2)
             {
                 if (//focusIdent is ['/', '/', .. var rest]
```
HebaruSan:
```suggestion
                if (//focusIdentArg is ['/', '/', .. var rest]
```

Let's keep the comment consistent with the code in case we are ever able to update past C# 9.0.

GUI/Main/Main.cs
```diff
@@ -68,22 +68,23 @@ public Main(string[]             cmdlineArgs,
...
-                    focusIdent.Length > 2)
+                    focusIdentArg.StartsWith("//"))
                 {
-                    focusIdent = focusIdent[2..];
+                    focusIdent = focusIdentArg[2..];
                 }
                 else if (//focusIdent is ['c', 'k', 'a', 'n', ':', '/', '/', .. var rest2]
```
HebaruSan:
```suggestion
                else if (//focusIdenArg is ['c', 'k', 'a', 'n', ':', '/', '/', .. var rest2]
```

Another comment that should track the correct variable name.

GUI/Main/Main.cs
```diff
@@ -68,22 +68,23 @@ public Main(string[]             cmdlineArgs,
             log.Info("Starting the GUI");
             if (//cmdlineArgs is [_, string focusIdent, ..]
                 cmdlineArgs.Length > 1
-                && cmdlineArgs[1] is string focusIdent)
+                && cmdlineArgs[1] is string focusIdentArg
+                && focusIdentArg.Length >= 2)
```
HebaruSan:
```suggestion
                && cmdlineArgs[1] is string { Length: >= 2 } focusIdentArg
```

Pattern matching can do this in one step.

HebaruSan (dismissed):
Added a couple of code review comments. Thanks again! It'll be great to have this fixed (and it's embarrassing that it broke).

GUI/Main/Main.cs
Halbann: Thanks and sorry for nearly introducing a free bug with my fix there. I agree that TrimEnd would be clearer. Has the added benefit of removing multiple forward slashes if they should be there for some reason.

Halbann:
Thank you for making those suggestions. Any other suggestions not withstanding, shall we leave this PR open to include fixes for linux and mac? I fixed an issue with the Linux URL handler this morning and found [something](https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleurltypes) that could work on Mac (unable to test it myself; if it's not too much trouble, perhaps @cheesebellies wouldn't mind helping when it's ready?).

HebaruSan:
> Any other suggestions not withstanding, shall we leave this PR open to include fixes for linux and mac?

That's fine with me, take all the time you like!

HebaruSan:
Another suggestion for a fixhancement to the existing functionality while we're working on it: the identifier matching should probably be case insensitive. I initially tested with ckan://astrogator/ and was confused when it didn't work, until I tried ckan://Astrogator/ which did work.

Halbann:
Agreed. I ran into that exact same confusion the first time I tried it out.

HebaruSan:
I believe this is the line to change for that:

https://github.com/KSP-CKAN/CKAN/blob/23dc9f874890342eff4e0ddbe77d859a1926ddb2/GUI/Controls/ManageMods.cs#L1344

cheesebellies:
I would love to help test this on Mac. I downloaded the code but can't for the life of me figure out how to build it. When I try to run `build.sh` I get a build error, and when I try to `make` I also get an error. Also, once it's built, how should I test whether the url works?

```
Build succeeded with 11 warning(s) in 2.0s
An error occurred when executing task 'Build'.

All netkan indexing stops for a prolonged period (days).
The errors thrown are non-conclusive and the bots keep
getting stuck on what appears to be a very standard .netkan

<paraphrase>
(Daz) So err... I found an extra space in one of the recommends
(pjf) That's it!
(Daz) https://github.com/KSP-CKAN/NetKAN/pull/1062
</paraphrase>


Task                               Duration            Status              
---------------------------------------------------------------------------
Setup                              00:00:00.0003242    Succeeded           
Restore                            00:00:13.0957345    Succeeded           
Generate-GlobalAssemblyVersionInfo 00:00:00.0221451    Succeeded           
Build                              00:00:02.1382656    Failed              
Teardown                           00:00:00.0011095    Succeeded           
---------------------------------------------------------------------------
Total:                             00:00:15.2575789                        
Error: Could not resolve MSBuild.

```

`make: *** No rule to make target `../_build/out/CKAN-CmdLine/Debug/bin/net10.0/osx-x64/publish', needed by `../_build/osx/dmg/CKAN.app/Contents/MacOS/x86_64'.  Stop.`

HebaruSan:
Hi @cheesebellies, that's great, thanks for your interest!

However, there are no Mac-related changes here yet. My understanding is that @Halbann intends to add Mac support for `ckan://identifier` style URLs from scratch (there is no code for this today). I don't know what the timeline would be for that, but if and when it happens, the CKAN.dmg file will be in the Artifacts dropdown on this page:

- <https://github.com/KSP-CKAN/CKAN/pull/4603/checks>

```
Error: Could not resolve MSBuild.
```

This might mean that you don't have the dotnet SDK installed...?

- <https://dotnet.microsoft.com/en-us/download/dotnet/10.0>

There are some guides for doing dev, but they might be a bit out of date since a lot has changed over time:

- <https://github.com/KSP-CKAN/CKAN/blob/master/doc/building.md>
- <https://github.com/KSP-CKAN/CKAN/wiki/Building-CKAN>
- <https://github.com/KSP-CKAN/CKAN/wiki/Build-Environment>
- <https://github.com/KSP-CKAN/CKAN/wiki/New-Developers%27-Guide>
- <https://github.com/KSP-CKAN/CKAN/wiki/Developers%27-Guide>

cheesebellies:
Got the DMG, thanks.

I just installed dotnet 10 on homebrew so I'm pretty certain that's not the issue. It might be an x86_64 vs arm64 issue, but since I can just download the DMG from the github build, it's not really a problem. Thanks for the dev guides as well, while they're out of date for sure (asks for net 8 lol) they're still really helpful, maybe I'll have something to add in the future.

Let me know whenever you need my help!

HebaruSan:
> I just installed dotnet 10 on homebrew so I'm pretty certain that's not the issue.

It might also mean that you don't have Mono's build tools installed; both are needed to build on Linux, and I think the same build steps are inherited on the Mac:

- <https://www.mono-project.com/docs/getting-started/install/mac/>

Halbann:
> shall we leave this PR open to include fixes for linux and mac?

On second thought, if we can get another Linux user or two to verify that the URL handler registration works for them, it would be useful to wrap up this PR as a fix for current behaviour. That way I can make structural changes in a separate improvement PR more easily.

GUI/Main/Main.cs
```diff
@@ -68,22 +68,28 @@ public Main(string[]             cmdlineArgs,
             log.Info("Starting the GUI");
             if (//cmdlineArgs is [_, string focusIdent, ..]
                 cmdlineArgs.Length > 1
-                && cmdlineArgs[1] is string focusIdent)
+                && cmdlineArgs[1] is string { Length: >= 2 } focusIdentArg)
```
HebaruSan, line 71: I think one other thing that broke in #4176 and is still broken here is `ckan.exe gui Astrogator` without any trappings of URL syntax. We used to assign the first argument to `focusIdent` and then massage it, but now it only gets assigned in the inner `if` blocks:

https://github.com/KSP-CKAN/CKAN/blob/3f578cc47600eaf6ecd0faa77706bce25f51d631/GUI/Main/Main.cs#L64-L80

HebaruSan:
> On second thought, if we can get another Linux user or two to verify that the URL handler registration works for them

What are you expecting to work as a method of testing? I tried passing a URL to `xdg-open` since that's a way of asking the desktop environment to open something (and works with `https://` URLs), and it doesn't seem to work:

```
$ xdg-open ckan://Astrogator/
gio: ckan://Astrogator/: The specified location is not supported
```

Halbann:
`xdg-open ckan://Astrogator/` is working as expected for me. It could be a discrepancy in how `xdg-mime` and `update-desktop-database` work between distros.

Which distro are you on?

For convenience, anybody who wants to help out can use this HTML file:
[url_handler_demo.html](https://github.com/user-attachments/files/27539361/url_handler_demo.html)

Halbann:
Oh and please would you mind having a look inside `~/.local/share/applications` to see if `ckan-handler.desktop` and `mimeinfo.cache` exist after running CKAN at least once.

The handler file should look like this, only with whatever your executable path is.

```ini
[Desktop Entry]
Version=1.0
Type=Application
Exec=mono "/home/lawrence/repos/CKAN/_build/ckan.exe" gui %u
Icon=ckan
StartupNotify=true
NoDisplay=true
Terminal=false
Categories=Utility
MimeType=x-scheme-handler/ckan
Name=CKAN Launcher
Comment=Launch CKAN
```

The cache file should look like this.

```ini
[MIME Cache]
x-scheme-handler/ckan=ckan-handler.desktop;
```

GUI/Main/Main.cs
Halbann: Thanks. Added that line back in and it works as expected.

HebaruSan (approved):
Thanks for fixing this! 👾

I haven't done a full testing pass on Linux, but I'm confident from a close audit that these changes will at least be no worse than what was already there. (I checked the `ini-parser` project to see if there was a way to turn off BOMs, since it's always nice to use a library when interacting with a specialized domain, but as far as I could tell there is no such option.)

PR #4606, opened 2026-05-06
Automatically pull focus for the Terminal window on MacOS

cheesebellies:
When you open CKAN in MacOS, it doesn't automatically focus the Terminal window. This means it gets hidden behind other windows, and you have to alt-tab to find and focus it.

This PR automatically focuses the Terminal window.

### DIFF
--- a/macosx/CKAN
+++ b/macosx/CKAN
@@ -23,6 +23,7 @@ cd "$MACOS_PATH"
 
 osascript <<END
 tell application "Terminal"
-    do script "cd \"`pwd`\"; ./CKAN-CmdLine consoleui" in front window
+    activate
+    do script "cd \"`pwd`\"; ./CKAN-CmdLine consoleui"
 end tell
 END
--- a/macosx/CKAN
+++ b/macosx/CKAN
@@ -23,6 +23,6 @@ cd "$MACOS_PATH"
 
 osascript <<END
 tell application "Terminal"
-    do script "cd \"`pwd`\"; ./CKAN-CmdLine consoleui"
+    do script "cd \"`pwd`\"; ./CKAN-CmdLine consoleui" in front window
 end tell
 END


HebaruSan:
Hey @cheesebellies, thanks for the contribution!

I don't have a Mac handy for testing, so I have to rely on online documentation to confirm that this will work.

So far it looks like most people use the `activate` command to make Terminal the active application after running a command in it.

Are you sure that `in front window` is better? I am concerned that it might attempt to run the command in an existing window if the user is already using Terminal for something else, possibly interfering with their other programs.

cheesebellies:
I originally tried to use `activate`, but it would open an extra window every time. `in front window` fixed that but as you said might run the command in an open window. And of course, as it turns out, the original command opens two windows by default anyways. I looked into it and it has to do with how the Terminal application handles focus, and can't be fixed without some really hacky AppleScript.

In the end, it is actually better just to use `activate`, and just deal with the extra window. I've made sure to rigorously test the script this time... 😅

```osascript
tell application "Terminal"
    activate
    do script "cd \"`pwd`\"; ./CKAN-CmdLine consoleui"
end tell
```

HebaruSan:
What happens if we put the `activate` call _after_ the `do script`? Maybe then it would open one window with CKAN in it, and just make Terminal the foreground application without opening another window?

```osascript
tell application "Terminal"
    do script "cd \"`pwd`\"; ./CKAN-CmdLine consoleui"
    activate
end tell
```

cheesebellies:
It does the same as putting `activate` before, unfortunately. You'd think `do script` opens the window, but it puts it in a weird background state, and switching to the app in any way triggers the `activate` behavior.

HebaruSan (approved):
Ah well, it was worth a try, thanks for giving it a shot.

And thanks again for contributing! Please don't hesitate to submit more changes if you think of anything else you'd like to add. 🎈

PR #4638, opened 2026-05-19
Truncate date columns

KerballOne:
The CKAN GUI has two columns that can be truncated.  Release date and Install date.  If the date is today, just show the time, if not, just show the date without the time.

Screenshot 2026-05-18 174251
image

### DIFF
--- a/GUI/Model/ModList.cs
+++ b/GUI/Model/ModList.cs
@@ -205,16 +205,16 @@ namespace CKAN.GUI
                     Value = "-"
                 };
 
-            var name           = new DataGridViewTextBoxCell { Value = ToGridText(mod.Name)                                  };
-            var author         = new DataGridViewTextBoxCell { Value = ToGridText(string.Join(", ", mod.Authors))            };
-            var installVersion = new DataGridViewTextBoxCell { Value = mod.InstalledVersion ?? ""                            };
-            var latestVersion  = new DataGridViewTextBoxCell { Value = mod.LatestVersion                                     };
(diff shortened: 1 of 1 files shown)


HebaruSan:
- Column width changes reverted to not mess up the changes from #4671
- See #4488 for why install date is already fine and doesn't need `.ToLocalTime()`

KerballOne:
> * Column width changes reverted to not mess up the changes from [Optimize new upgradeability logic #4671](https://github.com/KSP-CKAN/CKAN/pull/4671)

Okay, I figured there was some more changes to column width in the pipeline.


> * See [Display release date in local time zone #4488](https://github.com/KSP-CKAN/CKAN/pull/4488) for why install date is already fine and doesn't need `.ToLocalTime()`

Not sure if maybe I was using a version that did not have #4488... but it was an inconsistent issue where InstallDate was not always using Local time.  Like, it would be correct as Local initially (when installed)... but after a restart of ckan, it would go back to UTC.

HebaruSan:
> Not sure if maybe I was using a version that did not have #4488... but it was an inconsistent issue where InstallDate was not always using Local time. Like, it would be correct as Local initially (when installed)... but after a restart of ckan, it would go back to UTC.

Oh, OK. Since there was no explanation of that change, I guessed you were just adding it because you saw it for release date, not for a motivated reason. Timezones are extremely tricky, so we need to know why changes are happening so we can make sure they won't break something.

Trying the steps you describe, I do see a change in the displayed timestamp after relaunching CKAN. I'll add a fix for this in a different pull request so this one can remain focused on the format.

PR #4677, opened 2026-06-19
Support Windows text scaling setting on net10

### LINKED ISSUE #4673: [Bug]: CKAN v1.36.4 UI scaling regression on Windows 10

### Operating System

Win10 22H2 19045.4717

### CKAN Version

1.36.2 -> 1.36.4

### Games

KSP 1

### Game Version

1.12.5.3190

No

### Describe the bug

I have a 4k monitor with 200% UI Scale
On the latest version some kind of regression resulted in UI being scaled up only partly (Buttons and Menus have correct scale, the grid and side pane has 1x scale. Text is barely readable.

### Steps to reproduce

Upgrade 1.36.2 -> 1.36.4 using Auto-update or manually.

### Relevant log output

v1.36.4 scaling broken:
Image
v1.36.2 scaling works as it should
Image

### LINKED ISSUE #4676: [Bug]: CKAN removed the Squad folder in GameData after requesting a spacesuit mod located inside of it be removed

### Operating System

Windows 11

### CKAN Version

1.36.4.26132

### Games

KSP 1

### Game Version

112.5

No

### Describe the bug

I was trying to remove mods from a copy of KSP I had, but when I removed them all using CKAN the Squad folder was also removed. I investigated and found that when removing the mod "KSP Artemis Suit" made by Souptime, which was located inside the Squad folder, CKAN seems to just remove the folder in its entirety. Tried this multiple times with the same result.

### Steps to reproduce

I had the mod "KSP Artemis Suit" by Souptime downloaded from CKAN, and trying to have CKAN remove it removes the Squad folder.

### Relevant log output

_No response_

HebaruSan:
## Problems

- If you install `KSPArtemisSuit` and then uninstall it, CKAN offers to delete `GameData/Squad`:
  Image
- If you install a mod, the value in the Install Time column is correct initially, but changes to UTC if you close and re-open CKAN.
  Reported by @KerballOne in #4638.
- When the system DPI is higher than 96, some places in the GUI appear too tall, such as the description in mod info.
- The Metadata tab in mod info still has some alignment and wrapping issues, especially with the resource links.
- I encountered a scenario where the GUI froze on loading in the current dev build.
- Resizing can be a bit jittery in the current dev build.
- If you install some mods, the Play button in the toolbar often steals focus and shows its dropdown afterwards.
- The button for adding a new search should be aligned and sized to match up nicely with the bottommost search's dropdown button, but it isn't quite.

## Causes

- Originally, CKAN would simply ignore folders that had files it did not install when uninstalling, and this protected `GameData/Squad`, since it would always have such files. (Notably, if you test with an empty `Squad` folder, even the old code deletes it!)
  Later, #2962 added the ability to ask the user what to do about such folders at uninstallation. For typical mod folders, this means asking about leftover settings or cached files. However, no special provisions were added to protect `GameData/Squad`, so it was possible for it to appear in the list of directories the user might choose to delete.
- The install timestamp was serialized with a timezone marker, then deserialized into UTC and rendered as-is, which shows the UTC version.
- `Util.StringHeight` included logic to account for DPI scaling, but this was redundant with the underlying `Graphics.MeasureString` function on Windows, which already handles DPI scaling.
- `ModInfo` uses a 2-column `TableLayoutPanel` to arrange its labels and values. The columns were both set to `AutoSize` mode, which meant their widths depended on their contents. But the values (especially the links) can wrap, which meant that the widths of the contents also depended on the widths of the columns! This circular dependency meant there was not really a stable configuration of the overall system. There were also some quirks of some values having different heights, and the logic to wrap the links relied on setting the table's row heights, which was kind of crazy.
- The new column auto sizing on load from #4671 was sometimes doing GUI operations outside the GUI thread, which is a no-no in WinForms.
- `ManageMods.OnResize` was setting the auto size mode twice for each column, which probably incurred extra layout processing.
- Clicking a button in the toolbar gives it focus, and when installation finishes, the Apply button is disabled, which transfers focus to the first button in the toolbar.
- The new-search button's size was static, whereas the dropdown's size is set to match the adjacent `TextBox`'s height, which is unpredictable. The new-search button's position relied on anchoring to the right and bottom, which moved it up and down appropriately, but didn't achieve pixel-perfect alignment.

## Motivations

- Author names in mod info are shown with just space between them. Since they can contain spaces internally, this is a bit visually ambiguous.
- #4543 added DPI-based UI scaling, so text would always appear the same physical size regardless of monitor specifications. Windows also has a separate text scaling setting, and it would be nice to support that as well.

## Changes

- Now `ModuleInstaller.Uninstall` excludes everything in `IGame.StockFolders` from its `possibleConfigOnlyDirs` output parameter, and will also not delete them if empty.
  Tests are included to exercise this that fail on the old code and pass on the new code.
  Fixes #4676.
- Now we serialize dates in UTC and use `.ToLocalTime()` to convert the install timestamp to local time.
- Now the DPI scaling in `Util.StringHeight` is Mono-only.
- Now the `ModInfo` table's left column has a static width, and the right column takes up the remaining space, which finally clearly defines who is in charge of setting the widths. The links are then much easier to wrap by simply setting `MaximumSize.Width`, which we do in a new handler for the table's `Resize` event. Many other quirks were ironed out as well.
- Now the column auto sizing on load is moved inside of a `Util.Invoke` call to put it on the GUI thread.
- Now the column auto sizing on resize is debounced, so we set the appropriate columns to `Fill` mode immediately, and then set them back to `NotSet` 250 milliseconds after the last resize event. This way the normal sizing logic can handle rapid events without worrying about modes being toggled.
- Now the toolbar is taken out of the tab order, and when it receives focus, we set focus back to the grid.
- Now `EditModSearch` exposes `ButtonLocation` and `ButtonSize` properties to tell its parent control where its button is, which `EditModSearches` uses in its `Resize` handler to move the new-search button to the immediate right edge of the dropdown button at a matching size.
- Now the author name links have commas after them.
- Now if the Windows text scaling setting (registry key `HKEY_CURRENT_USER\Software\Microsoft\Accessibility\TextScaleFactor`) is set to something greater than 100%, the .NET 10 build uses `Application.SetDefaultFont` to scale the default font, and `Form.Scale` is used to scale the various forms by the same amount. An exception is made for `Main` because mod info's layout breaks if we scale the main form. We can enable that in the future if we figure out the reason, but for now it's not really needed because the main form doesn't use static layout the way the settings dialog does.
  The net481 build is not changed because `Application.SetDefaultFont` is not available in that version.
  Fixes #4673.

### DIFF
--- a/Core/Extensions/EnumerableExtensions.cs
+++ b/Core/Extensions/EnumerableExtensions.cs
@@ -231,8 +231,6 @@ namespace CKAN.Extensions
             {
                 result = func(result, item);
                 yield return result;
-
-
             }
         }
 
--- a/Core/Games/KerbalSpaceProgram.cs
+++ b/Core/Games/KerbalSpaceProgram.cs
@@ -101,17 +101,17 @@ namespace CKAN.Games.KerbalSpaceProgram
         /// Checks the path against a list of reserved game directories
         /// </summary>
         /// <param name="inst">Game instance we're checking</param>
-        /// <param name="path">Path to check</param>
+        /// <param name="absPath">Absolute path to check</param>
         /// <returns>True if reserved, false otherwise</returns>
-        public bool IsReservedDirectory(GameInstance inst, string path)
-            => path == inst.GameDir || path == inst.CkanDir
-            || path == PrimaryModDirectory(inst)
-            || path == Missions(inst)
-            || path == Scenarios(inst) || path == Tutorial(inst)
-            || path == Ships(inst)     || path == ShipsThumbs(inst)
-            || path == ShipsVab(inst)  || path == ShipsThumbsVAB(inst)
-            || path == ShipsSph(inst)  || path == ShipsThumbsSPH(inst)
-            || path == ShipsScript(inst);
+        public bool IsReservedDirectory(GameInstance inst, string absPath)
+            => absPath == inst.GameDir || absPath == inst.CkanDir
+            || absPath == PrimaryModDirectory(inst)
+            || absPath == Missions(inst)
+            || absPath == Scenarios(inst) || absPath == Tutorial(inst)
+            || absPath == Ships(inst)     || absPath == ShipsThumbs(inst)
+            || absPath == ShipsVab(inst)  || absPath == ShipsThumbsVAB(inst)
+            || absPath == ShipsSph(inst)  || absPath == ShipsThumbsSPH(inst)
+            || absPath == ShipsScript(inst);
 
         public bool AllowInstallationIn(string name, [NotNullWhen(true)] out string? path)
             => allowedFolders.TryGetValue(name, out path);
--- a/Core/IO/ModuleInstaller.cs
+++ b/Core/IO/ModuleInstaller.cs
@@ -1007,19 +1007,20 @@ namespace CKAN.IO
                 // before parents. GH #78.
                 foreach (string directory in directoriesToDelete.OrderByDescending(dir => dir.Length))
                 {
+                    var relPath = instance.ToRelativeGameDir(directory);
                     log.DebugFormat("Checking {0}...", directory);
                     // It is bad if any of this directories gets removed
                     // So we protect them
                     // A few string comparisons will be cheaper than hitting the disk, so do this first
-                    if (instance.Game.IsReservedDirectory(instance, directory))
+                    if (instance.Game.IsReservedDirectory(instance, directory)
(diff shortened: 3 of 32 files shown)


HebaruSan:
Oops; accidentally pressed <kbd>Ctrl</kbd>+<kbd>Enter</kbd> in vscode's pull request creation screen before the description was written...

Issue #4457, opened 2025-11-17
[Feature]: Support for Kitten Space Agency


HebaruSan:
### Motivation

[Kitten Space Agency](https://ahwoo.com/store/KPbAA1Au/kitten-space-agency) started its public pre-alpha a few days ago, and [there are already some mods on its forum](https://forums.ahwoo.com/forums/mod-releases.21/).

After KSP-SpaceDock/SpaceDock#518 is merged in approximately 19 hours, it will be possible to upload KSA mods to SpaceDock. People (including the KSA team themselves!) will expect CKAN support before too long.

### Suggestion

Here are the main steps needed to add a new game to CKAN, ordered approximately by functional dependencies, with links to how we did them for KSP2 for reference:

- [ ] Update the client to detect game instances, download metadata from the new repo, etc.
  - #3797
- [ ] Update the metadata tester to support the new game
  - KSP-CKAN/xKAN-meta_testing#91
- [ ] Make repos for the metadata and populate the `.github` folders:
  - <https://github.com/KSP-CKAN/KSP2-NetKAN>
  - <https://github.com/KSP-CKAN/KSP2-CKAN-meta>
- [ ] Update the back end to support the new game
  - KSP-CKAN/NetKAN-Infra#284
  - KSP-CKAN/NetKAN-Infra#289
- [ ] Add the new game to the status page filters:
  - KSP-CKAN/NetKAN-status#25
- [ ] Update/create wiki pages with game-specific info
  - <https://github.com/KSP-CKAN/CKAN/wiki/User-guide>
  - <https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN>
  - <https://github.com/KSP-CKAN/CKAN/wiki/How-to-Wrangle-Metadata-for-KSP2>

@cheese3660, @averageksp, and Discord user Plazma have expressed interest in working on this, although as far as I know no one has started coding yet.

sparr:
Does CKAN support mods for Juno New Origins? If not, I suspect the reasoning would also apply here.

HebaruSan:
Hi @sparr, no, nobody has added support for JNO yet, either. Would that interest you?

jokil123:
[...]

PR #4688, opened 2026-07-06
[Feature]: Support for Kitten Space Agency

averageksp:
## What

I'm opening this on behalf of [KSAModding](https://github.com/KSAModding) and everyone who worked on this fork over the past few days, adding Kitten Space Agency as a supported game in CKAN. This has been built, beta tested, and iterated on as a separate fork (KSAModding/CKAN-KSA), and we think it's ready to bring here.

Related: [Feature]: Support for Kitten Space Agency #4457, and following the same shape as #3797 (KSP2 support).

## The core change

At the architecture level, KSA needed four things added to `IGame` that no existing game did:

- `ModDirectoryIsExternal`. KSA doesn't install mods inside its own game folder like KSP and KSP2 do, it uses one shared, per-user mods folder under Documents, the same regardless of which install you're running. `GameInstance.ToAbsoluteGameDir`/`ToRelativeGameDir` map a game's mod root outside its own game directory when this is set.
- `AdjustCommandLine` gained a `GameInstance` parameter. KSA code mods need a separate loader, StarMap, instead of the game exe. This lets CKAN check the mods folder at launch time and swap in the right executable.
- `FormatVersion`. KSA shows `year.month.*` rather than a normal version string.
- A `ProcessLoadedModsBeforeGameStart` hook. Keeps KSA's own manifest.toml in sync with what's actually installed. No-op for KSP1 and KSP2.

KSA is the first game to use any of these, but all four are general `IGame` additions, not KSA special cases. Figured HebaruSan will want to look at the interface changes first before anything else.

## What's in this PR

- **Version handling.** KSA versions are `year.month.buildcounter.revision`, and the build counter is local to whichever machine produced that build, non-monotonic, meaningless for ordering. We pin it to 0 and order by the revision instead, matching how `KSA.VersionInfo.CompareTo` itself treats the same problem internally.
- **manifest.toml syncing.** KSA tracks enabled/disabled mods in its own manifest.toml, keyed by the on-disk folder name, not the CKAN identifier. CKAN now keeps that file in sync on install, uninstall, and upgrade.
- **StarMap launch integration.** Detects StarMap in the mods folder, swaps in the right executable at launch, and writes StarMap's own config so it points at the right install.
- **NetKAN/metadata changes.** `ksp_version`'s schema pattern is widened to 4 parts, `KsaVersionTransformer` normalizes declared compatibility at inflation time, and `GameVersionComponentsValidator` rejects a non-zero build counter for KSA and a 4-part version for every other game. We deliberately kept the shared `ksp_version` field name rather than renaming it, to avoid bundling a spec change into this PR. A game-neutral field name would be a good separate proposal, and we'd welcome it.
- **Schema widening** for KSA's `install_to` target.

## Evidence this works, not just that it compiles

- Full test suite green on current master, KSP1, KSP2, and KSA are all covered. We regression tested by hand against real KSP1/KSP2 installs too, not just by reading the diff.
- Two of the shared code paths we touched do change behavior for every game, worth calling out on their own: `IOExtensions.GetDrive` now resolves the drive from the nearest existing ancestor instead of the path as given, which fixes a `DriveNotFoundException` on Linux. And the install free-space check now measures the mod directory's drive rather than `GameDir`'s, which lands on the same drive as before for KSP1 and KSP2 and only actually differs for a game with an external mod root.
- The one real bug the beta turned up: the default install location for KSA on Windows is under Program Files, and that made the `<GameDir>/CKAN` state folder uncreatable, crashing with a raw permissions exception when adding the instance. That's fixed now, guarded with an actionable error message instead (KSAModding/CKAN-KSA#46, KSAModding/CKAN-KSA#47).
- A public beta [v1.36.4-ksa-beta-1](https://github.com/KSAModding/CKAN-KSA/releases/tag/v1.36.4-ksa-beta-1) has been out with real KSA testers running real mods. Feedback so far has been positive.
- Fork CI is green.

## Open design questions

KSA's shared mods folder means two registered KSA instances currently interfere with each other if you manage both through CKAN, since they'd share one physical folder and one manifest.toml but separate CKAN registries. Right now CKAN warns and asks for confirmation before letting that happen (see KSAModding/CKAN-KSA#19).

The real fix depends on whether the KSA developers can give us a way to point an install at its own mods directory, we've asked and are waiting on an answer. If that's not possible, the fallback is making CKAN's registry live next to the mod root itself rather than per-instance, so multiple instances become correct (shared) rather than silently broken, even though that's not the same as true isolation. We didn't want to hold this whole PR on that answer, since everything else here works with KSA.

Related to that: should a game with an external mod root keep its CKAN state (registry, downloads, etc.) next to that mod root instead of inside `GameDir`? That would be a bigger change than anything in this PR.

## Org and branding

With everything moving into the KSP-CKAN org, the project reads Kerbal end to end right now, the name expansion, the README, the in-app text, while it'd be hosting a second, unrelated game's community too. We don't have a concrete proposal for that (the "CKAN" name itself obviously isn't going anywhere), and it doesn't block this PR.

## What comes after this

Once this lands, a few more pieces follow in order: the metadata tester (xKAN-meta_testing, already coded and waiting on this to merge), the indexing bot (NetKAN-Infra, needs your team for the actual AWS deploy), the status page (already done), and wiki docs. Some of the deploy plumbing for the indexing bot is already real too, `InflatorKsa` is in `deploy.yml`/`bounce.yml` on our fork (KSAModding/CKAN-KSA#45), it just doesn't do anything yet since redeploy only works once the ECS service from the NetKAN-Infra stack actually exists. None of this blocks this PR, but wanted to be upfront about what's coming and what's already sitting there ready.

## Links

https://github.com/KSAModding/CKAN-KSA
https://github.com/KSAModding/KSA-NetKAN
https://github.com/KSAModding/KSA-CKAN-meta
https://github.com/KSAModding/xKAN-meta_testing
https://github.com/KSAModding/NetKAN-status
https://github.com/KSAModding/NetKAN-Infra

@Maximilian-Nesslauer and I are open to answer any questions, and thanks for taking the time to look at this.


Maximilian-Nesslauer:
@HebaruSan Hi, i just wanted to check in if you have any intermediate review feedback yet? :)

Let me know if i can provide more info somewhere.

averageksp:
Hey! We’re going to close this PR for now, as we’ve archived our repository and decided to move on with our own solution.

Of course, feel free to implement the changes from the PR if you still find them useful. We just won’t be maintaining this fork/PR going forward.

Thanks for taking the time to look at it!

PR #4464, opened 2025-11-28
Add KSP2 Redux support

dandoesgithub:
Preparations for KSP2 Redux

PR #4704, opened 2026-08-05
KSP2 Redux Support

cheese3660:
This adds support for KSP2 Redux to CKAN, by adding the `mods` folder (which is where KSP2 Redux puts its mods), and then the version number the latest KSP2 Redux stable version uses (0.2.8.5)

Supercedes #4464


cheese3660:
I dont know why that smoke test timed out - there is nothing in my code that shouldve touched that?

averageksp:
Probably just GitHub's new rate limit issues, try re-running the Workflow

cheese3660:
Gonna post this on here - there are KSP2 Redux mods that would benefit if they were accessible via CKAN, for example some of Falki's mods such as https://spacedock.info/mod/4424/SAS%20Extended

